PR #115, opened 2012-09-18
File system for attachments instead of idb. Orphaned attachment cleanup. Fix for deleting docs with attachments. Additional attachment tests.

chendricks:


mudcube:
Would it be possible for something like this to make it into the release version? This would be very useful for creating apps for iOS/Android using Phonegap/Cordova; the Apache cordova-plugin-file uses the same specs as the HTML5 FileSystemAPI; https://github.com/apache/cordova-plugin-file 

The benefit would be unlimited storage in iOS apps (at least for attachments, along with the 50MB limit on JSON entries) without taking the speed hit from using the SQLite plugin.

nolanlawson:
@mudcube I think this PR is pretty dead by now (3 years old), but for unlimited storage on iOS there is now the FruitDOWN adapter (http://pouchdb.com/adapters.html#pouchdb_in_the_browser), although it will probably still give you a performance hit because WebKit's IDB implementation is slow.

For storing large binary data on iOS we don't really have a good solution; sorry about that.

mudcube:
@nolanlawson Oh, that's totally fine! This project is awesome. I was hoping to help improve the situation with binaries on iOS. Would you be interested in an updated PR request making FileSystemAPI work as an option for developers?

nolanlawson:
@mudcube I think at this point it would make more sense to implement it as a separate adapter (ala [worker-pouch](https://github.com/nolanlawson/worker-pouch) or [socket-pouch](https://github.com/nolanlawson/socket-pouch)), but yeah, that would be awesome. :)

Issue #802, opened 2013-06-11
Feature Request - Method to Purge a document


jpgreen:
Feature Request - Method to Purge a document

Note: Purge is different from remove (or delete). It is as though the database had never heard of document at all. 

Info: [http://wiki.apache.org/couchdb/Purge_Documents](http://wiki.apache.org/couchdb/Purge_Documents)

doublerebel:
@ddouglascarr your purge addition looks good, is there anything missing?  I'm looking for this feature also and am willing to help.

ddouglascarr:
Hi @doublerebel. Sorry, I missed you mention in my events feed. 

There's quite a bit to do, and any help would be fantastic.
* I've just committed the websql adapter stuff, but that leaves the indexDB and levelDB adapters, which I have no experience with.
* purge_seq is not currently getting recorded. (This will probably involve modifying PouchDB.prototype.info()). This is my next mission.
* The view and find plugins will need updating. (I'm going to make a comment about this shortly)
* The above may mean this needs to be turned into a plugin. (Again, see next post).

ddouglascarr:
Implementing this feature will involve breaking backwards compatibility (and bumping the major version). Because of this, I think that it should perhaps be a plugin instead.

The problem is that any indexing plugins (pouchdb-view, pouchdb-find, etc) will need to check the databases purge_seq, react, and re-index from scratch if the purge sequence has advanced more than one since their last indexing event.

This is detailed in the couchdb documents here: http://docs.couchdb.org/en/latest/api/database/misc.html#updating-indexes

I think what I've done so far could be made into a plugin pretty easily. I don't know about the purge sequence though. I'm planning on looking into that next. Having it be a plugin would have the advantage of hiding the functionality from new users, who tend to struggle to differentiate between purge and remove (with potentially disastrous consequences). 

What are people's thoughts here?

stenit:
I'd like to have the plugin approach due to the "urgency" of this feature. Later on, this could be part of a new major version. To bridge this time, for me a plugin sounds like a good solution.

nolanlawson:
I propose we resolve this problem by adding a new event to the `changes()` API: `'purge'`. Then the mapreduce plugin can simply listen for purge events (as it's already listening to changes) and delete as necessary. It'll just be understood that http adapters don't emit this event. @daleharvey  @calvinmetcalf thoughts?

nolanlawson:
[...]

PR #1346, opened 2014-02-09
(#802) - adding purge

sygi:
First approximation of purge feature. Will have to add error handling and other stuff + I have no idea why my test fails (works on node shell).


nickcolley:
Is the view invalidation the only thing missing here?

PR #2676, opened 2014-08-24
Support Sync Gateway

ssured:
Adds support for:
- multipart responses
- _removed property in the _changes feed

### DIFF
--- a/lib/deps/ajax-browser.js
+++ b/lib/deps/ajax-browser.js
@@ -59,27 +59,10 @@ function ArrayBufferToString(buffer) {
         binary += String.fromCharCode( bytes[ i ] )
     }
     return binary;
-
-    // return BinaryToString(String.fromCharCode.apply(null, Array.prototype.slice.apply(new Uint8Array(buffer))));
 }
 
 function StringToArrayBuffer(string) {
     return StringToUint8Array(string).buffer;
-}
-
-function BinaryToString(binary) {
-    var error;
-
-    try {
-        return decodeURIComponent(escape(binary));
-    } catch (_error) {
-        error = _error;
-        if (error instanceof URIError) {
-            return binary;
-        } else {
-            throw error;
-        }
-    }
 }
 
 function StringToBinary(string) {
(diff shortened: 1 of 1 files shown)


gabrielnau:
@ssured Using Pouch with the Sync Gateway, I'm looking at your pull request and the [other](https://github.com/pouchdb/pouchdb/pull/3384) one.
I noticed you [removed the cache](https://github.com/pouchdb/pouchdb/pull/2676/files?diff=unified#diff-7c25e16427942d4ed6f34550f662b438L38) here, and having the same need as you to do so, I just wanted to let you in the case you missed it there is actually a way to use the options to do so : 

```javascript
local = new PouchDB('foobar')
remote = new PouchDB('http://foobar.com', { ajax: { cache: true } })
replication = localDb.replicate.from(remote)
```

I initially though ```cache``` option was not used at all, as I was using the main API instead of the shorthand, which does set default options on the remote part : 

```javascript
local = new PouchDB('foobar')
Pouch.replicate('http://foobar.com', local)
```

@nolanlawson I was going to update the documentation but this option is clearly documented ! I wonder though if the caveat explained above shouldn't be clearly explained ?

lib/adapters/http.js
```diff
@@ -664,7 +664,8 @@ function HttpPouch(opts, callback) {
 
       var keysAsString =
         'keys=' + encodeURIComponent(JSON.stringify(opts.keys));
-      if (keysAsString.length + params.length + 1 <= MAX_URL_LENGTH) {
+      // FIX for sync gateway, which does not accept multiple keys on a GET request
+      if (false) { //keysAsString.length + params.length + 1 <= MAX_URL_LENGTH) {
```
gabrielnau: @ssured I couldn't find anything documenting this, do you remember where does it come from please ?

nolanlawson:
What kind of custom behavior are you thinking? If it's small enough, I think we would rather just support CSG out-of-the-box with no extra config.

lib/adapters/http.js
ssured: I think I experimented and thus found out. SyncGateway only implements the replication protocol, not the full CouchDB API.
Try querying your SG instance: `http://...:5984/db/_all_docs?keys=["id1","id2"]` and see if it is implemented yet.

gabrielnau: You're right I missed that : http://developer.couchbase.com/mobile/develop/references/couchbase-lite/rest-api/database/get-all-docs/index.html

I believe though the ```keys``` parameter is simply ignored.

gabrielnau:
I was indeed referring to the work already done by @ssured for the support of ```_removed``` field. I will try to add some tests to document that.

PR #2681, opened 2014-08-25
(#2676) - WIP - support for Sync Gateway

nolanlawson:
Holy moley, this is way too much code to take in all at once.

I think we need to break this down into smaller steps:

1) support multipart. this is useful for CouchDB too, anyway.
2) support the other Sync Gateway crap (_removed, etc.)
3) Add tests for the Sync Gateway. (we can use the community edition, right?)

As goals, I think 1 is much more achievable than 2 and 3. This code needs a lot of work before we can get there, though.  For starters:

1) fixed the jshint stuff for ya
2) the arrayBufferToString stuff etc. should go into utils. I believe some of that is already implemented as well
2) `CSG does not accept multiple keys on a GET request`. I believe this is true for Cloudant as well, but it breaks Safari 6 last I checked, so we're doomed I guess. Maybe we can detect support by catching an error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,3 @@
-This repository add support for Sync Gateway to PouchDB. Use at your own risk.
-
 [PouchDB](http://pouchdb.com/) - The Javascript Database that Syncs
 ==================================================
 
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -666,7 +666,7 @@ function HttpPouch(opts, callback) {
       var keysAsString =
         'keys=' + encodeURIComponent(JSON.stringify(opts.keys));
       // FIX for sync gateway, which does not accept multiple keys on a GET request
-      if (false) { //keysAsString.length + params.length + 1 <= MAX_URL_LENGTH) {
+      if (keysAsString.length + params.length + 1 <= MAX_URL_LENGTH) {
         // If the keys are short enough, do a GET. we do this to work around
         // Safari not understanding 304s on POSTs (see issue #1239)
         params += (params.indexOf('?') !== -1 ? '&' : '?') + keysAsString;
--- a/lib/deps/ajax-browser.js
+++ b/lib/deps/ajax-browser.js
@@ -5,19 +5,20 @@ var errors = require('./errors');
 var utils = require("../utils");
 var hasUpload;
 
-// http://updates.html5rocks.com/2012/06/How-to-convert-ArrayBuffer-to-and-from-String
-var ArrayBufferToUtf8 = function (arrayBuffer) {
+// http://updates.html5rocks.com/2012/06/
+// How-to-convert-ArrayBuffer-to-and-from-String
+function arrayBufferToUtf8(arrayBuffer) {
   var result = "";
   var i = 0;
   var c = 0;
-  var c1 = 0;
   var c2 = 0;
   var c3 = 0;
 
   var data = new Uint8Array(arrayBuffer);
 
   // If we have a BOM skip it
-  if (data.length >= 3 && data[0] === 0xef && data[1] === 0xbb && data[2] === 0xbf) {
+  if (data.length >= 3 && data[0] === 0xef &&
+      data[1] === 0xbb && data[2] === 0xbf) {
     i = 3;
   }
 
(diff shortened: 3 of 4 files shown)


jchris:
If the multi part code is extracted, should we expect the existing tests to execute it, since it sets the Accept header for multipart?

If I put together a pull with just the stuff in Ajax-browser.js will we learn useful stuff from Jenkins?

nolanlawson:
Yeah, actually I started working on that, but then gave up due to complexity/lazyness: https://github.com/pouchdb/pouchdb/issues/2832

Adding accept:multipart should work fine, although we may want to make it user-configurable (JSON is better for debugging I imagine). PouchDB Server doesn't support it for downstream, but it does support it for upstream, so the tests should pass.

Also we use Travis rather than Jenkins now. :)

jchris:
I'm trying to figure out where the actual next step is. I think there are handful of implementations of multi-part, but I can't tell which is the best candidate to run via Travis. (And also for me to test locally.)

nolanlawson:
For testing locally, there are steps described in TESTING.md, but basically if you have CouchDB 1.6 running on localhost:5984 then it should be pretty easy:

```bash
export COUCH_HOST=http://localhost:5984 # if you want to point it somewhere else
npm install && npm run dev
# now open up localhost:8000/tests/integration
```

I would start with `ajax-browser.js` and just send `Accept:multipart` and see what CouchDB sends back, then process that in `ajax-browser.js` as well. Also apologies in advance; `ajax-browser.js` is a little messy, been meaning to refactor it. :stuck_out_tongue:

daleharvey:
My plan for `ajax-browser` was that we would switch all our `ajax` calls to node `request()` compatible calls and have ajax-browser be a browser compatible polyfill (or use `browser-request`, but I think thats likely a bit heavy), takes the pouchdb logic out of ajax and would make the seperation cleaner, I have a branch up somewhere that started this

gabrielnau:
@jchris About the ```_removed``` field : it's very simple to let Pouch ignore this field, but it doesn't seem to me the actual deletion is handled in existing [proposed](https://github.com/racido/pouchdb/commit/30d62fd6c3de4b6794162ff3745b4c2c8ed0ac26) patch. I've found Couchbase lite clients don't support it yet, so shall Pouch simply ignore it as well or actually cleanup the local db ?

jchris:
Removed can be ignored in a first pass implementation.

Dale should we be working to move to request, or adding multipart to Ajax
browser? Do you have a link to your topic branch? I don't mind rebasing and
testing

On Friday, January 16, 2015, Gabriel Nau <notifications@github.com> wrote:

> @jchris <https://github.com/jchris> About the _removed field : it's very
> simple to let Pouch ignore this field, but it doesn't seem to me the actual
> deletion is handled in existing proposed
> <https://github.com/racido/pouchdb/commit/30d62fd6c3de4b6794162ff3745b4c2c8ed0ac26>
> patch. I've found Couchbase lite clients don't support it yet, so shall
> Pouch simply ignore it as well or actually cleanup the local db ?
>
> —
> <https://github.com/pouchdb/pouchdb/pull/2681#issuecomment-70281923>.
>


-- 
Chris Anderson
http://www.couchbase.com
http://twitter.com/jchris

gabrielnau:
About multiple keys in ```GET _all_docs``` : 


@nolanlawson CGS [supports](http://developer.couchbase.com/mobile/develop/references/couchbase-lite/rest-api/database/post-all-docs/index.html) multiple keys on a POST ```_all_docs``` request. The breaking issue with the GET request is keys are simply ignored, which conduct to fetch everything.

I've read this thread #1239, but I still don't understand how ```test.all_docs.js``` specs can pass given ```keys``` argument isn't mentioned as being supported in neither [1.5.1](http://docs.couchdb.org/en/1.5.1/api/database/bulk-api.html) or [1.6.1](http://docs.couchdb.org/en/1.6.1/api/database/bulk-api.html). I also wonder if we shouldn't try to make the Safari bug workaround the exception instead of the rule ?

@jchris, if this is in the protocol, I wonder if Couchbase will try to match it ?

Being new to Couch and Couchbase in general, I'm sorry if I say anything dumb. I would like to settle on something for this problem that make sense to support Couch, CGS and Safari 6 ..

jchris:
I think it's not in the protocol, but I think the official answer is "we'll cross that bridge when we get to it."

Currently the multi-part stuff is the biggest cause of test case failures, so that's where I'm focussed now. I'm sure there will be other stuff (like this) that we have to address but I'm trying to take it one step at a time.

@daleharvey I forgot to tag you on my question about your request patch / branch. I can't tell which one it might be by the branch names, but if you can point me to it I'll take a pass at it.

gabrielnau:
@jchris I understand your point, and if @nolanlawson wants to support this, I would like to propose a patch to handle it on the PouchDB side instead of the Couchbase side, which seems simpler.

PR #2733, opened 2014-09-05
Use a real PNG dataurl to detect blob support

ssured:
Using an empty string results in a 404 error on Chrome 37, probably a security protection. Using a real png string (transparent one pixel) then no 404 happens.
`res.type` evaluates to `'none'`, so the detecting still works, detecting Chrome37 does not have blobsupport


nolanlawson:
@ssured I made a big mistake here. Your commit was right; blank pngs cause 404s in Chrome whereas the transparent 1x1 does not. This was causing #3326.

Issue #2890, opened 2014-10-17
DOM freezes when storing very large attachments


nolanlawson:
Steps to repro: store a huge attachment (>500KB), watch the DOM stutter. Animated GIFs are good for this.

Looks like we should probably store binaries in chunks rather than all as one big blob.

Issue #2818, opened 2014-10-02
Compaction should retain a single copy for each unique attachment


nolanlawson:
Thus spake the [CouchDB weekly news](http://blog.couchdb.org/2014/10/02/couchdb-weekly-news-october-02-2014/). Also I've been thinking about this: the dedup feature is nice, but it sucks that attachments just stick around for forever, even after you delete them.

nolanlawson:
This is long since fixed

PR #2900, opened 2014-10-19
(#2890) - websql: avoid hex() for binaries

nolanlawson:
See #2899 for backstory. Rebased on #2818 because I didn't want to bitrot myself.

For a demonstration, compare these two blocks: [before](http://bl.ocks.org/nolanlawson/raw/73f15f3fe612b8770e79/) and [after](http://bl.ocks.org/nolanlawson/raw/11b1af329e1f8f27fa13/). The difference should be obvious.

Using `hex()` was a nice hack, but decoding hex strings in JavaScript is slow. Apparently just using `replace()` to swap out the `\u0000` is much faster. I borrowed @neojski's trick from pouch-collate for this.

### DIFF
--- a/lib/adapters/websql.js
+++ b/lib/adapters/websql.js
@@ -4,8 +4,29 @@ var utils = require('../utils');
 var merge = require('../merge');
 var errors = require('../deps/errors');
 var vuvuzela = require('vuvuzela');
+var parseHexString = require('../deps/parse-hex');
+
 function quote(str) {
   return "'" + str + "'";
+}
+
+// escapeBlob and unescapeBlob are workarounds for a websql bug:
+// https://code.google.com/p/chromium/issues/detail?id=422690
+// https://bugs.webkit.org/show_bug.cgi?id=137637
+// The goal is to never actually insert the \u0000 character
+// in the database.
+function escapeBlob(str) {
+  return str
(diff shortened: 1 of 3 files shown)


jason-s:
[COBS](http://en.wikipedia.org/wiki/Consistent_Overhead_Byte_Stuffing) would be [more efficient for preventing embedded nulls](http://www.embeddedrelated.com/showarticle/113.php).

nolanlawson:
That's interesting! I would want to see a jsperf, though, to confirm that running COBS over strings is faster in most browsers than `replace()`ing a single character 3 times. E.g. I wonder if, by replacing a single character with more than one characters, this maybe causes an array shift under the hood, which would actually slow things down.

jason-s:
>to confirm that running COBS over strings is faster 

Faster: no. (though comparable, I would think, depending on how string manipulation is implemented.) You can run the COBS algorithm in place; it creates a string with 1 additional byte (or word) in length. 

More efficient in storage: yes. The thing about replacing unicode `0000`, `0001`, and `0002` is that although these are unusual string characters, they are likely to occur in binary data, so if you have data with lots of `0000`, then the approach used here essentially doubles the data size.

nolanlawson:
Ah, that's true. I forgot that we're replacing a single character with two characters here. OK, it's worth giving it a shot! :)

Issue #3262, opened 2014-12-26
Minor style fixes


yaronyg:
When I was studying leveldb.js I ran across some minor style issues. In my bid to fight entropy I submit some truly minor improvements.

PR #3263, opened 2014-12-26
(#3262) - Minor style fixes

yaronyg:
Inlined some unnecessary function calls, extraced some re-used logic,
indented a few things.


marten-de-vries:
@yaronyg There is a queue. Only an x amount of jobs are run in parallel, if there are more they wait.

Issue #2915, opened 2014-10-23
Replication busted if skipSetup true


rymohr:
Just upgraded to 3.0.6 from 2.x and I'm running into replication issues.  The pouch constructor promise never resolves so the replication waits indefinitely.

Pretty sure there should be a `callback(null, self);` call after this line yeah?

https://github.com/pouchdb/pouchdb/blob/275913631077f880a08f2524e8515034f468e579/lib/constructor.js#L147

daleharvey:
This landed, thanks

PR #3284, opened 2014-12-28
(#2915) - Replication busted if skipSetup true

lokeshj:
Fix by deferring the callback as per  https://github.com/pouchdb/pouchdb/issues/2915#issuecomment-60303047


marten-de-vries:
Was green, so merged: a54a35d6431a7161be06ff8107fde9c53b45a036. Thanks!

PR #3285, opened 2014-12-28
(#3285) - refactor idb changes

nolanlawson:
Also fix a small thing I noticed with blob support:
if the DB is cached, then we should use the `api._blobSupport`
from the cached DB. Looks like before, we would just assume
no blob support the second time you opened the database.

Also used `utils.Map()` instead of `{}` where
appropriate.


nolanlawson:
Closing for now. I need to rebase and try again when IE10 is more stable in Saucelabs.

PR #3294, opened 2014-12-29
Specify message in error object when abortReplication is called.

robertkeizer:
If a message hasn't been specified, use the reason we're aborting.

Noticed that the message was empty, and the reason for aborting was getting lost.


daleharvey:
So I think this is helpful for now, but I would like to follow up and document how people should be handling errors, improving message vs name vs reason etc

Thanks a lot :) - https://github.com/pouchdb/pouchdb/commit/dd3e3b755d3bac26767d7fda387db57ae06619a4

Issue #136, opened 2012-10-02
Get tests passing against cloudant


daleharvey:
As title

I believe the different checkpoint used by bigcouch is to be considered opaque and in both implementation of plain couchdb and bigcouch we should just be sending transferring the token without assuming a type

daleharvey:
Ok, there will be follow ups, running against a cluster, removing the isCouchDBMaster() checks, but we now have fully passing test runs for both browser and node against master, closing the oldest bug we have out, thanks for all the hard work, especially @willholley

PR #3297, opened 2014-12-29
(#136) - Fix "Testing successive new_edits.."

willholley:
Fix "Testing successive new_edits to the same doc, different content" against CouchDB 2.0. Sequence numbers in CouchDB 2.0 cannot be compared for equality and the assertions around last_seq don't seem to be a core part of this test - just remove them.


daleharvey:
Looks good to me https://github.com/pouchdb/pouchdb/commit/e99ce4fa03dded9687bf3ab71717ccf0dddb54ae

PR #3301, opened 2014-12-30
(#136) - call done() when skipping changes last seq

willholley:
When skipping the "changes last seq" test against CouchDB 2.0 (see #3271) call done() instead of returning true. Without this, the test never completes.


daleharvey:
Crap sorry should have caught that 

https://github.com/pouchdb/pouchdb/commit/8d0cfd444f2e2f19f35d447508631851cf360140

PR #3302, opened 2014-12-30
(#136) - Fix "descending changes" / CouchDB 2.0

willholley:
_changes in CouchDB does not guarantee order so skip the test which asserts this when testing against CouchDB master.


daleharvey:
makes sense, cheers - https://github.com/pouchdb/pouchdb/commit/d4cf55bbbc303fda510f5e06e229504564261c96

PR #3303, opened 2014-12-30
(#136) - Do not assume order in "changes limit and filter"

willholley:
_changes in CouchDB 2.0 does not guarantee order so do not assume we can find elements by index in tests. Instead, look up a specific document by id.


daleharvey:
Yup cheers - https://github.com/pouchdb/pouchdb/commit/4b0f7a897957e9fd24dc30f2187f3f8e57da99d9

PR #3304, opened 2014-12-30
(#136) - Skip compaction tests against CouchDB 2.0

willholley:
CouchDB 2.0 does not expose the _compact endpoint via the public interface. Instead, compaction can be triggered via the backdoor admin
port. In future we may want to test against this but skip for now.


daleharvey:
Hrm yeh definitely seems we will want to revisit this

https://github.com/pouchdb/pouchdb/commit/57d0a8522665382109e77b906a9a7ea35882802d

PR #3306, opened 2014-12-30
(#136) - Skip "Revs diff with empty revs" / CouchDB 2.0

willholley:
CouchDB 2.0 currently fails to respond when an empty revision set is passed to _revs_diff (COUCHDB-2531). Until this is fixed, skip the test.


daleharvey:
Thanks - https://github.com/pouchdb/pouchdb/commit/123cb573cd062f3a71288aad97fe3bd9379dc524

PR #3307, opened 2014-12-30
(#136) - Skip update_seq assertions in replication tests

willholley:
When testing against CouchDB 2.0, we cannot assume that the sequence number will be updated incrementally - skip the assertion.

This PR adds a helper function to consolidate the assertions against update_seq. In the helper, we skip the assertion if running against CouchDB master.

### DIFF
--- a/tests/integration/test.replication.js
+++ b/tests/integration/test.replication.js
@@ -60,6 +60,13 @@ adapters.forEach(function (adapters) {
       return changes;
     }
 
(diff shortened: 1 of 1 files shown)


daleharvey:
+1, but could you rebase and get a green run before merging please, thanks

willholley:
Merged in https://github.com/pouchdb/pouchdb/commit/66b19cdb9024681d65ff6070b0b97e5de76c5f36

PR #3308, opened 2014-12-30
(#136) - Fix "Test pull replication with many changes"

willholley:
CouchDB 2.0 represents sequence numbers as arrays which should be treated as opaque blobs. This prevents us using simple comparison operators on them (e.g. last_seq < current_seq).

Previously we were using a similar test to determine whether new changes were returned in a call to _changes (if params.since < result.last_seq, fetch more changes). This no longer holds so, instead, I've added a new field to the onChangesComplete callback, "finished", which indicates that the _changes call returned no results (i.e. we're finished).

### DIFF
--- a/lib/replicate.js
+++ b/lib/replicate.js
@@ -440,6 +440,10 @@ function replicate(repId, src, target, opts, returnValue, result) {
     if (state.cancelled) {
       return completeReplication();
     }
+    var filter = utils.filterChange(opts)(change);
+    if (!filter) {
+      return;
+    }
     if (
       pendingBatch.changes.length === 0 &&
       batches.length === 0 &&
@@ -458,7 +462,10 @@ function replicate(repId, src, target, opts, returnValue, result) {
     if (state.cancelled) {
       return completeReplication();
     }
-    if (changesOpts.since < changes.last_seq) {
+
+    // if no results were returned then we're done,
+    // else fetch more
+    if (changes.results.length > 0) {
       changesOpts.since = changes.last_seq;
       getChanges();
     } else {
@@ -515,10 +522,11 @@ function replicate(repId, src, target, opts, returnValue, result) {
         batch_size: batch_size,
         style: 'all_docs',
         doc_ids: doc_ids,
-        returnDocs: false
+        returnDocs: true // required so we know when we're done
       };
       if (opts.filter) {
-        changesOpts.filter = opts.filter;
+        // required for the client-side filter in onChange
+        changesOpts.include_docs = true;
       }
       if (opts.query_params) {
         changesOpts.query_params = opts.query_params;
--- a/tests/integration/test.replication.js
+++ b/tests/integration/test.replication.js
@@ -137,8 +137,10 @@ adapters.forEach(function (adapters) {
             result.ok.should.equal(true);
             result.docs_written.should.equal(docs.length);
             new PouchDB(dbs.name).info(function (err, info) {
-              info.update_seq.should.equal(numDocs, 'update_seq');
-              info.doc_count.should.equal(numDocs, 'doc_count');
+              verifyInfo(info, {
+                update_seq: numDocs,
+                doc_count: numDocs
+              });
               done();
             });
           }


daleharvey:
As far as I understand, the sequence numbers should not be opaque blobs, they can be complex objects that collate similiarly to view keys, so I think we should be using collate() for this test right?

willholley:
Worth checking this with the couchdb devs but I don't believe this to be the case with BigCouch. My understanding is that the sequence number is an encoded representation of the shard identity/sequence numbers which make up the database (where the shard identities which contribute to this could change over time e.g. if a shard copy servicing _changes is unavailable).

willholley:
also see http://stackoverflow.com/questions/20446673/update-sequence-changed-semantics-in-cloudant-db

willholley:
Aha - CouchDB 2.0 (and Cloudant) adds a new field to the _changes result - "pending". When present, this can be used as an indicator that there is nothing left to fetch (pending === 0).

willholley:
I've updated the PR to now use the "pending" field, if available, to determine whether further calls to _changes are required.

willholley:
rnewson: wilhol: daleharvey: the correct way to find if there are pending changes is to call _changes?since=blah and see if you get any rows back.

willholley:
As Bob suggested, I'm ignoring the "pending" field and just flagging that we're done when no additional results are returned from a call to _changes.

willholley:
This is going to need more thought (currently lots of broken tests). The principle works against Cloudant / CouchDB but the implementation currently adds an additional field into the changes result which is not ideal. We'd also need to implement the same field for the other storage adapters.

willholley:
OK - I think the latest PR represents a reasonable solution. Comments welcome :)

willholley:
@daleharvey @nolanlawson any feedback on this? With this change I think sync with CouchDB 2.0 / Cloudant will be fixed.

daleharvey:
So I would really like not have to expose anything more via the changes api, especially if its only for the replicator to function. the call to changes should only return empty documents if its a `live` call, or there are no more documents to fetch, so cant the replicator just check for an empty result set and assume complete?

willholley:
@daleharvey my reading of the changes processing in the http adapter (https://github.com/pouchdb/pouchdb/blob/master/lib/adapters/http/http.js#L879) is that the raw changes result is potentially filtered on the client before being passed to the "complete" callback where the decision to make further requests is made (https://github.com/pouchdb/pouchdb/blob/master/lib/replicate.js#L448). So, an empty result could be passed to onChangesComplete because all the changes were filtered out on the client, yet we would still want to make another server request.

The finished flag is the indication that no results were returned from the server (before client filtering occurs). That said, I would strongly favour a solution which doesn't change the output of the changes call - I just can't see a way to pass the information through to onChangesComplete without it.

daleharvey:
What about we do the filtering inside the replicator, we can leave the filter functionality in the changes api but if we never set it in the replicator we will know if we got a truly empty result set

willholley:
Y sounds possible (though maybe a little dirty). The other option is to
completely rearchitect batching of requests to changes - possibly a cleaner
long term solution but quickly gets into the weeds when I looked at it
yesterday.

What about we do the filtering inside the replicator, we can leave the
filter functionality in the changes api but if we never set it in the
replicator we will know if we got a truly empty result set

—
Reply to this email directly or view it on GitHub
<https://github.com/pouchdb/pouchdb/pull/3308#issuecomment-69453670>.

daleharvey:
Yeh doing the filtering inside the replicator isnt the nicest of solutions, but would avoid adding to the API, I dont mind having a few hacky solutions around but I would really prefer them being exposed to the user

willholley:
Cool - here's a (much simpler) commit which moves the filtering to the replicator.

daleharvey:
Woot, this looks great, and fixes almost half the tests left, will leave for others to look for a second but big :+1: from me

daleharvey:
Actually, I know how annoying it is to keep working with open PR's

https://github.com/pouchdb/pouchdb/commit/fe43de04557fd4df72ea9fe6b2e01c88e1ef8260

This is awesome, cheers for sticking with it

PR #3309, opened 2014-12-30
(#136) - Skip assertions which compare update_seq / CouchDB 2.0

willholley:
CouchDB 2.0 represents sequence numbers as arrays and they are not incremental. When testing against CouchDB master, skip assertions
that the update_seq should be above a specific integer.


daleharvey:
So I would like to come back to these ones very much, they arent incremental but they are supposed to collate we are using .above, but using pouchdb.collate we should get the same test that works across master / 1.6

daleharvey:
But will be much easier when -master is green, so merging for now, thanks

https://github.com/pouchdb/pouchdb/commit/8356f3b9ccf87288dbe5d1e4190643353f9add63

PR #3310, opened 2014-12-30
(#136) - Fix _changes ordering assertions in replication tests

willholley:
In CouchDB 2.0, _changes is not guaranteed to be ordered. Sort the response before making assertions on the contents.

I've also removed duplicate definitions of the simplifyChanges function so this sort only occurs in one place.


daleharvey:
Looks nice, thanks

https://github.com/pouchdb/pouchdb/commit/85ffee354b1734a6793d3fb61963bfe1797222f4

PR #3312, opened 2014-12-30
(#136) - Fix "Testing allDocs with some conflicts" / CouchDB 2.0

willholley:
When testing against CouchDB 2.0, skip assertion that update_seq should be within a known range. CouchDB 2.0 sequence numbers are not guaranteed to be incremental.


daleharvey:
Cheers - https://github.com/pouchdb/pouchdb/commit/65a3d95aab32585e344d806058a1662b7b0e99d0

PR #3320, opened 2014-12-31
(#2915) - Replication busted if skipSetup true

daleharvey:


marten-de-vries:
Was green, so merged: a54a35d6431a7161be06ff8107fde9c53b45a036

PR #3321, opened 2014-12-31
DONT MERGE try a bunch of firefoxes

nolanlawson:
I'm thinking we can nail down the firefox version as a last resort, if we cannot figure out what's causing the travis failures.


daleharvey:
Also looking at the code, fairly sure this will just reuse whatever firefox has been installed on the path :( I have a plan for testing fx versions

nolanlawson:
heh, yeah, they are all green. no worries, just wanted to test.

PR #3322, opened 2014-12-31
(#3322) - document _local and _design docs

nolanlawson:
I mostly just wanted to document `_local` docs, so `_design` docs are just filler because otherwise the section would have been short. :P

### DIFF
--- a/docs/_guides/compact-and-destroy.md
+++ b/docs/_guides/compact-and-destroy.md
@@ -107,3 +107,8 @@ Related API documentation
 
 * [compact()](/api.html#compaction)
 * [destroy()](/api.html#delete_database)
+
+Next
+----
+
+To wrap up, let's look at a special class of documents in PouchDB &ndash; local docs.
--- a/docs/_guides/local-documents.md
+++ b/docs/_guides/local-documents.md
@@ -1,0 +1,47 @@
+---
+index: 15
+layout: guide
+title: Local documents
+sidebar: guides_nav.html
+---
+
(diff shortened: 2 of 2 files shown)


marten-de-vries:
A user can only modify design docs when in admin party or using an admin account. The only case I can think of where that is the case is developing a Fauxton-like tool on top of PouchDB. In other words, the chance that PouchDB users will actually get some use out of the design document section is very small, I'm afraid.

Pointing to pouchdb-validation would help a bit (provides ``validate_doc_update`` offline), but I don't think we want to start duplicating plug-in documentation & the plugin page. Also, it's not the most used plug-in by far. Others would be more 'deserving' of a spot in the guide first.

In short: I like the /_local section a lot, but think the _design section should be scrapped. It's a shame that it makes the page very short, but I think it'll be too confusing otherwise.

daleharvey:
Agree with @marten-de-vries, I think the additional information about functionality not inside pouchdb is confusing, the extra docs are good though

nolanlawson:
OK, I'll just rewrite it as a "local docs" section.
Also in general I wouldn't mind some kind of "useful plugins" page or something. There is value in highlighting good ones rather than just dumping a big list of all of them.

nolanlawson:
Updated!

nolanlawson:
The commit message is wrong, but the content should be right.

marten-de-vries:
Great! :+1:

marten-de-vries:
Merged: b900d441044aaca8152d51a4ee05a4a534c6a050

nickcolley:
:+1: for this! Been referencing this to a few people so good to be able to link to some docs :)

PR #3323, opened 2014-12-31
DONT MERGE test ie

nolanlawson:
IE10 seems perma faily for my refactor branch, but I can't reproduce locally. So I want to see if master is even passing.


nolanlawson:
Wow, so IE10 only passed 1 out of 9 tests. This is the failiest I've ever seen it.

PR #3325, opened 2015-01-01
(#3325) - fix blob support w/ duplicate idbs

nolanlawson:
Just fixing this dumb bug, since my big refactor might not end up working in IE10.

FWIW, in Chrome 39 blobSupport is still false for some reason. Should probably investigate that separately.

### DIFF
--- a/lib/adapters/idb/idb.js
+++ b/lib/adapters/idb/idb.js
@@ -897,7 +897,7 @@ function init(api, opts, callback) {
   if (cached) {
     idb = cached.idb;
     instanceId = cached.instanceId;
-    idStored = cached.idStored;
+    api._blobSupport = cached.blobSupport;
     process.nextTick(function () {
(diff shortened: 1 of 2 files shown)


lib/adapters/idb/idb.js
```diff
@@ -1039,8 +1039,7 @@ function init(api, opts, callback) {
             };
           };
         }).catch(function (err) {
-          api._blobSupport = false;
-          checkSetupComplete();
+          return false; // error, so assume unsupported
```
nolanlawson: I'm not sure why we had this; it could potentially call `checkSetupComplete()` twice, which is wrong.

daleharvey:
Thanks, merged @ https://github.com/pouchdb/pouchdb/commit/54736a8fb512048e9071566ad222d8d474d8ac7b

Issue #3095, opened 2014-12-01
Selenium easy setup with chrome/firefox/phantomjs support proposal


vvo:
Hi, reading http://pouchdb.com/2014/11/27/testing-pouchdb.html I saw this:

```
I would love to see a project wrap up Selenium / ChromeDriver and possibly Cordova / Appium into a nice, well-documented module that installs and boots your browser(ish) platform of choice, ready to be driven by Selenium tests.
```

Well I made one: https://github.com/vvo/selenium-standalone

It downloads the right standalone selenium and chrome driver for you then exposes a command line or programmatic API to launchy/close it.

Do you think it would fit into the pouchdb test suite? So that we can remove this horrible 20mb vendor .jar.

Thanks.

daleharvey:
https://github.com/pouchdb/pouchdb/pull/3233 for some node-webkit work

daleharvey:
Merged in https://github.com/pouchdb/pouchdb/commit/bbf577d369763608f42d6eb8a7e67f2c79e0eb04

PR #3327, opened 2015-01-02
(#3095) - Use selenium-standalone to manage selenium

daleharvey:


nolanlawson:
If this works, I would love to start fixing all the plugins as well. It's a PITA to remove the selenium jar just for the benefit of bower users.

daleharvey:
@vvo any idea about whats happening here? locally this is running fine, but dying in travis

daleharvey:
Looks like its the same version issue I was talking about earlier, assumed if I did `export SELENIUM_VERSION=2.42.0` it would apply to all scripts, but doesnt seem so

nolanlawson:
Gonna play it safe and wait for a green in iphone+ie as well. Restarting those jobs.

daleharvey:
So iphone is faily randomly depending on what time of the day it is, it doesnt go through this at all as it uses the sauceClient and not the selenium server (https://github.com/pouchdb/pouchdb/blob/master/bin/test-browser.js#L206)

Its a fairly consistent error these days (security exception) so I will take a look at getting it back green soon, but I dont think it needs to block this

nolanlawson:
IE is green, so fair enough. +1

I'm mostly just paranoid about IE failures. iPhone failures are usually also caught by Phantom or Safari.

daleharvey:
Great thanks - https://github.com/pouchdb/pouchdb/commit/bbf577d369763608f42d6eb8a7e67f2c79e0eb04

PR #3328, opened 2015-01-02
(#136) - Fix "replication since" tests

willholley:
Two fixes here:
1. Ensure we find checkpoint documents which match array-based sequence numbers.
2. Alter the "replication since" test to not assume incremental sequence numbers.

### DIFF
--- a/lib/checkpointer.js
+++ b/lib/checkpointer.js
@@ -1,6 +1,8 @@
 'use strict';
 
 var utils = require('./utils');
+var pouchCollate = require('pouchdb-collate');
+var collate = pouchCollate.collate;
 
 function updateCheckpoint(db, id, checkpoint, returnValue) {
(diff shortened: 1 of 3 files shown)


daleharvey:
So I dont think we should ever be using JSON.stringify to test complex objects, the encoding is undefined in json so `{a: b, c:d} !== {c:d, a:b}` https://github.com/pouchdb/collate can do this (we already use it in map reduce) or there may be another way but not seeing it in utils

nolanlawson:
Yeah, you'll need to require collate.

willholley:
thanks - updated to use collate

daleharvey:
Awesome, merged, thanks

https://github.com/pouchdb/pouchdb/commit/bc21cf75ccbdea47c840b3e4b9bf1a60083df23f

PR #3329, opened 2015-01-02
(#136) - Fix "Replicates deleted docs w/ delay" against CouchDB 2.0

willholley:
Alter "Replicates deleted docs w/ delay" to not depend upon an incremental database sequence number. Instead, test incoming changes against the specific document properties that we're interested in (in this case, id and deleted).


daleharvey:
Looks good, thanks

https://github.com/pouchdb/pouchdb/commit/7473df990550db52423bcc4f4a09c100e318f6c2

PR #3330, opened 2015-01-02
(#136) - Fix "Reporting write failures if whole saving fails"

willholley:
Removed the assertion on last_seq. I'm not sure what this adds to the test and we have other coverage which ensures last_seq is set correctly.

In CouchDB 2.0 this assertion is invalid because sequence numbers are not integers.


daleharvey:
Makes sense - https://github.com/pouchdb/pouchdb/commit/7b6431c4719f838e415265a2a25956868b521f60

Issue #942, opened 2013-09-25
replicate returns all docs in docs_written even if there was an error


anowak:
`docsWritten` in `replicate` does not check if there were any errors, so even if saving all documents failed (as e.g. in #941) it returns all documents as successfully written to the target database.

PR #3333, opened 2015-01-02
(#136) - Fix "Reporting write failures (#942)"

willholley:
Change the "Reporting write failures (#942)" test to avoid dependencies on the _changes feed order. This was failing against CouchDB 2.0 because sometimes document "b" would come before "a" in the feed. The mocking function would return a write error for document "b" which caused the replication to abort at which point the assertions around documents read / written fail.

The mocked bulk_docs endpoint now returns failure for the second document, regardless of the _id.

### DIFF
--- a/tests/integration/test.replication.js
+++ b/tests/integration/test.replication.js
@@ -2482,6 +2482,7 @@ adapters.forEach(function (adapters) {
       var remote = new PouchDB(dbs.remote);
       db.bulkDocs({docs: docs}, {new_edits: false}, function (err, _) {
         var bulkDocs = remote.bulkDocs;
+        var bulkDocsCallCount = 0;
         remote.bulkDocs = function (content, opts, callback) {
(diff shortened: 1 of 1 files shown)


daleharvey:
Looks like a tricky one, thanks - https://github.com/pouchdb/pouchdb/commit/11880385363f7b44ebafc22b784a584b5f30cef5

marten-de-vries:
My latest express-pouchdb PR fails Travis on the test this changed, that's a bit too much of a coincidence:
https://travis-ci.org/pouchdb/express-pouchdb/builds/45806679

willholley:
@marten-de-vries I can't reproduce the timeouts locally - are you able to?

marten-de-vries:
@willholley I did further testing, turns out it is a problem with my express-pouchdb PR and unrelated to this except that it happens to show up using the same test as this PR edited. Sorry for the noise.

Issue #3326, opened 2015-01-01
blobSupport is false in Chrome 39


nolanlawson:
Follow-up to #3325.

I noticed blob support is still false even in Chrome Canary. Seems to be due to the 404, although I'm not sure why the blob url is giving us a 404. In Firefox/IE blob support is true.

daleharvey:
Fixed by https://github.com/pouchdb/pouchdb/commit/fce1c97e6bfaa483e8af21b5244b437f040f43fb

PR #3334, opened 2015-01-02
(#3285 #3326) - refactor idb, fix blob support in Chrome

nolanlawson:
This PR is a two-for-one.

The old 404 we were getting for blobs in Chrome was due to using an empty PNG. I switched to using a transparent 1x1 transparent PNG as originally advised by @ssured (but stupidly ignored by me).

Why a transparent PNG? Because an empty PNG causes the 404 in Chrome, whereas we have to use a PNG because apparently plaintext doesn't suss out the error in Chrome 37.

I verified manually that the 404 is no longer thrown in Chrome 39 and that `blobSupport` is set to true.

**Please wait for a green in IE10 before merging.**

### DIFF
--- a/lib/adapters/idb/idb-bulk-docs.js
+++ b/lib/adapters/idb/idb-bulk-docs.js
@@ -17,7 +17,7 @@ var decodeMetadata = idbUtils.decodeMetadata;
 var encodeMetadata = idbUtils.encodeMetadata;
 var idbError = idbUtils.idbError;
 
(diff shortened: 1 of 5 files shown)


nolanlawson:
IE10 in Win 7 in VirtualBox passes on my machine.

nolanlawson:
Dunno why this is failing in IE10, but it seems genuine. Something to do with the `idb` caching. Better not to put IE10 in perma-fail.

PR #3336, opened 2015-01-03
(#3334) - DON'T MERGE - test IE many times

nolanlawson:
I want to see if https://github.com/pouchdb/pouchdb/pull/3334 will pass even once on IE10.

I'm only nervous because I think the PR might introduce a bug that my local tests aren't catching for some reason.


nolanlawson:
Not a single success in IE out of 10 tries. I think that's my answer.

PR #3337, opened 2015-01-03
(#3326) - fix blob support in chrome

nolanlawson:
Verified manually that `api._blobSupport` is true in Chrome 39 and Firefox 36. Also the 404 is gone from Chrome, so there's no need for the info message any more.

Please wait for a green in IE10 to be safe.

### DIFF
--- a/lib/adapters/idb/idb.js
+++ b/lib/adapters/idb/idb.js
@@ -1003,16 +1003,22 @@ function init(api, opts, callback) {
       if (!blobSupportPromise) {
 
         // make sure blob support is only checked one
-        blobSupportPromise = new utils.Promise(function (resolve) {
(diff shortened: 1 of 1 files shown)


nolanlawson:
Explanation of why this works is in the other, slightly over-ambitious PR which I closed: https://github.com/pouchdb/pouchdb/pull/3334#issue-53284470

nolanlawson:
Green in IE!

marten-de-vries:
Code looks good, nice work. :+1:

daleharvey:
Nice, thanks https://github.com/pouchdb/pouchdb/commit/fce1c97e6bfaa483e8af21b5244b437f040f43fb

Issue #3155, opened 2014-12-12
Clarify events


daleharvey:
The documentation isnt particularly clear about the events, what data they get and in what circumstances they get fired

I was using `replication.on('error', ....` to display an error when replication failed, however it also gets triggered on conflicts which confused me somewhat

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/fb3a8e0e0621dadcc565bb5f47ca21dfd18fa04d

PR #3341, opened 2015-01-04
(#3155) - Replace syncStopped / started with active / paused

daleharvey:


nolanlawson:
Need to actually document `retry`; otherwise the `active`/`paused` documentation doesn't make much sense.

So are we going to release 3.3.0 and formally unveil `retry`? I would be -1 on that, given that we have a scheduled release tomorrow, and I was hoping to release a 3.2.1 with bugfixes only, so that we could have some more time to really get the retry stuff hammered out.

daleharvey:
Yeh I am good with waiting for the next release to merge this, it would be better to have retry enabled but active / paused is still useful without retry (its how you would show a notification that indicates whether documents are currently replicating, which is a little problematic right now), I think we can merge this (with your fixes, after release) then get retry: true enabled

daleharvey:
Closed in favour of https://github.com/pouchdb/pouchdb/pull/3354

PR #3343, opened 2015-01-05
(#3155) - spellcheck/document retry

nolanlawson:
Added documentation for `retry` itself, and also fixed some comma splices, etc.

One thing that's still unclear to me is this line:

> * `paused` - This event fires when the replication is paused, either because a live replication is waiting for changes, or because a live+retry replication is waiting to retry a request due to an error.

It seems to me that it would be clearer to keep `uptodate` for live replication and reserve `paused` for live+retry replication, unless there is some technical reason against it. I find it confusing that `paused` applies to two somewhat different events, and that we need to deprecate `uptodate` to do so.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -503,10 +503,11 @@ Replication is an event emiter like `changes()` and emits the `'complete'`, `'ac
 
 All options default to `false` unless otherwise specified.
 
+* `options.live`: If `true`, starts subscribing to future changes in the `source` database and continue replicating them.
+* `options.retry`: If `true`, replication will be retried when an error occurs (e.g. when the user goes offline). Only used during `live` replication: if you don't also set `{live: true}`, then `retry` will be considered `false`.
 * `options.filter`: Reference a filter function from a design document to selectively get updates.
 * `options.query_params`: Query params sent to the filter function.
 * `options.doc_ids`: Only replicate docs with these ids (array of strings).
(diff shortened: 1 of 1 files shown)


daleharvey:
I was avoiding documenting retry since its currently an experimental option, I want to fix up the tests, and turn its default to true then have it documented, we can do this in the next release.

`uptodate` is redundant if you can do the same thing (and more) with paused, its pretty broken (there are numerous bugs filed against it) and I would prefer to have a smaller API surface that we make very sure works than the loads of not really documented functionality we have now

nolanlawson:
Good points. If we're going to introduce a deprecation, though, then I would suggest releasing 4.0 and making `retry` default to `true`, or maybe getting rid of the option altogether and assuming people will always want it to be true. It would be a lot easier to explain the events system if there were only two modes (live/non-live).

daleharvey:
Well the `4.0` would be at the point it was removed, I definitely want to enable `retry: true` asap which I dont think required a major version as the actual behaviour / api doesnt change, just an improvement on how we fall back on errors, I think its fine to just explain how the events work in the default case (live / non live), we should still keep the `retry` option in the code base in case we want to revert, I would be fine with not documenting it

daleharvey:
Closing as merged in https://github.com/pouchdb/pouchdb/pull/3354

Issue #2632, opened 2014-08-17
Trim down the pouchdb.min.js size


nolanlawson:
Even gzipped, it's gotten a little big. (37.55KB for pouchdb-3.0.1.min gzipped.)

I can't really think of any way to make it smaller, except by removing functionality, though.

daleharvey:
So I have been careful everytime I assume new code has been added, but somehow our build size has increased dramatically, I think this shall be a fat christmas release but I will be looking into getting us back down to a reasonable size for the next release

daleharvey:
>  Yeah I am really shocked to see that we are up to 80KB. Last I checked it was 50. Something big must have been added recently to balloon the size.

Something is forcing buffers to be included, tried https://github.com/pouchdb/pouchdb/commit/9da2b479b303f47aa7fc1eb93a8ac2aab6f56119 but still getting included

nolanlawson:
Ah no, it turned out the 30kb was from `leveldb.js`; that's fixed now. As for buffers, I dunno.

nolanlawson:
Closing. It's as small as it's gonna get for now.

Issue #2597, opened 2014-08-13
Use closure compiler as well when minifying


nolanlawson:
I recall John-David Dalton saying on a podcast that for lodash he actually uglifies, then closure-compiles, then *uglifies again*, and apparently this gives the best compression. I tested it out, and it seems that he's right:

```
36475	gz/pouchdb-ccjs-ug.min.js.gz
37757	gz/pouchdb-ccjs.min.js.gz
37066	gz/pouchdb.min.js.gz
117506	js/pouchdb-ccjs-ug.min.js
117795	js/pouchdb-ccjs.min.js
120341	js/pouchdb.min.js
```

So we could shave half a K off the gzipped minified size with this method.

nolanlawson:
Nobody seems to be a fan of ccjs except for me. Also I was surprised to see that the "advanced optimizations" didn't buy us much, except headaches due to having to rewrite our code to avoid breakage.

nolanlawson:
hence closing

PR #3347, opened 2015-01-05
(#2597 #2632) - ccjs+uglifyify+bundle-collapser for smaller min

nolanlawson:
min+gzip size before: 86003.

After: 80828

Without using `ccjs` (i.e. just using uglifyify and bundle-transform): 82875

I think if `ccjs` is buying us 2 extra KB, then it's worth using.

### DIFF
--- a/bin/min.sh
+++ b/bin/min.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+BROWSERIFY=./node_modules/.bin/browserify
+DEREQUIRE=./node_modules/.bin/derequire
+UGLIFY=./node_modules/.bin/uglifyjs
+CCJS=./node_modules/.bin/ccjs
+
+# use uglifyify and bundle-collapser for best compression
+# for the pre-minified bundle
+
+# then uglify, ccjs, then uglify again for best compression
+# (yes really)
+
+$BROWSERIFY \
+    -p bundle-collapser/plugin \
+    -p uglifyify \
+    -s PouchDB . \
+    | $DEREQUIRE \
+    | $UGLIFY - -mc \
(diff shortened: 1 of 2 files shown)


daleharvey:
So I would still try to avoid this, we can still reduce the build size by a fair amount by reducing complexity, we have a 20KB file that is generating random strings for us, I think we can remove vuvuzela, we have a lot of code that is support for outdated things, I would like us to get to the minimal size possible, but I think the best way to do that would be to reduce complexity where we can, not add to the build steps

nolanlawson:
I'm not so optimistic that we can remove the things you mention and get a big boost in min size:

* I tested removing `vuvuezela` from the build; it only adds 539 bytes gzipped+minified (out of 50KB, so 51730 -> 51191)
* `spark-md5` is huge, but we also need it for deduping of attachments. Maybe instead we should ship a version of pouchdb without attachments support? Lots of folks aren't using them.

Also I feel like this is free KB that we can save without having to change our code at all, which is awesome. If you and @calvinmetcalf are against the `ccjs` dependency, though, I'd be happy to take it out and just use `bundle-collapser` and `uglifyify`; those alone help a lot.

daleharvey:
spark-md5 itself is only 6KB, so not sure why its hitting 20KB, it may be the browserify setup that needs looked at more, vuvuzela was first thing I thought of, sure there are other areas, but yeh would I like to avoid more deps

nolanlawson:
Just using the bundle-collapser alone saves a full 1MB. if you're cool with that, I'll open a new PR with just the bundle-collapser.

nolanlawson:
closing this, opened: https://github.com/pouchdb/pouchdb/pull/3360

PR #3348, opened 2015-01-05
(#2632) - remove leveldb from browserify build

nolanlawson:
044761a504f3cb0493804c6cefdb90813de527fe caused us to start including all of `leveldb.js` in the browserify build, adding about 30KB, up to ~80KB total. Now it's back down to ~50KB minified+gzipped.

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,7 +98,7 @@
     "./deps/buffer": false,
     "request": false,
     "crypto": false,
-    "./adapters/leveldb": false,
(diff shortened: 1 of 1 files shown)


nolanlawson:
"+1 when green" from @daleharvey on IRC, so gonna merge when green and then release 3.2.1 later today.

nolanlawson:
5c8b384e3d731f150bfaa77654ea4868e16f51c8

Issue #3350, opened 2015-01-06
Regression in .compact?


tlvince:
In 3.2.1, given:

```js
var db = new PouchDB('db');
db.compact()
  .then(function(response) {
    console.log(response);
  })
  .catch(function(error) {
    console.error(error);
  });
```

… logs `undefined`. Same if I use callbacks. In 3.2.0, this is returning the `_local/compaction` response as expected.

* [3.2.0 test case](http://bl.ocks.org/tlvince/924b4b59aea604160f35)
* [3.2.1 test case](http://bl.ocks.org/tlvince/832bf0f9140d15df7f5d)

NB, also note I'm seeing a possible migration error if I run the 3.2.0 test case after 3.2.1:

```
// Promise
[Error] TypeError: undefined is not a function (evaluating 'tasks[i - 1](tx, nextMigration)')
	(anonymous function) (pouchdb.js, line 3949)

// Callback
[Error] {"status":500,"name":"Object","message":"the statement callback raised an exception or statement error callback did not return false"}
	(anonymous function) (d97da61674b027c83f161854bbbd97139efcb9a8, line 24)
```

nolanlawson:
Yep, I forgot to mention in the blog post that there is a migration. I'll fix that.

As for `compact`, it's supposed to resolve with `{ok: true}` because that's what CouchDB does. Neither the 3.2.0 behavior nor the 3.2.1 behavior is correct. A pull request would be welcome. :)

marten-de-vries:
So ``compact()`` has always resolved to ``undefined`` until 3.2.0. For 3.2.1 it was changed back (#3111).

Changing it to the CouchDB-like ``{ok: true}`` makes sense, but is technically a backward-incompatible api change (so 4.0)?

tlvince:
Thanks for the heads up.

3.2.0 is resolving something like:

```js
{
  id: '_local/compaction',
  ok: true,
  rev: '0-2'
}
```

… which makes more sense to me than a straight `undefined`, but agreed, mirroring what CouchDB does is best.

nolanlawson:
@tlvince `_local/compaction` is an implementation detail; I think it was a bug that we ever exposed it.

@marten-de-vries AFAIK [the docs](http://pouchdb.com/api.html#compaction) have always been silent about what the promise is supposed to resolve with. So I don't see it as a breaking change; just a bug.

BTW I confirmed that `{ok:true}` is what CouchDB does:

```
[...]

PR #3353, opened 2015-01-06
(#3350) - http compact() also returns {ok:true}

nolanlawson:
@marten-de-vries I noticed that the test wasn't being run for `http`. Also the `http` adapter needed a fix.


marten-de-vries:
Ah, good  catch. :+1: when green.

daleharvey:
Green in https://github.com/pouchdb/pouchdb/pull/3355, merged in https://github.com/pouchdb/pouchdb/commit/beaddadb113484f67c35e96ca6cc42a7165846cb thanks

PR #3354, opened 2015-01-06
(#3155) - Replace syncStopped / started with active / paused

daleharvey:


daleharvey:
@nolanlawson How about this, I took out the explicit references to `retry` while its not enabled yet and tried to explain the events in terms of their semantics, first thing I will be doing this is fixing tests so we can get retry in

daleharvey:
Also sorry, usually try to preserve extra commits but with us both editing the same small piece I just had to squash them

nolanlawson:
No harm, no foul. The docs look good; I think the most important thing will be to add a new section to the guide with code samples.

nolanlawson:
So yeah, +1. We've got a month to figure this out; I think the new `retry` stuff should be the focus for 3.3.0. :)

daleharvey:
woot - https://github.com/pouchdb/pouchdb/commit/fb3a8e0e0621dadcc565bb5f47ca21dfd18fa04d

will work on retry now

PR #3358, opened 2015-01-07
Change events with docs property

gr2m:
adds `event.docs` property to `change` events of replications.

```js
var docs = [{_id: '0'}, {_id: '1'}, {_id: '2'}];
db.bulkDocs({ docs: docs }, {})
.then(function(results) {
  var replication = db.replicate.to(dbs.remote);
  replication.on('change', function(change) {
    console.log('replicated:', change.docs);
  });
})
// replicated: [{_id: '0'}]
// replicated: [{_id: '1'}]
// replicated: [{_id: '2'}]
```

Once it's good to go, I'll squash and rebase the commits.

### DIFF
--- a/lib/replicate.js
+++ b/lib/replicate.js
@@ -156,9 +156,11 @@ function replicate(repId, src, target, opts, returnValue, result) {
   };
   var checkpointer = new Checkpointer(src, target, repId, state);
   var outResult;
+  var allErrors = [];
 
   if (result) {
     result.docs = [];
+    result.errors = [];
     outResult = utils.clone(result);
   } else {
     outResult =  {
@@ -194,12 +196,14 @@ function replicate(repId, src, target, opts, returnValue, result) {
           errorsById[res.id] = res;
         }
       });
-      outResult.errors = outResult.errors.concat(errors);
+      outResult.errors = errors;
+      allErrors = allErrors.concat(errors);
       outResult.docs_written += currentBatch.docs.length - errors.length;
       var non403s = errors.filter(function (error) {
         return error.name !== 'unauthorized' && error.name !== 'forbidden';
       });
 
+      outResult.docs = [];
       docs.forEach(function(doc) {
         var error = errorsById[doc._id];
         if (error) {
@@ -393,6 +397,7 @@ function replicate(repId, src, target, opts, returnValue, result) {
     outResult.ok = false;
     outResult.status = 'aborting';
     outResult.errors.push(err);
+    allErrors = allErrors.concat(err);
     batches = [];
     pendingBatch = {
       seq: 0,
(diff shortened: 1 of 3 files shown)


daleharvey:
+1 to this, I think its awesome, I will give it today to let others get a look at this before merging

daleharvey:
Actually one little nit, can you keep a `var docs = [];` and `docs.concat` instead of using result.docs, givent we are passing the full documents through this event I would definitely prefer to avoid utils.cloning each document needlessly (with a large batch size that could be a large number of docs)

gr2m:
I'm sorry but I don't understand where the `result.docs` or the respective docs are cloned in the current code. Can you point me to the respective lines? I must miss something

daleharvey:
https://github.com/gr2m/pouchdb/blob/change_events_with_doc/lib/replicate.js#L302 where the event is output, the entire object (including the docs) will get cloned which can be expensive with large doc bodies (theres another thread talking about raising the 1MB limit for a bulkdocs request)

```
var outResult = utils.clone(result);
outResult.docs = changeDocs;
changeDocs = [];
returnValue.emit('change', outResult);
```

would avoid that

daleharvey:
You are right, we should also reset the errors, its less likely to blow up but makes more sense semantically if we are given the error one time

gr2m:
Okay, the errors are now also per batch in the `change` event. But I had to keep track of all errors, so it can be passed to `complete`. I've also removed the `.docs` property before triggering the `complete` event, as it would be the last batch only. 

It could be confusing to some that `"change"` events has a `.docs` property, while `"complete"` event does not. We could keep track of all docs like we do with the errors, but I'm not sure if that could run into memory trouble again. Your call

I _think_ besides that it's good to go now.

daleharvey:
Yeh the `change` events dont need to match the complete one, in live replication having objects that continually grow is a bad idea since it should run in constant space, we can live with `errors` being collected for now but it may be an idea to only do that for non live replication

daleharvey:
And apologies, but I think I confused you with the `outResult` thing

So when we `emit('change', utils.clone(result));`, we do that so if the user modifies the result object, we dont see any of those modifications internally, https://github.com/gr2m/pouchdb/commit/d93db495ed874a2d3d114304cc2b1f6f02e080ea breaks that since we emit `outResult` directly (we should probably write a test for it)

We should keep `result = result || {` as it was working before, but inside finishBatch do

```
      result.last_seq = last_seq = currentBatch.seq;
      var outResult = utils.clone(result);
      outResult.docs = changedDocs;
      returnValue.emit('change', outResult);
      currentBatch = undefined;
```

Sorry for the confusion

gr2m:
Oh man, I was confused indeed. Thanks for all your patience with me here. I think I got it now, see 94d887c

daleharvey:
I cancelled the test run on this, sorry the tests got slightly screwed up and should be fixed shortly, but looking over this I think its good to merge, sorry for getting you mixed up, and cheers for persisting, a huge amount of people have been asking for this

Will squash and rebase on top of the test fixes to get a proper test run

daleharvey:
Doing a full test run @ https://github.com/pouchdb/pouchdb/pull/3375

daleharvey:
Yay, all green, merged in https://github.com/pouchdb/pouchdb/commit/929c1bd8bec2704ad32b1b6a9465d563c5dbccbe

Thanks :)

nickcolley:
Great addition thanks @gr2m :) :+1:

PR #3359, opened 2015-01-08
(#2632) - remove browserified Buffer from build

nolanlawson:
So apparently Browserify does this thing where,
if you do `"./mydep": "false"`, you need to reference
all possible relative paths used to that dependency,
whereas if you just swap it out for an empty
implementation then it works regardless of the relative
path used.

This solution is dumb, but it cuts down the min+gz size
from 51185 to 46298.

### DIFF
--- a/lib/deps/buffer-browser.js
+++ b/lib/deps/buffer-browser.js
@@ -1,0 +1,2 @@
+// hey guess what, we don't need this in the browser
+module.exports = {};
--- a/lib/adapters/http/http.js
+++ b/lib/adapters/http/http.js
@@ -13,6 +13,7 @@ var utils = require('../../utils');
 var errors = require('../../deps/errors');
 var log = require('debug')('pouchdb:http');
 var isBrowser = typeof process === 'undefined' || process.browser;
+var buffer = require('../../deps/buffer');
 
 function encodeDocId(id) {
   if (/^_(design|local)/.test(id)) {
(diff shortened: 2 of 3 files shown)


lib/deps/buffer-browser.js
```diff
@@ -0,0 +1,2 @@
+// hey guess what, we don't need this in the browser
+module.exports = {};
```
marten-de-vries: This line is probably not necessary? (module.exports is already an object by default, I mean.)

marten-de-vries: ``module.exports`` is already an empty object by default, making this line unnecessary, right?

marten-de-vries:
That's a very nice bundle size improvement. :+1: when green (and the nit above has been taken into account).

daleharvey:
wow nice find, but ugh, @substack / @calvinmetcalf is there a nicer way to do this, I expect this will get regressed / confused in future as is

daleharvey:
Actually I was being stupid, I thought you were referencing a non existent file as a hack, having an empt `module.exports = {};` I actually prefer, its explicit over depending on default behaviour

daleharvey:
Checked with @marten-de-vries over IRC and was good with landing as is, thanks: https://github.com/pouchdb/pouchdb/commit/7ceebc8f4790ed48ab238007204ba62e975a6668

PR #3361, opened 2015-01-08
(#3361) - remove PouchDB.extend

nolanlawson:
`PouchDB.extend` was never an official API, and `extend`
is already exposed in `utils`. Looks like this was
only ever used in the integration tests.


marten-de-vries:
Nice. :+1: when green.

daleharvey:
+1, yay for cleaner api surface - https://github.com/pouchdb/pouchdb/commit/c4c39aa6866278d8e005abda677ddb77b81272ca

Issue #966, opened 2013-10-24
Persistent Replications


jo:
In CouchDB (as well as in TouchDB) there is the [`_replicator` database](https://gist.github.com/fdmanana/832610). When initiating replications through the `/_replicator` endpint (as opposed to the `/_replicate` endpoint) the replications overdue a restart.

This feature is extremely important to PouchDB, as its replications cancel on each page reload.

I have implemented such persistent replications in my [Todos Sync Example](http://jo.github.io/backbone-pouch/examples/todos-sync/), where replications gets stored and resumed on reload.

I would suggest supporting a `persistent: true` option. When this option is set, PouchDB would store replication documents inside a `_replicator` database and resume on startup.

daleharvey:
Ok, I have been testing this locally, it is working good but I think it needs a little time to settle and some more testing before being enabled by default, there seems to be a few bugs (I get a lot of event listener warnings), and active / paused events seem off

I am proposing documenting this as an option, warn that its default is value is possible to change and get some thorough testing of it

daleharvey:
Ok, its been in a while now, but finally documented - https://github.com/pouchdb/pouchdb/commit/ecc03209bea7e603ec8cf2d241be207ffa2c0cb8

tdermendzhieva:
I am working on ember project in which we are using ember- pouch, I have serious problem with it that’s why I decided to contact you. The problem occurs when refresh the browser and continue to use the app, my logic is to retrieve some data  with AJAX requests and save them in pouchdb.
If the db has data with some of the ids returned I am trying to  update it and if not, I am creating new record in DB.
But when I refresh the browser and check if there are some data that I should update it returns nothing , so when I try to save new records  “status 409 document conflict problem occurs”. It seems when browser refreshes there is no data in db (at the moment I check for records to update) and when I am trying to create new one they are in db and my logic breaks. How could I fix this? Any ideas?

nolanlawson:
Hi @tdermendzhieva - please read [the conflicts guide](http://pouchdb.com/guides/conflicts.html) - basically, you need to either be doing `upsert()` or handling 409s manually yourself.

tdermendzhieva:
Thanks @nolanlawson  I handled the error when browser refreshes  and it works

PR #3362, opened 2015-01-08
(#966) - Fix tests that depend on retry: false

daleharvey:


daleharvey:
This doesnt set retry: true by default yet, but fixes all the tests so they pass when it is, after this there is one test acting strangely when its retry: true but it shouldnt be effected so looking into it seperately

marten-de-vries:
Looks fine. +1.

marten-de-vries:
Thanks! Merged: dce9526035421bc31ec87f5e5c4ce6e541883917

PR #3366, opened 2015-01-09
(#136) - Remove test of deprecated local_seq parameter

glynnbird:
(#136) - Remove test of deprecated local_seq parameter

Removing a test that was using a ?local_seq=true parameter to ensure that local sequence numbers are sequential. It turns out, post CouchDB 2.0 that local_seq numbers are no longer sequential and become meaningless in a sharded environment.
 
CouchDB 2.0 is set to remove this parameter anyway (https://issues.apache.org/jira/browse/COUCHDB-2537) , so removing this test shouldn't do any harm.

### DIFF
--- a/tests/integration/test.get.js
+++ b/tests/integration/test.get.js
@@ -102,21 +102,6 @@ adapters.forEach(function (adapter) {
       });
     });
 
-    it('Get local_seq of document', function (done) {
-      var db = new PouchDB(dbs.name);
-      db.post({ test: 'somestuff' }, function (err, info1) {
-        db.get(info1.id, { local_seq: true }, function (err, res) {
-          res._local_seq.should.equal(1);
-          db.post({ test: 'someotherstuff' }, function (err, info2) {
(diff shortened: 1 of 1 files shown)


daleharvey:
We are aiming for compatibility with CouchDB (including 2.0), so a first step is to skip tests that are failing against master, but ideally `isCouchDBMaster()` will not exist, once its green I would like to go back and check all the tests that use it, rework them when when needed and remove them if they no longer apply

If there is no `local_seq` at all in couchdb master, I think this is one relatively unused property we can remove the test for and deprecate or plain remove

daleharvey:
we include it in the documentation, so we would need to deprecate it and remove it in the next release, the next release is going to have a few deprecations and the one after will be a major version bump, so it would be good to get these in early

daleharvey:
Sweet thanks, we need to remember to mark this as deprecated before the release - https://github.com/pouchdb/pouchdb/commit/605d496dca5ac7f08b1f40940cd565f08cb465ee

PR #3368, opened 2015-01-09
(#136) - Fix "Test test of invalid open_revs parameter"

glynnbird:
(#136) - Fix "Test test of invalid open_revs parameter"

When giving an invalid parameter to ?open_revs, CouchDB 1.6 and 
CouchDB 2.0 provide different error messages. The former says 
"unknown_error" the latter "function_clause". I've modified the test to 
handle both scenarios.

### DIFF
--- a/tests/integration/test.get.js
+++ b/tests/integration/test.get.js
@@ -724,7 +724,9 @@ adapters.forEach(function (adapter) {
             'not an array': 'or all string'
           }
         }, function (err, res) {
-          err.name.should.equal('unknown_error', 'correct error');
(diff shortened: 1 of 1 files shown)


willholley:
ah - looks like this was just fixed in CouchDB master...

daleharvey:
lol, I just merged this, but its backwards compat so either way we can fix it up later so we dont need to worry with the exact couch version we are testing against

https://github.com/pouchdb/pouchdb/commit/8bb5b50c26e98d8cf2a73dd41e600c337cd9a5b4

Issue #3369, opened 2015-01-09
Regression in Chrome 39-41: local replication of attachments


nolanlawson:
Somehow we're failing a local replication of attachments test. Probably related to https://github.com/pouchdb/pouchdb/commit/fce1c97e6bfaa483e8af21b5244b437f040f43fb, so unfortunately it probably slipped into 3.2.1.

Gist: https://gist.github.com/jchris/0a647d8d68321c078b66

jchris:
I was in Chrome 39

nolanlawson:
There doesn't seem to be a workaround for this. The bug is that the `FileReader` `readAsBinaryString`, `readAsArrayBuffer`, and `readAsDataURL` methods all return `e.target.result === null` sometimes. I could try to detect that the `blob.size !== resultString.length`, but then I get stuck in an endless loop because it seems impossible to recover, even if I create new blobs from the source one or use `setTimeout`.

Additionally, there doesn't seem to be a way to programmatically detect this bug, because it's a Heisenbug. The test passes when the dev tools are open, and fails when the dev tools are closed. I'm testing in Chrome 39 on a Mac.

nolanlawson:
[Tried to write a small test case](http://bl.ocks.org/nolanlawson/raw/800589439770c027aa8d/), not hitting the bug yet though. Hopefully I won't have to just send the Chrome team the entire PouchDB suite.

nolanlawson:
Bug filed on Chromium: https://code.google.com/p/chromium/issues/detail?id=447836

For us, I think the workaround will be reverting fce1c97, which in a roundabout way will simply guarantee that blobSupport is always set to false in Chrome/Android. I'll see if I can find a better way to feature-test this bug.

nolanlawson:
Also filed an issue on Chromium for the 404 bug. Although we might end up relying on it to detect Chrome... >_< https://code.google.com/p/chromium/issues/detail?id=447916

natcohen:
When adding multiple attachments, some attachments are empty (the size is equal to 0) in couchdb. This bug is only in 3.2.1 and not in 3.2.0... 

Is this behavior related to this issue?

nolanlawson:
If you replicated from Chrome, then yes.

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/76912586ae54481ecc38fd5df8875c5e3aabfac8

PR #3370, opened 2015-01-10
(#3369) - DON'T MERGE - see if travis can find failure

nolanlawson:


nolanlawson:
Nice, Chrome 38 and 39 fail with exactly this failed test. So we would have caught it if we had actually been testing 38 or 39. :)

I suspect that, without the version number, Sauce is actually giving us 36 or 37 or something.

PR #3373, opened 2015-01-10
Upgrade selenium standalone

daleharvey:


marten-de-vries:
Perfect, that seems to make https://github.com/pouchdb/express-pouchdb/pull/181 & https://github.com/pouchdb/pouchdb-server/pull/92 unnecessary. :+1: when green & the PR number is added to the commit message.

marten-de-vries:
Travis is green, and the allowed failures it gives are because the tests fail, not because of selenium problems. Thanks for this! Merged: e2675ec0ae3ce494905721567a6746d115f47719

Issue #3367, opened 2015-01-09
Deprecate local_seq


daleharvey:
PR to remove its failing test in https://github.com/pouchdb/pouchdb/pull/3366, its a documented option so we should put a deprecation warning on the docs, include it in the release notes, put a console warning when its used then we can remove it in the next major version bump

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/da13c71ab8e965e3c0efe3c6f71b975fb05cba93

PR #3377, opened 2015-01-11
(#3367) - Give notice of local_se deprecation

daleharvey:


lib/utils.js
```diff
@@ -671,6 +671,12 @@ exports.explain404 = function (str) {
   }
 };
 
+exports.info = function (str) {
+  if (process.browser && 'console' in global && 'info' in console) {
```
nolanlawson: this is probably not safe for node-webkit. you could do `typeof console !=== 'undefined'` instead of `'console' in global`

nolanlawson:
+1 after nit

daleharvey:
heh wow, green in ie + iphone without restarts

daleharvey:
Fixed nit and green, thanks

https://github.com/pouchdb/pouchdb/commit/da13c71ab8e965e3c0efe3c6f71b975fb05cba93

PR #3380, opened 2015-01-12
(#3369) - disable blobSupport in Chrome, fix Travis

nolanlawson:
This essentially reverts fce1c97, meaning we're back to
using the 404 test to disable blob support in Chrome/Android/Opera.

This is somewhat flakey because we're using one bug to detect
another, but I'm really hoping to avoid UA sniffing here.

I also moved `idb-blob-support.js` to its own file and changed
the Travis tests so that we test Chrome 39 explicitly. I confirmed
that 38 and 39 are currently failing in master, so I don't know
which version Sauce is giving us when we don't specify the version. It
seems pointless to test it without the version, if we aren't actually
getting the latest stable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,9 +53,9 @@ env:
   - AUTO_COMPACTION=true CLIENT=selenium:phantomjs ES5_SHIM=true COMMAND=test
 
   # Testing in saucelabs
-  - CLIENT=saucelabs:chrome COMMAND=test
   - CLIENT=saucelabs:chrome:36 COMMAND=test
   - CLIENT=saucelabs:chrome:37 COMMAND=test
+  - CLIENT=saucelabs:chrome:39 COMMAND=test
   - CLIENT=saucelabs:safari:6 COMMAND=test
   - CLIENT="saucelabs:iphone:7.1:OS X 10.9" COMMAND=test
   - CLIENT="saucelabs:internet explorer:10:Windows 8" COMMAND=test
--- a/lib/adapters/idb/idb-blob-support.js
+++ b/lib/adapters/idb/idb-blob-support.js
@@ -1,0 +1,63 @@
+'use strict';
+
+var utils = require('../../utils');
+var idbConstants = require('./idb-constants');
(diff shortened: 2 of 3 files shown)


nolanlawson:
As a side note, this is a supremely unsatisfying commit for me. I would like to use blobs in Chrome 38+ because of storage efficiency. But it seems blobs are still shaky in Chrome, and I couldn't find a workaround (even `setTimeout` didn't work - once the faulty blob is stored, it's always empty when you try to read it with `FileReader`).

daleharvey:
So I am not sure we are getting anything out of trying to feature detect out of bugs here, we use feature detection so we automatically work when the bug gets fixed however this means it will automatically break when the bug gets fixed.

If this is going to break when the 404 bug gets fixed, should we just do a UA test?

nolanlawson:
They are two separate bugs, but I am hoping that the chromium devs will fix them both in the same release. If they don't, then yes, we will have to use UA testing, which will get nasty due to Android, Opera, and Chrome all sharing this bug in various versions.

Arguably, by feature testing this 404 bug, we are preventing another bug, which is a 404 for empty attachments... But I should probably add a PouchDB test for that.

daleharvey:
So I am not super comfortable with this, but will defer to your better judgement since you have been the one dealing with all the crazy chrome + blob bugs.

If we get to a point where we cant actually detect whether a feature works properly programmatically, then I think we are ok to do UA sniffing, but yeh between opera and android sharing this UA sniffing is messy

I am going to try helping out on the android tests in travis now, once we get android passing in travis we can be specific about what versions we support and maybe kill off some of this hopefully

daleharvey:
https://github.com/pouchdb/pouchdb/commit/76912586ae54481ecc38fd5df8875c5e3aabfac8

nolanlawson:
@daleharvey You should be able to run the Android tests using that branch I wrote. Locally everything was peachy; I just wasn't able to start up a headless emulator in Travis.

nolanlawson:
As for the better judgment thing, this is really just a stopgap. This commit isn't great, but I think it's preferable to UA sniffing.

PR #3381, opened 2015-01-12
(#3381) - remove useless md5 code

nolanlawson:
Two things:

* `size` isn't correct; it's `length` or `byteLength`
* both `slice` and `substring` are clamped anyway, so
  this is useless


daleharvey:
Sweet, merged - https://github.com/pouchdb/pouchdb/commit/6d11ad2756af41f71718daa4137c5b3dadb78f01

PR #3383, opened 2015-01-12
(#2955) - Passing test!

### LINKED ISSUE #2955: PouchDB doesn't replicate deleted conflicted documents.
Hello, I think I might have found a bug. It is my understanding that CouchDB bidirectional replication keeps two databases perfectly in sync which would include all deleted revisions and documents. PouchDB, on the other hand, fails to replicate any deleted conflicting revisions. This makes conflict resolution hard when syncing with a remote CouchDB.

Here is a JSFiddle that shows it failing:
* v3.0.6 - http://jsfiddle.net/appleifreak/h5arfeb3/1/
* v3.0.7-prerelease - http://jsfiddle.net/appleifreak/h5arfeb3/2/

This doesn't work with IndexedDB or CouchDB databases, nor does it work while using continuous replication. I might be seriously confused here, so please correct me if I am wrong.

I have a side question about Pouch, slightly unrelated to this issue. Pouch seems to choose the local copy as the winner over the remote copy during a conflict. Is there a way to reverse this while syncing so that the remote is always the winner?

tyler-johnson:
I'm not sure when this was fixed, but the test for #2955 is now passing. I am submitting this pull request to get the test integrated so it's run in the future. I think this means we can close #2955 now, too.

Thanks for all your hard work guys. PouchDB is starting to feel really *stable*!

### DIFF
--- a/tests/integration/test.bulk_docs.js
+++ b/tests/integration/test.bulk_docs.js
@@ -782,5 +782,62 @@ adapters.forEach(function (adapter) {
       });
     });
 
+    it('Bulk docs two different revisions to same document id', function(done) {
+      var db = new PouchDB(dbs.name);
+      var docid = "mydoc";
+
+      function uuid() {
+          return PouchDB.utils.uuid(32, 16).toLowerCase();
+      }
+
+      // create a few of rando, good revisions
+      var numRevs = 3;
+      var uuids = [];
+      for (var i = 0; i < numRevs - 1; i++) {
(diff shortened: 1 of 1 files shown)


daleharvey:
Awesome, glad it was fixed as there was a few fixes around that area, more tests for this are great

Merged in https://github.com/pouchdb/pouchdb/commit/ca1de673832aa35addbea93bee75412970f87113

Thanks

PR #3386, opened 2015-01-14
(#136) - Fix "doc count after multiple replications"

willholley:
Remove the test dependency on temporary views (not supported in CouchDB 2.0 / Cloudant).

### DIFF
--- a/tests/integration/test.replication.js
+++ b/tests/integration/test.replication.js
@@ -2747,13 +2747,6 @@ adapters.forEach(function (adapters) {
         });
       }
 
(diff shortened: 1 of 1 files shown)


daleharvey:
Sweet, we are going to have to make a decision about temp queries once these tests are done with, which is soon - https://github.com/pouchdb/pouchdb/commit/c25cee00273305be88408203207f1ed4d79355d3

nolanlawson:
I'm working on the indexes plugin. Would be happy to deprecate the map/reduce plugin entirely after that. :)

PR #3388, opened 2015-01-14
(#136) - Fix "retry stuff"

willholley:
This test seems to occasionally use more than 2 retries. I think the behaviour is correct - just non-deterministic - so changed the assertion around the number of pauses to prevent transient failures.


daleharvey:
Makes sense, thanks - https://github.com/pouchdb/pouchdb/commit/1c90836fc90d32a789523258ec7e200a710e8e75

PR #3390, opened 2015-01-14
(#136) - Fix "Syncing should stop if one replication fails (issue 838)"

willholley:
Ensure this test copes with non-integer sequence numbers. We can still test for insertions by looking at the first element of the string or array (should equal 1).


daleharvey:
I think this would be better if we just removed the assertion, it doesnt look to be adding much

willholley:
sure - done :)

daleharvey:
Awesome thanks - https://github.com/pouchdb/pouchdb/commit/141dcc6d439bac5fb70d58d214f58649aff1532f

PR #3392, opened 2015-01-14
(#136) - Skip taskqueue "Query" test

willholley:
CouchDB 2.0 doesn't support temporary views so skip this test against it.


daleharvey:
Yay thanks - https://github.com/pouchdb/pouchdb/commit/2e03f8c6789c93739cb1f0b368c4fd8ca0e57a40

PR #3393, opened 2015-01-14
(#136) - Fix taskqueue / Info test against CouchDB 2.0

willholley:
Initial update_seq is deterministic for a new CouchDB 2.0 database. Skip this assertion when running this test against it.


daleharvey:
I would just remove the assertion, the doc_count assertion is fine for this test

daleharvey:
I just ammended the commit to do that, thanks - https://github.com/pouchdb/pouchdb/commit/93d533d8c560375469ee4e88b47acf421beae7ca

PR #3394, opened 2015-01-15
(#136) - Skip Test get with revs_info on compacted tree

willholley:
The _compact endpoint is not exposed via the standard port in CouchDB 2.0. Skip this test until we figure out how to handle that.


daleharvey:
Looks good cheers - https://github.com/pouchdb/pouchdb/commit/ca1c61c4e5a6db4f510c2a5fb657c2fa3f261cc6

Issue #3395, opened 2015-01-15
Match CouchDB handling of GET somedoc?open_revs=true when somedoc doesn't exist


willholley:
In CouchDB 1.x a call to open_revs=true on a non-existent document results in a crash - the response is a successful (200) response with junk in the body. In CouchDB 2.x this is fixed and the call correctly returns a 404 (see https://issues.apache.org/jira/browse/COUCHDB-2517) as you might expect.

PouchDB seems to match the broken behaviour of CouchDB 1.x and returns a 200 response but should probably return a 404 in this situation as well.

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/45908f4fe888797ae600e4f591da770cab22be38

PR #3396, opened 2015-01-15
(#3395) - Return 404 when open_revs is requested for missing doc

willholley:
CouchDB 1.X crashes and returns a 200 response when a request for open_revs on a non-existent document is made. In CouchDB 2.0 this is
fixed (COUCHDB-2517) and it now returns a 404 which seems like the correct response.

We were previously matching the CouchDB 1.X behaviour and hiding this error condition. This commit allows the error to propagate up, matching the new CouchDB behaviour.

The fix in the tests is a little ugly as we need to skip the assertion when testing against CouchDB 1.X.

### DIFF
--- a/lib/adapter.js
+++ b/lib/adapter.js
@@ -510,9 +510,7 @@ AbstractPouchDB.prototype.get =
     if (opts.open_revs === "all") {
       this._getRevisionTree(id, function (err, rev_tree) {
         if (err) {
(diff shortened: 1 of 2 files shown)


daleharvey:
Awesome, thanks - https://github.com/pouchdb/pouchdb/commit/45908f4fe888797ae600e4f591da770cab22be38

PR #3400, opened 2015-01-16
(#2676) Ignore _removed field

gabrielnau:
This is needed to support Couchbase Sync Gateway, which adds this field when a document previously accessible by a user in a channel isn’t accessible anymore.

A complete implementation of this feature would effectively delete the document locally and emit proper ```delete``` event.

I didn't add any test for this, since we would need in my opinion to run an integration test against a Couchbase Sync Gateway instance.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -43,7 +43,9 @@ var reservedWords = toObject([
   '_replication_state',
   '_replication_state_time',
   '_replication_state_reason',
-  '_replication_stats'
+  '_replication_stats',
+  // Specific to Couchbase Sync Gateway
+  '_removed'
 ]);
 
 // List of reserved words that should end up the document


gabrielnau:
This is based on this https://github.com/racido/pouchdb/commit/30d62fd6c3de4b6794162ff3745b4c2c8ed0ac26#commitcomment-9290056 commit, but I didn't include other changes, which would make sense if the document is actually removed from the local Db I think. For now we just want to ignore this.

willholley:
What's the situation where this should be ignored (can you add a test to demonstrate)? I guessing a bit but I would have expected that we'd see "removed" events in the _changes feed in which case PouchDB should purge the document (as opposed to deleting) and we wouldn't care about this field in a document body.

daleharvey:
So I dont mind small patches to facilitate Couchbase functionality, however it doesnt seem like this by itself would help all that much?

I was under the assumption that couchbase sync gateway was designed to allow couchdb compatible clients to be able to deal with replication into couchbase, if thats so then it seems like we shouldnt need support for `_removed` inside pouchdb itself, we should be able to setup a travis build that tests couchbase sync gateway, see the failures (like multipart/mixed) and fix them individually.

For this commit I dont mind it going in as is, but would like to understand what the plan is for supporting couchbase sync going forward

nolanlawson:
Agree, the sooner we set up Couchbase Community Edition + CSG on Travis, the sooner we'll know exactly what we need to fix. /cc @jchris

jchris:
I've been working on it locally and have configuration for SG that works as
admin party. Currently passes the first seven or eight replication tests.

Removing ajax-browser may be one way to resolve multipart, that's what I'm
working on now.

I guess it's not too soon to add SG to Travis. I'll send a pull req with my
config server for admin party

On Sunday, January 18, 2015, Nolan Lawson <notifications@github.com> wrote:

> Agree, the sooner we set up Couchbase Community Edition + CSG on Travis,
> the sooner we'll know exactly what we need to fix. /cc @jchris
> <https://github.com/jchris>
>
> —
> <https://github.com/pouchdb/pouchdb/pull/3400#issuecomment-70413060>.
>


-- 
Chris Anderson
http://www.couchbase.com
http://twitter.com/jchris

nolanlawson:
:tada: :balloon: :confetti_ball:

jchris:
SG doesn't do open_revs=all so I had to comment out some assertions.

On Sunday, January 18, 2015, Nolan Lawson <notifications@github.com> wrote:

> [image: :tada:] [image: :balloon:] [image: :confetti_ball:]
>
> —
> <https://github.com/pouchdb/pouchdb/pull/3400#issuecomment-70422555>.
>


-- 
Chris Anderson
http://www.couchbase.com
http://twitter.com/jchris

nolanlawson:
We don't use `open_revs=all` in the replicator, so that's no prob.

For the tests, though, instead of commenting, let's add a function like `isSyncGateway()` (see [isCouchMaster()](https://github.com/pouchdb/pouchdb/blob/207cb4e7b9420c9b094eadaa8bc75dd5ca8e39dd/tests/integration/utils.js#L15-L18) for an example).

gabrielnau:
@willholley once we have a CSG on Travis, I will add a spec describing the scenario.
The goal of this pull request is only to "not break" Pouch when used as a CGS client, which adds a field not allowed by Pouch. The implementation of this CGS "document removed from channel" feature is missing but I think the sooner anyone trying Pouch with CGS will have no error thrown, the better. That's the reasoning behind this PR which just allows Pouch not throw.

daleharvey:
So since this is a minor patch that just make sure everything doesnt blow up then I am fine with it being in core, merged in https://github.com/pouchdb/pouchdb/commit/f2d84ec86941cb9b49175f12493be6848f1def5f

As for future support, It will be situational depending on the feature but in general in Pouch we are only likely to support the common subset between couchbase, couch master, couch 1.6, cloudant type functionality. So for multipart/mixed that is some common functionality that we would like to see in core, however having support for some extra 'channel removed' type document property that only exists in couchbase, thats likely something that will have to be supported by a plugin (or couchbase lite)

If couchbase sync gateway is supposed to act as gateway for couchdb compatible clients to replicate to couchbase, then we should not need any of these modifications (couchdb will reject these documents as well) and we should still be able to fully support the gateway. 

If couchbase sync gateway is an different implementation with seperate features and an incompatible protocol, then couchbase lite most definitely should be a seperate project

Merged in https://github.com/pouchdb/pouchdb/commit/f2d84ec86941cb9b49175f12493be6848f1def5, Cheers

nolanlawson:
Users can also replicate from PouchDB to CouchDB, and then from CouchDB to CSG/Cloudant. CouchDB (1.x) seems to be the lingua franca between all these different libraries, so at the very least they should all be compatible with CouchDB.

daleharvey:
Yeh that is part of my point, if CouchDB can replicate to CSG then we can / should be able to as well and will be happy to merge any improvements that make that easier / possible.

However if CouchDB cant replicate to CSG, and only Couchbase lite specific clients can, which is indicated by `_removed` which is not supported in CouchDB then we probably wont want to merge much functionality into core that supports an incompatible protocol, couchbase lite is going to happen to support that anyway, and getting everything working together to support one common protocol is hard enough, supporting more makes it drastically harder

jchris:
CSG syncs using the Couch protocol (with a couple of extensions like removed) and we work to push any relevant stuff to CouchDB also (like the `_bulk_get` optimization).

It would make sense that any behavior based on the the removed flag would be part of a plugin, when we eventually get around to writing it.

nolanlawson:
If `_removed` is broadly equivalent to `_deleted`, then we can just add a `utils.isDeleted` helper function to check for either one, yes?

jchris:
Removed is deleted for you and not me. If we treat it as a regular delete it could replicate peer to peer. It's more like a local purge, but maybe not exactly. Probably interesting enough to be worth a plug, when we finally implement it.

PR #3401, opened 2015-01-16
Fix a comment

gabrielnau:
The ```conflicts``` option is actually supported by Couchbase.

### DIFF
--- a/lib/adapters/http/http.js
+++ b/lib/adapters/http/http.js
@@ -623,9 +623,9 @@ function HttpPouch(opts, callback) {
     var body;
     var method = 'GET';
 
-    // TODO I don't see conflicts as a valid parameter for a
-    // _all_docs request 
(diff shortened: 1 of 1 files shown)


nolanlawson:
That doesn't seem to be true. Looks supported in CouchDB too: http://docs.couchdb.org/en/latest/api/database/bulk-api.html

nolanlawson:
The conflict is definitely wrong, though. Just not sure what to replace it with :stuck_out_tongue:

daleharvey:
can just remove it?

daleharvey:
Ok, I just amended the commit to remove it since its obviously wrong / not clearing anything up, thanks @gabrielnau  - https://github.com/pouchdb/pouchdb/commit/89584a7d9b86b5735e4ba363b55d51bd9f552408

PR #3402, opened 2015-01-17
Clarify code & info about Persistent Queries

BigBlueHat:
Just a few things that tripped me up that today. Hopefully this'll help others. :smiley: :

### DIFF
--- a/docs/_guides/queries.md
+++ b/docs/_guides/queries.md
@@ -51,26 +51,29 @@ First, you create a **design document**, which describes the `map` function you 
 ```js
 // document that tells PouchDB/CouchDB
 // to build up an index on doc.name
-var myIndex = {
+var ddoc = {
   _id: '_design/my_index',
   views: {
-    'my_index': {
+    by_name: {
       map: function (doc) { emit(doc.name); }.toString()
     }
   }
 };
 // save it
-pouch.put(myIndex).then(function () {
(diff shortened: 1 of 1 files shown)


nolanlawson:
I think that for veteran users of CouchDB, things like `mydesigndocname/myviewname` are obvious, but for new users I think it's just too overwhelming. That's why I made it so that `foo/foo` is equivalent to `foo`.

Can you think of a way that would have made it easier for you to understand without introducing the `/`?

daleharvey:
I actually think the opposite, the fact that `foo` refers to the view `foo` inside the design doc `_design/foo` is somewhat a big of magic, whereas `ddocname/viewname` is consistent and more explicit. I mean neither are a particular nice api, but I think it will make things clearer to document the plain old dumb api as opposed to the slightly magic one

daleharvey:
Also ignore the errors, if you rebase they will be fixed

nolanlawson:
Still disagree; newbies don't know what design docs are, don't care, and just want the shortest path to making an "index."

But yeah, that'll be better fixed by the indexes plugin ([making progress, btw](https://github.com/nolanlawson/pouchdb-indexes)). So sure, +1

nolanlawson:
207cb4e7b9420c9b094eadaa8bc75dd5ca8e39dd

PR #3404, opened 2015-01-17
(pouchdb/express-pouchdb#194) - fix callback called twice

nolanlawson:
The `leveldb.js` fixes are actually unrelated; I was just doing some code cleanup over there.

### DIFF
--- a/lib/adapter.js
+++ b/lib/adapter.js
@@ -425,44 +425,34 @@ AbstractPouchDB.prototype.compact =
   });
 });
 AbstractPouchDB.prototype._compact = function (opts, callback) {
-  var done = false;
-  var started = 0;
-  var copts = {
-    returnDocs: false
-  };
   var self = this;
-  var lastSeq;
-  function finish() {
-    upsert(self, '_local/compaction', function (doc) {
-      if (!doc.last_seq || doc.last_seq < lastSeq) {
-        doc.last_seq = lastSeq;
-        return doc;
-      }
-      return false; // somebody else got here first, don't update
-    }, function () {
-      callback(null, {ok: true});
-    });
-  }
-  if (opts.last_seq) {
-    copts.since = opts.last_seq;
-  }
-  function afterCompact() {
-    started--;
-    if (!started && done) {
-      finish();
-    }
-  }
+  var changesOpts = {
+    returnDocs: false,
+    last_seq: opts.last_seq || 0
+  };
+  var promises = [];
+
   function onChange(row) {
-    started++;
-    self.compactDocument(row.id, 0).then(afterCompact, callback);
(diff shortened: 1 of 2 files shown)


lib/adapter.js
```diff
@@ -425,44 +425,36 @@ AbstractPouchDB.prototype.compact =
...
-  }
-  function afterCompact() {
-    started--;
-    if (!started && done) {
-      finish();
-    }
-  }
-  function onChange(row) {
-    started++;
-    self.compactDocument(row.id, 0).then(afterCompact, callback);
```
nolanlawson: ^ this was the source of the bug. We were calling `callback` for every `'change'` event, plus during the `'complete'` event.

This code was essentially just trying to do the compactions in order and then sequentially run `finish` at the end. So why not use a promise queue for that?

lib/adapters/leveldb/leveldb.js
```diff
@@ -1113,7 +1113,7 @@ function LevelPouch(opts, callback) {
         }
         if (++numDone === revs.length) { // done
           if (overallErr) {
-            return callback(err);
+            return callback(overallErr);
```
nolanlawson: This was clearly a programming error. None of our tests catch it, though, because `compact` never throws an error.

lib/adapters/leveldb/leveldb.js
```diff
@@ -1192,9 +1192,6 @@ function LevelPouch(opts, callback) {
 
       revs.forEach(function (rev) {
         var seq = seqs[rev];
-        if (!seq) {
-          return;
-        }
```
nolanlawson: `seq` should never be 0 or undefined or null. And even if it is, this should be `return checkDone(someError)`. I prefer to just remove it.

lib/adapter.js
```diff
@@ -425,44 +425,34 @@ AbstractPouchDB.prototype.compact =
...
-  }
+  var changesOpts = {
+    returnDocs: false,
+    last_seq: opts.last_seq || 0
+  };
+  var promises = [];
+
   function onChange(row) {
-    started++;
-    self.compactDocument(row.id, 0).then(afterCompact, callback);
```
nolanlawson: ^ this was the source of the bug. We were calling callback for every 'change' event, plus during the 'complete' event.

This code was essentially just trying to do the compactions in parallel and then sequentially run finish at the end. So why not use promises for that? I think the code is cleaner now.

nolanlawson:
OK, this should be good now.

daleharvey:
Looks good, thanks - https://github.com/pouchdb/pouchdb/commit/b4f79e215e9f96f0685a4612f62a613d9b609ea6

PR #3405, opened 2015-01-17
(#3405) - remove pouchdb-express-router test

nolanlawson:
So now that we have the `minimumForPouchDB` mode
in express-pouchdb, I think we can deprecate
pouchdb-express-router. Having two parallel
codebases always made me nervous anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,9 +40,6 @@ env:
   - CLIENT=selenium:firefox SERVER=pouchdb-server COMMAND=test
   - SERVER_ADAPTER=memory LEVEL_ADAPTER=memdown SERVER=pouchdb-server COMMAND=test
 
-  # Test against pouchdb-express-router
-  - CLIENT=node SERVER=pouchdb-express-router COMMAND=test
-
   # Test in firefox/phantomjs running on travis
   - CLIENT=selenium:firefox COMMAND=test
   - CLIENT=selenium:phantomjs ES5_SHIM=true COMMAND=test


daleharvey:
I am not ready to deprecate this, I want to take another shot at splitting out the express-pouchdb functionality into seperate modules, `pouchdb-express-router` is a clean decoupling with almost no pouchdb logic contained, its almost all routing / http logic.

express-pouchdb currently has a bug I havent had a chance to backport that breaks live replication, I still need to add a test and backport that, but even with the refactoring its still a large chunk of code with a lot going on that can get in the way, the last time I tried to use it for pouch.host it broke in a variety of ways, I attempted fixing them but ultimately it was easier to make a more contained module

tldr;

I would like to get express-pouch / pouchdb-express-router sharing all the code they can, but without having the time to do so I think its best to keep both tests running, ill obviously make sure pouchdb-express-router is kept up to date

nolanlawson:
If you're committed to maintaining express-pouchdb-router, then OK, but we did have a big discussion with @marten-de-vries (https://github.com/pouchdb/express-pouchdb/issues/122) and I thought the conclusion was that it didn't make any sense to try to split out express-pouchdb into its own module?

TBH I still tend to see pouchdb-server as being the atomic unit of this thing. Especially if people grow tired of Express and want to use Koa/Hapi/whatever, we don't want them to feel like they need to rewrite the thing from scratch. I enjoy just spinning up a pouchdb-server and using that in my app via HTTP; for me that's as modular as I need it. :)

Issue #3382, opened 2015-01-12
SecurityError DOM Exception 18 on Chrome for iOS on iPhone


jkndrkn:
Hi there. I really appreciate the work that goes into PouchDB. Working with it has been great so far. I've had one major issue with it so far, though.

I am using PouchDB 3.2.1 on a web application targeted for Chrome on iOS. The below error also appears in PouchDB 3.2.0.

When trying to create a connection to a local database on Chrome on iOS, PouchDB uses that browser's WebSQL database. This, however, fails intermittently with SecurityError: DOM Exception 18 when running the application on both iPhone 5 and iPhone 5s. I have not been able to reproduce this error on my iPad 5 yet.

I am creating a new PouchDB database as follows:

    var localDBName = "dbd8008497c6f368d";
    self.localPouch = new PouchDB(localDBName, {size: 50});

I have also tried setting size to 49, 1, and 0 and not setting size at all.

    Here is a stacktrace sourced from BugSnag:
    
    [native code]:4232 openDatabase
    http://192.168.1.144:8000/app/bower_components/pouchdb/dist/pouchdb.js:4232:48 openDB
    http://192.168.1.144:8000/app/bower_components/pouchdb/dist/pouchdb.js:4275:18 WebSqlPouch
    http://192.168.1.144:8000/app/bower_components/pouchdb/dist/pouchdb.js:5655:40
    [native code]:0 Promise
    http://192.168.1.144:8000/app/bower_components/pouchdb/dist/pouchdb.js:5570:28 PouchDB

Here is the code near line 4232 of pouchdb.js:

    var cachedDatabases = {};
    
    var openDBFunction = (typeof navigator !== 'undefined' &&
          navigator.sqlitePlugin &&
          navigator.sqlitePlugin.openDatabase) ?
        navigator.sqlitePlugin.openDatabase.bind(navigator.sqlitePlugin) :
          (typeof sqlitePlugin !== 'undefined' && sqlitePlugin.openDatabase) ?
        sqlitePlugin.openDatabase.bind(sqlitePlugin) :
          (typeof openDatabase !== 'undefined') ?
        openDatabase :
        null;
    
    function openDB(name, version, desc, size) {
      var db = cachedDatabases[name];
      if (!db) {
        db = cachedDatabases[name] = openDBFunction(name, version, desc, size);
      }
      return db;
    }

[...]

PR #3406, opened 2015-01-17
(#3382) - mark websql as invalid when errored

nolanlawson:
The hardest thing about this commit is figuring out what to call the stupid "test" database that we can never possibly delete.

I feel like the shortest possible name will confuse people the least. Seeing a database called `__pouchdb_meta` should not be too shocking or confusing (e.g. in phonegap this would become a `__pouchdb_meta.db` file).

This commit makes me sad, but let's do it.

### DIFF
--- a/lib/adapters/websql/websql.js
+++ b/lib/adapters/websql/websql.js
@@ -993,7 +993,28 @@ function WebSqlPouch(opts, callback) {
 }
 
 WebSqlPouch.valid = function () {
-  return !!openDBFunction;
+  // Some environments (WKWebView, Chrome for iOS, Android WebView
+  // when mis-configured) will throw Security Exception 11 or
+  // Security Exception 18 here. We can catch that by opening
+  // up a test database.
+  if (!openDBFunction) {
+    return false;
+  }
+  // PhantomJS fails our tests for some baffling reason if we
+  // open up this extra DB: https://github.com/pouchdb/pouchdb/pull/3406
+  // Since this test is really just for Android WebView, iOS Chrome,
+  // and WKWebView, assume PhantomJS is fine.
+  if (/PhantomJS/.test(navigator.userAgent)) {
+    return true;
+  }
+  try {
+    var dbName = '__pouchdb_meta';
+    openDBFunction(dbName, 1, dbName, 1);
+    return true;
+  } catch (err) {
+    // ignore
+  }
+  return false;
 };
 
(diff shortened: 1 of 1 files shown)


daleharvey:
failures look genuine here too

nolanlawson:
Yeah, it's failing in phantomjs and iOS for reasons that are kinda baffling.

nolanlawson:
maybe fixed now

nolanlawson:
PhantomJS is not a happy camper.

nolanlawson:
@daleharvey This seems to be some kind of intermittent related to `sync`. Now that I've rebased on your recent changes, I'm getting a timeout instead of an error.

nolanlawson:
Here was the previous error (now the same test is timing out):

```
{ failures: 
   [ { message: 'timeout of 100000ms exceeded',
       title: 'Syncing should stop if one replication fails (issue 838)',
       stack: null } ],
  passed: 750,
  failed: 1,
  lastPassed: 'Doesn\'t have a memory leak (push)' }
```

nolanlawson:
I'm going to pray this fixes itself if I just keep rebasing. Everything seem to be `sync`-related.

nolanlawson:
wtf, this makes no sense

nolanlawson:
I think it's more important that we throw a proper error on WKWebView and Chrome for iOS than that we run correctly in PhantomJS. Gonna throw in the towel and use user-agent sniffing for this.

nolanlawson:
OK OK, I know it sucks, but I added user-agent sniffing for PhantomJS. I just don't feel like expending more than the bare minimum of effort for Phantom: ¯\\\_(ツ)\_/¯

@daleharvey For the record, the reason this PR is important in the first place is that currently you cannot even use the `localstorage` adapter on Chrome for iOS, because the `websql` adapter reports itself as valid and then throw a synchronous error when it starts up, so the promise never resolves. With this fix, both `idb` and `websql` will report themselves as invalid. Incidentally, this will also work for misconfigured Android WebViews, e.g. [this error](http://pouchdb.com/errors.html#android_pre_kitkat).

nolanlawson:
Should probably wait for a green in iPhone to be safe.

daleharvey:
Safari error is looking genuine, I am fine with ua sniffing phantom

nolanlawson:
Ugh yeah, forget this. I filed a bug on Chrome. *insert tableflip ascii*

PR #3412, opened 2015-01-18
(#3412) - pouchdb-express-router should be devDependency

nolanlawson:


daleharvey:
Good catch - https://github.com/pouchdb/pouchdb/commit/e3cec59521ac87190592ae0e4b469b08406606a3

PR #3385, opened 2015-01-14
(#3197) - Last PouchDB.destroy call gone

daleharvey:

PR #3414, opened 2015-01-19
(pouchdb/mapreduce#251) - fix idb destroy() race

nolanlawson:


nolanlawson:
Better wait for a success from IE before merging. Also I might be able to rewrite the test from mapreduce and move it here.

daleharvey:
Hrm, this could possibly be the issue that was causing https://github.com/pouchdb/pouchdb/pull/3385 to fail

nolanlawson:
Yeah I'm also seeing another race related to Chrome+IDB+destroy. Trying to track it down.

daleharvey:
Green in iphone and ie :) 

https://github.com/pouchdb/pouchdb/commit/0bc2c4d6e203bf225483483ec47a2cefdfc41cec

Cheers

PR #3423, opened 2015-01-20
Serve admin party config to Sync Gateway.

jchris:
This requires that CSG be configured to look for database config from
the HTTP service that the dev server starts on port 8001.

Example Sync Gateway config: https://github.com/couchbase/sync_gateway/blob/master/examples/config_server.json

There are going to be some small changes like this that don't pay off until the whole thing comes together to allow CSG to sync with PouchDB (and pass the tests). I hope test/build stuff like this can get merged along the way instead of waiting for some mega merge in the future.


daleharvey:
Yeh doing small patches like this is perfect. We have some existing switches to support this type of functionality, we should avoid just starting a whole bunch of servers if we only need one of them.

in the tests we use SERVER=couchdb-master|express-pouchdb|pouchdb-express-router, currently we use that to start the server in https://github.com/pouchdb/pouchdb/blob/master/bin/run-test.sh, but it could also be picked up in https://github.com/pouchdb/pouchdb/blob/master/bin/dev-server.js to provide the alternative config and we use COUCH_HOST= to specify where pouchdb (or the cors proxy) sends its requests

jchris:
I think I've addressed Dale's advice with these updates. I also added a `testUtils.isSyncGateway()` function but this pull request doesn't include the tests that call it.

That commit log includes a bunch of noise, I'm happy to do a new flattened pull req once the code is reviewed.

daleharvey:
Yo chris, any chance you could rebase the branch against master? I cant actually see whats changed since its split across and joined together with all the master commit

jchris:
I think the noise is b/c I rebased. I'll just make a new flat pull request from my topic branch.

jchris:
Closing this in favor of https://github.com/pouchdb/pouchdb/pull/3464

Issue #3200, opened 2014-12-17
Remove ajax-browser.js


daleharvey:
I would like to remove ajax-browser.js, have http.js call request() directly, then replace request with a version of ajax-browser / browser-request that doesnt contain any pouchdb logic mostly to clean things up and make consistent error handling easier

jchris:
@daleharvey this is something I can help with, maybe you can help me move faster. It looks like `utils.js` requires `./deps/ajax` and ajax-browser is swapped out via it's entry in `package.json` in the top-level "browser" field.

If I remove `ajax-browser` from the build by deleting that entry, and instead modify `package.json`'s browser field to use browser-request instead of request, would that be worth running the test suite against?

daleharvey:
@jchris I looked around for the branch that I had started with this on but I cant find it, sorry. 

So the first thing I did was remove `require("ajax")` from ``http.js`, and and made all the `ajax(` calls go through a function that just sent it direct to `request`, it needed a little bit of logic from `ajax.js` like the default parameters but it should be mostly compatible. Then you can run the tests in node and get them passing.

I am not sure we will want to use "browser-request", if I remember it was a fair bit larger than our own wrapper with a bunch of polyfills for ie6 etc that we dont need, but when its passing in node its probably worth doing a test run with it so see how close it gets to passing and to check the build size to see if I am wrong about that. then we can either use browser-request or write our own

jchris:
The new release of request is supposed to be browserify friendly, so I started out down that path but ran into an issue upgrading request. Also I think we would not used request via browserify anyway, b/c it is so heavy. (Lots of proxy logic etc that we don't need.)

So maybe it makes sense for me to test the browser-request path instead. But I'll be surprised if any of these turn out to have multipart support either.

I'm trying to exhaust options other than enhancing ajax-browser.js with multipart. It seems at least that we should only have to write the multipart code once, for the browser and node.

daleharvey:
Yeh fully agree at not writing twice, have you got it to a point where it works in node with just request? at that point we can test browser-request but I expect we will want to write our own and it shouldnt be a huge amount of work

jchris:
It already works with request. Do you mean remove ajax.js also?

Right now I think testing ajax.js plus browser-request is one step to
simplifying the dependency graph.

On Tuesday, January 20, 2015, Dale Harvey <notifications@github.com> wrote:

> Yeh fully agree at not writing twice, have you got it to a point where it
[...]

PR #3425, opened 2015-01-21
(#3200) - Use a request polyfill for xhr

daleharvey:


daleharvey:
Actually fairly surprised here, this passes all but 3 tests in firefox at least

daleharvey:
All but 2, those 2 are intermittent, really hard to get reproducing in a sane way, may be related to the deletion race

daleharvey:
Ok so this should be passing tests, not implementing the cache busting made it break in curious and fun ways, one issue is that it adds 16K to the minified build size despite no extra modules and almost the same code

daleharvey:
Ok found it, there was a `new Buffer` in there

So this doesnt technically change much, request-browser is pretty much what used to be in `ajax-browser.js`, it just simplifies the dependency tree so the xhr logic is behind request, if there is a better request browser polyfill that is compatible where we need it and a smaller build size then it can be replaced much easier than before

Will likely want to publish request-browser at some point but I would like to get android testing then hopefully remove createBlob before worrying about that

nolanlawson:
Android 4.0-4.3 is really crucial to manually test before merging any XHR changes. I remember there being a bizarre http bug in that version, and sadly [4.0-4.3 is still a big chunk of Android users](https://developer.android.com/about/dashboards/index.html).

daleharvey:
ok think I have found the last of the issues for this, all worked fine except attatchments broke in various ways in various browsers

daleharvey:
Ok this is passing in every local browser I can test, should be good to go now

nolanlawson:
I'll manually test in android 4.3

daleharvey:
I have been between trying to manually test and get https://github.com/pouchdb/pouchdb/pull/3425 up and running

tests/integration/test.attachments.js
```diff
@@ -1749,7 +1749,7 @@ adapters.forEach(function (adapter) {
                 cache: true,
                 binary: true
               }, function (err, res) {
-                if (err && err.status === 405) {
+                if (err && err.status === 500) {
```
nolanlawson: ^ what's the deal here? I know firefox had fixed the 405 bug in nightlies, but for older versions of firefox, won't this cause us to use base64 when actually blobs are okay?

nolanlawson:
Nice, min+gz size only increased from 46035 to 46132.

tests/integration/test.attachments.js
daleharvey: As per the comment in the test, this check only to skip the tests in firefox, we previously had done https://github.com/pouchdb/pouchdb/blob/master/lib/deps/ajax-browser.js#L80 to always force a 405 which was causing various ways xhr could fail so misreport as cors problems, afaik it should affect blob detection

We are testing in fx 31 in travis which is older than we support anyway

nolanlawson: okay, I remember now. lgtm

nolanlawson:
100% in android 4.2.2. ship it :+1:

tests/integration/test.attachments.js
nolanlawson: For the record, it seems Firefox 35+ has the blob fix, so it doesn't report any kind of error.

As for previous versions of Firefox, I'm glad we're testing it, but to be fair, our tests won't capture if we get a false negative and revert to base64 when actually blob would be appropriate. But given that base64 works universally, and that Firefox is so good about auto-updating, I'm not really concerned.

nolanlawson:
a82c273d5e96146b366ce55780337d629eb169ba

Issue #3426, opened 2015-01-21
Uncaught missing: Test live push replication


daleharvey:
This test is failing regularly locally

daleharvey:
Fixed

PR #3427, opened 2015-01-21
(#3426) - Fix intermittent test

daleharvey:


daleharvey:
This isnt a fix, trying to find a reproducable test for the bug, but this testing being intermittent has been getting in the way of a lot of other work

nolanlawson:
This look equivalent to me, so +1

daleharvey:
They are supposed to be but in practise arent, however we use this pattern a ton in the tests so really need to find the fix, gonna close this one out

Issue #3197, opened 2014-12-17
Deprecate PouchDB.destroy


daleharvey:
I think @calvinmetcalf argued in favour of this a while ago and I may have objected, however just been debugging an issue with a deleted database and its pretty complex tracking where we actually destroy stuff from, it has several methods of indirection and its a very minor use case.

I think we should mark it as deprecated in the next release, then remove it in the one after

daleharvey:
```
      // return new PouchDB('mydb').then(function (db) {
      //   return db.destroy();
      // }).then(function () {
      //   return new PouchDB('mydb', {db: require('memdown')});
      // }).then(function (db) {
      //   return db.destroy();
      // });
      return PouchDB.destroy('mydb').then(function () {
        return PouchDB.destroy('mydb', {db: require('memdown')});
      });
```
So I am dumb, the commented out code fails with a circular promise resolution chain, they seem equivalent, any ideas what I am doing wrong? (cc @calvinmetcalf)

nolanlawson:
I basically stopped using the "PouchDB is a promise for itself" style the first time I ran into this.

Why not just:

```js
return new PouchDB('mydb').destroy().then(function () {
  return new PouchDB('mydb').destroy();
})
```

daleharvey:
Ah cheers, that worked great

daleharvey:
Merged in https://github.com/pouchdb/pouchdb/commit/13b9fa3ae552a5b18b5c564ddaf8630d4890283d

PR #3428, opened 2015-01-21
(#3197) - Fix test that left trailing changes feed

daleharvey:


daleharvey:
Found while trying to fix one of the many strange intermittent errors blocking my patches, we never cancel the changes feed in this test which means it will interfere with follow on tests

nolanlawson:
good find, +1

daleharvey:
Merged @ https://github.com/pouchdb/pouchdb/commit/83ae9e54ecb739657adc389abb4436cdb570fa12, cheers

PR #3434, opened 2015-01-22
Deprecate PouchDB.destroy

daleharvey:


daleharvey:
So the reasoning for this, when the use `db.destroy` currently, it goes to https://github.com/pouchdb/pouchdb/blob/master/lib/constructor.js#L115, makes an `info()` call to figure out its own name, then calls the static `PouchDB.destroy` in https://github.com/pouchdb/pouchdb/blob/master/lib/setup.js#L74 which then constructs another `new PouchDB` instance to then call the internal `destroy` on itself and get rid of dependantDbs

When we remove `PouchDB.destroy` we can put most of the code from https://github.com/pouchdb/pouchdb/blob/master/lib/setup.js#L111 into `db.destroy`, avoid the info call and the several setup requests, it will be vastly less complex and may actually be a fairly large reduction in the test suite time

lib/sync.js
```diff
@@ -95,6 +97,11 @@ function Sync(src, target, opts, callback) {
     };
   }
 
+  if (opts.live) {
+    this.push.on('complete', self.pull.cancel.bind(self.pull));
+    this.pull.on('complete', self.push.cancel.bind(self.push));
+  }
+
```
nolanlawson: ^ Why is this necessary?

daleharvey: This is based on top of https://github.com/pouchdb/pouchdb/commit/ada62f35b7fa1badab9dcf8e9b50689dc13e4000 which is already merged

nolanlawson: Shoot, missed that. I thought it was in a test instead.

Why does `sync` need to auto-cancel but `replicate` doesn't?

daleharvey: This cancels the other replication in the sync, so if I `pull.cancel()` this catches it and calls `pull.cancel()`

nolanlawson: That's weird; it seems like that ought to be handled already by [these lines](https://github.com/pouchdb/pouchdb/blob/ada62f35b7fa1badab9dcf8e9b50689dc13e4000/lib/sync.js#L178-L184).

daleharvey: thats if you call `sync.cancel()`

nolanlawson: How is it possible for a user to call `pull.cancel()` when `sync()` only gives them the `Sync` object?

daleharvey: the Sync object has push / pull properties pointing to the replciation objects

nolanlawson: ooh, okay, I getcha.

nolanlawson:
+1 when green. Let's get rid of `PouchDB.destroy`.

daleharvey:
ugh, this failing generally means some obscure broken test that is a genuine bug, however this particular configuration being the only failure is very strange

daleharvey:
hrm, so looks like the suite actually passes, however the selenium client stops being able to read

also seeing 

```
Possibly unhandled Error: Can't set headers after they are sent.
    at ServerResponse.OutgoingMessage.setHeader (http.js:691:11)
    at Object.exports.sendJSON (/Users/daleharvey/src/pouchdb/node_modules/express-pouchdb/lib/utils.js:185:9)
    at Object.exports.sendError (/Users/daleharvey/src/pouchdb/node_modules/express-pouchdb/lib/utils.js:177:11)
    at /Users/daleharvey/src/pouchdb/node_modules/express-pouchdb/lib/routes/compact.js:15:20
    at tryCatch1 (/Users/daleharvey/src/pouchdb/node_modules/pouchdb-server/node_modules/pouchdb/node_modules/bluebird/js/main/util.js:63:19)
    at Promise$_callHandler [as _callHandler] (/Users/daleharvey/src/pouchdb/node_modules/pouchdb-server/node_modules/pouchdb/node_modules/bluebird/js/main/promise.js:695:13)
    at Promise$_settlePromiseFromHandler [as _settlePromiseFromHandler] (/Users/daleharvey/src/pouchdb/node_modules/pouchdb-server/node_modules/pouchdb/node_modules/bluebird/js/main/promise.js:711:18)
    at Promise$_settlePromiseAt [as _settlePromiseAt] (/Users/daleharvey/src/pouchdb/node_modules/pouchdb-server/node_modules/pouchdb/node_modules/bluebird/js/main/promise.js:868:14)
    at Promise$_settlePromises [as _settlePromises] (/Users/daleharvey/src/pouchdb/node_modules/pouchdb-server/node_modules/pouchdb/node_modules/bluebird/js/main/promise.js:1006:14)
    at Promise$_rejectPromises [as _rejectPromises] (/Users/daleharvey/src/pouchdb/node_modules/pouchdb-server/node_modules/pouchdb/node_modules/bluebird/js/main/promise.js:999:10)
    at Async$_consumeFunctionBuffer [as _consumeFunctionBuffer] (/Users/daleharvey/src/pouchdb/node_modules/pouchdb-server/node_modules/pouchdb/node_modules/bluebird/js/main/async.js:74:12)
    at Async$consumeFunctionBuffer (/Users/daleharvey/src/pouchdb/node_modules/pouchdb-server/node_modules/pouchdb/node_modules/bluebird/js/main/async.js:37:14)
    at process._tickCallback (node.js:415:13)
``` 

in the output

daleharvey:
Ugh, accidently merged this, will try to get the pouchdb-server red fix or backout, but for now merged in https://github.com/pouchdb/pouchdb/commit/13b9fa3ae552a5b18b5c564ddaf8630d4890283d

marten-de-vries:
I just realised that the semantics for ``PouchDB.destroy()`` and ``db.destroy()`` aren't the same. The first makes a db which isn't removed when the second fails, the second 'just removes' (at least, from the user perspective). Since .destroy() can't fail on PouchDB level, it doesn't seem to matter much, but pouchdb-wrappers allows to add extra functionality to ``PouchDB.destroy()`` which can make it fail. pouchdb-security actively does this (``.destroy()`` fails if the user isn't an admin or something similar).

Because of this semantic difference, I will keep the function around in pouchdb-wrappers. It will just do something like this when it's asked to wrap a PouchDB object:
```
PouchDB.destroy = function (name, opts, callback) {
  var db = new PouchDB(name, opts);
  return db.destroy(callback);
}
```

This means that the function isn't 'going away' completely. E.g. express-pouchdb, which uses pouchdb-security, will still call it.

Does anyone see any ways around this? Or any things that are going to go wrong because of this approach?

marten-de-vries:
> Or any problems with this approach?

Found one. While PouchDB.destroy() is deprecated, it's still used internally so can't just be replaced with a non-warning function. To feature check around this I'm going to have to rely on:

- PouchDB.version >= '3.2.2' -> the function needs to be replaced - this should be safe because semver.
- *IF* PouchDB calls ``PouchDB.destroy()`` (which it still will until 4.0.0 at least), I will need it to pass in ``opts.internal`` as it currently does.

That's kind of brittle, but nothing to be done about that. So no changing the mechanics of showing the warning message please...

nolanlawson:
This is starting to seem kind of gnarly. Maybe we should just stop documenting `PouchDB.destroy` instead of removing it entirely? I'm kind of confused as to why we need it in express-pouchdb though @marten-de-vries.

daleharvey:
Debugging bugs with destroy is extremely complicated at the moment, its implementation is extremely messy because we need to maintain PouchDB.destroy + db.destroy, getting rid of it may be tricky, but we have done the majority of the work and I would rather work hard to remove technical debt than to work hard building on top of it

Not sure why you would need to maintain a `PouchDB.destroy` in express-pouchdb, can you point us to where it is used? it seems like something else should be used to deal with this

marten-de-vries:
@nolanlawson pouchdb-security can make ``db.destroy()/PouchDB.destroy()`` return an error, if the user is unauthorized to remove the current db. That breaks the ``new PouchDB(name).destroy()`` 'replacement': if the db that's deleted didn't exist it now does.

[Relevant line](https://github.com/pouchdb/express-pouchdb/blob/master/lib/routes/db.js#L48).

@daleharvey I'm completely ok with removing ``PouchDB.destroy()``, I don't need it - I'm just pointing out that semantics between: ``db.destroy()`` and ``PouchDB.destroy()`` are different and thus I'm going to maintain both in pouchdb-wrappers.

The main problem is that, although ``PouchDB.destroy()`` is deprecated, it actually is the implementation of ``db.destroy()`` internally. This is why things get hacky. I need to take into account ``opts.internal`` like PouchDB does too. Which is annoying, but I already have the code so it's possible.

---

As for alternate solutions, there are two I can think of:
- stop using pouchdb-security in express-pouchdb & roll something custom not dependent on wrappers (lot of work, + I like the fact I can re-use pouchdb-security offline so I'm not in favour of this.)
- check if the db exists using allDbs(). That's now an in-memory operation so reasonably efficient, but technically adds a possible race condition. (There's no guarantee the allDbs list changed).

nolanlawson:
If `allDbs()` has a race condition, then we should fix that, no? Also I'm pretty sure the tests in `allDbs` confirm that there are no races; in fact, when we went all in-memory, I recall having to fix lots of bugs due to race conditions, so it should be good AFAIK....

daleharvey:
He means if another request created the database between the allDbs call.

Ok so you have a wrapper around PouchDB.destroy that may or may not fail (prior to PouchDB.destroy being called), now that `PouchDB.destroy` is going away its not so much a wrapper but introducing a new API to PouchDB. I think this is a reasonable use case for PouchDB.destroy and one of the reasons I wasnt particularly for removing it in the first place.

The main reason `destroy` is so complicated is the `dependentDbs` work, this is map reduce functionality baked into pouchdb and I dont love the coupling involved, it may be possible to take `dependentDbs` out of pouchdb and into the plugins with the appropriate hooks `beforeDelete`? I remember there was problems with this before, but I may revisit seeing how possible it is as it seems like a much cleaner way to implement it

daleharvey:
And also wanted to mention there isnt any rush on removing PouchDB.destroy, I wanted to put the deprecation in early so we had the flexibility to do what we wanted in future.

If we come up with a nice solution for hooks, I vaguelly remember CouchDB discussing the same functionality and it may be a cleaner solution for implementing pouchdb-wrapper as well

nolanlawson:
Adding some kind of `beforeDelete` or `beforeDestroy` hook and letting the plugins take care of cleaning themselves up would fine, yes.

For the record, though, the dependentDbs stuff is not just required by map/reduce, but also pouchdb-find, pouchdb-quick-search, geopouch, and potentially others that create sub-databases.

Maybe plugins could define a magical `cleanup` method that PouchDB would use?

Issue #3220, opened 2014-12-19
PouchDB should not assume that update_seq is numeric


glynnbird:
In CouchDB 1.6, a database's update_seq is represented as an integer e.g.

```
curl  http://127.0.0.1:5984/test
{"db_name":"test","doc_count":1,"doc_del_count":0,"update_seq":1,"purge_seq":0,"compact_running":false,"disk_size":4188,"data_size":287,"instance_start_time":"1418989168372529","disk_format_version":6,"committed_update_seq":1}
```

In Cloudant, update_seq is a string: e.g.
```
{"update_seq":"60-g1AAAAG9eJzLYWBg4MhgTmGQS0lKzi9KdUhJMtXLTMo1MLDUS87JL01JzCvRy0styQGqY0pkSJL___9_ViInqg4TnDqSFIBkkj1YEztIkzxck6EBbl0OIF3xYF1ojrPArSkBpKkeYRUxPspjAZIMDUAKqG8-hm24PQbRuACicT8pnoPoPADReR_Drbg9CNH4AKIR5MksAJ6Xjlg","db_name":"grid_42","sizes":{"file":394600,"external":920,"active":5941},"purge_seq":0,"other":{"data_size":920},"doc_del_count":0,"doc_count":4,"disk_size":394600,"disk_format_version":5,"compact_running":false,"instance_start_time":"0"}
```

As of CouchDB 2.0  update_seq is an array e.g.:
```
{"db_name":"ab","update_seq":[1,"g1AAAAFDeJzLYWBg4MhgTmHgz8tPSTV0MDQy1zMAQsMcoARTIkOS_P___7MSGXAqSVIAkkn2hFQ5gFTFE1KVAFJVT0BVHguQZGgAUkCF8wmrXABRuZ-wygMQlfcJq3wAUQl0J2MWAB5TVnc"],"sizes":{"file":38064,"external":101,"active":311},"purge_seq":0,"other":{"data_size":101},"doc_del_count":0,"doc_count":1,"disk_size":38064,"disk_format_version":6,"data_size":311,"compact_running":false,"instance_start_time":"0"}
```

----------------------

As it stands, PouchDB does things like this (changes.js):
```
opts.since = info.update_seq  - 1;
```
which clearly won't work with strings or array sequence numbers.

willholley:
Just noting a couple of places where this breaks existing behaviour:
https://github.com/pouchdb/pouchdb/blob/master/lib/changes.js#L48
[...]

Issue #2342, opened 2014-06-04
Wrong update_sequence with WebSQL Adapter


PhilReinking:
Hi,
it seems to be that PouchDB have problems to update the 'update_sequence' to the correct value after receiving changes on safari and iOS devices by using the WebSQL adapter.
1. We use a design doc to query the database for documents and save the result in an array.
2. We fetch all future changes with the `db.changes` method and push them to the existing array.

As a result by using the WebSQL adapter we get duplicate entries everytime you run the code, because some documents are fetched in the `db.query` and `db.changes`.

I created that jsfiddle document to reproduce the issue: http://jsfiddle.net/PhilReinking/Me3Z6/7/

![04-06-2014 10-53-22](https://cloud.githubusercontent.com/assets/2592996/3171393/f40686ec-ebc5-11e3-8c8c-e076ed48ec7a.png)

Issue #2800, opened 2014-09-26
db.changes({since: 'now', live: true}).on('change', Always fires a change


daleharvey:
Every time my app starts up a change gets fired, pretty sure it shouldnt

oliverdm:
I was about to ask why `db.changes({since: 'now'})` also returns the last change when up-to-date before I saw this issue. From the docs I assumed otherwise.

In 9a33d72ba921320bda42157ab563b8834c408d29 `lib/changes.js` lines 143-166 `now` is mapped to `info.update_seq  - 1;`:

    if (opts.since === 'now') {
        ...
        opts.since = info.update_seq  - 1;
        ....
    }

Would that explain it? And is it intended to behave that way?

daleharvey:
This should be fixed by https://github.com/pouchdb/pouchdb/issues/3450

PR #3439, opened 2015-01-23
(#136) - Fix "#2342 update_seq after replication"

willholley:
Do not use sequence number arithmetic when calculating the sequence number for use with since=now/latest or continuous changes feeds.

CouchDB 2.0 represents sequence numbers as opaque blobs so the previous approach would fail. Rather than deduct 1 from the current sequence number, this commit fetches the latest 2 results from the changes feed and uses the second as the starting point for the original call. If less than two results are returned then we start from 0.

This also fixes most issues identified in #3220.

### DIFF
--- a/lib/changes.js
+++ b/lib/changes.js
@@ -141,20 +141,32 @@ Changes.prototype.doChanges = function (opts) {
     opts.since = 0;
   }
   if (opts.since === 'now') {
-    this.db.info().then(function (info) {
+    this.db.changes({
+      descending: true,
+      limit: 2
+    }).then(function (resp) {
       if (self.isCancelled) {
         callback(null, {status: 'cancelled'});
         return;
       }
-      opts.since = info.update_seq  - 1;
+      opts.since = resp.results.length > 1 ?
+        resp.last_seq : 0;
       self.doChanges(opts);
     }, callback);
     return;
   }
 
   if (opts.continuous && opts.since !== 'now') {
-    this.db.info().then(function (info) {
-      self.startSeq = info.update_seq - 1;
+    this.db.changes({
+      descending: true,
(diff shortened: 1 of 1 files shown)


daleharvey:
What happens if you take out the `- 1`? it seems like if `update_seq` is foo, since=foo should give you everything after it no? it seems like the -1 is a bug elsewhere, I remember we had to go back and forward on it previously

willholley:
Taking out the -1 changes the semantics of since=now/latest (currently they
always return with the latest change but if we took out the -1 they would
return empty until a new change occurs). I did consider it but thought it
was better to have a fix which didn't change the behaviour. If your happy
with that behaviour change then we can explore the impact as in alternative
PR?

willholley:
I should also mention that in the current PR, since=latest is
non-deterministic when running against CouchDB 2 due to the ad-hoc ordering
in the changes feed. This means the tests for this occasionally fail and
will need further consideration.

daleharvey:
It is a bug that `since: "now"` returns the most recent change, couchdb doesnt do that and not all the browsers do last time I checked, I forgot why the -1 was added but may have been an incorrect workaround of an existing issue, or taking it out may show up the issue, no tests should depend on that behaviour

daleharvey:
https://github.com/pouchdb/pouchdb/issues/2800 is the bug filed for it

daleharvey:
@willholley also if you make your PR's as branches against master on this repo as opposed to your own, then the full suite including the saucelabs tests will run, I think thats a good idea for some of these things which may have cross browser issues

willholley:
OK makes sense. I'll take a stab at removing the -1 then.

willholley:
closing this in favour of #3442

PR #3442, opened 2015-01-23
(#3220) - remove sequence number arithmetic in changes.js

willholley:
Remove the seq - 1 logic from since=now/latest and continuous changes implementation. This commit is just to observe/document the test failures.

### DIFF
--- a/lib/changes.js
+++ b/lib/changes.js
@@ -146,7 +146,7 @@ Changes.prototype.doChanges = function (opts) {
         callback(null, {status: 'cancelled'});
         return;
       }
-      opts.since = info.update_seq  - 1;
+      opts.since = info.update_seq;
       self.doChanges(opts);
     }, callback);
     return;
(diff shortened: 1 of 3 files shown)


willholley:
In Firefox it seems that https://github.com/pouchdb/pouchdb/blob/master/tests/integration/test.events.js#L63 adds the event listener asynchronously so it is not listening when the following db.put call is made in the test. Not sure yet why FF differs from other browsers here but inserting a delay before the PUT makes it pass reliably.

daleharvey:
The create event test also needs fixed, thats the current failure, I did the same as you did to the change event tests in https://github.com/pouchdb/pouchdb/pull/3451, also needed a rebase, fingers crossed for a green run

daleharvey:
Also cancelled the test run since it was based on a bunch of bugs in master, all my fault sorry, will get tested in the other run

daleharvey:
Merged :) https://github.com/pouchdb/pouchdb/commit/fc8abbea4ea762435428389bb1ff70e39761a9cf

Issue #3415, opened 2015-01-20
destroy() race condition causing IndexedDB to hang


nolanlawson:
Still haven't tracked this down, but opening up an issue so it doesn't fall through the cracks.

I've managed to get IndexedDB in a state where it just doesn't respond in Chrome after a few tests. Seems to be related to `stale: update_after` in map/reduce causing some parallelism. Weirdly, the tests complete when the Dev Tools are open, so we may have another Heiseinbug on our hands.

https://github.com/pouchdb/pouchdb/pull/3414 does not fix this particular bug.

nolanlawson:
Oh okay, in both Firefox and Chrome it gets stuck around 93%. Live demo: http://bl.ocks.org/nolanlawson/dc5d9d1f2de083963689

nolanlawson:
Cannot repro in Safari.

nolanlawson:
Cannot repro in IE. Also cannot repro in Firefox anymore; maybe it was a fluke.

daleharvey:
Hmm, I am sure there is some problem with https://github.com/pouchdb/pouchdb/blob/master/lib/constructor.js#L118 interfering with the task queue

nolanlawson:
Hmmm, I'm kinda convinced it's an IDB bug, else why would it work in WebSQL/Level?

daleharvey:
I think we may have different bugs

nolanlawson:
I can occasionally repro in Firefox 35. It's very inconsistent though.

nolanlawson:
OK, Chrome is saying `Unhandled promise rejection` with stacktrace;

```
"Error: Failed to execute 'transaction' on 'IDBDatabase': The database connection is closing.
    at Error (native)
    at startTransaction (http://bl.ocks.org/nolanlawson/raw/dc5d9d1f2de083963689/test-bundle.js:19678:15)
    ...
```

The line throwing an error is [this one](https://github.com/pouchdb/pouchdb/blob/ada62f35b7fa1badab9dcf8e9b50689dc13e4000/lib/adapters/idb/idb-bulk-docs.js#L63).

So I think we need two fixes:

1) Whenever `idb.transaction` is called, convert a sync error into an async error
2) Mapreduce should catch errors and do something smart if the database is closing

nolanlawson:
OK, I get it. The issue is that `idb.transaction` throws a sync error from within a function that is not a promise, so it doesn't call `reject`.

nolanlawson:
[...]

PR #3443, opened 2015-01-24
(#3415) - open idb transactions safely

nolanlawson:
I cannot for the life of me find a failing test for this. It's super inconsistent because the IndexedDB has to be closed at exactly the moment that the `bulkDocs`'s transaction is created.

This PR will really unblock me on the Cloudant queries thing, so I would appreciate a merge even without a new test.

### DIFF
--- a/lib/adapters/idb/idb-bulk-docs.js
+++ b/lib/adapters/idb/idb-bulk-docs.js
@@ -16,6 +16,7 @@ var compactRevs = idbUtils.compactRevs;
 var decodeMetadata = idbUtils.decodeMetadata;
 var encodeMetadata = idbUtils.encodeMetadata;
 var idbError = idbUtils.idbError;
+var openTransactionSafely = idbUtils.openTransactionSafely;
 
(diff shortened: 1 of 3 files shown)


daleharvey:
I really dont like the defensive code here, but if it unblocks you with stuff, hopefully this makes it possible to find the actually issue and we can improve on this / take out and fix it properly

https://github.com/pouchdb/pouchdb/commit/af89eba96ac2ab83a39b5959a77469305b4ba30a

nolanlawson:
I know the actual issue; I just can't find a test to repro it.

Basically we need a test that hits the `catch` inside the `try/catch`, but it's a race condition so hard to repro.

daleharvey:
`idb.transaction` should not throw though? should really be an upstream bug it seems like, but understand its a hard one to pin down

nolanlawson:
it throws if the database is closing, yeah. So basically if you try to destroy a database while doing a map/reduce update (e.g. with `update_after`).

PR #3445, opened 2015-01-24
(#3197) - less logging for deprecation warning

nolanlawson:
This is a bit much while I'm trying to test.
Can we tone it down?

### DIFF
--- a/lib/setup.js
+++ b/lib/setup.js
@@ -71,6 +71,8 @@ PouchDB.parseAdapter = function (name, opts) {
   throw 'No valid adapter found';
 };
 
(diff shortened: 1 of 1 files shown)


marten-de-vries:
Makes sense to me. :+1: when green.

daleharvey:
So this notice really should be annoying, its going to be pretty easy to miss, I would really prefer it to be left as is, I filed a patch to take the destroy usage out of mapreduce, can we go that way instead? 
https://github.com/pouchdb/mapreduce/pull/257

nolanlawson:
sure, I'm fine with that. closing

Issue #3446, opened 2015-01-24
pouchdb-server failure


daleharvey:
Introduced by https://github.com/pouchdb/pouchdb/issues/3197, the tests actually pass but reported as a failure, maybe the slowness adds a timeout

daleharvey:
This magically fixed itself, to be reenabled in https://github.com/pouchdb/pouchdb/pull/3461

PR #3448, opened 2015-01-24
(#3446) - Add pouchdb-server + firefox to expected fails

daleharvey:


daleharvey:
So I dont particularly want to do this, usually test failures should be backed out, however this isnt actually failing tests, they pass however the connection with selenium is lost, also the delete patch may have fixed the ie tests, so we might be able to get ie out of expected failures, which is more important

daleharvey:
But looking into this, I want to get the rest of the expected failure out of there and in the green list

nolanlawson:
So I take it this is happening because `new PouchDB().destroy()` is slower than `PouchDB.destroy()`? (Due to the implicit create-then-delete?) Kinda sad then to do this then; I would almost prefer to just roll back the `PouchDB.destroy` changes entirely. TBH I don't think anybody is using `destroy` (either variant) except for us, so messing with this seems to be causing more problems than it's solving.

daleharvey:
I dont know what is causing it, I dont think its because it is slower in particular but when we do actually remove `pouchdb.destroy` we can vastly improve both its complexity and its performance, right now its going through a whole bunch of indirection so I would like to figure out why this is going wrong without backing it out

nolanlawson:
OK, fair enough, let's move forward. +1 when green

daleharvey:
Merged in https://github.com/pouchdb/pouchdb/commit/31b0821453744d41423a74ba5b86df7fdb219328

Issue #3453, opened 2015-01-25
Reenable ie


daleharvey:

daleharvey:
Closing in favour of https://github.com/pouchdb/pouchdb/pull/3461

PR #3454, opened 2015-01-25
(#3453) - Remove ie from allowed failures

daleharvey:


daleharvey:
This is pretty green in https://travis-ci.org/pouchdb/pouchdb/builds/48252525

daleharvey:
3/20 is around the rate we generally get from saucelabs, it isnt perfect but better than being an expected failure

nolanlawson:
I'm used to seeing about 75% failure rate, so that is phenomenal

daleharvey:
Surprisingly it was the db.destroy() commit that got this more stable

PR #3463, opened 2015-01-26
3176 - fix changes/last_seq in local storage adapters

willholley:
Matching CouchDB behaviour, the last_seq field in _changes should represent the last active change processed.
    
The previous implementation was correct for the WebSQL adapter. However, the IndexedDB and LevelDB adapters would update last_seq for each change regardless of whether there was a more recent version of the document in the database (which would subsequently be filtered out of the feed).
This worked for ascending changes but when querying the feed in descending order, the bug was exposed.

In this commit, we increment last_seq after checking that the current change represents the max revision of the current document.

TODO: check the behaviour when conflicts are present - possibly we need to set last_seq for the winning revision rather than the max revision?

### DIFF
--- a/tests/integration/test.changes.js
+++ b/tests/integration/test.changes.js
@@ -1336,7 +1336,6 @@ adapters.forEach(function (adapter) {
         return db.changes();
       }).then(function (result) {
         normalizeResult(result);
-        delete result.last_seq;
         result.should.deep.equal({
           "results": [
             {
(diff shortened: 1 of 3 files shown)


nolanlawson:
> TODO: check the behaviour when conflicts are present - possibly we need to set last_seq for the winning revision rather than the max revision?

I am 99% sure it's the max revision rather than the winning revision.

tests/integration/test.changes.js
```diff
@@ -1358,77 +1357,71 @@ adapters.forEach(function (adapter) {
               "changes": [{ "rev": "3-f"}
               ]
             }
-          ]
-          //, "last_seq": seqs[5]
+          ],
+          "last_seq": seqs[5]
```
nolanlawson: ahhhh this explains it! I could not figure out what was wrong with the descending changes. nice one.

nolanlawson:
+1 when green. Glad the fix was so simple.

daleharvey:
Failure looks genuine, broken in couch master

willholley:
according to http://docs.couchdb.org/en/latest/api/database/changes.html, changes will return the "winning" revision by default. I'll have a think around devising a test...

willholley:
if this passes I think we're good

daleharvey:
Sweet good catch - https://github.com/pouchdb/pouchdb/commit/50ebc3c4a30294fcba9efd886bd2c1cec7bcf664

PR #3464, opened 2015-01-27
Sync gateway config

jchris:
This is a cleaned up version of https://github.com/pouchdb/pouchdb/pull/3423

I refactored the config server to it's own file / process so that it can be managed by run-test.sh, and now everything seems to work great for testing node.

In the browser you have to append `?SERVER=sync-gateway` to your URL to get `testUtils.isSyncGateway()` or `testUtils.isCouchMaster()` to behave properly during tests. I think this is a distinct issue and can be handled later.


daleharvey:
> In the browser you have to append ?SERVER=sync-gateway to your URL to get 
> testUtils.isSyncGateway() or testUtils.isCouchMaster() to behave properly during tests. 
> I think this is a distinct issue and can be handled later.

This is how we pass through the options from the env var to the browser, its generally used for test automation, could be improved maybe but not an issue with this patch

This looks good, thanks for the edits, +1 will let it go green before merging though

daleharvey:
All green, squashed them together and merged in https://github.com/pouchdb/pouchdb/commit/0c34765dfe1b77f0d4f68f284b43a71074c9c5b1, cheers

PR #3469, opened 2015-01-29
(#136) - Enable assertions previously blocked by COUCHDB-2523

willholley:
COUCHDB-2523 is now fixed (applies only to CouchDB 2.0) so enable the assertions which this was blocking.


daleharvey:
Awesome, I would love to get the isCouchMaster checks cleaned up and removed from the tests, great start

https://github.com/pouchdb/pouchdb/commit/763462531f612bb10480f8ec856742ae301405a5

Issue #3058, opened 2014-11-28
iphone has gone permafail


daleharvey:

daleharvey:
Ok, so to describe the failure where it is now, somewhere around 915 tests in, we stop getting results back from saucelabs, on the client side we see `{ [Error: [eval("window.results")] ESOCKETTIMEDOUT] code: 'ESOCKETTIMEDOUT' }` (https://travis-ci.org/pouchdb/pouchdb/jobs/48998583). This is happening fairly consistently.

On the saucelabs side these runs are generally failing with `Test did not see a new command for 599 seconds. Timing out.`, I also see `"ERROR user closed connection while waiting for command to complete"` in the saucelabs run log https://saucelabs.com/tests/a9ab99ef884c47bda72d7f4020963783, the appium log for this is @ https://saucelabs.com/tests/a9ab99ef884c47bda72d7f4020963783

We set up the tests with max-duration = 45 minutes, command-timeout and idle-timeout are 10 minutes (https://github.com/pouchdb/pouchdb/blob/master/bin/test-browser.js#L163), while the test is running we run an eval every 10 seconds to check the progress of the tests @ https://github.com/pouchdb/pouchdb/blob/master/bin/test-browser.js#L173

It seems something is dying, iphone only at around 30 minutes, going to run a test that will complete earlier to verify

daleharvey:
So its migration tests that are killing the browser somehow

https://github.com/pouchdb/pouchdb/pull/3477/files

Can skip them easily, would like to figure out whats broken about them as well first though

daleharvey:
Attempted to debug locally, am seeing similiar issues to https://github.com/pouchdb/pouchdb/issues/3382, I have seen the migrations pass sometimes but others I see 1.1.0 just throw up complete

We cant change 1.1.0, so going to just do the simple skip migration patch for this

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/0e619ad8a36e7e6317911d46f5875490be50a04a

PR #3476, opened 2015-01-31
(#3058) - Update iPhone tested, more stable in saucelabs

daleharvey:
Not passing yet, but I did 6 runs in https://travis-ci.org/pouchdb/pouchdb/builds/48996085, the 2 based on 7.1 failed immediately, the 4 based on 8.X failed due to some timeout

8.X is 70% of iOS, so we may or may not want to go back and fix 7.X, but for now just working on getting this green

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,7 +57,7 @@ env:
   - CLIENT=saucelabs:chrome:37 COMMAND=test
   - CLIENT=saucelabs:chrome:39 COMMAND=test
   - CLIENT=saucelabs:safari:6 COMMAND=test
-  - CLIENT="saucelabs:iphone:8.1:OS X 10.10" COMMAND=test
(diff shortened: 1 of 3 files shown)


nolanlawson:
8.x is impossible to run without skipping the migration tests.

daleharvey:
> 8.x is impossible to run without skipping the migration tests.

Do you know why? 1.1.0 looks like it is intermittently crashing in the constructor but it seems to crash reliably when the rest of the suite has run. 

Happy with just getting the tests green though, should be on this commit fingers crossed :)

nolanlawson:
It's impossible because the old idb adapters don't know enough to declare themselves invalid for Apple's IndexedDB. So they try to use Apple's IndexedDB and then fail.

nolanlawson:
:+1:

daleharvey:
Sweet, merged in https://github.com/pouchdb/pouchdb/commit/0e619ad8a36e7e6317911d46f5875490be50a04a

PR #3482, opened 2015-02-01
Improved the example of `allDocs`

kennyki:
When using `attachments: true` it's returning base64-string as data.


daleharvey:
Looks like a good change to me, +1, will let others comment on it

I dont think the base64 needs to be in brackets, "Include attachment data as base64-encoded strings." seems good to me

kennyki:
Sure, no problem.

nolanlawson:
Fixed your commit message on the way in to match CONTRIBUTING.md, thanks for the PR! :)

4bbdaef61168d16e7a450ffbf891432c8cd74eab

Issue #3039, opened 2014-11-24
Improve map/reduce performance


nolanlawson:
I don't necessary love map/reduce, but it's currently our weakest point. And it makes sense to improve perf for two reasons:

1) It's all we got in terms of secondary indexes, for now.
2) Since it's built on `allDocs`/`bulkDocs`, any improvements there will spill over to map/reduce.

Been profiling the `temp-views` test, trying to figure out what's up. This seems to affect all three adapters. ([perf report](https://github.com/nolanlawson/pouchdb-perf-report-3.10))

*Edit:* tracking

* [X] autoCompaction can be done without an explicit doCompaction(). we can just do it in the same transaction as bulkDocs (map/reduce DBs are auto-compacted)
* [ ] doing allDocs(keys) in a single transaction could be beneficial for IndexeDB. I just learned from yathit that sorting the keys and then using a skipping cursor is actually faster than parallel get()s. and for websql of course we can use WHERE id IN (...)
* [X] putLocal could be faster in IndexedDB if we used a unique index rather than keypath, because we can let the database throw ConstraintErrors and catch them rather than doing an explicit get() before the put() *actually we can use add()*
* [ ] IndexedDB might be faster if we tried to merge the doc-store and by-sequence-store into a single store and use string concatenation for keys. then we could use a skipping cursor for allDocs/changes. also for IndexedDB, put()s seem incredibly slow. WebSQL doesn't have bulk insertions. maybe if we make some assumptions about the underlying store, we can improve perf there. (e.g. is it faster if we put() from lowest to greatest keys? need to ask a LevelDB expert)
* [ ] if the user is not emitting values, then map/reduce could do allDocs({include_docs: false}) and avoid fetching from the by-seq store at all. however currently there's no way to tell before we query, so we just query with include_docs: true. could be a way to encode that info in the _id
* [x] one promise queue per view would be nice for people who have >1 view

nolanlawson:
A lot of progress has been made on this (e.g. https://github.com/pouchdb/mapreduce/issues/242 and https://github.com/pouchdb/mapreduce/issues/240). But we're still too slow.

As a real-world example, check out this attempt to create a full-text search index for Magic the Gathering cards: https://github.com/nolanlawson/pouchdb-quick-search/issues/22

Using the WebSQL adapter, it takes 11324ms to do the initial `bulkDocs` and then 227158ms to build the index: http://jsfiddle.net/0kzcd5sh/8/

[...]

PR #3485, opened 2015-02-01
(#3039) - WIP DON'T MERGE - use a bloomfilter for idb

nolanlawson:
When we're doing `bulkDocs`, one of the things that slows us down a lot is checking for existing documents by id. This is especially unfortunate for generation-1 docs, because the docs don't exist and so the `get()` is pointless.

This adds a bloom filter to keep track of doc ids, so that we can bail out early if we know a document has never been inserted before. 

Performance impact (temp-views test, 5 iterations):

    Firefox 37: 52444ms -> 49611ms (5% improvement)
    Chrome 40: 65949ms -> 46770ms (29% improvement)
    IE 11: 57911ms -> 54317ms (6% improvement)

Footprint size change:  46170 before, 47136 after (min+gz)

I can eventually implement this for WebSQL and LevelDB as well. Opening this PR early for feedback.

### DIFF
--- a/tests/integration/test.bulk_docs.js
+++ b/tests/integration/test.bulk_docs.js
@@ -685,6 +685,12 @@ adapters.forEach(function (adapter) {
     });
 
     it('handles simultaneous writes', function (done) {
+      if (testUtils.isCouchMaster()) {
+        // couch might give a 201 for both writes
+        true.should.equal(true);
+        return done();
+      }
+
       var db1 = new PouchDB(dbs.name);
       var db2 = new PouchDB(dbs.name);
       var id = 'fooId';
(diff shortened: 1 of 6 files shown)


tests/integration/test.events.js
```diff
@@ -47,7 +47,7 @@ adapters.forEach(function (adapter) {
     });
 
     it('emit creation event', function (done) {
-      var db = new PouchDB(dbs.name).on('created', function (newDB) {
+      var db = new PouchDB(dbs.name).once('created', function (newDB) {
```
nolanlawson: This was a preexisting racy test, which was revealed to be racy because now IDB takes much longer to start up.

tests/integration/test.bulk_docs.js
```diff
@@ -708,6 +708,30 @@ adapters.forEach(function (adapter) {
...
+        if (++numDone === 2) {
+          errorNames.should.deep.equal(['conflict']);
+          ids.should.deep.equal([true]);
+          done();
+        }
+      }
+      db1.bulkDocs({docs : [{_id : id}]}, callback);
+      db2.bulkDocs({docs : [{_id : id}]}, callback);
+    });
+
```
nolanlawson: ^test added because I am paranoid

nolanlawson:
small fixup

nolanlawson:
Well that's weird. The new test is failing in CouchDB master because apparently you can use two simultaneous `bulkDocs` to PUT the same `_local` doc, and you'll get a 201 for each one. I understand the whole clustering thing, but shouldn't that not apply to `_local` docs @willholley?

If nothing else, even the original test is flakey in CouchDB master, so I can just add an `isCouchMaster()` check to both.

nolanlawson:
added the `isCouchMaster` test.

nolanlawson:
Hm, not improving the perf for WebSQL, which makes me nervous. Also we might want to play around with different cache sizes: 8192 actually gives us a really high false positive rate.

willholley:
@nolanlawson - _local docs do not have MVCC and are not replicated around a
cluster. If you write two copies simultaneously then I think it will be
last write wins (which might not be the same as last request wins). This
actually might cause some strange behaviour in a cluster  (a discussion for
#couchdb-dev) but my understanding is that's how it's worked in Cloudant
forever...

nolanlawson:
Closing because I don't want the extra complexity in exchange for modest perf gains. LevelDB already has a bloom filter, let the browser vendors make IndexedDB faster.

PR #3486, opened 2015-02-01
(#3486) - move parseDoc() to own file

nolanlawson:
This is mostly just code cleanup, but also
partially motivated by performance, because
profiling of map/reduce shows that this function
gets called a lot (once per `emit()` basically).

Below I've noted where I made some micro-optimizations.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -12,6 +12,7 @@ var EventEmitter = require('events').EventEmitter;
 var collections = require('pouchdb-collections');
 exports.Map = collections.Map;
 exports.Set = collections.Set;
+var parseDoc = require('./deps/parse-doc');
 
 if (typeof global.Promise === 'function') {
   exports.Promise = global.Promise;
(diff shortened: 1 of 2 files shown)


lib/utils.js
```diff
@@ -179,103 +122,8 @@ exports.filterChange = function filterChange(opts) {
...
-  var newRevId;
-  var revInfo;
-  var error;
-  var opts = {status: 'available'};
-  if (doc._deleted) {
-    opts.deleted = true;
-  }
-
-  function parseRevisionInfo() {
-    revInfo = /^(\d+)-(.+)$/.exec(doc._rev);
```
nolanlawson: simplified the regex a bit. no need for the capture groups or `(.+)$` to search all the way to the end of the string

lib/utils.js
```diff
@@ -179,103 +122,8 @@ exports.filterChange = function filterChange(opts) {
...
-  var nRevNum;
-  var newRevId;
-  var revInfo;
-  var error;
-  var opts = {status: 'available'};
-  if (doc._deleted) {
-    opts.deleted = true;
-  }
-
-  function parseRevisionInfo() {
```
nolanlawson: ^ functions-within-functions are bad for perf. moved this out

lib/utils.js
```diff
@@ -179,103 +122,8 @@ exports.filterChange = function filterChange(opts) {
...
-    if (doc._revisions) {
-      doc._rev_tree = [{
-        pos: doc._revisions.start - doc._revisions.ids.length + 1,
-        ids: doc._revisions.ids.reduce(function (acc, x) {
-          if (acc === null) {
-            return [x, opts, []];
-          } else {
-            return [x, {status: 'missing'}, [acc]];
-          }
-        }, null)
```
nolanlawson: `reduce()` is bad for perf. replaced with a for-loop.

lib/utils.js
```diff
@@ -179,103 +122,8 @@ exports.filterChange = function filterChange(opts) {
...
-      doc._rev_tree = [{
-        pos: nRevNum,
-        ids: [newRevId, opts, []]
-      }];
-    }
-  }
-
-  exports.invalidIdError(doc._id);
-
-  doc._rev = [nRevNum, newRevId].join('-');
```
nolanlawson: ^ unnecessary array allocation. replaced with string concatenation

nickcolley:
@nolanlawson do you have the perf stats with this change?

nolanlawson:
Just ran 'em. :)

`temp-views` with 5 iterations:

    Chrome: 77057ms -> 67663ms
    Firefox: 49785ms -> 50868ms

Re-ran the tests just to be safe:

    Chrome: 63215ms -> 62547ms
    Firefox: 52308ms -> 49003ms

I didn't expect a big measurable change, but at least we haven't regressed. I think I would need to do many more iterations to actually figure out how much it's improved. (The first Chrome run was probably a fluke I think.)

daleharvey:
This is cleaner and should be marginally faster, cheers - https://github.com/pouchdb/pouchdb/commit/8d0728682fb73d76e3d2198a1bf9bb30f6b3be35

PR #3487, opened 2015-02-02
(#3039) - WIP DON'T MERGE - add bloomfilter to websql

nolanlawson:
This actually causes the `temp-views` test to be a little slower (!). I think it's worth investigating why, so please don't merge this just yet.

`temp-views`, 5 iterations:

    Safari 8: 17541ms -> 18102ms
    Chrome 40: 27151ms -> 29252ms

### DIFF
--- a/lib/adapters/idb/idb-bulk-docs.js
+++ b/lib/adapters/idb/idb-bulk-docs.js
@@ -118,7 +118,13 @@ function idbBulkDocs(req, opts, api, idb, Changes, callback) {
         checkDone(); // skip local docs
         continue;
       }
-      var req = docStore.get(docInfo.metadata.id);
+      var id = docInfo.metadata.id;
+      if (!api._docIdsFilter.test(id)) { // doesn't exist
+        api._docIdsFilter.add(id); // for races with simultaneous bulkDocs
+        checkDone();
+        continue;
+      }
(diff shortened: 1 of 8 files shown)


nolanlawson:
And BTW leveldb actually uses a bloom filter under the hood, and WebSQL/SQLite probably has its own caching mechanisms, but this ought to be a perf boost because we're making fewer context switches from JavaScript to native. Not sure why that's not panning out for WebSQL.

nolanlawson:
Still seeing temp-views being slightly slower, but bulk-inserts is slightly faster:

bulk-inserts (500 iterations)

```
Chrome: 41067ms -> 40955ms
Safari: 29433ms -> 29305ms
```

So overall this seems like a wash, which makes me question why we need it for IndexedDB. I have no idea why a bloomfilter should be such a clear win for IndexedDB but a mixed bag for WebSQL.

nolanlawson:
Hm, well 8192 may be too low to really give us gains in these tests. 1700 true negatives vs 48000 false positives.

nolanlawson:
Mm, even upping the size to where we get 100% true negatives, the test time doesn't improve.

nolanlawson:
see https://github.com/pouchdb/pouchdb/pull/3485#issuecomment-73372841

Issue #3488, opened 2015-02-02
Get rid of manualDescEnd in idb.js


nolanlawson:
Checking the behavior of IDBKeyRange, it seems we can just create a bound where the endkey is first and the startkey is second. Manually checking the end value is not necessary. Chalk it up to me being an IDB newbie when I wrote that code. :stuck_out_tongue:

daleharvey:
Merged in https://github.com/pouchdb/pouchdb/commit/904c812b225ba8250cfd433a5ce74c20354ae030

PR #3489, opened 2015-02-02
(#3488) - get rid of manualDescEnd

nolanlawson:
Seems to be passing in Firefox and Chrome. Better
wait to make sure IE works.

### DIFF
--- a/lib/adapters/idb/idb.js
+++ b/lib/adapters/idb/idb.js
@@ -367,6 +367,35 @@ function init(api, opts, callback) {
     };
   };
 
+  function createKeyRange(start, end, inclusiveEnd, key, descending) {
+    try {
+      if (start && end) {
+        if (descending) {
+          return IDBKeyRange.bound(end, start, !inclusiveEnd, false);
+        } else {
+          return IDBKeyRange.bound(start, end, false, !inclusiveEnd);
+        }
+      } else if (start) {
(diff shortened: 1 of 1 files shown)


lib/adapters/idb/idb.js
```diff
@@ -367,54 +367,56 @@ function init(api, opts, callback) {
     };
   };
 
-  function allDocsQuery(totalRows, opts, callback) {
```
nolanlawson: The diff is a little misleading. I moved a lot of logic into its own function, to avoid try/catch deoptimizations.

nolanlawson:
To be safe, I'd also like to manually run the map/reduce tests on this, so please don't merge until someone has done that.

daleharvey:
Cool, all passing and looks good to me, :+1: for when you are ready

nolanlawson:
wait for this to be green: https://github.com/pouchdb/mapreduce/pull/259

daleharvey:
Gonna just merge this after the release, its a cleanup that shouldnt change behaviour, but just in case it does :)

nolanlawson:
Yup, better safe than sorry.

daleharvey:
ok release done, safe now - https://github.com/pouchdb/pouchdb/commit/904c812b225ba8250cfd433a5ce74c20354ae030

PR #3493, opened 2015-02-03
Be robust about changes entries which don't have documents

jchris:
Sync Gateway occasionally consumes a sequence number and sends a
changes entry that does not correspond to a document. It is safe
to ignore these. This patch prevents them from blowing up filters.

Issue #3136, opened 2014-12-08
changes() with style=all_docs is wrong, causing replication loss


nolanlawson:
(Another follow-up to #3128).

We didn't implement `changes()` with `style=all_docs` correctly, so if a winning rev has a lower `seq` than some other rev, the other rev will fail to get replicated.

I've got a bunch of failing tests but few solutions for now, but I'm working on it.

PR #3495, opened 2015-02-03
[Don't Merge] please review: skip tests / assertions for Sync Gateway

jchris:
With these patches, `test.replication.js` runs to completion against Sync Gateway. This is more of a scoping exercise about what it will take to get to full compatibility. Note that this requires https://github.com/pouchdb/pouchdb/pull/3493 to be merged before it will pass.

The issues that jump out most:

* SG allocates sequence numbers for non-document changes (usually access control changes) which don't change sync behavior but do mess with assertions.
* In certain cases the PouchDB replicator doesn't think that it's completed, leading to tests that time out.
* SG doesn't have a replicator, making the `http-local` and `http-http` tests a little odd. We might add a replicator, in which case they'll make more sense.
* SG doesn't support `open_revs=all` which doesn't matter for sync, but a lot of the assertions need it.
* There are other little things, like the formatting of `all_docs` or the way conflicts are presented in the changes feed, which are probably split evenly between actual bugs, vs things that just break the tests but not actual sync. These areas are the ones that need most investigation.
* I'm curious why `docs_read` and `docs_written` are acting up.

Please take a look at this patch, and imagine similar patches to most of the other test files, followed by a process to slowly back out the skipped tests / assertions to maybe 50% of this amount (by actually fixing bugs). Then tell me if you think it's a good approach or if there's another way to do it that doesn't litter the tests with so many SG-specific conditionals.

It would also help to make sure that this patch doesn't cause havoc to non-SG testing environments, so please run it in Travis just to catch that stuff early. Thanks!

It doesn't apply to this test as much, but most of the failures in the other tests are on the "remote-client" `http` adapter side, so if we cut that out of scope, it would be a small patch to each of those tests to just not run the `http` adapter tests when testing against SG.

### DIFF
--- a/tests/integration/test.replication.js
+++ b/tests/integration/test.replication.js
@@ -68,9 +68,16 @@ adapters.forEach(function (adapters) {
 
     function verifyInfo(info, expected) {
       if (!testUtils.isCouchMaster()) {
-        info.update_seq.should.equal(expected.update_seq, 'update_seq');
+        if (typeof info.doc_count === 'undefined') { 
+          // info is from Sync Gateway, which allocates an extra seqnum
+          // for user access control purposes.
+          info.update_seq.should.be.within(expected.update_seq, 
+            expected.update_seq + 1, 'update_seq');
+        } else {
+          info.update_seq.should.equal(expected.update_seq, 'update_seq');
+        }
       }
-      if (!testUtils.isSyncGateway()) {
+      if (info.doc_count) { // info is NOT from Sync Gateway
         info.doc_count.should.equal(expected.doc_count, 'doc_count');
       }
     }
@@ -409,12 +416,14 @@ adapters.forEach(function (adapters) {
                 if (!testUtils.isCouchMaster()) {
                   info.update_seq.should.be.above(2, 'update_seq local');
                 }
-                info.doc_count.should.equal(3, 'doc_count local');
+                if (info.doc_count) {
+                  info.doc_count.should.equal(3, 'doc_count local');
+                }
                 remote.info(function (err, info) {
                   if (!testUtils.isCouchMaster()) {
                     info.update_seq.should.be.above(2, 'update_seq remote');
                   }
-                  if (!testUtils.isSyncGateway()) {
+                  if (info.doc_count) {
                     info.doc_count.should.equal(3, 'doc_count remote');
                   }
                   done();
@@ -553,7 +562,9 @@ adapters.forEach(function (adapters) {
           }
           db.replicate.to(dbs.remote, {
             complete: function (err, details) {
-              details.docs_read.should.equal(0); // SG bug
+              if (!testUtils.isSyncGateway()) {
+                details.docs_read.should.equal(0);
+              }
               db.info(function (err, info) {
                 verifyInfo(info, {
                   update_seq: 3,
@@ -628,6 +639,10 @@ adapters.forEach(function (adapters) {
             doc._rev = results.rev;
             doc.count++;
             db.put(doc, {}, function (err, results) {
+              // console.log("rep", db, remote);
+              if (testUtils.isSyncGateway() && db.adapter === 'http') {
+                return done();
+              }
               PouchDB.replicate(db, remote, {}, function (err, result) {
                 result.ok.should.equal(true);
                 result.docs_written.should.equal(1);
@@ -885,7 +900,9 @@ adapters.forEach(function (adapters) {
             return db2.changes({style: 'all_docs'});
           }).then(function (res2) {
             changes2 = simplifyChanges(res2);
-
+            if (testUtils.isSyncGateway() && adapters[0] === 'http') {
+              return true;
+            }
             changes1.should.deep.equal(changes2, 'same changes');
           });
         });
@@ -1090,7 +1107,9 @@ adapters.forEach(function (adapters) {
         return PouchDB.replicate(db, remote);
       }).then(function (result) {
         result.ok.should.equal(true);
-        result.docs_written.should.equal(1);
+        if (!(testUtils.isSyncGateway() && adapters[0] === 'http')) {
+          result.docs_written.should.equal(1);          
+        }
         db.info(function (err, info) {
           verifyInfo(info, {
             update_seq: 3,
@@ -1104,8 +1123,8 @@ adapters.forEach(function (adapters) {
     });
 
     it('Testing allDocs with some conflicts (issue #468)', function (done) {
-      if (testUtils.isSyncGateway()) {
-        // something about update-seq
+      if (testUtils.isSyncGateway() && adapters[0] === 'http') {
+        // looks interesting
         return done();
       }
       var db1 = new PouchDB(dbs.name);
@@ -1152,10 +1171,12 @@ adapters.forEach(function (adapters) {
                                   // if auto_compaction is enabled, will
                                   // be 5 because 2-c goes "missing" and
                                   // the other db tries to re-put it
-                                  if (!testUtils.isCouchMaster()) {
-                                    info.update_seq.should.be.within(4, 5);
+                                  if (!testUtils.isSyncGateway()) {
+                                    if (!testUtils.isCouchMaster()) {
+                                      info.update_seq.should.be.within(4, 5);
+                                    }
+                                    info.doc_count.should.equal(1);
                                   }
-                                  info.doc_count.should.equal(1);
                                   db2.info(function (err, info2) {
                                     verifyInfo(info2, {
                                       update_seq: 3,
@@ -1191,7 +1212,9 @@ adapters.forEach(function (adapters) {
         remote.put(doc2, function (err, remoteres) {
           db.replicate.to(dbs.remote, function (err, _) {
             remote.get('adoc', { conflicts: true }, function (err, result) {
-              result.should.have.property('_conflicts');
+              if (!testUtils.isSyncGateway()) {
+                result.should.have.property('_conflicts');
+              }
               db.info(function (err, info) {
                 verifyInfo(info, {
                   update_seq: 1,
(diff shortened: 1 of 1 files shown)


daleharvey:
> SG doesn't have a replicator, making the http-local and http-http tests a little odd. We might add a replicator, in which case they'll make more sense.

We dont expect or use a replicator on the adapter, Pouch handles all of its own replication so the adapters only need to expose enough api to be able to replicate.

> SG doesn't support open_revs=all which doesn't matter for sync, but a lot of the assertions need it.

As discussed in the other bug we use it for replication, it may be possible to patch our replicator to not use it, however as far as I can tell it really does look like CouchDB uses it for replication as well.

I am totally fine with getting the tests against csg running, liberally skipping the failing ones then going back and fixing other individually either on the server or in pouch (or in the tests). Thats the exact process we are going through with couch master right now

daleharvey:
One issue is, is it possible to sync CouchDB (1.6 or 2.0) with CSG / Couchbase? is that tested anywhere? looking at a few of the incompatibilities its seeming like it shouldnt really work

tests/integration/test.replication.js
```diff
@@ -217,6 +232,9 @@ adapters.forEach(function (adapters) {
         result.docs_written.should.equal(0, 'correct # docs written (1)');
         return db.info();
       }).then(function (info) {
+        if (testUtils.isSyncGateway()) {
+          return [{"ok": {"_id": "doc", "_rev": doc1._rev}}];
+        }
```
nolanlawson: ^ I think this is a case where I would prefer to smooth out the differences in `http.js`. We already do this for Cloudant; they don't actually return `{ok: true}`, whereas I like to be reassured that everything is ok. :)

tests/integration/test.replication.js
```diff
@@ -229,6 +247,12 @@ adapters.forEach(function (adapters) {
         result.docs_written.should.equal(0, 'correct # docs written (2)');
         return db.info();
       }).then(function (info) {
+        if (testUtils.isSyncGateway()) {
+         return [
+            {"ok": {"_id": "doc", "_rev": doc1._rev}},
+            {"ok": {"_id": "doc", "_rev": doc2._rev}}
+          ];
+        }
```
nolanlawson: ditto

tests/integration/test.replication.js
```diff
@@ -293,13 +321,21 @@ adapters.forEach(function (adapters) {
     });
 
     it('Test pull replication with many conflicts', function (done) {
+      if (testUtils.isSyncGateway()) {
+        return done();
+      }
       var remote = new PouchDB(dbs.remote);
 
-      var numRevs = 200; // repro "url too long" error with open_revs
+      var numRevs = 10; // repro "url too long" error with open_revs
```
nolanlawson: So the point of the 200 here is to catch a browser error where the URL is too long. Can't recall which browsers throw exactly, but I believe it's IE and Safari.

tests/integration/test.replication.js
```diff
@@ -293,13 +321,21 @@ adapters.forEach(function (adapters) {
...
       for (var i = 0; i < numRevs; i++) {
-        var rev =  '1-' + PouchDB.utils.uuid(32, 16).toLowerCase();
-        docs.push({_id: 'doc', _rev: rev});
+        // var rev =  '1-' + PouchDB.utils.uuid(32, 16).toLowerCase();
+        docs.push({
+          _id: 'doc', 
+          _revisions: {
+            start : 1, ids : [PouchDB.utils.uuid(32, 16).toLowerCase()]
+          }
+        });
```
nolanlawson: Let's go ahead and include both `_rev` and `_revisions` when we do `bulkDocs`; the behavior is undefined otherwise, and I know PouchDB handles it weirdly compared to CouchDB. Much clearer if we just add both.

tests/integration/test.replication.js
```diff
@@ -57,14 +57,29 @@ adapters.forEach(function (adapters) {
...
+
       return changes;
     }
 
     function verifyInfo(info, expected) {
       if (!testUtils.isCouchMaster()) {
-        info.update_seq.should.equal(expected.update_seq, 'update_seq');
+        if (typeof info.doc_count === 'undefined') { 
+          // info is from Sync Gateway, which allocates an extra seqnum
+          // for user access control purposes.
```
nolanlawson: ^ let's use `isSyncGateway()` then to be clearer. `should.be.within` is a fine test.

tests/integration/test.replication.js
```diff
@@ -377,12 +416,16 @@ adapters.forEach(function (adapters) {
...
+                  info.doc_count.should.equal(3, 'doc_count local');
+                }
                 remote.info(function (err, info) {
                   if (!testUtils.isCouchMaster()) {
                     info.update_seq.should.be.above(2, 'update_seq remote');
                   }
-                  info.doc_count.should.equal(3, 'doc_count remote');
+                  if (info.doc_count) {
+                    info.doc_count.should.equal(3, 'doc_count remote');
+                  }
```
nolanlawson: again, `isSyncGateway` would make it clearer

tests/integration/test.replication.js
```diff
@@ -839,7 +900,9 @@ adapters.forEach(function (adapters) {
             return db2.changes({style: 'all_docs'});
           }).then(function (res2) {
             changes2 = simplifyChanges(res2);
-
+            if (testUtils.isSyncGateway() && adapters[0] === 'http') {
+              return true;
+            }
```
nolanlawson: Surprised that this one fails. this was a really tricky bug that we managed to fix recently (https://github.com/pouchdb/pouchdb/pull/3165 #3136). Maybe should check to make sure this isn't a bug in CSG?

tests/integration/test.replication.js
```diff
@@ -3489,6 +3714,10 @@ adapters.forEach(function (adapters) {
     });
 
     it('#3270 triggers "change" events with .docs property', function(done) {
+      if (testUtils.isSyncGateway() && adapters[0] === 'http') {
+        // SG doesn't have a replicator
+        return done();
+      }
```
nolanlawson: The rest of these look good; seems to be testing very CouchDB-specific auth/validation/ddoc stuff.

nolanlawson:
> One issue is, is it possible to sync CouchDB (1.6 or 2.0) with CSG / Couchbase? is that tested anywhere? looking at a few of the incompatibilities its seeming like it shouldnt really work

I brought this up on IRC, but I started working on this thing: https://github.com/nolanlawson/couch-community-compat-table. Please add sync pairs to the list and link to the docs/tests/issues.

Also I'm +1 for adding @jchris to the pouchdb org; it would enable him to actually run Saucelabs tests and see if any tests are failing in exotic browsers, or to re-run a false negative.

nolanlawson:
Also @jchris would you be interested in PRs to CSG for stuff like open_revs=all or even JSON attachments?

daleharvey:
> I brought this up on IRC, but I started working on this thing: https://github.com/nolanlawson/couch-community-compat-table. Please add sync pairs to the list and link to the docs/tests/issues.

Yeh getting some coordination on this will be awesome.

> Also I'm +1 for adding @jchris to the pouchdb org; it would enable him to actually run Saucelabs tests and see if any tests are failing in exotic browsers, or to re-run a false negative.

Done

jchris:
Thanks y'all I'll try to reply in detail soon!

On Thursday, February 5, 2015, Dale Harvey <notifications@github.com> wrote:

> I brought this up on IRC, but I started working on this thing:
> https://github.com/nolanlawson/couch-community-compat-table. Please add
> sync pairs to the list and link to the docs/tests/issues.
>
> Yeh getting some coordination on this will be awesome.
>
> Also I'm +1 for adding @jchris <https://github.com/jchris> to the pouchdb
> org; it would enable him to actually run Saucelabs tests and see if any
> tests are failing in exotic browsers, or to re-run a false negative.
>
> Done
>
> —
> <https://github.com/pouchdb/pouchdb/pull/3495#issuecomment-73108817>.
>


-- 
Chris Anderson
http://www.couchbase.com
http://twitter.com/jchris

tests/integration/test.replication.js
jchris: This patch is about running as many assertions as possible without actually touching implementation unless I have to. So my goal is to build a process around cutting down the number of `isSyncGateway` calls over time. I agree this is an important one to work on

tests/integration/test.replication.js
jchris: I forgot I changed that, I'll put it back and see what happens.

tests/integration/test.replication.js
jchris: Cool idea. What I see in general is that PouchDB mostly works with Sync Gateway for sync, but some of the tests that manipulate SG are failing for things like this. So I'm planning to skip the full 'http' adapter api in many of the other test files. We can address PouchDB as a Sync Gateway remote client in another pass.

tests/integration/test.replication.js
jchris: This has different semantics. If I use isSyncGateway everywhere it skips too many assertions. Here I'm detecting the case where the doc count isn't passed so even when it's sync gateway I still verify the doc count PouchDB gives, since `verifyInfo` is used both for local and remote responses.

tests/integration/test.replication.js
jchris: The reason I'm skipping HTTP in this case is because there's stuff worth investigating behind that flag. Thanks for reminding me.

jchris:
@nolanlawson  `open_revs=all` would be a good start, maybe here: https://github.com/couchbase/sync_gateway/blob/ed8e9fb7c04202798a5b71d24aa3c252eece1c65/src/github.com/couchbaselabs/sync_gateway/rest/doc_api.go#L67

tests/integration/test.replication.js
nolanlawson: ah okay, missed that

nolanlawson:
Whoa, I didn't realize CSG was written in Go. That's really cool.

jchris:
I'm taking a different approach with test by test pull requests and making changes on the SG side when needed. I'm making progress:

more better

Issue #3505, opened 2015-02-05
There was not enough remaining storage space in Cordova App


ScreenDriver:
I am using Apache Cordova to build my app. I installed PouchDB and, as explained [here],(http://pouchdb.com/adapters.html) added the SQLite Plugin.
New instances of PouchDB will be created with 

```javascript
new PouchDB('myDB', {adapter: 'websql', size: 10});
```

Unfortunately I get a error if I made some batch (bulk) inserts:
 
```"there was not enough remaining storage space, or the storage quota was reached and the user declined to allow more space"```

I can increase the limit how much I like, it seems to be ignored. I've tested it with a some mock JSON objects with the same size. The first inserts run fine and then when I reach the limit I get this error message. If I increase the size, nothing happens and I still get immediately this exception.

Could it be that the size parameter will be ignored? Or does PouchDB don't use the natively installed SQLite Plugin?

**Comment by nolanlawson** [comment]
So it sounds like PouchDB is not actually using the SQLite plugin. You should never see that kind of error if you are using the SQLite plugin.

Another thing is that the `size` parameter is totally ignored if you are using the SQLite plugin. It's only for WebSQL.

Unfortunately it's kind of hard to tell what PouchDB is actually using (open issue somewhere...), but if you use the non-minified code, and then search for `sqlitePlugin`, and set a breakpoint, you should be able to see if PouchDB is actually using the SQLite Plugin. Another way is to check that `window.sqlitePlugin` is actually defined **before** you load PouchDB (so be sure to include the plugin in your `index.html` first).

Another tool: I just created a [PouchDB Cordova guide](https://github.com/nolanlawson/pouchdb-phonegap-cordova) which contains a sample app that actually uses the SQLite plugin (I checked!). So you can compare your code against that and see if it's working.

**Comment by ScreenDriver** [comment]
Thank you for your help but I don't get it. I did everything explained [here](https://github.com/nolanlawson/pouchdb-phonegap-cordova) but still the same.
 
```window.sqlitePlugin``` is defined **before** I use PouchDB

the script order is ok
```html
<script src="cordova.js"></script>
<script src="bower_components/pouchdb/dist/pouchdb.js"></script>
```
I added following code
```javascript
var openDBFunction = (typeof navigator !== 'undefined' &&
      navigator.sqlitePlugin &&
      navigator.sqlitePlugin.openDatabase) ?
    navigator.sqlitePlugin.openDatabase.bind(navigator.sqlitePlugin) :
[...]

PR #3509, opened 2015-02-07
(#3505) - fix sqlite plugin support

nolanlawson:
My goal: to get PouchDB to use the SQLite Plugin without any manual intervention from the user.

I succeeded! There were just a few things to change:

* `window.sqlitePlugin` is not declared until after Cordova's `deviceready` event has fired, so we cannot use it to detect `valid()`. (I.e. because of Windows Phone 8, which has sqlitePlugin support but not `openDatabase`.) But luckily the SQLitePlugin leaks a global `SQLitePlugin` variable, so we can use that to detect SQLitePlugin support. Flakey, but it fixes Windows Phone 8.
* the `openDB` function has to be re-generated every time a new database is created, because the global variables might have changed post-`deviceready`, in which case we want to use the SQLite plugin instead of `openDatabase`.

So basically what this requires users to do is to not call `new PouchDB()` until after the `deviceready` event (which they are probably accustomed to doing. I verified in [my sample app](https://github.com/nolanlawson/pouchdb-cordova-hello-world-with-sqlite-plugin) that it was actually working and actually using the SQLite Plugin.

### DIFF
--- a/lib/adapters/websql/websql.js
+++ b/lib/adapters/websql/websql.js
@@ -24,6 +24,7 @@ var select = websqlUtils.select;
 var compactRevs = websqlUtils.compactRevs;
 var unknownError = websqlUtils.unknownError;
 var getSize = websqlUtils.getSize;
+var openDB = websqlUtils.openDB;
 
 function fetchAttachmentsIfNecessary(doc, opts, api, txn, cb) {
   var attachments = Object.keys(doc._attachments || {});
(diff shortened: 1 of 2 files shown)


lib/adapters/websql/websql.js
```diff
@@ -60,28 +61,6 @@ function fetchAttachmentsIfNecessary(doc, opts, api, txn, cb) {
   });
 }
 
-var cachedDatabases = {};
-
-var sqlitePluginOpenDBFunction =
-  ((typeof navigator !== 'undefined' &&
-    navigator.sqlitePlugin &&
-    navigator.sqlitePlugin.openDatabase) &&
-    navigator.sqlitePlugin.openDatabase.bind(navigator.sqlitePlugin)) ||
```
nolanlawson: This `navigator.sqlitePlugin` stuff seems to date from an older version of the SQLite Plugin. I'm happy to just require that users use the latest vesion of the SQLite Plugin.

daleharvey:
Looks great, cheers

https://github.com/pouchdb/pouchdb/commit/7a7ee7d1dcad2ddfb79baaf4bc00b7d3551df8c4

PR #3510, opened 2015-02-07
(#3510) - update docs with true size

nolanlawson:
We are 46166 bytes min+gz as of 3.3.0.


daleharvey:
Now we need to get the size back down again :) cheers, https://github.com/pouchdb/pouchdb/commit/3b7f3f2de4aec81178fa5d85ec08c4c1a10aa0f1

Issue #3179, opened 2014-12-14
Removing conflicting revision doesn't sync from local to remote


cschwarz:
It works from remote to local.

I'm opening a PR with a failing test replicating the issue.

cschwarz:
I have updated pouchdb to 3.2.1 and still have the same problem, even though the test works fine now. I could not reproduce it in a unit test, but I have isolated the issue in a jsbin.

http://jsbin.com/fahigagumu/1/

Maybe it is also a mistake in the way I resolve the conflict.

nolanlawson:
OK, so it looks like we've finally found a bug due to our non-deterministic rev-generation. I have some tests that reproduce the error.

nolanlawson:
Actually, there are two issues at play here: 1) live sync, which is breaking where non-live sync succeeds, and 2) duplicate docs, which are hashed to the same rev in CouchDB but not PouchDB.

nolanlawson:
Found a fix.

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/572539869b1595e579c53c71a7f55b0cfa27cdb2

PR #3511, opened 2015-02-07
(#3179) - fix live replication of conflicts

nolanlawson:
This was one of those one-line bugfixes that
took me a few hours to figure out. :P

The tests fail before the fix but succeed after.
Pretty straightforward.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -192,6 +192,7 @@ Changes.prototype.addListener = function (dbName, id, db, opts) {
     }
     inprogress = true;
     db.changes({
+      style: opts.style,
       include_docs: opts.include_docs,
       attachments: opts.attachments,
       conflicts: opts.conflicts,
--- a/tests/integration/test.issue3179.js
+++ b/tests/integration/test.issue3179.js
@@ -7,6 +7,10 @@ var adapters = [
   ['local', 'local']
 ];
 
+if ('saucelabs' in testUtils.params()) {
+  adapters = [['local', 'http'], ['http', 'local']];
+}
+
 adapters.forEach(function (adapters) {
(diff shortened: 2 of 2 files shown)


nolanlawson:
So here's what happened:

We weren't correctly passing `style=all_docs` to the changes API for *live* changes (non-live changes were fine). This meant that, in the case of non-winning revisions like e.g. deleted conflicts, the revs weren't being mentioned in the `changes` list on the emitted `change` object because hey, they weren't the winning revision, and it wasn't `style=all_docs`, so non-winning leafs could be omitted. So the target database would never get notified about later non-winning revisions.

In those tests, there were 8 variations (live/nonlive, plus all the local/http pairs), but only 4 were failing: non-winning and local-originating (i.e. local-local and local-http).

nolanlawson:
increased the timeout to 10 seconds because it seemed faily in travis

nolanlawson:
The tests are faily. @daleharvey do you have any suggestions? It all works fine locally; can't figure out how to fix this without `setTimeout`... :/

daleharvey:
https://github.com/pouchdb/pouchdb/pull/3511/files#diff-070382d6862d30174d29271e33abbfeeR180 looks suspect, cancelling replication isnt synchronous, needs to wait for a `.on('complete'` before returning from the test

nolanlawson:
third attempt

nolanlawson:
green! and not faily at all; it only failed in one test (couchdb-master), so I re-ran it.

daleharvey:
That is a good amount of tests for a 1 line patch :) we need more tests for this stuff so this is awesome, thanks

https://github.com/pouchdb/pouchdb/commit/572539869b1595e579c53c71a7f55b0cfa27cdb2

Issue #2806, opened 2014-09-28
Web Workers and Service Workers


nolanlawson:
PouchDB currently does not work within a Web Worker or Service Worker thread across all browsers.

This is not something we can really solve, because the browser bugs are kinda unfixable from our end. But they're worth tracking.

| Browser | Status | Issue | Details |
| ---- | ---- | --- | ---- |
| Chrome | :heavy_check_mark:   | | |
| Chrome + WebSQL | :heavy_check_mark:/:x: | [434740](https://code.google.com/p/chromium/issues/detail?id=434740) | Retired in v39 |
| Firefox | :heavy_check_mark: | [701634](https://bugzilla.mozilla.org/show_bug.cgi?id=701634) | Fixed in v37 |
| IE | :x: | [866495](https://connect.microsoft.com/IE/feedbackdetail/view/866495) | No known workaround |
| Safari | :heavy_check_mark:/:x: | [434740](https://code.google.com/p/chromium/issues/detail?id=434740) | To be retired circa Safari v9 |

**Edit:** updated 2015-03 to note Chrome/Safari deprecations and Firefox fixes.

AdrianoFerrari:
Does anyone know the status of this?

As far as I can tell, Firefox now supports idb in workers (https://bugzilla.mozilla.org/show_bug.cgi?id=701634 marked as RESOLVED).

nolanlawson:
Apparently WebSQL worked before, but will soon not work in both Blink and Webkit: https://code.google.com/p/chromium/issues/detail?id=434740

nolanlawson:
I updated the table above. Looks like the Firefox fix will land in v37!

daleharvey:
Firefox and Chrome are good, ie and safari are plain unsupported at the moment so closing out

nolanlawson:
The readme for worker-pouch also clarifies these incompatibilities: https://github.com/nolanlawson/worker-pouch

swapnil0545:
Hi, I am using ydn-db library which is similar to pouch db.
We were using websql inside the webworker and it seems to not working since the Safari 9 update.
I can see that chrome has already dropped it support for websql inside webworker.
But I do not see any official apple/safari link saying that.
The indexedDB apis in Safari 8 were buggy, do we have a buglist which are there for Safari 9 indexedDB apis.
Any issues the pouch db users are facing for websql inside webworkers.

nolanlawson:
[...]

PR #3512, opened 2015-02-08
Worker blob attachment

JakeChampion:
## Problem:
When inside of a web worker and adding a blob attachment, pouchDB assumes `btoa` does not exist globally and so tries to user `Buffer` which is available in Node/IO.

## Solution:
Check if `btoa` exists in the global scope first and if so, use it.
Align `atob` logic along with the logic for `btoa` to avoid using the polyfill.

## Warning: 
This PR also edits the version of the debug module to pull from the master branch on github. This was to get the tests for web working running.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -246,7 +246,11 @@ Changes.prototype.notify = function (dbName) {
   this.notifyLocalWindows(dbName);
 };
 
-if (typeof window === 'undefined' || typeof window.atob !== 'function') {
+if (typeof atob === 'function' || typeof window !== 'undefined') {
+  exports.atob = function (str) {
+    return atob(str);
+  };
+} else {
   exports.atob = function (str) {
     var base64 = new buffer(str, 'base64');
     // Node.js will just skip the characters it can't encode instead of
@@ -255,10 +259,6 @@ if (typeof window === 'undefined' || typeof window.atob !== 'function') {
       throw ("Cannot base64 encode full string");
     }
     return base64.toString('binary');
-  };
-} else {
-  exports.atob = function (str) {
-    return atob(str);
   };
 }
 
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -246,7 +246,7 @@ Changes.prototype.notify = function (dbName) {
   this.notifyLocalWindows(dbName);
 };
 
-if (typeof atob === 'function' || typeof window !== 'undefined') {
+if (typeof atob === 'function') {
   exports.atob = atob;
 } else {
   exports.atob = function (str) {
@@ -260,7 +260,7 @@ if (typeof atob === 'function' || typeof window !== 'undefined') {
   };
 }
 
-if (typeof btoa === 'function' || typeof window !== 'undefined') {
+if (typeof btoa === 'function') {
   exports.btoa = btoa;
 } else {
   exports.btoa = function (str) {
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -247,7 +247,9 @@ Changes.prototype.notify = function (dbName) {
 };
 
 if (typeof atob === 'function') {
-  exports.atob = atob;
+  exports.atob = function (str) {
+    return atob(str);
+  };
 } else {
   exports.atob = function (str) {
     var base64 = new buffer(str, 'base64');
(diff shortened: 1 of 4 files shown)


daleharvey:
Awesome, is https://github.com/visionmedia/debug/commit/988592b640bf10faac2d8ebafa3135c0f3351123 the commit we need in debug? I think we should ask them for a release if so, would prefer to avoid having a git dep

This is failing on firefox, we test firefox 31 in travis, its ok if they dont work, but need to be skipped properly

daleharvey:
I think a good approach would to just do an if !/chrome/ skip all the tests so get the tests enabled at least for a single browser, then try to backfill the other browsers individually @nolanlawson will likely know more about this stuff

JakeChampion:
@daleharvey That sounds like a good solution for now, would help avoid future regressions with web workers in one browser at least :-)

Also @nolanlawson has already made a request for a new release of the debug module: https://github.com/visionmedia/debug/issues/178

JakeChampion:
@daleharvey I added a check to run the tests only in chromium browsers. Also, 3 of the tests pass in Firefox Stable, should I enable them to run on CI?

nolanlawson:
It's not clear to me which version of Firefox Travis uses, so they might fail. But you can give it a shot. :)

lib/utils.js
```diff
@@ -246,7 +246,11 @@ Changes.prototype.notify = function (dbName) {
   this.notifyLocalWindows(dbName);
 };
 
-if (typeof window === 'undefined' || typeof window.atob !== 'function') {
+if (typeof atob === 'function' || typeof window !== 'undefined') {
```
nolanlawson: I think the `typeof window !== 'undefined'` is unnecessary here

lib/utils.js
```diff
@@ -256,19 +260,15 @@ if (typeof window === 'undefined' || typeof window.atob !== 'function') {
...
     return base64.toString('binary');
   };
-} else {
-  exports.atob = function (str) {
-    return atob(str);
-  };
 }
 
-if (typeof window === 'undefined' || typeof window.btoa !== 'function') {
+if (typeof btoa === 'function' || typeof window !== 'undefined') {
```
nolanlawson: ditto

package.json
```diff
@@ -16,12 +16,12 @@
     "pouchdb"
   ],
   "dependencies": {
-    "pouchdb-upsert": "^1.0.2",
-    "pouchdb-collections": "^1.0.0",
     "argsarray": "0.0.1",
     "bluebird": "^1.2.4",
+    "debug": "git://github.com/visionmedia/debug.git",
```
nolanlawson: having to do this is annoying, but I'd prefer not to be held up by another project

nolanlawson:
After those few little nits, I'm fine with merging. `window.chrome` is a good test for now; we can add Firefox later if it passes in Travis, ditto for IE and Safari. (see https://github.com/pouchdb/pouchdb/issues/2806 for all the gory details)

lib/utils.js
```diff
@@ -246,7 +246,11 @@ Changes.prototype.notify = function (dbName) {
   this.notifyLocalWindows(dbName);
 };
 
-if (typeof window === 'undefined' || typeof window.atob !== 'function') {
+if (typeof atob === 'function' || typeof window !== 'undefined') {
+  exports.atob = function (str) {
+    return atob(str);
+  };
```
nolanlawson: also this can be simplified to `exports.atob = atob;` feel dumb for missing that :P

JakeChampion:
@nolanlawson The test actually fails if I do `exports.btoa = btoa` with a "Uncaught TypeError: Illegal invocation" message.

nolanlawson:
oh interesting, didn't realize that. okay, no prob then. :)

nolanlawson:
+1 when green, although unfortunately it will need another PR from a collaborator due to non-collabs not getting chrome tested in saucelabs >_<

daleharvey:
Full tests are passing, I filed a new bug to revert back to using release version of debug (https://github.com/pouchdb/pouchdb/issues/3517), squashed the commits

Thanks a lot @JakeChampion, merged - https://github.com/pouchdb/pouchdb/commit/490e8fc11b418d7854d3f0f697f4f2620312bd4b

PR #3521, opened 2015-02-10
run Sync Gateway on Travis

jchris:
The `COUCH_HOST` question is not part of this patch so I'll remove it before this is ready to merge.

Will my making this pull request automatically start Travis trying to run Sync Gateway?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ before_script:
   - "sudo sed -i 's/^127\\.0\\.0\\.1.*$/127.0.0.1 localhost/' /etc/hosts"
 
     # Lets know what CouchDB we are playing with
-  - "curl -X GET http://127.0.0.1:5984/"
+  - "curl -X GET http://127.0.0.1:5984/" # should this use COUCH_HOST variable?
 
 
 script: npm run $COMMAND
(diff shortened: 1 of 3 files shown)


nolanlawson:
Yes, it should! If you want to save time while you iterate, you can also make a travis file that *only* runs CSG, just to check if it's green.

And yes, your comment is correct; it should probably use the COUCH_HOST variable.

BTW this is super exciting. :)

jchris:
do you mean .travis-jchris.yml or just alter the existing travis file for my testing pull reqs? I think I will also start with a GREP in the command.

.travis.yml
```diff
@@ -20,7 +20,7 @@ before_script:
   - "sudo sed -i 's/^127\\.0\\.0\\.1.*$/127.0.0.1 localhost/' /etc/hosts"
 
     # Lets know what CouchDB we are playing with
-  - "curl -X GET http://127.0.0.1:5984/"
+  - "curl -X GET http://127.0.0.1:5984/" # should this use COUCH_HOST variable?
```
daleharvey: COUCH_HOST is defined as part of the test scripts so wouldnt be picked up here, this was mostly for debugging previous issues so not entirely needed anymore

daleharvey:
I think @nolanlawson means while you are testing the PR's (not one preparing to be merged) you can take out the rest of the tests and only run the CSG ones so you can see the results without waiting for every test to complete.

This looks great, can you add the csg run to the expected failures and then we can get it merged

jchris:
How do I add it to expected failures?

daleharvey:
https://github.com/pouchdb/pouchdb/commit/a46601208372b5bbab95b8c5a19984b3f7f71775 is where  I took out some expected failures, just need to readd that section with csg in there, then csg can be failing but we still get a green build

jchris:
closing in favor of #3526

mrjumpy:
```
Couchbase Sync Gateway support is in progress. It will work, but you may run into issues, especially with attachments. Drupal 8 has also announced support for PouchDB, and there is rcouch as well, but these are both untested by PouchDB.

If you are ever unsure about a server, consider replicating from PouchDB to CouchDB, then from that CouchDB to the other server.```

PR #3522, opened 2015-02-10
allow _security docs to be put w/out an error

slang800:
previously, I was getting a "Only reserved document ids may start with underscore." error whenever I tried to `put` a `_security` doc

### DIFF
--- a/lib/deps/parse-doc.js
+++ b/lib/deps/parse-doc.js
@@ -53,7 +53,7 @@ exports.invalidIdError = function (id) {
     err = errors.error(errors.MISSING_ID);
   } else if (typeof id !== 'string') {
     err = errors.error(errors.INVALID_ID);
-  } else if (/^_/.test(id) && !(/^_(design|local)/).test(id)) {
+  } else if (/^_/.test(id) && !(/^_(design|local|security)/).test(id)) {
     err = errors.error(errors.RESERVED_ID);
   }
   if (err) {


nolanlawson:
LGTM, but just for my own edification, what's a security doc? I see that there's a ["security object"](http://docs.couchdb.org/en/latest/api/database/security.html), but it doesn't seem to be the same as a document that can start with the prefix `_security/`.

marten-de-vries:
@nolanlawson A security object defines what rights you have in a CouchDB database. Who's allowed to access, write, and delete it? To support it properly, you need an authentication system and a bunch of extra logic. I approximately implemented that in my [pouchdb-security](https://www.npmjs.com/package/pouchdb-security) plug-in.

That said, security documents are NOT regular CouchDB documents for performance reasons. (No revisions e.g., it goes even further than local docs IIRC). It doesn't make much sense to treat them as such as this PR does. pouchdb-security instead adds custom getSecurity()/putSecurity() methods, which internally use the document _local/_security.

So: I'm currently -1 on this.

slang800:
What is the extra logic that's needed? (the docs linked to in that npm page don't seem to mention pouchdb-security, and the plugin doesn't seem to be on your github account)... Couchdb already handles authentication internally, and pouchdb allows you to authenticate when you call the `new PouchDb()` constructor. All I'm doing with this  is:

```
db.put(
  _id: '_security'
  admins:
    names: []
    roles: []
  members:
    names: []
    roles: ['admin']
)
```

...and that seems to work w/out any extra logic.

slang800:
Oh, I should elaborate: I'm working with bootstrapping a remote CouchDB instance, not trying to add `_security` docs to the local PouchDB database. Maybe that's part of the confusion.

marten-de-vries:
My main problem with this is that it does make it possible to store a _security doc in a local database. That doc will follow normal doc conventions (like having conflict management, _id, _rev, etc.), which is unexpected (because in CouchDB, it doesn't). Also, storing such a document offline doesn't actually change access restrictions like in CouchDB, which is also unexpected.

The fact that your code example works is in fact a coincidence: if we reimplemented .put() to call _bulk_docs for the http adapter, it wouldn't work anymore e.g., while other things (probably) wouldn't be affected.

pouchdb-security does provide more CouchDB-like conventions (including 'own' methods, .putSecurity() and .getSecurity()). Its code [is here](https://bazaar.launchpad.net/~marten-de-vries/python-pouchdb/0.x/view/head:/js/pouchdb-security/index.js). But, if pouchdb-security is too heavy for your needs, something like this would be a possibility too (uses undocumented API):

````javascript
db.request({
  method: 'PUT',
  url: '_session',  
  body: {...}
}).then(...);
```

Or just a plain equivalent XHR/request call.

nolanlawson:
I agree with @marten-de-vries; this should be a plugin or an explicit `db.request` call.

By the way, `db.request` is not documented, but it should be. This is the 2nd time in a week this has come up. :P

nolanlawson:
I think the other issue here is that it would be really weird if you put a `_security` doc in PouchDB and it tried to sync that doc to CouchDB. (Which would fail due to the `_bulk_docs` issue cited above.) Another reason to make this an explicit HTTP request.

slang800:
yeah, that works - thanks @marten-de-vries

Issue #3528, opened 2015-02-12
Memory and Localstorage adapters don't work


corps:
Hey guys,

Ever since about 3.1, I've noticed that the memory and localstorage adapter were not working at all in my phantomjs tests.  Versions 3.2 - current head have the same issue.  It seems that both adapters actually end up just using the LevelUP indexeddb implementation, meaning that they neither work as fallbacks for older browsers nor are they usable in certain test environments (like phantomjs).

I have a simple demonstration here. http://bl.ocks.org/corps/1d285ffac2b6d67e895f

 It's kinda tricky because it's not possible to "remove" the native indexedDB from the environment to invoke the exceptions I'm seeing in CI, but the issue can be demonstrated in two ways:

1.  View the page.  Now reload.  Notice the 409, despite being "in memory"?  Check your resources tab on your browser: somedb and pouch-memory are now indexeddb databases.

2.  Open a debugger and reload the page.  I've added a single debugger in the offending constructor call that leads to an idb.open.  You can see that the opts.adapter is, in fact, memory, and that PouchDB.adapters['memory'] is set to use this the LevelPouchAlt that leads to using the LevelUP implementation.

Hopefully this is an easy issue to fix.  To be honest, I've tried digging deeper but the way that the actual adapter code gets assembled is somewhat confusing to me.

nolanlawson:
Yup, I can confirm that the memory/localstorage plugins are both just using level.js. Probably a bug we introduced in d3bb6ea349371b4ff202927ea7c86cb001df62e6

nolanlawson:
Hm, browserify seems to ignore `-r memdown:leveldown`, but it uses the `{browser: {leveldown: 'level-js'}}` option in the `package.json`.

nolanlawson:
OK, figured it out. Got a fix.

nolanlawson:
Scratch that, I didn't figure it out. And I don't have time to look into this right now. @calvinmetcalf any ideas?

Things I did figure out:

* I can get it to work if I put `{browser: {leveldown: 'memdown'}}` in the package.json
* `-r memdown` includes memdown, but doesn't link it to `leveldown`
* both `-r memdown:leveldown` and `-r leveldown:memdown` don't do anything anymore

nolanlawson:
Nevermind, found a hack that fixes it.

nolanlawson:
[...]

PR #3530, opened 2015-02-12
(#3528) - fix memory/localstorage plugins

nolanlawson:
This is sort of a nasty hack, but it works.

I can confirm that this doesn't include anything we don't want (e.g. level-js in the built `pouchdb.js`, or memdown in `pouchdb.localstorage.js` etc.), and I can also confirm that the plugins are actually using in-memory/localstorage/level.js as they should.

### DIFF
--- a/bin/build-plugin.sh
+++ b/bin/build-plugin.sh
@@ -12,7 +12,7 @@ fi
 DEREQUIRE=./node_modules/.bin/derequire
 
 ./node_modules/.bin/browserify lib/plugins/index.js \
-    -r $LEVEL_BACKEND:leveldown \
+    -r $LEVEL_BACKEND \
     -x pouchdb \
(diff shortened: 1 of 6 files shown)


nolanlawson:
BTW if we want to avoid this kind of bug in the future, the only thing I can think of would be to add a PhantomJS+localstorage or PhantomJS+memory test (which would fail if it tried to used IndexedDB).

I would be happy to add the latter, especially since I've been recommending PhantomJS+memory for unit tests.

nolanlawson:
Green! I'll try to include this in 3.3.1.

daleharvey:
+1, cheers for jumping on this quickly

nolanlawson:
3229218b8633a41d9c2af5bdae849a52dcb84209

Issue #3466, opened 2015-01-28
Views are not correctly supported in changes() apparently


nolanlawson:
See [this SO question](http://stackoverflow.com/questions/28059854/listening-to-pouchdb-changes-in-particular-view-returns-all-documents/28071662?noredirect=1#comment44739531_28071662).

robertkeizer:
Some what related #2928 ; I would absolutely love this feature. Not enough time in the day to get back to it though.

nolanlawson:
they are supported, just poorly documented: 5b0d4d7cb6d005b61583808c623bd1c1422f7d74

PR #3545, opened 2015-02-17
(#3466) - Update docs for using view to filter changes.

dahjelle:
Updated the documentation to reflect what appears to be in the code to use a `map` function to filter the `changes` feed.


nolanlawson:
Ah yeah, this is probably something we could fix in the code as well. But I believe this is how CouchDB works by default, so it's probably better to just document it.

nolanlawson:
cheers, 5b0d4d7cb6d005b61583808c623bd1c1422f7d74

PR #3546, opened 2015-02-17
(#3546) - Fix api docs rendering

nickcolley:
Yeah the docs broke when we included the alert html...

wat

I don't even understand the fix but it works so.. Jekyll moves in mysterious ways with whitespace.. :P


nolanlawson:
Looks good, thanks!

Issue #3543, opened 2015-02-16
Headers not getting set in sync and replications in v3.3.1


alanquigley:
I'm setting the following options object

      var options = {
        ajax: {
          headers: {
             Authorisation : 'my token value'
          }
        }
      }; 

and passing it into the PouchDB constructor.

this.db = new PouchDB(_id, options);

But my sync and replication requests don't contain the header.

nolanlawson:
I think you mean `Authorization` not `Authorisation`.

alanquigley:
Uk spelling, but I've tried both, am I missing something?

nolanlawson:
Looks like you may be missing the "Basic" keyword. Here's a live working example: http://bl.ocks.org/nolanlawson/61c9c615516582436e22

alanquigley:
I'm trying to pass a jwt auth token through a proxy, the format is: 
Authorization:Bearer eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ1c2VySUQiOiJvcmcuY291Y2hkYi51c2VyOmllX2FsYW5xdWlnbGV5X2dtYWlsX2NvbSJ9.X_qWaFXVdxMEhxsyfURALZrLbklTq03Gh33vA00Hbjg
Is it possible to set custom headers?

alanquigley:
https://github.com/alanquigley/pouchdb/commit/c76c2a596ba97d2f58db7bd4ef3f153d503c8a23

The above tweak seems to fix my issue.

alanquigley:
Got it working, the global settings don't seem to be working, If I put the headers in into the sync options it attaches them.

 
          var token = 'Bearer '+ AuthTokenFactory.getToken();
          var options = {
            ajax: {
              headers: {
                 Authorization : token
              }
            },
            filter : filter
          };

nolanlawson:
Great, glad it's working!

alanquigley:
The documentation seems to indicate that attaching the headers globally should work, is this a bug? or an error in the docs?

nolanlawson:
You're right; that looks like a bug.

alanquigley:
https://github.com/alanquigley/pouchdb/commit/c76c2a596ba97d2f58db7bd4ef3f153d503c8a23

When I said I got it working I was mistaken, I actually required the above code change to allow me to add  the options sync method.

daleharvey:
So you can pass these options to the PouchDB constructor

[...]

PR #3549, opened 2015-02-19
(#3543) - allow set Auth headers in replications

nolanlawson:
This fix seems obviously correct to me, and I can't think of a test for it since we don't use authentication in our test CouchDB. So I'd be fine with merging it as-is (assuming it's green).

### DIFF
--- a/lib/replicate.js
+++ b/lib/replicate.js
@@ -575,7 +575,7 @@ exports.toPouch = toPouch;
 function toPouch(db, opts) {
   var PouchConstructor = opts.PouchConstructor;
   if (typeof db === 'string') {
-    return new PouchConstructor(db);
+    return new PouchConstructor(db, opts);
   } else if (db.then) {
     return db;
   } else {


daleharvey:
We do some tests of recieving headers in test/components/, I would like to see us ensuring this via a test because the functionality can be confusing between whats supported / how it behaves, will try to take a look at how to get a nice test for this 

(it doesnt need to specifically test auth, just that the opts pass)

daleharvey:
:+1: 

My new policy is to not block on things that make sense that dont have tests, if I want to make sure it doesnt break then i will write a test for it :)

yaronyg:
@daleharvey Unless you have infinite time you can't write all those tests! I hate writing tests but it's one of the things that keeps PouchDB so damn reliable!

nolanlawson:
2e995c080c7b273fe3215d0e4cd289ac9b9a9203

PR #3551, opened 2015-02-19
Document the `timeout` option for `changes`

### LINKED ISSUE #3550: HTTP+filtered changes fails with ESOCKETTIMEDOUT on large CouchDB changeset
If CouchDB contains a large changeset (in my case, 170K+ deletions), filtered replication will take some time to process in CouchDB (which is OK), but PouchDB's `.changes` will fail with an `ESOCKETTIMEDOUT` error:

    $ node -e 'P = require("pouchdb"); db = new P("http://127.0.0.1:5984/provisioning"); db.changes({include_docs:true,filter:"royal-thing/global_numbers",since:1897043}).on("error",function(error){console.log(error.toString());});'
    Error: ESOCKETTIMEDOUT

This is repeatable; the error shows up 31 seconds after the start of the request.
(This is PouchDB 3.3.1 against CouchDB 1.4.0. `since` in the query is at 1897043, while the `update_seq` of the database is 2020286. CouchDB logs show it running the filter.)

Doing a plain `curl` at the command prompt with the same parameters as the PouchDB request, `curl` just waits for the data to come in:

    $ curl 'http://127.0.0.1:5984/provisioning/_changes?timeout=25000&include_docs=true&feed=longpoll&filter=royal-thing/global_numbers&since=1897043&limit=25'
    {"results":[

I tried playing with `ajax.timeout` since this is the only timer available, but unsurprisingly it doesn't affect the outcome, since the TCP connection is established (there's no timeout at the network level).

Note: this might be related to the second problem described in #3210

shimaore:
Fixes #3550.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -426,6 +426,7 @@ All options default to `false` unless otherwise specified.
 * `options.doc_ids`: Only show changes for docs with these ids (array of strings).
 * `options.since`: Start the results from the change immediately after the given sequence number, you can also pass 'now' if you want only new changes.
 * `options.live`: Uses the  `_longpoll_` feed.
+* `options.timeout`: Request timeout (in milliseconds).
 * `options.limit`: Limit the number of results to this number.
 * `options.style`: Specifies how many revisions are returned in the changes array. The default, `'main_only'`, will only return the current "winning" revision; `'all_docs'` will return all leaf revisions (including conflicts and deleted former conflicts).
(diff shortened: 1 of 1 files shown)


nolanlawson:
+1

nolanlawson:
fixed the commit message on the way in to match our style. thanks for the contribution!

cf88f72ed3d4d3ee5acf85d58417cadf7a34fbf3

nolanlawson:
Site updated.

PR #3552, opened 2015-02-19
Make a test pass for Sync Gateway

jchris:
This makes test.replication.js-local-http by skipping assertions
that expect database info to contain a doc count. If there is a
doc count we make sure it's correct.

Also Sync Gateway sends an additional _revisions field which was
messing up deep equal, so instead now I'm testing directly for
the field we expect.

### DIFF
--- a/tests/integration/test.replication.js
+++ b/tests/integration/test.replication.js
@@ -217,10 +217,13 @@ adapters.forEach(function (adapters) {
         result.docs_written.should.equal(0, 'correct # docs written (1)');
         return db.info();
       }).then(function (info) {
-        info.doc_count.should.equal(1, 'doc_count');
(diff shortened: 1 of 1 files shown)


jchris:
passes

It's alive!

jchris:
This shouldn't pass on the Travis version of SG yet, the changes in SG to make it happen are still on the SG master branch.

jchris:
I think that failure was spurious, maybe merge anyway?

nolanlawson:
I tend to restart the failed jobs just to be safe. It's also good motivation to fix the intermittents :P.

You should have received an invite to join the pouchdb org; after that if you log into Travis, you can restart the jobs too. I've restarted this one, though.

nolanlawson:
Green and merged. :) cd656e001c89e04a1b8e95f126a5e95bc7b1f287

PR #3553, opened 2015-02-20
(#3553) - clarify docs on destroy()

nolanlawson:


nickcolley:
Need to make sure the html didn't break the markdown rendering again..

nickcolley:
Looks fine! @nolanlawson for the guide improvement, couldn't we just change those snippets to use db.destroy?

nolanlawson:
Personally I feel like it's more confusing when documentation deprecates something by just removing all mentions of it, as if it never existed. :/

daleharvey:
I think the API docs are the source for information about deprecated functionality, I do think @nickcolley is right about this, having a guide tell you to use this function, then saying right afterward that it may go away is more confusing imo.

nolanlawson:
OK, I'm outvoted. I'll close this and make the fix elsewhere, in https://github.com/pouchdb/pouchdb/pull/3577 and related PRs.

PR #3556, opened 2015-02-21
local db documents

jchris:
SG doesn't pass doc_count, but when it is passed, we want to test it.


nolanlawson:
sweet, d0cc4c730be4217ca2281e67df0b8af135056e59

PR #3557, opened 2015-02-21
(#3557) - idb allDocs() refactor, perf improvements

nolanlawson:
- keep docCount in memory, ala leveldb.js
- use metadata.deleted instead of calculating
  with utils.isDeleted()
- fetch docs asynchronously for include_docs=true

### DIFF
--- a/lib/adapters/idb/idb-all-docs.js
+++ b/lib/adapters/idb/idb-all-docs.js
@@ -1,0 +1,184 @@
+'use strict';
+
+var merge = require('../../merge');
+var errors = require('../../deps/errors');
+var idbUtils = require('./idb-utils');
+var idbConstants = require('./idb-constants');
+
+var ATTACH_STORE = idbConstants.ATTACH_STORE;
+var BY_SEQ_STORE = idbConstants.BY_SEQ_STORE;
+var DOC_STORE = idbConstants.DOC_STORE;
+
+var decodeDoc = idbUtils.decodeDoc;
+var decodeMetadata = idbUtils.decodeMetadata;
+var fetchAttachmentsIfNecessary = idbUtils.fetchAttachmentsIfNecessary;
+var postProcessAttachments = idbUtils.postProcessAttachments;
+var openTransactionSafely = idbUtils.openTransactionSafely;
+
+function createKeyRange(start, end, inclusiveEnd, key, descending) {
+  try {
+    if (start && end) {
+      if (descending) {
+        return IDBKeyRange.bound(end, start, !inclusiveEnd, false);
+      } else {
+        return IDBKeyRange.bound(start, end, false, !inclusiveEnd);
+      }
+    } else if (start) {
+      if (descending) {
+        return IDBKeyRange.upperBound(start);
+      } else {
+        return IDBKeyRange.lowerBound(start);
+      }
+    } else if (end) {
+      if (descending) {
+        return IDBKeyRange.lowerBound(end, !inclusiveEnd);
+      } else {
+        return IDBKeyRange.upperBound(end, !inclusiveEnd);
+      }
+    } else if (key) {
+      return IDBKeyRange.only(key);
+    }
+  } catch (e) {
+    return {error: e};
+  }
+  return null;
+}
+
+function handleKeyRangeError(api, opts, err, callback) {
+  if (err.name === "DataError" && err.code === 0) {
+    // data error, start is less than end
+    return callback(null, {
+      total_rows: api._meta.docCount,
+      offset: opts.skip,
+      rows: []
+    });
+  }
+  callback(errors.error(errors.IDB_ERROR, err.name, err.message));
+}
+
+function idbAllDocs(opts, api, idb, callback) {
+
+  function allDocsQuery(opts, callback) {
+    var start = 'startkey' in opts ? opts.startkey : false;
+    var end = 'endkey' in opts ? opts.endkey : false;
+    var key = 'key' in opts ? opts.key : false;
+    var skip = opts.skip || 0;
+    var limit = typeof opts.limit === 'number' ? opts.limit : -1;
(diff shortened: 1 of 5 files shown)


nolanlawson:
Currently this is showing a big improvement for Chrome and an almost equal un-improvement for Firefox :(

```
temp-views test, 10 iterations
Chrome 42: 89105ms -> 78397ms
Firefox 37:  80500ms -> 86922ms
Firefox 37 (test 2): 82754ms -> 89679ms
```

I've been trying to figure out what's going on, because it doesn't make any sense to me that it would be so much slower in one but faster in the other.

lib/adapters/idb/idb-all-docs.js
```diff
@@ -0,0 +1,184 @@
+'use strict';
```
nolanlawson: I moved the `allDocs()` function into its own file, same as I did before with `bulkDocs()`.

lib/adapters/idb/idb-all-docs.js
```diff
@@ -0,0 +1,184 @@
...
+      };
+      var deleted = metadata.deleted;
+      if (opts.deleted === 'ok') {
+        results.push(row);
+        // deleted docs are okay with "keys" requests
+        if (deleted) {
+          row.value.deleted = true;
+          row.doc = null;
+        } else if (opts.include_docs) {
+          fetchDocAsynchronously(metadata, row, winningRev);
```
nolanlawson: ^ this is new. instead of blocking the cursor, fetch the doc in parallel.

lib/adapters/idb/idb-all-docs.js
```diff
@@ -0,0 +1,184 @@
...
+        }
+        if (--limit === 0) {
+          return;
+        }
+      }
+      cursor.continue();
+    }
+
+    function onGetCursor(e) {
+      docCount = api._meta.docCount; // do this within the txn for consistency
```
nolanlawson: ^ this is new. we can just grab the doc count from in-memory instead of caculating/caching it like before.

lib/adapters/idb/idb-all-docs.js
```diff
@@ -0,0 +1,184 @@
...
+
+    function allDocsInner(cursor, winningRev, metadata) {
+      var row = {
+        id: metadata.id,
+        key: metadata.id,
+        value: {
+          rev: winningRev
+        }
+      };
+      var deleted = metadata.deleted;
```
nolanlawson: ^ this is new. we used to be doing utils.isDeleted(), when we really didn't need to, because the info was already there.

lib/adapters/idb/idb-utils.js
```diff
@@ -63,7 +63,7 @@ exports.decodeMetadata = function (storedObject) {
   }
   var metadata = utils.safeJsonParse(storedObject.data);
   metadata.winningRev = storedObject.winningRev;
-  metadata.deletedOrLocal = storedObject.deletedOrLocal === '1';
+  metadata.deleted = storedObject.deletedOrLocal === '1';
```
nolanlawson: I was tired of the metadata exposing this internal implementation detail of the IDB adapter: that our "deleted" index is named "deletedOrLocal" for backwards-compatibility reasons. Prefer not to expose this weird name anymore.

lib/adapters/idb/idb.js
```diff
@@ -46,14 +47,10 @@ function IdbPouch(opts, callback) {
...
-  var name = opts.name;
+  var dbName = opts.name;
 
-  var instanceId = null;
-  var idStored = false;
   var idb = null;
-  api._docCount = -1;
-  api._blobSupport = null;
-  api._name = name;
+  api._meta = null;
```
nolanlawson: We were stocking so much data in this cache, it was starting to look pretty ugly. So I just folded everything into a single `_meta` variable.

lib/adapters/idb/idb.js
```diff
@@ -46,14 +47,10 @@ function IdbPouch(opts, callback) {
 
 function init(api, opts, callback) {
 
-  var name = opts.name;
+  var dbName = opts.name;
```
nolanlawson: ^ this generic name `name` was annoying me, especially because I frequently screw up and accidentally use the global `window.name`, causing hard-to-debug bugs.

lib/adapters/idb/idb.js
```diff
@@ -999,57 +819,87 @@ function init(api, opts, callback) {
...
+      //
+      // count docs
+      //
+
+      var index = txn.objectStore(DOC_STORE).index('deletedOrLocal');
+      index.count(IDBKeyRange.only('0')).onsuccess = function (e) {
+        docCount = e.target.result;
+        checkSetupComplete();
+      };
+
```
nolanlawson: we can just count the docs once now, when the database starts up. we might also want to reconsider this `deletedOrLocal` index entirely, since this is the only place we're using it anymore IIRC.

nolanlawson:
Some more numbers from Chrome:

```
78558ms -> 83279ms
79023ms -> 78912ms
```

I think what this is revealing is just that our performance tests aren't fine-grained or consistent enough to tell us anything useful. For this PR, I'm happy to go with my gut.

nolanlawson:
OK, I've added a new performance test, which I think more accurately tests what I'm trying to improve here: `all-docs-include-docs`:

```
Chrome: 11256ms -> 6982ms
Firefox: 7388ms -> 5645ms
```

daleharvey:
Slightly wary about the caching, but with tests passing and the performance improvements I think this is good, cheers - https://github.com/pouchdb/pouchdb/commit/f010d2eb959e045d9c94462780207f1ea16e45c0

nolanlawson:
The caching of `docCount` is the exact same trick used by `leveldb.js`, so it has precedent. I'd also like to add it to WebSQL eventually, but these days I'm more concerned about IDB improvement.

Issue #3558, opened 2015-02-22
Publish blob, ajax, binary-utils as separate libs


nolanlawson:
This is starting to lead to issues like this one (https://github.com/nolanlawson/pouchdb-load/issues/11) because I have to copy-paste in the `ajax.js` files into any plugins, which then inevitably gets out of date.

nolanlawson:
Note that there's also a `pouchdb-promise` module published on npm. I think for now, though, I'm going to punt on this, because it just made the project structure of pouchdb too weird.

PR #3559, opened 2015-02-22
(#3558) - modularize ajax/blob/bin-util/errors

nolanlawson:
I've created `pouchdb-binary-util`, `pouchdb-ajax`, and `pouchdb-errors` to support this. They are in the pouchdb org and I've added daleharvey, cwmma, and commandoline as npm owners.


nolanlawson:
min+gz size was 48K before and is 48K after, so looks like we're avoiding any new code bloat.

daleharvey:
I am not particularly comfortable with this level of modularisation, none of these are particularly reusable and so this only really gives us the downsides of modularising (we run into issues fixing bugs in the deps) with none of the updates (people reuse the modules and they get better tested)

I already reuse browser-request in another project, but I just inlined the utils I needed, a tiny bit of extra code but much more reusable, ill put up an example tomorrow

nolanlawson:
I've already started reusing these in `pouchdb-load`. I'm also looking to modularize the checkpointer so I can reuse it there as well...

Would it make you feel more comfortable if I moved those to the pouchdb org, and we had the same rules about +1s as we do for core libs like mapreduce? How about if we kept them in the same git repo, but still published them separately on npm?

daleharvey:
> How about if we kept them in the same git repo, but still published them separately on npm?

I think this would be perfect

nolanlawson:
OK, I will try to code that up soon. They're already published on npm, so it should be easy.

Any objection to putting the checkpointer in a separate module as well? I just fixed a bug in it, but `pouchdb-load` is using my copypasta version... >_<

daleharvey:
Nah thats totally fine with me, if your using it externally it makes sense, but I think we should keep these in the same repo for now and extract them to a seperate repo if they are more useful outside pouchdb (and very closely related modules)

nolanlawson:
Hm, so the awkward thing here is that, since the modules have dependencies on each other, I'm not sure whether they should reference each other directly (e.g. `require('../deps/pouchdb-ajax')`) or via npm (`require('pouchdb-ajax')`). Either way it's kinda weird. They both have downsides.

Another option is to take the lodash approach and have third-party libraries do e.g. `require('pouchdb/deps/ajax')`, but that probably requires some kind of commitment on our part that we won't change that part of the API.

All in all, it creates a very complicated system, and I'm not sure this benefits any plugins other than pouchdb-load and pouchdb-authentication. I think I might just close this for now, out of conservatism.

I'll also note that @calvinmetcalf published a pouchdb-promise module, which we are also not using. But in the future, if we figure out this modularity thing, we might want to use that too.

nolanlawson:
Reopening this because there has been a LOT of discussion in the plugins about this (mostly `pouchdb-ajax`), and it needs to get done soonish or likely PouchDB will start to diverge from the plugin implementations.

nolanlawson:
Ugh, this is already so old that I need to just rewrite it from scratch. :/

nolanlawson:
On second thought, I'm reopening as-is. Some arguments for doing it this way:

* this is reusable; `pouchdb-ajax` is already in demand by pouchdb-load, pouchdb-authentication, and the pouchdb hoodie plugin
* hopefully things like `pouchdb-ajax`, `pouchdb-binary-utils`, and `pouchdb-errors` should not change much
* doing it in the `require('pouchdb/ajax')` style seems like an unholy mess because you literally need to have `ajax/index.js` or `ajax.js` at the root of the project structure (i.e. outside `lib/`). Looks like lodash does a separate build process for Node in order to accomplish this, which adds way too much complexity for my taste.

I'd also like to do this for `pouchdb-checkpointer`, but that will have to wait.

nolanlawson:
rebased

nolanlawson:
Ugh, can't extract the checkpointer because it depends on `utils.explain404` and `utils.Promise`.

I don't even know anymore. Maybe I should just unpublish `pouchdb-ajax` and people should use `require('pouchdb/lib/deps/ajax')`?

nolanlawson:
Changed my mind; we can create a top-level `ajax.js` file that just exports `lib/deps/ajax`.

PR #3560, opened 2015-02-22
(#3543) - fix filtered replication with a ddoc

nolanlawson:
This should fix the problem where filtered replication
with a ddoc was being ignored. Apparently we had no tests
for ddoc-filtering in `replicate()`.

I tested this against Cloudant and it passed, so hoping it will
pass against Couch master as well.


daleharvey:
Nice, I thought this was going to be far more complicated, looks great

daleharvey:
https://github.com/pouchdb/pouchdb/commit/9c093980d4bc03b19d9dfdb338f656849ee01a6b

Forgot to close out

PR #3561, opened 2015-02-23
verifyInfo for CSG

jchris:
This duck types db infos from Sync Gateway and makes allowances
for things like access control sequences and lack of doc_count.

We don't switch on isSyncGateway() explicitly b/c we don't want
to drop these assertions for the whole environment, just when
the info is from Sync Gateway. If we want to make sure it's not
being lenient in other environments we could refactor the verifyInfo
function to look more like a switch statement. Maybe the time to
do this is on the NEXT environment that gets added. :)

### DIFF
--- a/tests/integration/test.replication.js
+++ b/tests/integration/test.replication.js
@@ -62,9 +62,18 @@ adapters.forEach(function (adapters) {
 
     function verifyInfo(info, expected) {
       if (!testUtils.isCouchMaster()) {
-        info.update_seq.should.equal(expected.update_seq, 'update_seq');
+        if (typeof info.doc_count === 'undefined') { 
+          // info is from Sync Gateway, which allocates an extra seqnum
+          // for user access control purposes.
+          info.update_seq.should.be.within(expected.update_seq, 
+            expected.update_seq + 1, 'update_seq');
+        } else {
+          info.update_seq.should.equal(expected.update_seq, 'update_seq');
+        }
       }
-      info.doc_count.should.equal(expected.doc_count, 'doc_count');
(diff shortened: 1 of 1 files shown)


jchris:
Since it's already flattened into one commit with a detailed assertion, and the tests pass, I'm tempted to click the green merge button. Anyone tried that in a simple case like this? Can it work?

nolanlawson:
@jchris Sadly, it will definitely create an explicit merge commit if you press the Green button. :( Don't worry, though, I got ya covered. I'll merge it.

tests/integration/test.replication.js
```diff
@@ -62,9 +62,18 @@ adapters.forEach(function (adapters) {
 
     function verifyInfo(info, expected) {
       if (!testUtils.isCouchMaster()) {
-        info.update_seq.should.equal(expected.update_seq, 'update_seq');
+        if (typeof info.doc_count === 'undefined') { 
+          // info is from Sync Gateway, which allocates an extra seqnum
+          // for user access control purposes.
```
nolanlawson: Duck typing is fine, but the comments are much appreciated. I worry the tests may end up getting pretty gnarly if we have a lot of special logic for CSG vs Couch master vs Couch 1.x vs Pouch Server, but so far it hasn't been too bad.

nolanlawson:
ca5dfd4571c9318345b4f32b573f27b2e48d09a3

One trick I often use to get the right issue number for a commit is to just march through `github.com/../issues/<issuenum>` until I find a 404. I think @daleharvey commits and then amends the commit message. :stuck_out_tongue:

PR #3562, opened 2015-02-23
docs read range for SG

jchris:

PR #3563, opened 2015-02-24
(#3562) Cleanup CSG assertions

jchris:
Instead of a range assertion, assert for the specific case.


jchris:
merged ee340e414f7ca8f6b4bbee7cd782b3c40f8c2fd9

nolanlawson:
@jchris This looks good, but just for next time, we prefer to get a +1 from another contributor before merging stuff, even if it's tiny. :)

PR #3564, opened 2015-02-24
(#3495) filter changes for CSG

jchris:
CSG has extra entries in the changes feed, which we can ignore
in tests.

### DIFF
--- a/tests/integration/test.replication.js
+++ b/tests/integration/test.replication.js
@@ -56,7 +56,12 @@ adapters.forEach(function (adapters) {
           return a.id > b.id;
         });
       }
-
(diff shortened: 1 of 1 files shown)


jchris:
For something as small as this, if it passes locally, would you do a pull request / Travis run before merge to master, or just go ahead and commit to master?

nolanlawson:
@jchris Yep, please wait on another contributor to +1 it, even if it's tiny. Also you forgot some hyphens in your commit message ;)

nolanlawson:
+1 when green, nice to see that CSG isn't really too different from CouchDB after all. Nice work!

jchris:
merged in e9a3c87

Issue #99, opened 2012-08-06
Store incremental data for views


pixelcort:
It appears each time a view is used, all data is rerun through the map and reduce functions.

Instead, it would be nice if PouchDB cached incremental data for the output of map and reduce functions, like CouchDB does, so that future uses of the view would be faster, at the expense of additional storage space.

PR #3565, opened 2015-02-24
(#3493) be robust about missing docs

jchris:
Sync Gateway occasionally consumes a sequence number and sends a
changes entry that does not correspond to a document. It is safe
to ignore these. This patch prevents them from blowing up filters.

This is a redo of an earlier pull request. Now we have tests for
Sync Gateway.

However those tests won't start behaving correctly until we cut a new release
of Sync Gateway and point Travis at it.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -106,6 +106,11 @@ exports.filterChange = function filterChange(opts) {
   req.query = opts.query_params;
 
   return function filter(change) {
+    if (!change.doc) {
+      // CSG sends events on the changes feed that don't have documents,
+      // this hack makes a whole lot of existing code robust.
+      change.doc = {};
+    }
     if (opts.filter && hasFilter && !opts.filter.call(this, change.doc, req)) {
       return false;
     }


nolanlawson:
I'm super-excited about disabling the "allowed failures" in Travis, but this seems like a pretty bad hack. Could you explain more about what the issue is and maybe point to the corresponding issue on the CSG side?

jchris:
Sync Gateway sends an entry sometimes when documents you can't see have changed an Access Control setting somewhere. Basically your client needs to know a sequence was consumed, even if it doesn't get a corresponding document.

So they are safe to ignore. The same rows I'm ignoring here: https://github.com/pouchdb/pouchdb/pull/3564/files

jchris:
There are other times the document could come up missing, like if you are fetching them async and it is purged before you get it. To me it makes sense that you'd want to be robust here anyway.

nolanlawson:
Okay, that makes sense in the context of how we "smooth out" the different server APIs (e.g. Cloudant/Couch master don't always set `ok` to `true`). Please fix the commit message before you merge (missing hyphen), but otherwise LGTM.

jchris:
I'm not trying to be dense. What needs a hyphen?

On Tuesday, February 24, 2015, Nolan Lawson <notifications@github.com>
wrote:

> Okay, that makes sense in the context of how we "smooth out" the different
> server APIs (e.g. Cloudant/Couch master don't always set ok to true).
> Please fix the commit message before you merge (missing hyphen), but
> otherwise LGTM.
>
> —
> <https://github.com/pouchdb/pouchdb/pull/3565#issuecomment-75771797>.
>


-- 
Chris Anderson
http://www.couchbase.com
http://twitter.com/jchris

jchris:
Oh is see, I missed it in Contributing: `(#99) -` will fix

jchris:
merged in d58a3b7

nolanlawson:
I've forgotten the hyphen many a time; no worries. The effort is appreciated, though. :)

PR #3566, opened 2015-02-24
Ignore changes feed include_docs conflicts from CSG

jchris:
This just normalizes some of the Sync Gateway differences. (We don't list conflicts in the include_docs doc, just in the changes feed list, etc.)

Creating pull request to verify test run before committing. I really wish I could green button to commit, once it's passing.


nolanlawson:
This looks good, although I would probably add a comment before committing just to explain the CSG difference.

Not having the big green button is an annoyance, but looking back on the nice, linear commit log, I think it's worth the extra effort. Thanks for your help on this! :)

jchris:
Pushed in b355fb6e645d6750af943090bcd53fb021b6e2d7

Issue #3569, opened 2015-02-25
replicator throws a 409 if you try to run multiple replications simultaneously


nolanlawson:

daleharvey:
We have a bunch of new tests for this now right? closing but reopen if needed

PR #3570, opened 2015-02-25
(#3569) - catch 409s in checkpointer

nolanlawson:

I'm kinda surprised this bug has hung around for so long. I thought we were more diligent about catching 409s and retrying.


daleharvey:
Good catch, cheers

https://github.com/pouchdb/pouchdb/commit/8ba85d02cc0fb3ff1ad7cd0f43963f01d7432063

PR #3571, opened 2015-02-25
(#3571) - blog post about UI state

nolanlawson:
This came up recently as a question on IRC, and I thought
it was interesting enough to turn into a blog post. It's
been awhile since we had a non-release blog post. :)

### DIFF
--- a/docs/_posts/2015-02-25-efficiently-managing-ui-state-in-pouchdb.md
+++ b/docs/_posts/2015-02-25-efficiently-managing-ui-state-in-pouchdb.md
@@ -1,0 +1,128 @@
+---
+layout: post
+
+title: Efficiently managing UI state with PouchDB
+
+author: Nolan Lawson
+
+---
+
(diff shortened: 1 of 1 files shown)


nickcolley:
How about we inline your toy at the end of the post since we have Pouchdb global? :dancer:

nolanlawson:
I like the way you're thinking, but let's keep it simple for now. :)

nolanlawson:
At some point in the future I would like a consistent way to inline code samples (maybe a jekyll `include`?) Then we should definitely update this post to do it inline.

nickcolley:
I'm working on that currently, based on the demo @ http://nck.im/try/

daleharvey:
This is awesome, +1 to release when you think is good for you

nolanlawson:
I'll do it right now; don't want to steal the limelight from the Monday release. :)

nolanlawson:
0a523ed98c545232b761ba1203c57b6ddd393fb8

PR #3575, opened 2015-02-25
[Don't Merge] Test SG master

jchris:


jchris:
With SG master, test.replication.js runs to the end. So we have that to look forward to on the next release.

```
  262 passing (4m)
  20 pending
  51 failing
```

I thought about making it so we actually have to SG jobs, one is master and one is released. I need to do some stuff on the Couchbase side to get a stable master URL, but once I have that I'll revisit this.

nolanlawson:
That's awesome! I'm fine with adding two jobs - it's the same as what we're doing for CouchDB anyway.

jchris:
I'm working on some stuff on my end so we can use auto-updating master builds of Sync Gateway.

PR #3585, opened 2015-03-01
(#3585) - blog post about ES7 async funcs

nolanlawson:
@calvinmetcalf helped me out on IRC, and I finally figured out how to use PouchDB with Babel/Regenerator. It's exciting, so I couldn't wait to blog about it.

I think this might be a nice blog post to coincide with the new promise/callback switcher. Really helps clarify the whole "async API" story.

I've built the blog post, so you can read [the draft](https://nolanlawson.s3.amazonaws.com/pouchdb/www/20150301/post.html) without having to build it yourself.


daleharvey:
Awesome, Im not totally sure I agree with the post as I find cooperative threading hugely confusing but we dont need to agree on everything we blog, Its still very interesting

daleharvey:
https://github.com/pouchdb/pouchdb/commit/476e158a40b5f876da980fac6847891817f59c88

nolanlawson:
Interested to know what you mean. My understanding is that it's not cooperative threading (in the same sense that Java has threads), but instead just a bit of syntactic sugar.

Issue #2352, opened 2014-06-05
Convert dates to strings in IndexedDB


nolanlawson:
Otherwise things like this (https://github.com/pouchdb/pouchdb/issues/2351) can happen, where the developer tests storing raw Dates in IndexedDB and then it bites them when it doesn't work in CouchDB/WebSQL/LevelDB.

daleharvey:
Closing as dupe of https://github.com/pouchdb/pouchdb/issues/4126

Issue #2158, opened 2014-05-05
IDBObjectStore: An object could not be cloned (Ember.js integration)


jfgirard:
I just gave a try to 2.2.0.  With WebSQL, its working great.  But with IndexedDb (Chrome and FF),  if i insert a document and `query()` it, i receive this error:

Uncaught DataCloneError: Failed to execute 'put' on 'IDBObjectStore': An object could not be cloned. pouchdb.js:2140

I still use the "old" query way, passing along the map function.

It fails on `docInfo.data` with value:

     Object {value: Object, _id: "5542014-05-05T19:57:35.812Z4E1D619FA-C497-146F-     A636-8BAAEADF6F6F31", _rev: "1-28a58b028db3af911ec118c3aa8cd4a4", _doc_id_rev:      "5542014-05-05T19:57:35.812Z4E1D619FA-C497-146F-A…EADF6F6F31::1-28a58b028db3af911ec118c3aa8cd4a4"}
     _doc_id_rev: "5542014-05-05T19:57:35.812Z4E1D619FA-C497-146F-A636-   8BAAEADF6F6F31::1-28a58b028db3af911ec118c3aa8cd4a4"
     _id: "5542014-05-05T19:57:35.812Z4E1D619FA-C497-146F-A636-8BAAEADF6F6F31"
     _rev: "1-28a58b028db3af911ec118c3aa8cd4a4"
     value: Object
     id: "E1D619FA-C497-146F-A636-8BAAEADF6F6F"
     key: Array[1]
     0: "2014-05-05T19:57:35.812Z"
     ....

Using allDocs({include_docs:true}) works fine.

ronyrun:
I also currently get that error : 

Error: Uncaught DataCloneError: Failed to execute 'put' on 'IDBObjectStore': An object could not be cloned. Script: file:///android_asset/www/lib/pouchdb/dist/pouchdb.js Line: 2192 Column: 29 StackTrace: DataCloneError: Failed to execute 'put' on 'IDBObjectStore': An object could not be cloned. debug.js:5window.onerror

I'm implementing pouchDB on cordova / IonicFramework  with Crosswalk.
OS: Windows 8.1
Node Version: v0.12.0
Cordova CLI: 4.2.0
Ionic CLI Version: 1.3.11
Latest Ionic version: 1.0.0-beta.14  (released 2014-12-15)

nolanlawson:
Are you using Ember.js? Can you provide a live reproducible test case?

[...]

PR #3588, opened 2015-03-02
(#2352 #2158) - clone object before inserting in IDB

nolanlawson:
Revival of 6c6d95a, which used cloning via pouchdb-collate
to make sure that Dates become strings, etc. This also
has the benefit of cloning the doc, so we avoid weird errors
due to global Object prototype mangling.

### DIFF
--- a/lib/adapters/idb/idb-bulk-docs.js
+++ b/lib/adapters/idb/idb-bulk-docs.js
@@ -1,5 +1,6 @@
 'use strict';
 
+var collate = require('pouchdb-collate');
 var utils = require('../../utils');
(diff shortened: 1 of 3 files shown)


nolanlawson:
I don't know what the performance impact is going to be, but we might want to do a sanity check of the performance tests before merging.

nolanlawson:
Well, it turns out that this doesn't actually solve #2158, because the `collate.normalizeKey()` function doesn't remove `function`s. I'm not too keen to add performance-reducing code for spurious gains, so I'm going to close this. We need to just document this error better.

Issue #3578, opened 2015-02-27
Filtered replication ignores _deleted docs


sphaso:
Hi.            
I'm facing a weird issue using PouchDB 3.3.0. Basically when I set filtered replication using a ddoc the onChange doesn't immediately trigger at load time (but it does without filters!). I have to ctrl+R and then I finally see something.

This is my setup (with the filter parameters commented). I'm not sure if this is a bug or just dumbness on my part. I've seen a PR vaguely related to filtered replication so I thought I'd ask.

```javascript
localDB.sync(remoteDB, {
                        live: true,
                        retry: true
                })
            .changes({
            continuous: true,
            //filter: 'app/by_user',
            //query_params: { agente: agent },
            onChange: function(change){
                $rootScope.$apply(function(){
                    localDB.get(change.id, function(err, doc){
                        console.log("change detected!");
                        if(!change.deleted){
                            $rootScope.$apply(function(){
                                if(err) console.log(err);
                                else {
                                    console.log("broadcast");
                                    $rootScope.$broadcast('upsert', doc);
                                }
                            });
                        }
                        else if(change.deleted){
                            $rootScope.$apply(function(){
                                $rootScope.$broadcast('delete', {_id: change.id});
                            });
                        }
                    });
                });
            }
        });
```

nolanlawson:
You can do `include_docs: true` on changes to avoid having to fetch the `doc`.

There's probably a bug with filtered replication. I'll try to look into it when I get a chance!

Another tip: if you are using filtered replication to try to do user security, [you might be doing it wrong](https://github.com/nolanlawson/pouchdb-authentication#couchdb-authentication-recipes), because users can actually still read each other's documents. See that link for details.

sphaso:
Thanks for the tip about user security, but the filter is more of a usability solution. It might become a nice workaroud if we can't fix this in time and I get enough HD space.

I'm not sure about the exact behavior of include_docs. From the docs it looks like it actually includes conflicts and attachments, not really my case. Does it ignore filter? (will I have to filter client side? that's not really an option).

nolanlawson:
You also need to set `conflicts: true, attachments: true` in order to get conflicts and attachments. :)

You shouldn't need to filter client-side. I'll try to fix this bug.

nolanlawson:
Ah okay, I remember now. This is a documentation issue. I've updated the PouchDB web site with a recent fix for the documentation. Now it says:

[...]

PR #3594, opened 2015-03-04
(#3578) - clarify docs on change events

nolanlawson:
I'm not a big fan of these events (I think they add unneeded complexity), but it seems they were documented, so I guess the right course of action is to make sure they are well-documented.

### DIFF
--- a/docs/_includes/api/changes.html
+++ b/docs/_includes/api/changes.html
@@ -36,8 +36,14 @@ var changes = db.changes({
   since: 'now',
   live: true,
   include_docs: true
-}).on('change', function(change) {
-  // handle change
+}).on('create', function (info) {
+  // document was created
+}).on('update', function (info) {
+  // document was updated
+}).on('delete', function (info) {
+  // document was deleted
+}).on('change', function(info) {
+  // any document was created/updated/deleted
 }).on('complete', function(info) {
   // changes() was canceled
 }).on('error', function (err) {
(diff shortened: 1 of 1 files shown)


daleharvey:
No longer needed due to https://github.com/pouchdb/pouchdb/commit/800de8602796082e2b55dd4a12791478172f68db

nolanlawson:
Er, actually these are a separate set of events than 800de86. These events apply to the `db.changes()` emitter, not the `db` itself. They're very lightly documented [here](http://pouchdb.com/api.html#changes):

>  In addition to the 'change' event, any change will also emit a 'create', 'update', or 'delete' event.

That being said, I'd definitely be open to removing them for 4.0. You can do just fine without them, as I showed in [this blog post](http://pouchdb.com/2015/02/28/efficiently-managing-ui-state-in-pouchdb.html).

calvinmetcalf:
Yeah I added them before I realized couchdb dedups errors making them not
as useful as I thought

daleharvey:
Yeh I would be more tempted to remove these, they havent been well documented and it looks like they will cause problems with couch 2.0, @calvinmetcalf you feel strongly either way?

nolanlawson:
Yeah I'd be happy to remove these events and publish the next release as 4.0. (While also removing `destroy()`).

daleharvey:
Closing this and opening https://github.com/pouchdb/pouchdb/issues/3833

PR #3603, opened 2015-03-05
(#3578) - Tests for ddoc filtering, deleted docs not replicated

sphaso:
I've written two simple tests that will probably clarify things. Basically, when doing filtered replication:
- deleted documents that used to pass the filter are not replicated anymore (i.e. still present locally)
- changed documents that used to pass the filter (change is on the condition necessary for the doc to pass) are not replicated anymore (i.e. still present locally).

As mentioned in #3578 this breaks the idea of two-ways-syncing remote-local dbs.
Please let me know if \ how I can improve the tests or any misunderstanding on my part.

### DIFF
--- a/tests/integration/test.replication.js
+++ b/tests/integration/test.replication.js
@@ -3609,6 +3609,65 @@ adapters.forEach(function (adapters) {
       });
     });
 
+    it("#3578 replication with a ddoc filter remove server-side", function(){
+      var db = new PouchDB(dbs.name);
+      var remote = new PouchDB(dbs.remote);
+
+      return remote.bulkDocs([{
+        _id: '_design/myddoc',
+        filters: {
+          myfilter: function (doc) {
+            return doc._id === 'a' || doc._id === 'b';
+          }.toString()
+        }
+      },
+        {_id: 'a'},
+        {_id: 'b'},
+        {_id: 'c'}
+      ]).then(function () {
+        return remote.replicate.to(db, {filter: 'myddoc/myfilter'});
+      }).then(function(){
+        return remote.get('a');
+      }).then(function(doc){ 
+        return remote.remove(doc);
+      }).then(function () {
+        return db.allDocs();
+      }).then(function (docs) {
+        docs.rows.should.have.length(1);
+      });
+
+    });
+
+    it("#3578 replication with a ddoc filter change server-side", function(){
+      var db = new PouchDB(dbs.name);
+      var remote = new PouchDB(dbs.remote);
+
+      return remote.bulkDocs([{
+        _id: '_design/myddoc',
+        filters: {
+          myfilter: function (doc) {
+            return doc.replicate;
+          }.toString()
+        }
+      },
+        {_id: 'a', replicate: true},
+        {_id: 'b', replicate: false},
+        {_id: 'c', replicate: false}
+      ]).then(function () {
+        return remote.replicate.to(db, {filter: 'myddoc/myfilter'});
+      }).then(function(){
+        return remote.get('a').then(function(doc) {
+           doc.replicate = false;
+           remote.put(doc);
+        });
+      }).then(function () {
+        return db.allDocs();
+      }).then(function (docs) {
+        docs.rows.should.have.length(0);
+      });
+
+    });
+
   });
 });
 


tests/integration/test.replication.js
```diff
@@ -3609,6 +3609,65 @@ adapters.forEach(function (adapters) {
...
+      },
+        {_id: 'a', replicate: true},
+        {_id: 'b', replicate: false},
+        {_id: 'c', replicate: false}
+      ]).then(function () {
+        return remote.replicate.to(db, {filter: 'myddoc/myfilter'});
+      }).then(function(){
+        return remote.get('a').then(function(doc) {
+           doc.replicate = false;
+           remote.put(doc);
```
nolanlawson: ^ small nitpick: this should be `return remote.put(doc)`; otherwise you're not guaranteed that the promise will resolve before the next `then()` is run.

nolanlawson:
I see the bug. It's a pretty interesting case, and I notice that in [the CouchDB docs](https://wiki.apache.org/couchdb/Replication) they say:

>Note: When using filtered replication you should not use the DELETE method to remove documents, but instead use PUT and add a _deleted:true field to the document, preserving the fields required for the filter. Your Document Update Handler should make sure these fields are always present. This will ensure that the filter will propagate deletions properly. 

This would require a pretty big change on PouchDB's part, because up to now we have assumed that there is no difference between `put()`ing a doc with `_deleted: true` vs `remove()`ing a doc. In fact, I don't think we store the fields at all if the document is deleted.

This issue argues that we should, though. I can try to find some time to work on this if you are too busy, but otherwise I think changing the logic so that:

* `db.remove()` is equivalent to CouchDB's DELETE, whereas
* `db.put()`/`db.post()`/`db.bulkDocs()` with the `_deleted` flag keeps the fields around.

would be the way to go here. Tricky bug!

tests/integration/test.replication.js
sphaso: I'd say a pretty important note. Keep them coming :) I'll commit a better version adding a return, but let me know if there's anything else I should change.

sphaso:
I'm glad the CouchDB team addressed the issue of deleted documents in filtered replication. However, this is only one of the two issues I raised concerning filtered replication:
- deleted documents (bug acknowledged)
- documents changed server-side in such a way that they won't pass the filter anymore

What do you think of the second case? Should it be the responsability of the end-developer to track such cases manually? or do you think CouchDB (or PouchDB...) should replicate as to apply the filter to the latest two revs of the document instead of the latest only?

About the first issue, I deleted documents using Futon so I didn't think of the difference in the API methods invoked, good catch!
I'd like to work on a fix (at least for the first issue) but I'm uncertain about my knowledge of the codebase and my JS-fu. I'll look into it but don't count too much on me :)

nolanlawson:
@sphaso Good point; I missed the second point until now. Yes, I think it is the responsibility of developers not to change the documents such that they no longer pass the filter.

nolanlawson:
Another thing you might not know (based on reading your code): non-live replication only does a single-shot replication, not a continuous one. The docs are in the process of being updated to make that clearer ([read this](https://github.com/pouchdb/pouchdb/blob/a4ce4c68654e0059a10be11840b84c796f3d63b2/docs/_includes/api/replication.html#L68-L126)).

sphaso:
About the second issue:
Ok, I'll delete the second test then. 
This of course raises the issue of "how?". Do you have any recipe? Obviously if the document shouldn't pass the filter, you want to delete it from the end-point of the replication (without having the deletion replicate back to CouchDB!). I can think of some ways but they sound more like hacks than proper solutions (e.g. add the deleted value to another field and filter on both...).

About live \ single-shot replication:
Thank you. I was in doubt about that, unfortunately adding "live: true" was giving me a timeout error :\

nolanlawson:
Hm, and actually, the funny thing is that there's actually a bug in your "deleted" test, because you expect there to be 0 docs when there should be 1 - because `'b'` was never deleted. :)

So actually, if your filter function is relying on doc `_id`s, it doesn't matter whether you `remove()` or `put()` with `_deleted: true`, because the minimum field-less deleted doc looks like this:

```
{
  _id: 'a',
  _rev: '1-x',
  _deleted: true
}
```

nolanlawson:
My personal recipe would be to add a check for `_deleted` in your filter function. You can just assume you always want to replicate deleted docs, because you want the client to know if anything was deleted (and it's not a big overhead to replicate empty docs).

sphaso:
Haha brilliant. I shouldn't write UTs at midnight. I'll commit a better test ASAP.

Regarding the recipe: I cannot check for _deleted, I'm not talking about deleted documents in this case. E.g.
{
    replicateTo: ['a','b','c'] 
}

I am 'c', my filter checks if I'm in the array. If someone server-side removes 'c' from the array, I want to know so I can delete that document from my personal replica of the DB. This is the problem that's been bugging me :\

nolanlawson:
Right, so what I'm saying is: just make your filter function:

```js
function (doc) {
  if (doc._deleted) {
    return true;
  }
  // other filtering logic goes here
}
```
Simple. :) Also, here's a PR showing the passing tests: https://github.com/pouchdb/pouchdb/pull/3613

sphaso:
I'm not deleting the document. e.g.
{
   _id: "pleaseDontDeleteMe",
   replicateTo: ['a','b','c']
}

Someone server-side updates this document into:

{
   _id: "pleaseDontDeleteMe",
   replicateTo: ['a','b']
}

note the missing 'c'. My filter function is something like:

function (doc) {
  if(doc.replicateTo.indexOf('c') > 0){
     return true;
  }
  return false;
}

so the updated version won't pass the filter and the end-point 'c' will never know it's not supposed to see that doc. I don't want to delete the document because 'a' and 'b' still need it.

nolanlawson:
Right, so you need to not do that. ;)

I'm not 100% sure what your scenario is, but it sounds like it may be complex enough that you should be avoiding filtered replication entirely. PouchDB (and CouchDB for that matter) is really optimized for replicating entire databases. If you add some server-side logic to your app to divvy up databases into exactly what you want to replicate, then all of this could be handled more smoothly IMO.

For example, if you are trying to divide documents by user or by role or something, then you could use the ["db per user"](https://github.com/nolanlawson/pouchdb-authentication#couchdb-authentication-recipes) pattern, which is pretty easy to abstract into a "db per role" pattern if need be.

sphaso:
Thank you for the tips. Unfortunately I cannot use either DB per user nor DB per role. I have changing groups of user sharing documents, the users of each group can change each day (one user can be in one group or more than one group). I've tried to explain the scenario here http://stackoverflow.com/questions/28874087/couchdb-replication-strategy-with-dynamic-groups-of-users

Feel free to continue the discussion there :)

nolanlawson:
@sphaso I discovered there's actually no bug. We already have the correct behavior, despite my initial pessimism. https://github.com/pouchdb/pouchdb/pull/3615 :) Like Christmas morning!

I'll comment on your stack overflow question.

nolanlawson:
Closing this issue, though, because I think we've determined this is a bug in the documentation more than anything. PouchDB is working exactly like CouchDB.

sphaso:
This means that Futon "delete document" triggers "remove" and not a put {_delete: true}. Very good news! And we learned something in the process :)

PR #3605, opened 2015-03-07
(#3605) - fix immediate cancel of live replication

nolanlawson:
The example from the docs where the user immediately
calls `cancel()` on a live replication is actually
broken, because we don't detect it if it's canceled
in the same turn of the event loop. This fixes that.


daleharvey:
https://github.com/pouchdb/pouchdb/commit/f58dbf5f447dcda473b2da2254a5d04144c1fc9c

Issue #3608, opened 2015-03-07
Better documentation for filtering on filters/query_params/doc_ids/views


nolanlawson:

sphaso:
Hey Nolan. While you work on the docs, I was wondering about writing a little blog post detailing my experience with filtered replication, what do you think?

nolanlawson:
@spaso please do! Would love to have a guest blog post

nickcolley:
:+1:  Would be great.

sphaso:
Great! I have a first draft ready, will probably issue a PR in the weekend :)

SinanGabel:
Hi!
It seems that with a filtered replication this works:
```
{
  _id: '_design/mydesign',
  filters: {
    myfilter: function (doc, req) {
      return doc.type === req.query.type;
    }.toString()
  }
}

# but this does not
{
  _id: '_design/mydesign',
  filters: {
    myfilter: function (doc, req) {
      return doc.key === req.query.type;
    }.toString()
  }
}
```
even though (assuming) the .key would be supposed to match .type
Is this on purpose?

Thanks,

Best,
Sinan

nolanlawson:
@SinanGabel Yeah, there is no relationship between `req.query` and `doc`; you have to define that yourself.

PR #3612, opened 2015-03-07
(#3608) - clarify filtered replication/changes

nolanlawson:
Filtered replication is way too complicated. Our goal is consistency with CouchDB, though, so I'm glad we implemented all these different options.

But golly, there are a lot.

### DIFF
--- a/docs/_includes/api/replication.html
+++ b/docs/_includes/api/replication.html
@@ -14,11 +14,17 @@ All options default to `false` unless otherwise specified.
 
 * `options.live`: If `true`, starts subscribing to future changes in the `source` database and continue replicating them.
 * `options.retry`: If `true` will attempt to retry replications in the case of failure (due to being offline), using a backoff algorithm that retries at longer and longer intervals until a connection is re-established. Only applicable if `options.live` is also `true`.
-* `options.filter`: Reference a filter function from a design document to selectively get updates.
-* `options.query_params`: Query params sent to the filter function.
-* `options.doc_ids`: Only replicate docs with these ids (array of strings).
 * `options.since`: Replicate changes after the given sequence number.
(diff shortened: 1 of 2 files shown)


docs/_includes/api/replication.html
```diff
@@ -14,11 +14,17 @@
 
 * `options.live`: If `true`, starts subscribing to future changes in the `source` database and continue replicating them.
 * `options.retry`: If `true` will attempt to retry replications in the case of failure (due to being offline), using a backoff algorithm that retries at longer and longer intervals until a connection is re-established. Only applicable if `options.live` is also `true`.
-* `options.filter`: Reference a filter function from a design document to selectively get updates.
-* `options.query_params`: Query params sent to the filter function.
-* `options.doc_ids`: Only replicate docs with these ids (array of strings).
 * `options.since`: Replicate changes after the given sequence number.
-* `options.create_target`: Create target database if it does not exist. Only for server replications.
```
nolanlawson: `create_target` is not needed anymore, because we don't have server-initiated replication anymore

nolanlawson:
Closing #3616 because I can handle that in this PR.

nolanlawson:
Updated to include info about filtering deleted docs.

daleharvey:
Doc updates look great, thanks

https://github.com/pouchdb/pouchdb/commit/22c6b4be03b9a50792bee13f380952f335e899fc

docs/_includes/api/replication.html
gr2m: Hey Nolan, I totally missed that `create_target` has been removed. Does that mean that if I replicate from `db1` to `db2`, `db2` will now always be created if it does not exist yet? Or does `create_target` still have an effect for that? We use the option here: https://github.com/hoodiehq/pouchdb-hoodie-sync/blob/master/lib/connect.js

PR #3613, opened 2015-03-07
(#3578) - tweak tests, confirm filtered deletions work

nolanlawson:
These tests just confirm that deleted filtering works as expected, when the filter is purely based on the doc `_id`s. In this case, it doesn't matter whether `remove()` or `_deleted=true` is used, because the doc has an `_id` either way.


daleharvey:
Looks good, cheers

https://github.com/pouchdb/pouchdb/commit/725185a09d6173db6802d685fe9c2c36a222e1fb

PR #3614, opened 2015-03-07
(#3578) - failing test to demo deleted+filtered replication

nolanlawson:
I think I might have found a bug in CouchDB. I'll file it over there. This is currently failing for all pairs of http and local.


nolanlawson:
Of course it's not a bug in CouchDB; just a bug in my tests. :P

PR #3615, opened 2015-03-07
(#3578) - confirm filtered replication works for deletion

nolanlawson:
If you use `remove()`, then filtering on fields should no longer
work, because you have removed all fields.

If you use `put()` with `_deleted=true`, then filtering on fields
should continue to work, because the field is still there.

These tests confirm that CouchDB and PouchDB both exhibit the
correct behavior for this.

### DIFF
--- a/tests/integration/test.replication.js
+++ b/tests/integration/test.replication.js
@@ -3560,6 +3560,77 @@ adapters.forEach(function (adapters) {
       });
     });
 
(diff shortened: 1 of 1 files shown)


daleharvey:
Merged in https://github.com/pouchdb/pouchdb/commit/725185a09d6173db6802d685fe9c2c36a222e1fb

PR #3618, opened 2015-03-08
Support for new SqlitePlugin params

tannerlinsley:
This is working for me and my current build with Ionic + sqlitePlugin + PouchDB


nolanlawson:
Cool stuff. The jshint isn't passing, and the style doesn't quite match ours, but I can clean it up a bit for you. Looking good!

tannerlinsley:
Thanks! I literally needed it asap to submit an app to the appstore tonight, so it was a bit thrown together ;)  Thanks for taking a look at it!

tannerlinsley:
Sorry, I accidentally added to this commit.  I just did a revert on it.

nolanlawson:
What is the difference between `location` and `createFromLocation`? Cannot figure it out from reading the SQLite Plugin docs...

nolanlawson:
Also I'm trying to add documentation for this, but I'm struggling to understand why you would even need this. :/  (Android developer here.)

tannerlinsley:
The way I understand it is `location` is specifically for the location of the db on iOS (location is a big factor for if the database is backed up by iCloud/iTunes or not and this is usually the case for an app getting rejected in the app store for not following data storage guidelines).

`createFromLocation` looks like a flag to basically init the db from a saved database file in `www/`

I agree, I think it's dumb that it's a needed setting.  I've been battling iOS storage guidelines and rejections for weeks now, and finally things are coming together for the cordova/phonegap/ionic world on the issue.

nolanlawson:
Wait, so Apple will reject your app *because* it is backed up to iCloud? And there's a difference between iCloud and iTunes backup??

tannerlinsley:
Haha no.  My bad.  A lot of hybrid apps download assets or databases to operate on the phone, but these assets aren't exactly user specific or changing.  They're just their to lessen the load on a server and what not.  *That's* what apple doesn't allow, as it would create unnecessary storage on their servers of data that shouldn't be backed up per se.

tannerlinsley:
And yes, there is still a small difference between the two backups.  iTunes is local, and much less strict on storage guidelines than using iCloud for backup.  Most users prefer iCloud backup.

nolanlawson:
Good to know. In any case, I have an alternate version of this PR in #3619. Could you take a look and let me know if it works for you? I'll also test it out in my [SQLite Plugin sample app](https://github.com/nolanlawson/pouchdb-cordova-hello-world-with-sqlite-plugin).

PR #3621, opened 2015-03-09
upgrade derequire

calvinmetcalf:


calvinmetcalf:
its ~60 times faster

nickcolley:
At runtime?

nolanlawson:
tried this out recently in another project, can confirm way faster

nolanlawson:
BTW I was joking on Twitter about how you should increment the major version for every performance improvement. Hope that's not why you incremented it. ;)

calvinmetcalf:
haha no since it switches from esprima to acorn code that it could previously parse it could theoretically error on.

PR #3637, opened 2015-03-14
(#3585) - update to es7 post

nolanlawson:


calvinmetcalf:
Maybe point out its not finalized yet and subject to change

nolanlawson:
I do say

> ES7 is still very bleeding-edge. Async functions aren't supported in either Node.js or io.js, and you have to set some experimental flags to even get Babel to consider it.

nolanlawson:
Fixed a bug in the code pointed out on twitter, and also added a note about the async/await spec being in the the proposal stage, per Calvin. Also, because this article is getting a lot of attention on Twitter right now, Ima just go ahead and publish, but still leaving this PR open for review. :)

daleharvey:
Yay https://github.com/pouchdb/pouchdb/commit/ca1e79bb5ad86009fd88b8da937f442a4dbe51c1

Issue #3319, opened 2014-12-31
WebPack related error: LevelUp package.json produces errors


andru:
When packaging PouchDB in an app with WebPack, the following error gets output. The current PouchDB npm release doesn't build because of a bug recently fixed (see #3287). The current github master allows it to proceed but it then hits the error below.

Unless I'm mistaken LevelUP shouldn't be required in the client? Can it be left out of the package altogether in a similar manor to the fix for #3287 ?

Is it ok to just add `levelup: false` to the `browser` section of `package.json` or should it be replaced with a browser compatible module?
```
ERROR in ./~/pouchdb/~/levelup/package.json
Module parse failed: /Volumes/Files/Projects/Harvest/app2/node_modules/pouchdb/node_modules/levelup/package.json Line 2: Unexpected token :
You may need an appropriate loader to handle this file type.
| {
|   "name": "levelup",
|   "description": "Fast & simple storage - a Node.js-style LevelDB wrapper",
|   "version": "0.18.6",
 @ ./~/pouchdb/~/levelup/lib/util.js 102:30-56
```

svnlto:
@daleharvey I've picked this up again with more recent versions of `webpack` and `pouchdb`. I've had to make some addition to PouchDB's `package.json` file to get it working this time round:

```json
"browser": {
  "levelup": false,
  "leveldown": false
}
```

svnlto:
I had to remove the `noParse`. So now I'm facing another issue. Not quite sure if this is the right place tho..

```bash
WARNING in ./~/pouchdb/~/lie/~/immediate/dist/immediate.js
Critical dependencies:
1:440-447 This seem to be a pre-built javascript file. Even while this is possible, it's not recommended. Try to require to orginal source to get better results.
 @ ./~/pouchdb/~/lie/~/immediate/dist/immediate.js 1:440-447

WARNING in ./~/pouchdb-authentication/~/lie/~/immediate/dist/immediate.js
Critical dependencies:
1:440-447 This seem to be a pre-built javascript file. Even while this is possible, it's not recommended. Try to require to orginal source to get better results.
 @ ./~/pouchdb-authentication/~/lie/~/immediate/dist/immediate.js 1:440-447
````
/cc @nolanlawson @calvinmetcalf

nolanlawson:
Can you provide a PR? I'm not too familiar with webpack, but we definitely want it to at least work in webpack.

As for the warnings, they don't seem too severe. :)

[...]

PR #3639, opened 2015-03-14
(#3319) - explicitly set leveldown and levelup to false in browser env

svnlto:
more work related to webpack


svnlto:
Not sure about those test failures having anything to do with this change to be honest..

nolanlawson:
I would say it's a Travis intermittent, but those errors are all related to the browserification of the adapter "plugins" (memory, idb-alt, localstorage). So probably the browserify process just needs to be harmonized between the two.

package.json
```diff
@@ -111,7 +111,9 @@
     "./lib/adapters/preferredAdapters.js": "./lib/adapters/preferredAdapters-browser.js",
     "./lib/deps/buffer.js": "./lib/deps/buffer-browser.js",
     "bluebird": "lie",
-    "fs": false
+    "fs": false,
+    "levelup": false,
```
nolanlawson: Ah yeah `levelup: false` will cause problems when trying to build the adapter plugins. I assume if we set them from the browserify command line in the [plugin build script](https://github.com/pouchdb/pouchdb/blob/7a7d0310afcc6cfca0509e36d75986f6ce62d313/bin/build-plugin.sh), it'll ignore this value in the package.json, yes?

svnlto: > I assume if we set them from the browserify command line in the plugin build script, it'll ignore this value in the package.json, yes?

I'd assume so.

daleharvey:
Yup it looks like these failures are genuine, will take a look once I am caught up

nolanlawson:
This PR may or may not be made easier by https://github.com/pouchdb/pouchdb/pull/3786. @svnlto how are you checking that webpack works? I don't know anything about webpack.

nolanlawson:
OK, it doesn't make it any easier. The same thing that fixes webpack breaks browserify.

In short, I can add `"levelup": false, "leveldown": false` to the `package.json`, and this allows Webpack to build PouchDB without any errors (`npm install -g webpack && webpack ./ outfile.txt`). But then the adapter plugins don't build correctly because they are missing `levelup`.

I could fix this by adding `-r levelup` to the browserify build script for the plugins, but then that would break for people trying to use the plugins via the `extras` API, unless they knew to add `-r levelup` themselves, and I'd rather not oblige people to do that.

I think the real fix here is in either `levelup` or `leveldown`. They do not seem to support Webpack. For the record, the full error output is:

```
ERROR in ./~/levelup/package.json
Module parse failed: /Users/nolan/workspace/pouchdb/node_modules/levelup/package.json Line 2: Unexpected token :
You may need an appropriate loader to handle this file type.
| {
|   "name": "levelup",
|   "description": "Fast & simple storage - a Node.js-style LevelDB wrapper",
|   "version": "0.18.6",
 @ ./~/levelup/lib/util.js 102:30-56

ERROR in ./~/leveldown/~/bindings/bindings.js
Module not found: Error: Cannot resolve module 'fs' in /Users/nolan/workspace/pouchdb/node_modules/leveldown/node_modules/bindings
 @ ./~/leveldown/~/bindings/bindings.js 6:9-22
```

nolanlawson:
closing in favor of #3789; I think I got this figured out @svnlto

nolanlawson:
For those who are curious about the json fix, you just add a `webpack.config.js` file to your root with the following:

```js
module.exports = {
    context: __dirname,
    entry: "./lib/your-root-file.js",
    module: {
        loaders: [
            {
                test: /\.json$/,
                loader: "json-loader" 
            }
        ]
    }
}
```

And you actually **cannot** fix this at the module level. So e.g. even if I opened a pull request on levelup to add this file to their repo; it wouldn't fix the problem. Webpack only looks at the `webpack.config.js` in the top-level directory. By design, it's supposed to be app developers who control this stuff.

PR #3656, opened 2015-03-20
Fix #3646

### LINKED ISSUE #3646: websql by-sequence table sometimes stores deleted flag from wrong rev
We are having a problem where replication-to-WebSQL winds up with too many documents, while IndexedDB works fine. Once this happens replication will never result in the remote and local databases containing the same number of documents. The following demo shows the same problem, but with pouchdb-load instead of pouchdb-bulk-get.

https://github.com/dholth/pouchdb-websql-bug

Instead, WebSQL should wind up with the same number of documents as IndexedDB and should not care if an older dump is loaded into the database first.

dholth:
Please edit to make sure the test runs under both websql and idb when available.

### DIFF
--- a/lib/adapters/websql/websql-bulk-docs.js
+++ b/lib/adapters/websql/websql-bulk-docs.js
@@ -104,7 +104,8 @@ function websqlBulkDocs(req, opts, api, db, Changes, callback) {
 
     function finish() {
       var data = docInfo.data;
-      var deletedInt = deleted ? 1 : 0;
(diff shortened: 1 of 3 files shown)


nolanlawson:
Whoa, thanks a lot! You saved me like a full weekend of work; this looks amazing. :smiley:

Looks like JSHint is failing... did you run `npm run jshint` before submitting? Also `npm test` would catch that.

nolanlawson:
closing in favor of #3659

PR #3657, opened 2015-03-20
(#3656) - fix up jshint and promise test

nolanlawson:
@dholth I made some corrections to https://github.com/pouchdb/pouchdb/pull/3656, please take a look. I understand that our test suite is kinda complicated, but basically:

* tests should test both the http and local adapters, so that we can know what CouchDB says is the truth :)
* the Mocha promise format is waaay easier to work with than the callback format (i.e. "done")

I have good new and bad news. The good news is that your test is passing in CouchDB, IndexedDB, and WebSQL. The bad news is that it is failing in LevelDB, and that it doesn't actually capture the change you made to `websql-bulk-docs`, because it succeeds in WebSQL before and after the commit. :( And unfortunately I don't have time today to look into the LevelDB issue, but it seems you may have found some other bug in there.

Let me know if you need help with this; I'll be on IRC all day today. And thanks again for your help trying to track it down; it is *much much* appreciated.

### DIFF
--- a/lib/adapters/websql/websql-bulk-docs.js
+++ b/lib/adapters/websql/websql-bulk-docs.js
@@ -104,7 +104,8 @@ function websqlBulkDocs(req, opts, api, db, Changes, callback) {
 
     function finish() {
       var data = docInfo.data;
-      var deletedInt = deleted ? 1 : 0;
+      // #3646 deleted passed to writeDoc is not for BY_SEQ_STORE
+      var deletedInt = docInfo.metadata.deleted ? 1 : 0;
 
       var id = data._id;
       var rev = data._rev;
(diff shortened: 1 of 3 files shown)


nolanlawson:
Also some tips:

You can test just what you want in Node by doing

    GREP=mytestname npm run test-node

And you can test WebSQL in the browser by adding the query param

    localhost:8000/testsintegration?grep=mytestname&adapters=websql

nolanlawson:
Also another tip: when you do `Promise.all`, it executes all of those promises concurrently, so if you are trying to reproduce a bug where the documents are inserted in a particular order, you may want to do them sequentially instead. I have an example of sequential promises in [this blog post](http://pouchdb.com/2015/03/05/taming-the-async-beast-with-es7.html) (search for "promise chain").

dholth:
If you do take a hard look at it, don't forget https://github.com/dholth/pouchdb-websql-bug/

nolanlawson:
closing in favor of #3659

PR #3659, opened 2015-03-21
(#3646) - fix deleted conflicts in all 3 adapters

nolanlawson:
This bug appears when a deleted revision is added, which is not the winning
revision. It caused errors both in how `doc_count`/`total_rows` was
calculated, as well as how the winning doc was returned.

The bug appeared in all three adapters.

The test captures both failures. I can confirm that the test fails before
the fix and succeeds afterwards.

### DIFF
--- a/lib/adapters/idb/idb-bulk-docs.js
+++ b/lib/adapters/idb/idb-bulk-docs.js
@@ -185,8 +185,8 @@ function idbBulkDocs(req, opts, api, idb, Changes, callback) {
     });
   }
 
-  function writeDoc(docInfo, winningRev, deleted, callback, isUpdate,
-                    delta, resultsIdx) {
+  function writeDoc(docInfo, winningRev, winningRevIsDeleted, newRevIsDeleted,
+                    isUpdate, delta, resultsIdx, callback) {
 
     docCountDelta += delta;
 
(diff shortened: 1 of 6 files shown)


lib/adapters/idb/idb-bulk-docs.js
```diff
@@ -185,8 +185,8 @@ function idbBulkDocs(req, opts, api, idb, Changes, callback) {
     });
   }
 
-  function writeDoc(docInfo, winningRev, deleted, callback, isUpdate,
-                    delta, resultsIdx) {
+  function writeDoc(docInfo, winningRev, winningRevIsDeleted, newRevIsDeleted,
+                    isUpdate, delta, resultsIdx, callback) {
```
nolanlawson: ^ not having the callback at the end was bugging me

lib/adapters/leveldb/leveldb.js
```diff
@@ -498,22 +498,22 @@ function LevelPouch(opts, callback) {
       return complete();
     }
 
-    function writeDoc(doc, winningRev, deleted, callback2,
-                      isUpdate, delta, index) {
+    function writeDoc(docInfo, winningRev, winningRevIsDeleted, newRevIsDeleted,
+                      isUpdate, delta, resultsIdx, callback2) {
```
nolanlawson: I felt that `leveldb.js` should use the same variable names as idb/websql.

daleharvey:
Very nice catch

https://github.com/pouchdb/pouchdb/commit/a7b7c89bfbf3befa5276ef93685098cb47d9e9ea

Issue #3640, opened 2015-03-15
High RAM usage while storing large PDF as blob, results in a tab crash in Google Chrome


KSchreyer:
When storing large pdf documents as blob, the RAM usage rises up to 1.6GB and more. This issue report is the result of the chromium bug report https://code.google.com/p/chromium/issues/detail?id=412409 and an email conversation with Nolan Lawson.
Also in this chromium ticket, I've added a test script to reproduce the problem. This can be found here:
https://code.google.com/p/chromium/issues/detail?id=412409#c18

As Chris Mumford describes (https://code.google.com/p/chromium/issues/detail?id=412409#c20), the RAM usage rises up to 1,7GB in Firefox until the process finishes. In Google Chrome this currently results in a tab crash, because of a heap exhaustion.

nolanlawson:
[Test script](https://code.google.com/p/chromium/issues/detail?id=412409#hc18)

nolanlawson:
Interestingly, this does not crash Firefox either.

nolanlawson:
OK, the line that causes the crash is [this one](https://github.com/pouchdb/pouchdb/blob/b0b0200ee9a417233c83033538176e13fce3f4f4/lib/utils.js#L740).

Because we don't support blobs in Chrome < latest Canary, we throw an error when the input is too large, because we create gigantic base64 strings.

I suppose one fix is that we could probably get away with storing binary strings in IDB instead of base64 strings. It's never been tested, though.

nolanlawson:
Hm, binary string seems to crash it as well.

nolanlawson:
Developers who are using PouchDB can work around this issue by just breaking up their attachments into smaller chunks. This is probably a good idea anyway when it comes to replication, because we currently don't support multipart/mixed, and so attachments will be converted to base64 strings anyway in order to be replicated to CouchDB.

nolanlawson:
After profiling and actually spending a lot of time on this, I'm pretty sure this is not something we can fix.

The fundamental problem is that Chrome <43 has Blob bugs, so we cannot use Blobs. Hence we have to fall back to base64, which is fundamentally inefficient since we have to keep it as a string in memory.

This is a Chrome bug that will be fixed with 43. If your code needs to run on Chrome <43 I would recommend splitting up your attachment into many smaller attachments. Sorry about that.

PR #3661, opened 2015-03-21
(#3640) - simplify md5sum for large binaries

nolanlawson:
This doesn't fix the "crashiness" problem. (Interesting, Chrome seems to have solved that itself - it no longer crashes as of Chrome Canary when I try a very large file with the provided test.)

However, this does simplify the code, so that we can start actually fixing the root of the problem, which I believe is in SparkMD5, notably [these lines](https://github.com/satazor/SparkMD5/blob/0d8d2a772ab2ef1fd66fec41bb6cb3665e6e3ccc/spark-md5.js#L497-L499).

### DIFF
--- a/lib/deps/md5.js
+++ b/lib/deps/md5.js
@@ -5,49 +5,6 @@ var Md5 = require('spark-md5');
 var setImmediateShim = global.setImmediate || global.setTimeout;
 var MD5_CHUNK_SIZE = 32768;
 
-function sliceShim(arrayBuffer, begin, end) {
-  if (typeof arrayBuffer.slice === 'function') {
-    if (!begin && !end) {
-      return arrayBuffer.slice();
-    } else if (!end) {
-      return arrayBuffer.slice(begin);
-    } else {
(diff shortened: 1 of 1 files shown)


nolanlawson:
BTW apparently `ArrayBuffer.prototype.slice` is not only poorly supported; it also apparently [copies all the contents over](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer?redirectlocale=en-US&redirectslug=Web%2FJavaScript%2FTyped_arrays%2FArrayBuffer). One of the nice properties of `new Uint8Array(buff)` is that it doesn't actually appear to allocate any extra memory, if my reading of [this html5rocks article](http://www.html5rocks.com/en/tutorials/webgl/typed_arrays/) is correct.

daleharvey:
Always nice to delete some code 

https://github.com/pouchdb/pouchdb/commit/3183712def81d38c10748dae0da98fbc0b265979

Issue #3658, opened 2015-03-20
Replicate from local to remote with filters in ddoc


sphaso:
Hi, it's me again.           
I've solved (most of) my issues with filtered replication from remote to local, now I'm testing local to remote.
Some good ol' code.

Sync method:
```Javascript
localDB.sync(remoteDB, {
            live: true,
            retry: true,
            filter: 'app/by_user',
            query_params: { "agente": agent }
        })
        .on('change', function(result){
                result.change.docs.forEach(function(change) {
                    if(!change.deleted){
                        console.log("CHANGE!");
                        $rootScope.$apply(function(){
                            $rootScope.$broadcast('upsert', change);
                        });
                    }
                });
        });
```

Save method:
```Javascript
this.save = function(bolla){
            return localDB.get(bolla._id).then(function(doc){
                doc.DataEsposizione = bolla.DataEsposizione;

                return localDB.put(doc);
            });
        };
```

The problem:
this last promise returns "ok: true" and everything seems to be in order, but this local change never gets replicated to the remote DB. If instead of "localDB.put" I write "remoteDB.put" everything works. Of course I cannot use remoteDB.put because I'm writing an offline-first app.

Am I doing anything grossly wrong?
Thank you :)

nolanlawson:
Can you write a live reproducible test case please? I don't see anything obviously wrong here, so it would be helpful to have a live example.

sphaso:
Yes, of course, I just wanted to make sure I wasn't doing anything egregiously wrong first. Test coming up ASAP!

sphaso:
Victory! 
As you suggested in the PR, the ddoc NEEDS to be replicated to the local DB for the whole syncing process to work. This was not really intuitive to me, since simply doing remoteDB.put works instead of localDB.put and my thought process was "the ddoc stays on the server, the client replicates back whatever the server gives to it". Bi-injectivity of ddoc rules was not necessary nor imposed (still, in my skewed understanding).
After many failed attempts at replicating from \ to with different rules I simply threw the towel and relaxed the replicating rules in my filter. That's the result:

```Javascript
function(doc, req) { if(doc._id == '_design/app' || doc.Agents && doc.Agents.indexOf(req.query.agente) > -1){ return true; } return false; 
}
```

It would be nice to have it clearly stated in the docs, as I imagine other people might run into the same issue (or I might be exceptionally dumb).
Thanks!

nolanlawson:
[...]

PR #3662, opened 2015-03-21
failing test

sphaso:
#3658

### DIFF
--- a/tests/integration/test.replication.js
+++ b/tests/integration/test.replication.js
@@ -3625,7 +3625,7 @@ adapters.forEach(function (adapters) {
         {_id: 'b', dank: false},
         {_id: 'c', dank: false}
       ]).then(function () {
-	db.replicate.to(remote, {filter: 'myddoc/myfilter', live: true});
+	db.replicate.to(remote, {live: true});
         return remote.replicate.to(db, {filter: 'myddoc/myfilter'});
       }).then(function () {
         return db.get('a');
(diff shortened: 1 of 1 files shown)


tests/integration/test.replication.js
```diff
@@ -3609,6 +3609,37 @@ adapters.forEach(function (adapters) {
       });
     });
 
+    it('#3658 replication with ddoc filter, update doc', function() {
+      var db = new PouchDB(dbs.name);
+      var remote = new PouchDB(dbs.remote);
+
+	      return remote.bulkDocs([{
```
nolanlawson: It seems the bug is that you are putting the design doc in the remote database. If you are replicating *from* the local database, then that is where the ddoc should go. :)

sphaso:
I'll try to remove the filter instruction in the local -> remote replication. 
So when I use "sync", does the ddoc needs to be in both local and remote DB?

nolanlawson:
> So when I use "sync", does the ddoc needs to be in both local and remote DB?

Yes.

sphaso:
This is very counterintuitive (at least to me). I guess I could change the ddoc filter remote-side to have it replicated locally ( like "conditions... || doc.id.startsWith("_design"), then I could use "sync" without worrying about it.
I removed the "filter" option from local -> remote replicate but the test is failing anyway :/

daleharvey:
Filtered replication got fixed and looks like this got cleared up, closing, Thanks a lot

PR #3664, opened 2015-03-25
(#3664) - update docs on chrome blobs

nolanlawson:
I've tested and can confirm that Chrome
has finally resolved all the blob issues
as of Chrome v43 (currently Canary). This
will eventually make it into Android Lollipop,
and we can update the docs again when that
happens.


daleharvey:
Merged in https://github.com/pouchdb/pouchdb/commit/59654323e4b37620b412fea3de628d6ad2de7892

Cheers

PR #3669, opened 2015-03-26
Update queries.md

oliviertassinari:


daleharvey:
Good catch, thanks a lot for the PR 

Sorry for the delay, you caught a time when most of the maintainers were out of action, its usually much faster :)

https://github.com/pouchdb/pouchdb/commit/ae3c7042f99c9da2d6b56fb30088e8122636558a

PR #3678, opened 2015-03-30
Make sure PouchDB works in Service Workers

KlausTrainer:
In Chromium, `XMLHttpRequest` isn't exposed to Service Workers anymore.
See https://src.chromium.org/viewvc/blink?revision=178829&view=revision.

This changes `request-browser` to fall back to the new Fetch API if
`XMLHttpRequest` is `undefined`.

### DIFF
--- a/lib/deps/request-browser.js
+++ b/lib/deps/request-browser.js
@@ -5,28 +5,30 @@
 var createBlob = require('./blob.js');
 var utils = require('../utils');
 
-var wrappedFetch = utils.getArguments(function(args) {
-  var wrappedPromise = {};
+function wrappedFetch() {
+  return (utils.getArguments(function(args) {
+    var wrappedPromise = {};
 
-  var promise = new utils.Promise(function(resolve, reject) {
-    wrappedPromise.resolve = resolve;
-    wrappedPromise.reject = reject;
-  });
+    var promise = new utils.Promise(function(resolve, reject) {
+      wrappedPromise.resolve = resolve;
+      wrappedPromise.reject = reject;
+    });
 
-  wrappedPromise.then = promise.then.bind(promise);
-  wrappedPromise.catch = promise.catch.bind(promise);
-  wrappedPromise.promise = promise;
+    wrappedPromise.then = promise.then.bind(promise);
+    wrappedPromise.catch = promise.catch.bind(promise);
+    wrappedPromise.promise = promise;
 
-  fetch.apply(null, args).then(function(response) {
-    wrappedPromise.resolve(response);
-  }, function(error) {
-    wrappedPromise.reject(error);
-  }).catch(function(error) {
-    wrappedPromise.catch(error);
-  });
+    fetch.apply(null, args).then(function(response) {
+      wrappedPromise.resolve(response);
+    }, function(error) {
+      wrappedPromise.reject(error);
+    }).catch(function(error) {
+      wrappedPromise.catch(error);
+    });
 
-  return wrappedPromise;
-});
+    return wrappedPromise;
+  }))(arguments);
+}
 
 function fetchRequest(options, callback) {
   var wrappedPromise, timer, fetchResponse;
--- a/lib/deps/request-browser.js
+++ b/lib/deps/request-browser.js
@@ -3,9 +3,7 @@
 var createBlob = require('./blob.js');
 var utils = require('../utils');
 
-function wrappedFetch() {
-  var args = Array.prototype.slice.call(arguments); // arguments to array
-
+var wrappedFetch = utils.getArguments(function(args) {
   var wrappedPromise = {};
 
   var promise = new global.Promise(function(resolve, reject) {
@@ -26,7 +24,7 @@ function wrappedFetch() {
   });
 
   return wrappedPromise;
-}
+});
 
 function fetchRequest(options, callback) {
   var wrappedPromise, timer, fetchResponse;
--- a/lib/deps/request-browser.js
+++ b/lib/deps/request-browser.js
@@ -1,3 +1,5 @@
+/* global fetch */
+/* global Headers */
 'use strict';
 
 var createBlob = require('./blob.js');
@@ -15,7 +17,7 @@ var wrappedFetch = utils.getArguments(function(args) {
   wrappedPromise.catch = promise.catch.bind(promise);
   wrappedPromise.promise = promise;
 
-  global.fetch.apply(null, args).then(function(response) {
+  fetch.apply(null, args).then(function(response) {
     wrappedPromise.resolve(response);
   }, function(error) {
     wrappedPromise.reject(error);
@@ -29,7 +31,7 @@ var wrappedFetch = utils.getArguments(function(args) {
 function fetchRequest(options, callback) {
   var wrappedPromise, timer, fetchResponse;
 
-  var headers = new global.Headers();
+  var headers = new Headers();
 
   var fetchOptions = {
     method: options.method,
--- a/lib/deps/request-browser.js
+++ b/lib/deps/request-browser.js
@@ -6,7 +6,7 @@ var utils = require('../utils');
 var wrappedFetch = utils.getArguments(function(args) {
   var wrappedPromise = {};
 
-  var promise = new global.Promise(function(resolve, reject) {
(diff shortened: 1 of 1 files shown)


daleharvey:
Sorry for the delay, just catching up on a huge backlog of PR's

This change is awesome, thanks a lot for doing it. 

I am somewhat worried since the chromium commit is 8 months ago and we arent seeing failures in our test suite going to take a look into that now

We are going to split out request-browser into a new module and being able to use one simple api "request()" and have it work in the browser in or out of service workers is going to be super handy.

What is the wrappedPromise for? it looks like it tries to do exactly what calling fetch plain would do except handle exceptions?

I am going to take a look at getting the tests to fail in chrome and hopefully this can fix a test instead of needing a new failure, I dont think I have any issues with the code, understanding `wrappedPromise` would be nice, also @calvinmetcalf you have any thoughts on this, you are the promise expert

lib/deps/request-browser.js
```diff
@@ -3,7 +3,111 @@
...
+      if (options.binary) {
+        result = response.blob();
+      } else {
+        result = response.text();
+      }
+    } else {
+      result = response.json();
+    }
+
+    return result;
```
daleharvey: Generally lean towards early returns and it looks like it would be cleaner?

   if (response.status >= 200 && response.status < 300) { 
     return options.binary ? response.blob() : response.text();
   } 
   return response.json;

KlausTrainer: Sure, thanks. I'll change it accordingly.

KlausTrainer:
Sorry for not getting back to it earlier.  I had a few great holidays :)

> I am somewhat worried since the chromium commit is 8 months ago and we arent seeing failures in our test suite going to take a look into that now

It looks like the respective change is still a Chromium-only issue, i.e., the change doesn't seem to have landed in Google Chrome yet.

> What is the wrappedPromise for? it looks like it tries to do exactly what calling fetch plain would do except handle exceptions?

The `wrappedFetch` function returning the `wrappedPromise` is currently necessary as a workaround, as it is currently neither possible to set a timeout for `fetch` requests, nor to abort a request by using the promise that is returned from the `fetch` call.  (See e.g. [here](http://jakearchibald.com/2015/thats-so-fetch/#whats-missing) for more information on the `fetch` API's current limitations.)

The `wrappedFetch` just exposes the `reject` function (amongst others) to the caller.  Note that `reject` is normally only exposed to the `fetch` callbacks.  The exposed `reject` function is used in order to (optionally) set a timeout, and also for the `abort` value in the returned object (i.e., `{abort: wrappedPromise.reject}`).

daleharvey:
> It looks like the respective change is still a Chromium-only issue, i.e., the change doesn't seem to have landed in Google Chrome yet.

Ok cool that makes sense, I dont think we particularly need new tests here, just as long as we dont break existing ones and make sure our SW coverage is good. I expect we will be seeing more SW related issues.

The wrapper is a shame but sounds reasonable. I think instinctively there is a good chance that we will want to switch from using `request` to using `fetch` and then write wrappers around request, it seems like fetch will likely become the new client standard for xhr, and it has a promise api.

Im :+1: on this providing the tests are stable (and we shouldnt be touching the new code path so they really should be)

KlausTrainer:
I fully agree with all your above points.

nolanlawson:
It looks like this is only failing due to intermittents, so I've restarted the failed jobs. Sorry about that, @KlausTrainer, welcome to our Travis situation. :stuck_out_tongue:

nolanlawson:
+1 when green, "green" meaning passing in everything but couch-master due to current couch-master wonkiness.

lib/deps/request-browser.js
```diff
@@ -3,7 +3,105 @@
 var createBlob = require('./blob.js');
 var utils = require('../utils');
 
-module.exports = function(options, callback) {
+function wrappedFetch() {
+  var args = Array.prototype.slice.call(arguments); // arguments to array
+
```
nolanlawson: ^ this leaks the arguments array; please use `utils.getArguments` instead.

lib/deps/request-browser.js
```diff
@@ -3,7 +3,105 @@
...
+  var wrappedPromise = {};
+
+  var promise = new global.Promise(function(resolve, reject) {
```
nolanlawson: should use `PouchDB.utils.Promise` instead.

lib/deps/request-browser.js
```diff
@@ -3,7 +3,105 @@
...
+    wrappedPromise.resolve = resolve;
+    wrappedPromise.reject = reject;
+  });
+
+  wrappedPromise.then = promise.then.bind(promise);
+  wrappedPromise.catch = promise.catch.bind(promise);
+  wrappedPromise.promise = promise;
+
+  global.fetch.apply(null, args).then(function(response) {
```
nolanlawson: Referring to `global` at all is dangerous due to Node Webkit (currently untested... I know). Better just to say `fetch`.

nolanlawson:
+1 after nits actually

KlausTrainer:
Thanks, @nolanlawson. I've added three more commits addressing your points. Just tell me if you'd like to have them squashed together before merging this.

nolanlawson:
I can squash 'em for you, no worries. :)

nolanlawson:
Hm, Travis failures seem genuine. Sorry, may have broken your PR by nitpicking... :sweat:

KlausTrainer:
It was still me who broke it ;) The problem was `utils.getArguments` still being `undefined`. Please see the commit that I just added, and check if you can find a better solution.

nolanlawson:
Instead of getArguments (which has an async API), you can also just use a normal for loop and Array.prototype.push, which will also not leak the arguments object

KlausTrainer:
Thanks :) Changed it accordingly.

daleharvey:
Any chance you can rebase @KlausTrainer? apologies, I think everyones :+1: on this but the fail although doesnt look related looks permanent, maybe been fixed along the way

KlausTrainer:
Just rebased against current master.

nolanlawson:
Looks like it will probably pass after a few restarts. Sadly, restarting Travis tests is a regular part of merging PRs for us... :unamused:

daleharvey:
Nope looks like the failure is genuine, surprised since code path shouldnt change unless its in somewhere with fetch() but will take a look

daleharvey:
@marten-de-vries Hey Martin, do you have any idea how this could be happening, basically if I run the tests with ?grep=3606 against pouchdb server it fails on the 6th test run, it looks like either a _changes or an _all_docs call, possibly both are stalling within pouchdb server itself

Failing in firefox and chrome, where the code path for the pouchdb client should have changed in the slightest and pouchdb-server code hasnt changed at all

daleharvey:
I can reproduce this with 

```
~/src/pouchdb (master)$ git diff
diff --git a/lib/deps/request-browser.js b/lib/deps/request-browser.js
index 46a01f5..23f1520 100644
--- a/lib/deps/request-browser.js
+++ b/lib/deps/request-browser.js
@@ -3,7 +3,7 @@
 var createBlob = require('./blob.js');
 var utils = require('../utils');

-module.exports = function(options, callback) {
+function xhRequest(options, callback) {

   var xhr, timer, hasUpload;

@@ -108,3 +108,7 @@ module.exports = function(options, callback) {

   return {abort: abortReq};
 };
+
+module.exports = function(options, callback) {
+  xhRequest(options, callback);
+};
```

daleharvey:
ok figured it out, need to return the value of `xhRequest` so requests can be aborts, this should really be done for the `wrappedPromise` as well however I figure I will test the fix in https://github.com/pouchdb/pouchdb/pull/3799 land it so we can get a bit of this tested, then can either fix abort: after or by that time will hopefully be done with my fetch path

daleharvey:
Woot, #3799 is green, thanks so much @KlausTrainer 

Merged in https://github.com/pouchdb/pouchdb/commit/6c13541cb1f7ab0ea459574da0e1b50822d9ad7e

KlausTrainer:
Thank you all, and especially Dale for fixing this up :)

nolanlawson:
:tada: nice work!!

PR #3681, opened 2015-03-31
(fix #3680) - Properly encode special document ids

### LINKED ISSUE #3680: Http adapter should handle design documents containing another /
I have (accidently) created a design document with an id like `_design/foo/bar`, which completely broke the replication. Even after I deleted the document it still caused the replication to fail, understandably because the deleted revision has to be replicated, too.

The problem here is that the docid does not get properly escaped, resulting in a request to `/dbname/_design/foo/bar` instead of `/dbname/_design/foo%2Fbar`.

jo:
Correctly encode design and local document ids, like `_design/foo/bar` in http adapter.

### DIFF
--- a/lib/adapters/http/http.js
+++ b/lib/adapters/http/http.js
@@ -16,8 +16,11 @@ var isBrowser = typeof process === 'undefined' || process.browser;
 var buffer = require('../../deps/buffer');
 
 function encodeDocId(id) {
-  if (/^_(design|local)/.test(id)) {
-    return id;
+  if (/^_design/.test(id)) {
+    return '_design/' + encodeURIComponent(id.slice(8));
+  }
+  if (/^_local/.test(id)) {
+    return '_local/' + encodeURIComponent(id.slice(7));
   }
   return encodeURIComponent(id);
 }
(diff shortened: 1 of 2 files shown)


daleharvey:
Hey @jo 

Apologies for the delay, hit pouch at a quiet time :) looks like there is a small jshint error here, I fixed it and pushed a test run @ https://github.com/pouchdb/pouchdb/pull/3693

Otherwise looks perfect, especially thanks for the test

jo:
Thanks @daleharvey for the fix!
I have rebased the branch, picked your commit and squashed them. Lets see if the tests will now pass...

daleharvey:
Awesome thanks, all passing now. Tiny nit is that we just reference the issue # in the commit message, so (#1234) instead of (fix #1234), amended on the way in

Thanks - https://github.com/pouchdb/pouchdb/commit/49248fac87836978abf76cd08223c1376f8d67f0

jo:
Thanks, and sorry for the commit message mess.

Btw. I had to restart [some](https://travis-ci.org/pouchdb/pouchdb/jobs/57143584) [test](https://travis-ci.org/pouchdb/pouchdb/jobs/57143570) jobs several times, they sometimes failed on compactation.

PR #3690, opened 2015-04-02
Updated link for Felix's Node.js Style Guide

yaronyg:


yaronyg:
Waiting for a +1 before I commit it.

daleharvey:
Whats the reason for the switch? 

@felixge any reason to prefer one over the other?

yaronyg:
Because the current link says explicitly that it's outdated and one should instead check the second link. So I put in the second link.

nolanlawson:
63a0a10b11b59b3c6bb3c733f50c50585777cf90

PR #3691, opened 2015-04-02
(#3687) - return to pending state when live replication pauses

### LINKED ISSUE #3687: active event only fired once, the first time live replication starts
When using live replication, the 'active' event is correctly fired only the very first time replication starts actively processing changes.  This is eventually followed by a 'paused' event, but then the next batch of replication does not fire the 'active' event (the 'change' events and the following 'pause' both fire).  This seems like it may have been intentional, since the emit('active') in inside a once() handler in replicate.js, but it is at odds with the documentation (and also with what one might expect, that 'active'/'pause' events come in pairs).  From inspecting the code, perhaps the emit('active') (and the if statement around it) should be moved to somewhere in startNextBatch? If so, then the state will also need to be reset to "pending" somewhere, perhaps in completeReplication.

jjd314:
Fixes #3687.  As a side effect, though, it also slightly changes the order of events triggered by live replication: now 'active' always comes first, before the first 'change' of the replication batch (it used to be fired in a "once('change')" handler, so it was trigger just after).

### DIFF
--- a/lib/replicate.js
+++ b/lib/replicate.js
@@ -97,12 +97,6 @@ Replication.prototype.cancel = function () {
 
 Replication.prototype.ready = function (src, target) {
   var self = this;
-  this.once('change', function () {
-    if (this.state === 'pending' || this.state === 'stopped') {
-      self.state = 'active';
-      self.emit('active');
-    }
-  });
   function onDestroy() {
     self.cancel();
   }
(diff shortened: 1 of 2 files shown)


daleharvey:
Hey @jjd314, this PR looks perfect, thanks for the new test

Merged in https://github.com/pouchdb/pouchdb/commit/176321b9765982afb8599c4775e5f0272fea9416

PR #3692, opened 2015-04-03
(#3689) - Add ajax to GET and fix npm run dev on Windows

### LINKED ISSUE #3689: Add AJAX support to GET requests
This is a more limited version of #3491. I need AJAX (specifically headers and proxy) to work on GET requests so I can test some things out in a reproducible way for atts_since support. I'm finishing a PR for this now.

yaronyg:
Fixes #3689.

Needed the ability to actually use the advertised support for ajax on GET
requests.

test.headers.js - Fixed it up to use promises for the tests. This was
important for after() because otherwise for some reason done never
got called. Also added tests for new GET functionality.

npm run dev - The package.json was edited to use an environment variable
to set the client variable to 'dev'. But this doesn't work on Windows.
So I edited bin/run-test.sh to support an optional command line argument
and then edited package.json to use that argument to pass in 'dev'

### DIFF
--- a/tests/component/test.headers.js
+++ b/tests/component/test.headers.js
@@ -2,7 +2,10 @@
 
 var http = require('http');
 var PouchDB = require('../../lib');
-var assert = require("assert");
+var should = require("chai").should();
+require('bluebird').onPossiblyUnhandledRejection(function (e, promise) {
+  throw e;
+});
 
 describe('test.headers.js', function () {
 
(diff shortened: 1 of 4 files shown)


yaronyg:
If it passes tests and someone +1s then I'll try to remember how to properly commit to master. This is the price I pay for going too long between commits.

yaronyg:
Passed Travis. If I can get a +1 I'll merge. Thanks!

tests/component/test.headers.js
```diff
@@ -19,25 +22,43 @@ describe('test.headers.js', function () {
...
+    });
+  });
+
+  it('Test that we combine ajax options both globally and locally on GET',
+      function() {
+    var opts = { ajax: { headers: { aheader: "whyyes" } } };
+    var pouchDB = new PouchDB('http://127.0.0.1:' + PORT, opts);
+    var getOpts = { ajax: { headers: { ick: "slick", aheader: "override!" } } };
+    return pouchDB.get("fake", getOpts).
+    then(function() {
```
daleharvey: Minor nit, we arent super consistent about these, but I think its far clearer having the then on the previous line

    return pouchDB.get('fake', getOpts).then(function() { 
      should.equal....
    });

Another minor nits are ' over ", and `var db = new PouchDB` instead of `pouchDB`

daleharvey:
This looks great thanks, theres a few nits about the code style in the new tests but would have fixed them in the way in. The change to promises looks great.

Slightly worried about not being able to do `CLIENT=node` in windows, and in general it looks like we are really abusing the `CLIENT` var. Its probably my fault and I definitely dont know how to get things working on windows.

Could we split the changes, get the switch to promises and the actual fix for ajax in GET requests in and have this one split out so we can look at it seperately?

daleharvey:
I made a PR that split out the changes @ https://github.com/pouchdb/pouchdb/pull/3694, will merge when green and can look into fixing the CLIENT=dev issue on its own

daleharvey:
Ok I merged the `ajax` part of things, now we can take a look at CLIENT=dev

yaronyg:
Thanks for taking the time on the minor nits. Seriously. Because I haven't committed in a while I'm rusty. I'm hoping to have a substantial commit regarding atts_since (which is what motivated this commit) so I should get back up to speed. Splitting the commits is fine. The only reason I didn't is because then I couldn't have run the dev tests on the GET commit.

As for Windows. It's tricky. The configuration I use is that I run the Git Bash shell and run the commands from there. This creates a bizarre hybrid environment where the script command is run "unix style" but the individual commands are run in what appears to be a windows command line environment. In other words if the NPM script command is "foo && bar" then parsing that line will happen in "unix style" so the && will be recognized but foo and bar will both be exec'd (it appears) in windows command line. That's why, I'm guessing, the environment variable setting didn't work since that isn't how environment variables are set on the command line in Windows. In other words running "blah = 3 foo" will just confuse Windows.

That's why I went with a command line argument. That works the same everywhere. Also note that the change i proposed should still work just fine if someone manually sets the CLIENT environment variable. The script change I made looks to see if there is an argument. If there is then it sets CLIENT, otherwise it does nothing. So if someone sets CLIENT in the environment and doesn't pass in an argument then their setting should work.

yaronyg:
@daleharvey Are you waiting on me to split this up to just have the dev changes (which I'm happy to do) or did this PR just kind of get lost? Just checking in. Thanks!

nolanlawson:
@yaronyg So, I think this is something where we've been inconsistent in the past, and it's worth talking about.

The `ajax` options have been pretty poorly documented up to now. But my understanding was that they only applied to `new PouchDB()` as well as the exotic (and now deprecated) `PouchDB.destroy()`. If we start adding them to every possible API call, that feels like a slippery slope, when it's much simpler IMO to just pass it in to the constructor.

Is there some reason why you need specific ajax options for specific API methods? I.e. is there a reason you can't pass your ajax options in to the constructor? Or am I missing the issue entirely?

nolanlawson:
Also as far as I can tell your changes got merged in as https://github.com/pouchdb/pouchdb/commit/4481c954f0f2d003714f913f2d7ab8a8d9021ba3. Is there anything missing that wasn't in https://github.com/pouchdb/pouchdb/commit/4481c954f0f2d003714f913f2d7ab8a8d9021ba3?

daleharvey:
Yup, the `ajax` fix is merged, need this split out into the code that makes it run on windows, I wasnt happy with the `./run-test.sh dev` switch, `CLIENT dev` seems like it may have been a mistake as well. Will close this out since its mostly merged

yaronyg:
@daleharvey - Right now (I just tested with the latest bits) npm run dev won't work because the CLIENT environment variable is not going to get set on Windows. You said you didn't like the work around with passing in a command line argument. I'm happy to fix this however you would like but I need to know what you would like. Please let me know. Thanks!

PR #3694, opened 2015-04-04
(#3689) - Add ajax override to GET

daleharvey:
Needed the ability to actually use the advertised support for ajax on GET
requests.

test.headers.js - Fixed it up to use promises for the tests. This was
important for after() because otherwise for some reason done never
got called. Also added tests for new GET functionality.


daleharvey:
Merged in https://github.com/pouchdb/pouchdb/commit/4481c954f0f2d003714f913f2d7ab8a8d9021ba3

PR #3695, opened 2015-04-04
Prefer nextTick over setImmediate

ronag:


calvinmetcalf:
why? generally we prefer setImmediate as it doesn't block the event loop

ronag:
This is the only place in pouched where setImmediate is used over nextTick. So I'm not sure what you mean by generally?

In what way does nextTick block the event loop?

calvinmetcalf:
so the difference between setImmediate and process.nextTick is that process.nextTick is a micro task while setImmediate is a macrotask, this boils down to setImmediate allowing other events to be emitted before it is called while process.nextTick won't so if you use process.nextTick recursivly in the browser this will freeze the dom and node it will cause preemnt any streams, file io, or http requests from happening until it is done, contrast this with setImmediate which gives other events a chance to be fired before it runs it's callback.

In this case we wanted to make sure that the md5 calcualtions didn't freeze the dom while they running as it can take a bit of time for a large file, setImmediate does the job here.

nolanlawson:
Closing unless we have a good explanation for why we need this. @calvinmetcalf's explanation makes the most sense to me.

PR #3704, opened 2015-04-07
Add md5sum wikipedia link

RichardLitt:


daleharvey:
@RichardLitt this is markdown, so [MD5 sum](http://en.wikipedia.org/wiki/Md5sum) is a bit better than manually creating the HTML, do you mind ammending? otherwies looks good thanks a lot

RichardLitt:
Changed. The wiki link above for MIME types should also be changed, as it uses <a> tags and I figured that was some sort of weird style thing for this doc.

Kiwiroo:
'


Sent from my Samsung Galaxy S®4

-------- Original message --------

nolanlawson:
@RichardLitt long story, but that's because our alert snippets only accept HTML, not markdown. anyway this looks good to me, thanks!

nolanlawson:
4355f3404d73a905ddd496f03630da4b8a373c05

PR #3707, opened 2015-04-08
(#3707) - add bower install documentation

nolanlawson:
Just realized it's not clear that PouchDB is available from Bower. Many people are still using Bower (e.g. with Ionic).


daleharvey:
Thats nice, cheers

https://github.com/pouchdb/pouchdb/commit/76c676cfb6045fb565e9c3cba24aa1de44e4d8c3

Issue #3713, opened 2015-04-09
In Changes, call db.removeListener('destroyed', onDestroy) before emitting 'complete'


dholth:
During replication the 'complete' event causes the PouchDB live/retry replicator to request a new Changes. This can cause db to have too many listeners. Try calling removeListener() (but not self.removeAllListeners()) before firing 'complete'.

dholth:
Seems to help, but it would help more if replication would ensure all its changes feeds were cancelled before requesting a new one. Seems to be a bigger problem on iOS/WebSQL

dholth:
The real fix for this bug seems to be to avoid throwing any exceptions in replication event handlers.

daleharvey:
Also closing as dupe

PR #3718, opened 2015-04-09
call db.removeListener() before emit

dholth:
The Changes object's event listeners can add listeners to db; remove our listener before that happens.

### DIFF
--- a/lib/changes.js
+++ b/lib/changes.js
@@ -15,13 +15,13 @@ function Changes(db, opts, callback) {
   opts = opts ? utils.clone(opts) : {};
   var oldComplete = callback || opts.complete || function () {};
   var complete = opts.complete = utils.once(function (err, resp) {
+    db.removeListener('destroyed', onDestroy);
     if (err) {
       self.emit('error', err);
     } else {
       self.emit('complete', resp);
     }
     self.removeAllListeners();
-    db.removeListener('destroyed', onDestroy);
   });
   if (oldComplete) {
     self.on('complete', function (resp) {


dholth:
#3713

nolanlawson:
Sorry, could you explain more about what this fixes? Is there any way you can add a test to demonstrate what's being fixed here?

dholth:
When the Changes object fires ```self.emit('complete', resp);``` on L22 and we are doing live replication, a brand new Changes object calls `db.once('destroyed', ...)` before the complete Changes object has a chance to call `db.removeListener('destroyed', ...)`.

In the best case that will result in `db` briefly having 2 `destroyed` listeners. If a lot is going on then `db` will have more than 11 `destroyed` listeners; `EventEmitter` will complain and discard the listener.

By moving `db.removeListener('destroyed', ...)` to the top, `db` will briefly have *0* listeners in the best case instead of *2*. If a lot of listeners care about `destroyed` this could avoid adding too many `destroyed` listeners and breaking `EventEmitter`.

nolanlawson:
Ah okay, that makes sense. Is there any way to write a test to capture that? :smiley:

dholth:
It would look something like this. Either add a lot of listeners to `db0` and wait for EventEmitter to throw or measure the number of listeners for `destroyed` in the `newListener` handler (which may not get emitted at exactly the right place?). Unfortunately I did not see any difference from my change when using local to local replication. It might also be that the other EventEmitter bugs I've reported are still causing problems in this branch, making it harder to single out this particular problem.

Thankfully when the other EventEmitter problems are solved and no callbacks throw exceptions (the exceptions causing more EventEmitter problems) this change shouldn't matter.

```
'use strict';

describe("#3713 Unnecesary db 'destroyed' listener", function() {
  var dbNames = ['3713a', '3713b'], dbs;

  beforeEach(function (done) {
    testUtils.cleanup(dbNames, done);
  });

  after(function (done) {
    testUtils.cleanup(dbs, done);
  });

  it('should not add too many destroyed listeners', function(done) {
    var db0 = new PouchDB(dbNames[0]), db1 = new PouchDB(dbNames[1]), i = 0;
    dbs = [db0, db1];

    for(var n=0; n<5; i++) {
      db0.addListener('destroyed', function() {});
    }

    db0.addListener('newListener', function(type) {
      console.log('newListener', type);
      if(i++ === 4) { done(); }
    });

    db0.replicate.to(db0, {live: true, retry: true});
  });
});
```

nolanlawson:
Could you add the test to your PR? Would be a slam-dunk if the test fails before your fix but succeeds afterwards. :) Don't worry about the extra commits; we can squash them together when it gets merged.

nolanlawson:
Note: this is passing in everything but couch master, which means it is passing.

nolanlawson:
> By moving db.removeListener('destroyed', ...) to the top, db will briefly have 0 listeners in the best case instead of 2. If a lot of listeners care about destroyed this could avoid adding too many destroyed listeners and breaking EventEmitter.

It sounds like we are trading one bug for another one. Shouldn't there always be 1 `destroyed` listener, not 0 or 2? Sorry, I'm just really struggling to understand this PR.

dholth:
When we are in the cancel or complete handler, tasked with shutting down this Changes object, it does make sense to have 0 destroyed listeners. The other code decides to continue the replication, will immediately create a new Changes object, and the *new* Changes object is now the thing that needs to be notified if the database is destroyed.

nolanlawson:
Ah, that makes sense.

This seems possible to test. I might try my hand at writing a test case. Maybe something like this:

```
changes.on('complete', function () {
  var count = EventEmitter.listenerCount(changes, 'destroyed');
  count.should.equal(1);
});
```

daleharvey:
Sorry for not getting round to this, 2nd highest on my list now, should get to it today

nolanlawson:
closing in favor of https://github.com/pouchdb/pouchdb/pull/3825

PR #3720, opened 2015-04-10
(#3688) - correctly es3ify

### LINKED ISSUE #3688: PouchDB not working in IE8
I am testing PouchDB app in IE8, but I am getting following errors. I have included pouchdb.localstorage.js and es5.shim.js file in index.html.

You can find below the errors:

Expected Identifier, string or number for following line (pouchdb.localstorge.js):
Line no. 11019: , try: [
Expected Identifier for following line (main.js - local js file):
}).catch(function (err) { console.log('Unable to insert into DB. Error: ' + err.name + ' - ' + err.message); });

nolanlawson:
es3ify is not really doing what we want it to do. When used as a
browserify transform, it is unsafe, because it only applies to
OUR OWN code, not necessarily dependencies.

Furthermore, it's been causing problems from day one due to weirdnesses
where projects that depend on PouchDB and browserify it themselves
also need to require es3ify directly. That's not cool.

My fix is to use es3ify in the same way that we use derequire - i.e.
just as a command-line tool when building, which runs over the entire
file.

I've manually confirmed that this outputs es3ified code and fixes #3688.

### DIFF
--- a/bin/build-plugin.sh
+++ b/bin/build-plugin.sh
@@ -16,4 +16,5 @@ DEREQUIRE=./node_modules/.bin/derequire
     -x pouchdb \
     -r ./lib/plugins/config-$LEVEL_BACKEND.js:adapter-config \
     -r ./lib/plugins/migrate-browser.js:migrate \
+    | ./bin/es3ify.js \
     | $DEREQUIRE > ./dist/$OUTPUT_FILENAME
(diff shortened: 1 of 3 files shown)


nickcolley:
:+1: this reminds me of the 'dont ship a ES6 transpiler with your code' mantra

nolanlawson:
5d4265a8fd5738551d368cb2d69be4cb6bce5e88

Issue #3481, opened 2015-02-01
Get Android passing on Saucelabs


daleharvey:

daleharvey:
Did a trial run @ https://github.com/pouchdb/pouchdb/pull/3474, it looks like it passed the http tests then failed when testing the local database

daleharvey:
@nolanlawson suspects the webview may not have the correct permissions set to access the database as described in http://pouchdb.com/errors.html#android_pre_kitkat

daleharvey:
5.1 is now passing, can get to 4.4 now

https://github.com/pouchdb/pouchdb/commit/47944e2c507fc6904a565432e553ebc42e738cc1

PR #3726, opened 2015-04-11
(#3481) - try testing in android on saucelabs

nolanlawson:
Just to see if something changed in Saucelabs
and maybe the tests are passing finally.


nolanlawson:
Looks like Saucelabs has 5.0 Lollipop too, but there's little reason to test it imo because it's basically modern Chrome.

nolanlawson:
failed of course

Issue #3721, opened 2015-04-10
couch master tests are broken


nolanlawson:
Seems we need to explicitly set admin party now.

nolanlawson:
See https://github.com/apache/couchdb/pull/313

nolanlawson:
> 7:12 AM <•Kxepal> nolanlawson: https://github.com/pouchdb/pouchdb/blob/master/bin/run-couch-master-on-travis.sh#L34 add --with-admin-party-please to return admin party to the cluster

kxepal:
Since I'm the one to blame for, I'll leave some notes here why these changes were needed:
1. Push /_cluster_setup endpoint into more wide testing. It's not right that CouchDB has API for cluster setup, but developers cluster builds with workaround, but still legal, ways;
2. Prepare users that Admin Party will be no more available for cluster;
3. Take care about system databases which are creates automatically by /_cluster_setup;
4. However, CouchDB Javascript tests are addicted to Admin Party as well, so optionally there is a way to make a cluster with Admin Party, but this is not the default state of things.

Sorry for breaking things!

nolanlawson:
No prob; the fix seems easy. :)

nolanlawson:
@kxepal Sadly it's failing now for another reason, so I filed an issue: https://issues.apache.org/jira/browse/COUCHDB-2659

kxepal:
It's a pity. But do you have an exact error of that failure?

nolanlawson:
Looks like Travis isn't giving us much additional information unfortunately.

kxepal:
@nolanlawson ok, will try to run tests locally to see what's wrong there.

kxepal:
No, seems I would not: https://www.friendpaste.com/30mlrdJafnsrOOybdwcs8s
Have you already prepared environment to run a test and extract an error?

nolanlawson:
@kxepal sorry it's so wonky to get pouchdb to `npm install`. I guarantee if you run Ubuntu/Debian in a virtual machine it will work.

kxepal:
Yes, that's exactly what I did: on Ubuntu PouchDB build passed fine.
I'd updates CouchDB issue. TL;DR that's CouchDB bug caused by presence `_metadata` database (previously, it wasn't created automatically on cluster setup). To workaround the issue just drop it before run test suite. However, better not since PouchDB is able to detect this problem and track it down till it will be solved.

daleharvey:
[...]

PR #3727, opened 2015-04-11
(#3721) - add couch master to allowed failures

nolanlawson:
Until couch-master is fixed properly, we need
to unblock ourselves and get the tests back to green.


nolanlawson:
@daleharvey informally +1'ed this in IRC, so I'm gonna merge it to unblock us

Issue #3338, opened 2015-01-03
Modularize, publish separate npm modules


nolanlawson:
I know this is something that @jchris has mentioned, and it seems [pretty established as a best practice](https://twitter.com/Raynos/status/551243347994497024) for Node these days.

On the other hand, there is a limit to the usefulness of modularity, and it might end up being more trouble than it's worth. So I'm opening this issue to start the discussion.

As a first step, we could do something like:

* Publish `utils.js` as `pouchdb-utils`
* Publish `merge.js` as `pouchdb-merge`
* Then we can split out the adapters, so e.g. `pouchdb-idb-adapter`, `pouchdb-websql-adapter`, `pouchdb-http-adapter`, and `pouchdb-leveldb-adapter`
* Publish `replicate.js` as `pouchdb-replicate`

The only downside I can think of is the increased complexity when submitting new PRs (e.g. if it touches multiple repos) and also running the tests in Travis. For that, I think we could mimic the express-pouchdb and pouchdb-server tests, where `pouchdb` is a `devDependency` and then we just use `npm link` to run the PouchDB test suite.

daleharvey:
So I have wanted this for a long time, but I have also had a real problem with how to manage testing across dependencies, the solution for pouchdb-server actually looks great, or we can just keep all the modules in one repo and test them that way,

nickcolley:
Is this only worth doing for modules that have merit on their own?

daleharvey:
Yeh its only worth doing for modules reusable outside pouchdb, replicator and http adapters specifically, it would only make things harder if we split them into multiple repositories, seperate modules inside the same repository would be much the same as it is now but with the benefit or reusability)

For things that are highly reusable being in their own repository may actually make things easier, less overhead if that is the only thing you are working on

nolanlawson:
That's a good point; currently none of these things need to be used outside of PouchDB itself except maybe ajax/utils (speaking from experience writing some PouchDB plugins)

The adapters and replicator are currently not used by anything but PouchDB, although they may be used by PouchDB-like JavaScript libraries. So this would really be for the benefit of those libraries.

marten-de-vries:
I think keeping the modules in the same repo, at least at first, gives us most of the benefits without the drawbacks. We would get the reuse & more clearly documented/defined interfaces between them, but not the extra testing setups, repositories and the like.

jchris:
I like same repo also. Seen this movie before, same repo wins.

[...]

PR #3730, opened 2015-04-12
(#3338) - expose ajax/checkpointer/promise

nolanlawson:
This is my proposed solution to the problem of exposing
internal modules to plugin authors. I think this fix is
cleaner than publishing as separate npm modules.

Basically, plugin authors can require whatever they want
by using e.g. `require('pouchdb/extras/ajax')`. I did an
audit of current plugins, and I think the only things we
need are:

* `ajax` (pouchdb-authentication, pouchdb-load, Hoodie)
* `Promise` (just about every plugin)
* `Checkpointer` (pouchdb-load)

If I missed anything, please let me know.

### DIFF
--- a/docs/_includes/api/plugins.html
+++ b/docs/_includes/api/plugins.html
@@ -22,4 +22,38 @@ PouchDB.plugin({
 new PouchDB('foobar').sayHello(); // prints "Hello!"
 {% endhighlight %}
 
+### Extra APIs
 
+In order to reduce code size when your plugin is browserified/webpacked, PouchDB exposes some of its internal modules to plugin authors.
+
+To use them, you should save PouchDB as a required npm dependency (`npm install --save pouchdb`) and then require them like so:
+
+#### Ajax
+
+{% highlight js %}
+var ajax = require('pouchdb/extras/ajax');
+{% endhighlight %}
+
+The `ajax()` function as used by PouchDB. Essentially a shim in the style of `jQuery.ajax`.
+
+#### Checkpointer
+
+{% highlight js %}
+var Checkpointer = require('pouchdb/extras/checkpointer');
+{% endhighlight %}
+
+The `Checkpointer` function as used by PouchDB's replicator. Writes checkpoints as `_local` docs so that replication can resume where it last left off.
+
+#### Promise
+
+{% highlight js %}
+var Promise = require('pouchdb/extras/promise');
+{% endhighlight %}
+
+The ES6 `Promise` shim as used by PouchDB. Expect this to be `bluebird` in Node, `lie` in browsers without Promise support, and the global `Promise` in browsers with Promise support.
+
+#### Caveats
+
+These "extras" should not be considered stable parts of the API, and may change unpredictably between PouchDB versions. Refer to the [source code](https://github.com/pouchdb/pouchdb/tree/master/extras) for the details of the API.
+
+If your plugin depends on a particular version of PouchDB, please notify your users. Unless you like to live dangerously, you should also nail down the version of PouchDB you are using in your `package.json`. And if you need any other internal modules, please submit a pull request.
--- a/extras/ajax.js
+++ b/extras/ajax.js
@@ -1,0 +1,4 @@
+'use strict';
+
+// allow external plugins to require('pouchdb/extras/ajax')
+module.exports = require('../lib/deps/ajax');
--- a/extras/checkpointer.js
+++ b/extras/checkpointer.js
@@ -1,0 +1,4 @@
+'use strict';
+
+// allow external plugins to require('pouchdb/extras/checkpointer')
+module.exports = require('../lib/checkpointer');
--- a/extras/promise.js
+++ b/extras/promise.js
@@ -1,0 +1,4 @@
+'use strict';
+
+// allow external plugins to require('pouchdb/extras/promise')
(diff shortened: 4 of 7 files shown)


nolanlawson:
cc @janl @gr2m, please let me know if this seems like a good solution to you, wrt https://github.com/hoodiehq/pouchdb-hoodie-api/pull/16. If y'all agree on this approach, then I will deprecate `pouchdb-ajax` in npm (currently it's only used by `pouchdb-load` anyway AFAIK).

gr2m:
That looks great to me! /cc @boennemann @svnlto

svnlto:
i like it.

nolanlawson:
Travis failures are genuine; I'll fix them later.

jo:
I like it.

jo:
Does it work well with browserify? I mean that only the parts I require land in the buid?

gr2m:
@jo yes :) That's why it's so cool!

jo:
@gr2m cool. Do you have a working example?

nolanlawson:
@jo The lodash style of `require('lodash/object/clone')` etc. would be an example of this.

jo:
@nolanlawson yeah, that is the only lib that also came to my mind. But I haven't figured out how they do it until now, where I found the [modern build](https://github.com/lodash/lodash/tree/3.6.0-npm) - thanks ;)

gr2m:
Can we get this merged? I'd love to use it in https://github.com/boennemann/pouchdb-hoodie-api :)

gr2m:
Actually, maybe for plugins that have no dependency on `pouchdb`, it would make more sense to expose things like `ajax`, `Promise` etc directly on the `db` instances? Otherwise we'd run risk that my app requires a different version of PouchDB than my plugins, and it would load more code than necessary, which would be nice to avoid when used in browser.

nolanlawson:
Yep, just waiting for a green. :) Need to restart the Travis jobs, one sec...

nolanlawson:
> Otherwise we'd run risk that my app requires a different version of PouchDB than my plugins, and it would load more code than necessary, which would be nice to avoid when used in browser.

Hm, so the worst-case scenario is that you include slightly more code than otherwise. But this duplicated will get gzipped away anyhow, assuming it is exactly the same. So I think it's not so bad.

Also, if you just tell your users to use the same version of PouchDB as you (e.g. `^3.4.0`), then dups can be avoided. You'll want to do this anyway, to avoid the scenario where e.g. your plugin depends on PouchDB 3 and your users are using PouchDB 4.

jo:
@gr2m Glad you mention the use of `db.constructor.utils.Promise` which I also consider. The [MDP page on `Object.prototype.constructor`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/constructor) states this is standard.
@nolanlawson don't you think this approach is feasable?

boennemann:
@nolanlawson 
> ## What to tell your users
Below is some boilerplate you can use for when you want a real README for your users.

> To use this plugin, include it after pouchdb.js in your HTML page:
```
<script src="pouchdb.js"></script>
<script src="pouchdb.mypluginname.js"></script>
```

I know that this isn't meant for production, but if we're talking about the _worst-case_ then it's this. We neither get gzip nor deduplication and the bundled code inside the plugin might in some weird way be incompatible with the actual pouchdb code, due to differing versions.

With the pouchdb-hoodie-api we're using the passed instance, which is rather safe and compatible, but it produces some awfully long lines of code.

With peer-dependencies going away I'm not sure how to handle this right ¯\_(ツ)_/¯

nolanlawson:
> @nolanlawson don't you think this approach is feasable?

@daleharvey and I have been against that approach for awhile, because `PouchDB.utils.*` is undocumented and therefore you're not "supposed" to use it. (I haven't always been a good Boy Scout in my plugins, but I've tried.)

>  If we're talking about the worst-case then it's this.

@boennemann I'd be happy to amend the plugin-seed docs to explain how to use plugins with Browserify as well. It's a question I get a lot, and basically boils down to using it like you would Node:

```js
var PouchDB = require('PouchDB');
PouchDB.plugin(require('some-plugin'));
var db = new PouchDB('foo');
```

This ought to browserify and dedup correctly, if the plugin depends on the same PouchDB version.

Also, to be clear, I see this as a power-user feature. I think the vast majority of our users will use separate scripts as described above and maaaaybe concatenate and gzip with some kind of grunt/gulp build tool. And I think that's okay. :)

> bundled code inside the plugin might in some weird way be incompatible with the actual pouchdb code, due to differing versions.

This is something I would like to avoid. Maybe it would be most prudent to bump PouchDB's major version if something under `extras/` changes. Then I can remove the part from the API docs saying `these "extras" should not be considered stable parts of the API, and may change unpredictably between PouchDB versions.`.

nolanlawson:
Gonna merge due to clear community support for the general idea, and because I want to start using it in pouchdb-load etc.

nolanlawson:
a3007450183a0b9dcc9852e6f4376efc65793a2b

jo:
Thanks for this!

It seems this introduced a jshint problem:
```sh
lib/deps/promise.js: line 4, col 20, 'Promise' is not defined.
```

nolanlawson:
Yeah I've fixed that in another PR: https://github.com/pouchdb/pouchdb/pull/3752

PR #3737, opened 2015-04-18
Fix for BlackBerry 10 selecting WebSQL backend instead of IndexedDB

alax:
Apparently the BlackBerry 10 browser identifies itself as Mobile Safari, which causes the WebSQL adapter to be selected instead of IndexedDB, even though the BlackBerry 10 browser has a modern IndexedDB implementation.

This fix simply causes the IndexedDB backend to identify itself as valid on the BlackBerry 10 platform.

Thanks!

### DIFF
--- a/lib/adapters/idb/idb.js
+++ b/lib/adapters/idb/idb.js
@@ -913,7 +913,8 @@ IdbPouch.valid = function () {
   // to our knees.
   var isSafari = typeof openDatabase !== 'undefined' &&
     /(Safari|iPhone|iPad|iPod)/.test(navigator.userAgent) &&
-    !/Chrome/.test(navigator.userAgent);
+    !/Chrome/.test(navigator.userAgent) &&
+    !/BlackBerry/.test(navigator.platform);
 
   // some outdated implementations of IDB that appear on Samsung
   // and HTC Android devices <4.4 are missing IDBKeyRange


nolanlawson:
So does BlackBerry 10 use Chromium's IndexedDB implementation, or does it use WebKit's? I assume it uses Chromium's, in which case this PR would be very appropriate. :)

alax:
All I can offer is a guess, and that guess is that it probably uses Chromium's, since it doesn't exhibit any of the weird behavior present in Safari's implementation.

nolanlawson:
I have a BlackBerry at work, so I'm going to test this on Monday. Sorry, it's not that I don't trust you; it's just that we've made lots of mistakes with this kind of stuff in the past, and I want to be very sure. :)

alax:
Sure thing! Just ensure you're running the latest OS (10.3.1)  so that you have their latest browser build. I know 10.2 had some issues with IndexedDB, but 10.3 seems to be flawless. If there is anything you'd like me to test for you in the mean time, just let me know. I can post screenshots or whatever you'd like.

nolanlawson:
Hm, well now you're making me worried. :) How widespread is 10.3? If 10.2 still has IndexedDB issues, then maybe it's safer to keep people on WebSQL? Or maybe check for 10.3 in the user agent? Is there some good way to test these IndexedDB issues? E.g. for some broken Samsung/HTC browsers we are able to detect the presence of `window.IDBKeyRange` to determine if it's using an outdated version of the IndexedDB spec.

alax:
10.3 is still fairly new, and has only been released on some devices (BlackBerry Passport, BlackBerry Classic) with other devices in some countries receiving OTA updates. But, that being said, now that I'm thinking about it the issues may have been with 10.1, since most of our production devices are running 10.2 and using IndexedDB as their backend with no issues (albeit not with PouchDB).

If I remember correctly, the issue was if you made a call to IDBKeyRange, the browser would halt and then after a short delay crash altogether. I have some 10.2 devices around that I can test with today. I'll patch PouchDB on my end with the proposed change and run the integration tests on both a 10.2 and 10.3 device and let you know how it works. I don't have any 10.1 devices though; they have all long been upgraded.

The issue with WebSQL on BB10 is that after 20mb or so of data being stored, the backend starts rejecting writes and there is no way to request more quota. Even if you open a larger DB, or preemptively ask for a larger quota, the system will always stop accepting writes at about 20mb, which makes it impossible to write applications that intend to store more data than that.

nolanlawson:
Wow, that's pretty bad. Agreed that we should prefer IndexedDB in that case.

Does that apply to Cordova apps as well? Feel free to run the PouchDB integration tests in Cordova if you have time (instructions are in TESTING.md). Your help here is really invaluable. :)

alax:
Yes, I actually first noticed the limitation in Cordova, and then ran the app in the browser to confirm it wasn't a Cordova specific issue. I'll try to get around to the integration tests tomorrow, today actually ended up being fairly hectic for me.

nolanlawson:
@alax Any update on this? I'm tempted just to merge this, since at the very least it will fix 10.3 users, even if 10.2 users are left in the cold.

alax:
Oh, my apologies, I have been so busy I completely forgot to run the tests. I am away from my workstation for the next few days, but if you can wait until next Friday or so, I can do it then. I'm fairly certain it will work, since we have a fairly large user base using IndexedDB on 10.2, and there have been no issues.

I have been using the patched version of PouchDB on 10.3 pretty heavily since I originally posted this pull request, so I can say that at least on 10.3 everything does work as intended.

nolanlawson:
Ah okay, if 10.2 IndexedDB works, then I'll accept your say-so. :) I don't have a Blackberry for testing, and I don't know the platform, so I'll just trust the expert. :D

nolanlawson:
Cheers, merged as 56d4da155680ad960389b6ce2e924c91c41d32a8 to match our commit message format. Thanks!

Issue #3748, opened 2015-04-20
Remove double error stack


jo:
We currently forget write errors which happened in past batches: https://github.com/pouchdb/pouchdb/blob/master/lib/replicate.js#L187

jo:
Pull: #3749

jo:
It turns out that the assumption above is not true. The correspondent line is an artefact and can be removed.

nolanlawson:
a3efd41732983ed66400fc14d736f08121e688a9

PR #3749, opened 2015-04-20
(#3748) - Remove double error stack

jo:


kxepal:
Wouldn't this blow up the client when you replicate a lot of docs and accidentally most of them are failed to store?

jo:
Yes, you're right, this can happen and we should think about how to handle this.

This commit [reverts a presumably unintentional change](https://github.com/pouchdb/pouchdb/commit/929c1bd8bec2704ad32b1b6a9465d563c5dbccbe#diff-c169e1e72632697264bcd930cc444020R194).

Maybe we can leave it as it is now?

kxepal:
How about to keep last N errors and let the user to configure their number on replication start?

Another option is to group errors by type and limited set of document ids which suffers from them:

```
{
    "replication_errors": {
        "You are not a db or server admin.": [
            "_design/doc"
        ],
        "You are not document owner": [
            "foo",
            "bar"
        ]
    }
}
```

Note, that error keys are the values of "reason" fields in bulk_docs response, not "error" one since validate_doc_update may only raise "unauthorised" or "forbidden" kind of errors and the latter isn't much specific for reason of document update failure.

jo:
How does CouchDB handle this?

kxepal:
Nohow, but there is [COUCHDB-2626](https://issues.apache.org/jira/browse/COUCHDB-2626) about (:

jo:
I like your idea with the grouped errors the most.

kxepal:
Yes, but it has a flaw when you want to store two conflicted documents with same id and different revs - with plain array of ids you'll never know which one failed.

I'm not sure what's the right way to store explained errors since the result structure is will define the way to process them. Grouping by errors is friendly for producing user-end reports. Grouping by id/rev is friendly for developers to lookup and iterate over the failures they are interested in. Not sure where is balance lies here. So brainstorming is welcome (: I really want to have such feature in CouchDB as well.

nolanlawson:
This seems like a reasonable fix to me, but I would like to see a unit test added to describe the functionality. :)

jo:
As a result of a deeper look at the code (thanks @nolanlawson for bugging me towards a test) I found out that the entire line can be removed, because errors are stored elsewhere, which is already covered by the current tests.

jo:
[These travis failures](https://travis-ci.org/pouchdb/pouchdb/jobs/60034164) seems to have nothing todo with these changes. I have restarted them three times and they fail reliable. @nolanlawson do you have an idea?

nolanlawson:
@jo They all seem to be related to pouchb-server, which was just updated today. I'll file an issue over there.

marten-de-vries:
#203 is fixed, Travis is green now.

nolanlawson:
Woohoo, nice!

jo:
Thank you both!

nolanlawson:
If we can remove code and the tests still pass, then that's a win to me.

nolanlawson:
a3efd41732983ed66400fc14d736f08121e688a9

Issue #2543, opened 2014-07-28
iOS 7.1.x Safari: Maximum call stack exceeded when performing replicate.from() using remote CouchDB


smeinel:
I first brought this issue up in the message group: https://groups.google.com/forum/#!topic/pouchdb/pLvogMtICZI

Replicating a relatively small DB with a few conflicts errors out only on iOS Safari. All other OS and browser combinations appear to be unaffected.

There is a live Mocha + Chai test available at:
https://batmobile.caltech.edu:8080/db_test

```
Test output:
    Database
        . Should create a session 215ms
        . Should destroy old DB 247ms
        . Should create new DB
        X Should sync the remote DB
            Error: RangeError: Maximum call stack size exceeded. (https://batmobile.caltech.edu:8080/js/vendor/pouchdb-2.2.3.js:5284)

[Error stack]
    (anonymous function) (db_test, line 122)
    emit (pouchdb.js, line 7196)
    completeReplication (pouchdb.js, line 5898)
    abortReplication (pouchdb.js, line 5874)
    (anonymous function) (pouchdb.js, line 5828)
    (anonymous function) (pouchdb.js, line 7770)
    drainQueue (pouchdb.js, line 7801)

PouchError
line: 5695
message: "the statement callback raised an exception or statement error callback did not return false"
name: "Object"
result: {
    doc_write_failures: 2
    docs_read: 298
    docs_written: 296
    end_time: Mon Jul 28 2014 13:36:18 GMT-0700 (PDT)
    errors: Array[0]
    last_seq: 283
    ok: false
    start_time: Mon Jul 28 2014 13:34:44 GMT-0700 (PDT)
    status: "aborted"
    __proto__: Object
}
sourceURL: "https://batmobile.caltech.edu:8080/js/vendor/pouchdb.js"
stack: "https://batmobile.caltech.edu:8080/js/vendor/pouchdb.js:5695:16↵https://batmobile.caltech.edu:8080/js/vendor/pouchdb.js:7770:25↵drainQueue@https://batmobile.caltech.edu:8080/js/vendor/pouchdb.js:7801:18"

CouchDB info:
    Version: 1.5.0
    Size: 2.8 MB
    # Documents: 180
    Update Seq: 3751

Last requests to CouchDB:
    CouchDB proxy: /vdn_qunit/_local/e81a403483bed6460db04b698e2a0fdd?_nonce=wBXbOqWPlbrVxDpH
[...]

PR #3750, opened 2015-04-20
(#3750) - fix typo in replication test

nolanlawson:
This was a pretty dumb mistake, noticed by @jo here:
https://github.com/pouchdb/pouchdb/commit/c2ef4055d3b227150f724f3314d309b759c21c10#commitcomment-10801263

The point of the test was to cause a stack overflow, which it did
before #2543 was fixed, so in a sense the test was fine. But I think
it's way less confusing now.


daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/3b84d256597e7ebfa459b8cba00966b07ced8a4f

Thanks

PR #3754, opened 2015-04-21
(#3754) - small cordova fix

nolanlawson:
I know we don't have any automated way of testing
this, but if you follow the steps in TESTING.md to run
Cordova, it should be working now.


nolanlawson:
Actually Cordova tests are broken now (doesn't even install the app) and I don't have time to figure out what's going on.

Issue #3757, opened 2015-04-22
Alternating indexdb and websql adapters in the same browser


hakan1111:
Hi,
thought I would add another issue that we can not explain and get any new input on why/how it can happen. 
We are experiencing that pouchdb is using either indexeddb or websql in the same browser (of course running the same code).
As a result, data is saved in 2 different dbs and our users does not have access to all their data.

We are using PouchDB 3.3.1 right now.
Attaching a screen-shot from chrome to clarify the issue and to show that the same dbs are created in either idb or websql. 
Unfortunately I can not reproduce this, but as soon as I can I will update this issue.

@nolanlawson Can you give us any clues for when this could happen it would be great

BR /Håkan
![image](https://cloud.githubusercontent.com/assets/11521724/7274431/f43b9c12-e8fc-11e4-96fd-766258bef0d7.png)

nolanlawson:
This bug should only occur during testing. If you clear all browser data (including local storage) and then stick with either websql or IndexedDB it will go away.

Long answer: we have a local storage flag to tell us if a database was ever opened in websql. If a websql-only browser like Android 4.3 gets updated to include IndexedDB, then your existing users don't lose data because we stick with websql for that database.

Let me know if that describes your issue. If not, please provide a reproducible test case. :)

nolanlawson:
We should probably add a console message to let people know when we are forcibly downgrading to websql btw

hakan1111:
Thanks for quick answer there Nolan, don't you ever sleep?

The sticky behaviour is there, ie if ever websql has been used Pouchdb will always use websql. My problem (which I am trying to reproduce 8-) ) is why - suddenly - websql is used instead of idb when (re)initalising the dbs. We do not force either idb or websql in our code (we just use the browser default) so we are searching for reasons for why/how this can happen.

The "data loss" I was talking about above occurs when data has been stored in idb in one session, and then suddenly - in a later session - a websql adapter is used instead. So there is historical data in indexeddb and new data in websql making the historical data unavailable (for the users of our app at least).

Again, sorry for not being able to provide reproducible test case, will work on that.

hakan1111:
Another questions is what adapter(s) are currently supported for Chrome on iOS? It is not explicitly listed here -> http://pouchdb.com/adapters.html.
However the info note below suggests that Pouchdb does not use the idb adapter generally for iOS 8+. Is that still true and does it apply to Chrome on iOS?

nolanlawson:
[...]

PR #3759, opened 2015-04-22
(#3757) - notify when force-downgrading websql

nolanlawson:
I put this in console.log instead of the `debug` log,
because I'm going to assume casual users don't have
debug mode enabled.

### DIFF
--- a/lib/setup.js
+++ b/lib/setup.js
@@ -51,6 +51,9 @@ PouchDB.parseAdapter = function (name, opts) {
       adapterName = PouchDB.preferredAdapters[i];
       if (adapterName in PouchDB.adapters) {
         if (skipIdb && adapterName === 'idb') {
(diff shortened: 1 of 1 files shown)


daleharvey:
Yeh I dont think many people will need to see this, but when its applicable it will be handy

https://github.com/pouchdb/pouchdb/commit/655049730f0500a2461c01225642dd0ea7555586

rarkins:
@nolanlawson @daleharvey I am using PouchDB site-wide now, and I see this message every page/visit when I view console in Chrome for OSX. Is this expected/desired that this is shown so often, or does it indicate a problem in our setup?

nolanlawson:
No, we have a bug where this message gets printed out if you are deliberately forcing websql in environments that support both. We need to fix the message.

PR #3760, opened 2015-04-23
add hoodie-pouchdb-api to external.md

gr2m:
`hoodie-pouchdb-api` provides a hoodie-like API on top of PouchDB. It's going to be used internally for the `hoodie.store` API, but is useful for anyone how prefers a higher level API. All methods can be loaded separately, too.


nolanlawson:
42c86b481b604961561f2caf27fd91d01d8dc92f

PR #3765, opened 2015-04-26
(#3765) - checkpointer should browserify smaller

nolanlawson:
When the checkpointer is required externally from
`'pouchdb/extras/checkpointer'`, it's too big because
it pulls in all of utils. This makes it smaller.


daleharvey:
Ok this makes sense, cheers

https://github.com/pouchdb/pouchdb/commit/649e5ee9a5597597e1ad00b9c8e7605bc5f2da25

PR #3767, opened 2015-04-26
Fuzzy Replication tests

daleharvey:


daleharvey:
Not ready to be merged, I think this should probably be split out from the rest of the integration tests as its a different type of test. 

As commented, the general idea is to do a whole bunch of random things then check that at the end of it all, replication completes fine. It doesnt currently find any bugs, but the actions it triggers are currently fairly basic.

nolanlawson:
I love this! Great idea.

One thing I'd add is the ability to print out the steps to show how we got there, in case a test does fail. Otherwise we might go mad trying to reproduce it twice. :)

daleharvey:
So ready to merge this now I am seeing very occasional failures already which is nice, the first thing I want to do it work on the ability to replay a test because as you said its pretty much impossible to debug from the test output. But I want to keep this maintained so worth adding to the repo before it gets stale

nolanlawson:
:+1: when green, love the approach.

I also intend to do something similar for retry replication, because I'm sure there are cases where retry fails if it goes offline at a particular point in the replication process.

daleharvey:
I think that would be a good addition to this test, so at some (random) point we can just kill the server and start it up again later.

daleharvey:
Merged in https://github.com/pouchdb/pouchdb/commit/8df1beaa7cda5c5327ec041fd3ec391fa7365406

PR #3769, opened 2015-04-27
(pouchdb/express-pouchdb#203) - Deprecate PouchDB.destroy() properly.

marten-de-vries:
Does the heavy lifting for when ``PouchDB.destroy()`` is actually going to be removed. That should be as easy as just removing the ``PouchDB.destroy()``/``PouchAlt.destroy()`` code now. This also makes it possible to properly remove the deprecation notice in ``pouchdb-wrappers`` (background: pouchdb/express-pouchdb#203).

I think this is the first time I actually modified adapter-specific code, so a cautious review would be prudent.

### DIFF
--- a/lib/setup.js
+++ b/lib/setup.js
@@ -2,7 +2,6 @@
 
 var PouchDB = require("./constructor");
 var utils = require('./utils');
-var Promise = utils.Promise;
 var EventEmitter = require('events').EventEmitter;
 PouchDB.adapters = {};
 PouchDB.preferredAdapters = require('./adapters/preferredAdapters.js');
@@ -75,6 +74,8 @@ PouchDB.parseAdapter = function (name, opts) {
 };
 
 PouchDB.destroy = utils.toPromise(function (name, opts, callback) {
+  console.log('PouchDB.destroy() is deprecated and will be removed. ' +
+              'Please use db.destroy() instead.');
 
   if (typeof opts === 'function' || typeof opts === 'undefined') {
     callback = opts;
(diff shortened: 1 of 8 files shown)


marten-de-vries:
Ready for review. The last Travis job seems to time out pretty consistently (restarted 3x now).

lib/setup.js
```diff
@@ -85,56 +86,11 @@ PouchDB.destroy = utils.toPromise(function (name, opts, callback) {
...
-        //so we don't have to sift through all dbnames
-        PouchDB.emit(name, 'destroyed');
-        callback(null, resp || { 'ok': true });
-      }
-    });
-  }
-
-  var createOpts = utils.extend(true, {}, opts, {adapter : backend.adapter});
-  new PouchDB(baseName, createOpts, function (err, db) {
+  new PouchDB(name, opts, function (err, db) {
```
daleharvey: We have already had a deprecation notice for quite a while, how about we just remove it now and next version is a major version bump?

daleharvey:
I am mostly :+1: on this, however I think its probably time to just remove the PouchDB.destroy implementation entirely, its usage has already been warned against for quite a while, we can leave the console.log in there for anyone that still may use it, but have it do nothing. What do yall think?

nolanlawson:
I'd rather put it off until we can sneak something else into a 4.0 major bump (e.g. the `changes()` events, which I'm not a fan of).

I admit some of this is selfishness; I do have a lot of plugins depending on the `pouchdb/pouchdb` master branch and using `PouchDB.destroy()`, which I haven't had time to update yet. :stuck_out_tongue: 

In any case I am :+1: for this particular PR.

daleharvey:
Ok I am good with holding of complete removal until we have done a bit of a check on the plugins 

https://github.com/pouchdb/pouchdb/commit/ef247eced792f374c1632d9b55821c3c3e62e473

PR #3770, opened 2015-04-27
(#3770) - update quota documentation

nolanlawson:
@daleharvey would also appreciate some input from you
on FirefoxOS. Does it still have a hard 5MB limit for apps?

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -39,17 +39,17 @@ For mobile applications, you can use PouchDB within [PhoneGap](http://phonegap.c
 
 {% include anchor.html class="h3" title="How much data can PouchDB store?" hash="data_limits" %}
 
-In **Firefox**, PouchDB uses IndexedDB. Though Firefox has no upper limit besides disk space, if your application wishes to store more than 50MB locally, Firefox will [ask the user](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API) to confirm that this is okay.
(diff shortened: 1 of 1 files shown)


nolanlawson:
Also there's been some great research by Raymond Camden recently:

* http://www.raymondcamden.com/2015/04/17/indexeddb-and-limits
* http://www.raymondcamden.com/2015/04/24/indexeddb-and-limits-ie

As it turns out, if you try to abuse storage, you can get some weird (and possibly silent) errors in Firefox and IE. This seems to only affect using an obscene amount of data, though, so I didn't think it was necessary to document it in our FAQs.

daleharvey:
Apps should ask for a 'storage' permission and they will have unlimited storage, web content in the browser have a 5MB hard limit currently

nolanlawson:
That only applies to Mozilla-approved apps in the marketplace, right? I tried a normal self-hosted webapp with a manifest file, and it was still stuck at 5MB even with the permission. :(

daleharvey:
Nope that should be for all applications - https://developer.mozilla.org/en-US/Apps/Build/App_permissions, definitely file a bug if you can reproduce otherwise

daleharvey:
Awesome cheers, Ill see if I can test out fxos and get that added too 

https://github.com/pouchdb/pouchdb/commit/8f630a3ad9ca120b0b33116137c76439260ba44e

nolanlawson:
@daleharvey You can always test out http://www.npm-browser.com/manifest.webapp . It's a FirefoxOS app. :)

nolanlawson:
Just tested it in FXOS 3.0 on both a Nexus 5 and the simulator, seems to work great as a packaged app. 10000 npm modules synced, and still going. Nice work!

PR #3773, opened 2015-04-28
(#3773) - update docs to be more accurate

nolanlawson:
These updates are based on stuff I noticed from
this blog post:

http://gonehybrid.com/how-to-use-pouchdb-sqlite-for-local-storage-in-your-ionic-app/

1. The SQLite Plugin is actually not faster than
   the native IndexedDB/WebSQL, at least anymore.
   That statement about "often faster" is based on
   my own experience with the SQLite Plugin from back
   in Android 2.3 days, when a spinning GIF would be
   stuck while WebSQL was processing heavy DB operations,
   whereas it wouldn't with the SQLite Plugin.

   This is
   not an issue that I've seen with PouchDB anymore, and in
   fact the SQLite Plugin seems to have gotten slower in recent
   months, to the point where it's noticeably slower than either
   WebSQL or IndexedDB.

2. A lot lot LOT of people are using PouchDB with Ionic, and it
   pains me that they don't know about `$q.when()`. Makes your
   code much simpler.

### DIFF
--- a/docs/_includes/api/api_overview.html
+++ b/docs/_includes/api/api_overview.html
@@ -20,6 +20,14 @@ function(error, result) { /* ... */ }
 
 If you don't specify a `callback`, then the API returns a [promise][]. In [supported browsers](http://caniuse.com/#feat=promises), native promises are used, falling back to the minimal library [lie][] as needed. In Node.js, promises come from [Bluebird][].
 
+{% include alert/start.html variant="info"%}
+{% markdown %}
+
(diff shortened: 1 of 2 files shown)


nolanlawson:
And yeah, I know the SQLite Plugin has changed its name, but everybody still seems to be calling it the SQLite Plugin for the moment.

marten-de-vries:
Looks good to me. :+1:

daleharvey:
Merged cheers: https://github.com/pouchdb/pouchdb/commit/0ca28499c35765ecb96f98e809708b3bb8faa75d

jdnichollsc:
Eyy people, ¿What do you think about my example? Is correct?  https://gist.github.com/jdnichollsc/f82592d6f303aaebfb9f

Issue #3573, opened 2015-02-25
Test improving CORS performance by avoiding content-type on GETs


nolanlawson:
Per https://lists.w3.org/Archives/Public/public-webappsec/2015Feb/0347.html

nolanlawson:
This should be a pretty easy fix to get us a good performance boost. (If it works! :))

nickcolley:
Does this mean less requests? :octopus:

nolanlawson:
Yes, supposedly we can eliminate most of the OPTIONS requests during replication.

remy:
Is there any reason that [theses headers](https://github.com/pouchdb/pouchdb/blob/master/lib/deps/request-browser.js#L30) should be added if the request type is a `GET`? If I understand correctly, the `content-type` is only needed if there's a body being posted, and since pouch is talking JSON, and a default `GET` request accepts `*` (i.e. all) then you don't need the headers.

I've not tested this at all (and am happy to give it a try if I'm on the right lines), but the change might be as simple as:

```js
if (options.json && options.method !== 'GET') {
```

nolanlawson:
I tried @remy's suggestion, and although it passes the tests (I was worried CouchDB would send us `multipart/mixed` types we weren't prepared to accept), it also doesn't seem to solve the problem of extra OPTIONS requests (see 8a42b10e5241f47446335cf8a542c4bc5baac130). In Chrome at least, every GET still seems to be preceded by an OPTIONS. :confused:

nolanlawson:
OK, the good news is that I've tried another approach - avoiding the `'Content-Type'` header for GETs - and actually, this works great. (Removing the `Accept` does indeed cause an error because CouchDB starts sending us data we can't parse.)

With this approach, the tests pass and I can confirm that GETs no longer require so many OPTIONS. I'll open a PR. Thanks @remy!

nolanlawson:
PR: https://github.com/pouchdb/pouchdb/pull/3777

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/c14d52a3e5b8f4b8e6ae72455696e9e7b41d63ac

jo:
Fuck yeah, it works!

PR #3777, opened 2015-05-01
(#3573) - avoid setting Content-Type for GETs

nolanlawson:
This avoids extra OPTIONS requests for GETs
in the case of CORS requests. If this PR does
what I think it does, then even the Travis tests
should be significantly faster this time around.

### DIFF
--- a/lib/deps/request-browser.js
+++ b/lib/deps/request-browser.js
@@ -27,7 +27,9 @@ module.exports = function(options, callback) {
   xhr.open(options.method, options.url);
   xhr.withCredentials = true;
 
-  if (options.json) {
+  if (options.method === 'GET') {
+    delete options.headers['Content-Type'];
+  } else if (options.json) {
     options.headers.Accept = 'application/json';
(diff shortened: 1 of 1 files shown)


remy:
Looks like the test is failing due to an npm install issue rather than code. Are you able to re-run the test?

daleharvey:
Unfortunately our tests still have some intermittent failures and a full green run usually requires a restart or 2. This is probably all fine just restarting the test now.

Code looks good nice catching default content-types being set, I thought we already fixed this but I guess not, :+1: ill give this a manual check that we arent getting those extra cors on GETS and merge.

nolanlawson:
Yep, btw Travis doesn't seem any faster than our last build, but then again we never know what kind of hardware we're getting.

daleharvey:
Sweet, checked and hitting less OPTIONS, merged in https://github.com/pouchdb/pouchdb/commit/c14d52a3e5b8f4b8e6ae72455696e9e7b41d63ac, thanks

PR #3778, opened 2015-05-01
(#3778) - updates to changes/query docs

nolanlawson:
These are just some random changes to these docs
that I think aid in their readability. I also updated a
reference to 2014 to 2015.


daleharvey:
These changes look good to me, thanks

https://github.com/pouchdb/pouchdb/commit/ce8b25ad07e6b64242b244e2d120da273fdb3ca2

PR #3779, opened 2015-05-01
(#3754) - fix cordova tests

nolanlawson:
We still don't have automated Appium tests, but this at least
fixes the Cordova tests so that they're running again. I can confirm
that I can run the tests on both Android and iOS simulators,
with and without the SQLite Plugin. It also runs in FirefoxOS.


daleharvey:
Nice thank you 

https://github.com/pouchdb/pouchdb/commit/ce5ff1a8713a85848e4970a9249aa69657f65470

Issue #3082, opened 2014-11-30
Intermittent leveldb mapreduce failure: skip/limit and multiple keys/values


nolanlawson:
Seeing this in some failed mapreduce builds:

```
local with temp views: views should query correctly with skip/limit and multiple keys/values:

AssertionError: correctly return rows: expected [ Array(2) ] to have a length of 1 but got 2
```

slang800:
I'm getting something similar... I have a db with 2715 docs (using leveldb as a backend), and I'm requesting them in batches of 100 (to avoid loading everything into memory at once). But sometimes I get more than 100 docs back. Here's a list of results from incrementally requesting the database in chunks:

```
got 101 docs
got 102 docs
got 100 docs
got 100 docs
got 101 docs
got 100 docs
got 100 docs
got 103 docs
got 100 docs
got 100 docs
got 101 docs
got 100 docs
got 100 docs
got 100 docs
got 102 docs
got 100 docs
got 100 docs
got 100 docs
got 100 docs
got 100 docs
got 101 docs
got 102 docs
got 100 docs
got 101 docs
got 100 docs
got 100 docs
got 100 docs
got 1 docs
```

it returns the correct number of docs, but it is exceeding the limit I set

nolanlawson:
Can you write a test to reproduce? Also do you see this error only in LevelDB or also in IndexedDB/WebSQL?

slang800:
This is running only on nodejs, so I don't think I can use it with IndexedDB/WebSQL, right?

slang800:
I did test a direct connection to the main server (using the http adapter), and that gives no issues with being over the limit, and doesn't skip any docs. So I think this is a problem in the LevelDB adapter

slang800:
Oh, and it seems like `total_rows` in the response is correct for leveldb (because it matches what couchdb's /_utils reports) - it's somehow skipping rows.

nolanlawson:
Looks like we may have a race condition in the LevelDB adapter that allows it to return more results than requested. Again, a reproducible test case would be excellent here, because otherwise it's really hard to tell what's going on. :)

slang800:
I haven't been able to reproduce it consistently enough to make a test-case for it... It happens with like 20% of all requests, and that's with me requesting 100 docs each time. Maybe if we just made the test case that you showed above request a lot more docs? If I figure out how to reproduce it more reliably, I'll be sure to make a case, but at this point my only lead is "request a lot of data in big chunks".

[...]

PR #3785, opened 2015-05-03
(#3082) - fix for leveldb returning too many docs

nolanlawson:
I'm not actually able to reproduce #3082, but if
there was a race condition in some version of LevelDOWN
that caused more than `limit` docs to be returned, I imagine
this would fix it. Even if not, this is a harmless change.


daleharvey:
This looks like a sensible change and passing, cheers

https://github.com/pouchdb/pouchdb/commit/b59a2c12383091ac35146f1cddee70adbe180a28

Issue #2286, opened 2014-05-24
Browserify should work with adapter plugins


nolanlawson:
Currently they don't. This is a follow-up to discussion in #2190.

nolanlawson:
fixed by https://github.com/pouchdb/pouchdb/commit/30ee3ea58a3886c8eff43a15153754535e42d7a4

PR #3786, opened 2015-05-03
(#2286) - browserifiable adapter plugins

nolanlawson:
This finishes up the work @calvinmetcalf and I started on
this some time ago.

Now, you can either use the LocalStorage/memory/idb-alt plugins
as standalone scripts downloaded from the website, or via the
"extras" API, allowing them to be Browserified/Webpacked.

I tested this out and confirmed that, when `require()`d from
a separate project, Browserify does an intelligent job
deduplicating the code. E.g. PouchDB itself is only included once.

For instance, when using Browserify, PouchDB and the memory adapter
will add up to 710K. When using separate standalone scripts, you're
looking at 352K for PouchDB and 470K for the memory plugin script,
which adds up to 822K. So you save over 100K, which ain't bad.

This commit should also make it easier to create new browser
adapter plugins in the future. The API is much simplified and
requires a lot less command-line Browserify magic.

Note that the "browserify API" itself is not tested in CI; we're
still testing the standalone scripts the old-fashioned way. But
since the standalone scripts are just browserified versions of the
"extras" APIs (with `-x pouchdb` thrown in to avoid browserifying
PouchDB itself), I think this is pretty low-risk.


bin/build-all-plugins.sh
```diff
@@ -1,19 +1,14 @@
 #!/bin/sh
```
nolanlawson: The big change with this commit is that, instead of having a single `plugins/index.js` file that requires `leveldown` and gets swapped out for the appropriate *down adapter using command-line Browserify arguments, instead I've just got three different `index.js` files, one for each of memory/idb-alt/localstorage.

docs/_includes/api.html
```diff
@@ -20,4 +20,5 @@
 <li><a href="#events">Events</a></li>
 <li><a href="#defaults">Default settings</a></li>
 <li><a href="#plugins">Plugins</a></li>
+<li><a href="#extras">Extras</a></li>
```
nolanlawson: Didn't make sense to have this documented under "Plugins" anymore.

lib/setup.js
```diff
@@ -4,7 +4,7 @@ var PouchDB = require("./constructor");
 var utils = require('./utils');
 var EventEmitter = require('events').EventEmitter;
 PouchDB.adapters = {};
-PouchDB.preferredAdapters = require('./adapters/preferredAdapters.js');
```
nolanlawson: Another change was to how `preferredAdapters` work. Instead of explicitly setting this as a list of `['idb', 'websql']` in the browser and `['leveldb']` in Node.js, we can just add them to the array when we call `PouchDB.adapter()`.

The only change I had to make to `PouchDB.adapter()` was a third argument, `addToPreferredAdapters`, so we could avoid doing this for the http/https adapters. But this is an API change I wanted anyway, because I would like it for third-party adapter plugins such as `socket-pouch`. Eventually we should probably document the `PouchDB.adapter()` API, but for now it's still "internal."

lib/index.js
```diff
@@ -14,12 +14,11 @@ var httpAdapter = require('./adapters/http/http');
...
 
-PouchDB.adapter('idb', require('./adapters/idb/idb'));
-PouchDB.adapter('websql', require('./adapters/websql/websql'));
+PouchDB.adapter('idb', require('./adapters/idb/idb'), true);
+PouchDB.adapter('websql', require('./adapters/websql/websql'), true);
 PouchDB.plugin(require('pouchdb-mapreduce'));
 
 if (!process.browser) {
   var ldbAdapter = require('./adapters/leveldb/leveldb');
-  PouchDB.adapter('ldb', ldbAdapter);
```
nolanlawson: ^ I don't remember seeing anywhere where this `'ldb'` name was used. I think we can get rid of it.

nolanlawson:
Small fix - forgot to set it to not include the leveldb-specific migration code in the browser.

docs/_includes/api/extras.html
```diff
@@ -0,0 +1,104 @@
...
+
+You can also download this plugin [as a standalone script](https://github.com/pouchdb/pouchdb/releases/download/{{ site.version }}/pouchdb.idb-alt.js).
+
+{% include alert/start.html variant="info"%}
+{% markdown %}
+
+When you include these plugins in your project, they will be automatically picked up as a "third" option when you create a new PouchDB without an adapter specified (e.g. `new PouchDB('mydb')`).
+
+So for instance, if the user loads PouchDB in a browser that does not support WebSQL or IndexedDB, and if you are using the in-memory plugin, then PouchDB will automatically use the memory adapter instead of WebSQL or IndexedDB.
+
```
nolanlawson: @davidpfahler read this ^ regarding preferredAdapters

daleharvey:
This looks good to me, I would really like a :+1: from @calvinmetcalf just in case because a lot of this is over my head, but its all passing and mostly makes sense to me

calvinmetcalf:
:+1:

daleharvey:
Sweet merged - https://github.com/pouchdb/pouchdb/commit/30ee3ea58a3886c8eff43a15153754535e42d7a4

Issue #3787, opened 2015-05-03
react-native compatibility


davidpfahler:
I spent the last two days figuring out how to make PouchDB compatible with react-native. To make it work I had to work around a couple of things that I would like to address below. This is intended as a comprehensive list for discussion on how we could make this easier, e.g. by eliminating the need for workarounds. If you just need to get it working, skip to the TLDR at the very bottom.

## webpack
Issue #3319 is concerned with getting PouchDB working with webpack. It is to be expected that close to 100% of react-native users are going to also use webpack. react-native brings additional challenges to getting PouchDB working with webpack. As far as I understand, react-native requires some files ("native components", see https://github.com/roman01la/react-native-babel/tree/dep-graph#react-native-in-es6-with-webpack-and-babel for details) via the packager that should not be resolved by webpack beforehand. To avoid this, we need to define them as webpack externals. The most elegant way to do this, that I found, is to use the `DependencyGraph` that comes with react-native. 

This is relevant for PouchDB, because - get this - one of the dependencies spit out by DependencyGraph is called `merge`. Now, the way `roman01la/react-native-babel` currently handles externals is using a RegEx. If you employ this tactic, you match everything with `merge` in a require call. As you might now, PouchDB has quite some files called `merge.js` and is requiring them. To avoid webpack failure, you need to make sure that you specify the webpack externals as exact matches, as you can see in my [pull request](https://github.com/davidpfahler/react-native-babel/commit/d6c92ab0f65461e6f3c82b915a1fc62fe9e6873c).

I found this config to work for me: [`webpack.config.js`](https://github.com/davidpfahler/react-native-babel-example/blob/47265eb8b55c0a99e04fddfbea45f3c4e48aaeea/webpack.config.js), [`webpack-watch.js`](https://github.com/davidpfahler/react-native-babel-example/blob/c6129ef7bee58a5bf7b9dc9e623ea9be00c037ff/webpack-watch.js).

The plugins in the config are probably optional. What is important is that you have specified the loaders. Also make sure you `npm install`ed the loaders, in this case, `npm install -D babel-loader json-loader`. If the warning about `lie.js` bothers you, add `noParse: /lie.js/` to your config.
The `webpack-watch.js` not only provides watch functionality, but also handles the issue with the `DependencyGraph`.

## package.json of PouchDB
The above config will still fail. When webpack parses `leveldown`, it logs the following warnings and error:

```
[...]

PR #3788, opened 2015-05-03
(#3787) - fix adapter prefix logic

nolanlawson:
This bug had no real impact, except that custom adapters
couldn't specify use_prefix or not.

Since this is a really subtle and uncommon use case,
and it doesn't have any impact other than when people use
e.g. LocalStorage/IndexedDB directly alongside PouchDB,
I don't think it's worth adding a test for it.

Suffice it to say that, before this commit, all adapters
in the browser were having their `use_prefix` set to true,
because IDB was being chosen for its `use_prefix`. The correct
adapter will still eventually chosen; but the `use_prefix`
always came from IDB.

This was not really a big issue, since our
other browserified adapters all have `use_prefix` set to true
anyway, but this commit makes the code more correct. Thanks to
@davidpfahler for reporting it!

### DIFF
--- a/lib/setup.js
+++ b/lib/setup.js
@@ -43,9 +43,12 @@ PouchDB.parseAdapter = function (name, opts) {
     utils.hasLocalStorage() &&
     localStorage['_pouch__websqldb_' + PouchDB.prefix + name];
 
-  if (typeof opts !== 'undefined' && opts.db) {
+
+  if (opts.adapter) {
+    adapterName = opts.adapter;
+  } else if (typeof opts !== 'undefined' && opts.db) {
     adapterName = 'leveldb';
-  } else {
+  } else { // automatically determine adapter
     for (var i = 0; i < PouchDB.preferredAdapters.length; ++i) {
       adapterName = PouchDB.preferredAdapters[i];
       if (adapterName in PouchDB.adapters) {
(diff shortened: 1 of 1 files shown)


nolanlawson:
Also I failed to think of a test for this PR, except to actually open up an IndexedDB and start messing with it, to see if `use_prefix` was working or not... The test itself would be really brittle, so I figured it wasn't worth the bother.

lib/setup.js
```diff
@@ -61,16 +64,15 @@ PouchDB.parseAdapter = function (name, opts) {
...
   adapter = PouchDB.adapters[adapterName];
-  if (adapterName && adapter) {
-    var use_prefix = 'use_prefix' in adapter ? adapter.use_prefix : true;
 
-    return {
-      name: use_prefix ? PouchDB.prefix + name : name,
-      adapter: adapterName
-    };
-  }
+  // if adapter is invalid, then an error will be thrown later
```
nolanlawson: There IS a test already that confirms that if you do `new PouchDB('foo', {adapter: 'fake'})` then it will throw a useful error.

daleharvey:
Awesome thanks

https://github.com/pouchdb/pouchdb/commit/b26212cb2101da573393bce5003b2d17110f1e9f

PR #3789, opened 2015-05-04
(#3319) - set leveldown to false to fix webpack

nolanlawson:
Webpack does not like that `leveldown` is trying to use
`fs`. We can harmlessly set it to `false` in the package.json.

For `levelup`, the issue is that it's trying to directly load
its own package.json file, which is apparently a big contentious
issue in Webpack, and the standard solution is that app devs add their
own json loader, which is what Webpack devs themselves advocate (see
[here](https://github.com/webpack/webpack/issues/378#issuecomment-50258332) and [here](http://mattdesl.svbtle.com/browserify-vs-webpack)).

We can add a note about this Webpack config fix to our own wiki,
or we can just trust that Webpack-using devs will find it themselves
after 5 minutes of googling the JSON issue (I found it in about
as much time). In any case, the bare minimum we need to do is
set `leveldown` to `false`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,6 +111,7 @@
     "./lib/adapters/preferredAdapters.js": "./lib/adapters/preferredAdapters-browser.js",
     "./lib/deps/buffer.js": "./lib/deps/buffer-browser.js",
     "bluebird": "lie",
-    "fs": false
+    "fs": false,
+    "leveldown": false
   }
 }


nolanlawson:
cc @davidpfahler you may need to start cherry-picking a lot of recent PRs in order to get react-native working :stuck_out_tongue:

nolanlawson:
@davidpfahler never mind, I did it. The new branch `for-react-native` is referenced by `pouchdb-async-storage`.

calvinmetcalf:
god I don't understand webpack and it's 'we're going to tell everyone we're node compatible, but then refuse to implement 10% of what you need to be node compatible' stratagem

nolanlawson:
I hear ya, I hear ya.

daleharvey:
Cool cheers (ugh @ another distribution / loader / format) 

https://github.com/pouchdb/pouchdb/commit/6e95001a4fc97fa0cf33a00e7654c9a225274426

PR #3790, opened 2015-05-04
Test commit - don't merge!

nolanlawson:
This is just to verify that the PRs #3786 and #3789
play well together. I think they should both work
at the same time, but it's worth testing.


nolanlawson:
not useful anymore, we know it's green now

Issue #3791, opened 2015-05-04
unable to use with a *down that uses uri style db names


calvinmetcalf:
specifically [this line](https://github.com/pouchdb/pouchdb/blob/56d4da155680ad960389b6ce2e924c91c41d32a8/lib/setup.js#L29) prevents using `postgres://localhost/sqldown?table=_pouchdb_test_db_testdb` as a db name

nolanlawson:
I see that issue! One sec, got a fix...

nolanlawson:
OK, with that commit ^ you should be able to run:

    SERVER=pouchdb-server \
      SERVER_ADAPTER=sqldown \
      SERVER_PREFIX=server_side_prefix \
      LEVEL_ADAPTER=sqldown \
      LEVEL_PREFIX=client_side_prefix \
      npm test

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/f91b6e8e18af1a8b0d4c08e5ace30ad623b3b0dc

PR #3792, opened 2015-05-04
(#3791) - fix leveldown test args

nolanlawson:
This adds correct SERVER_ADAPTER and
SERVER_PREFIX variables that can be used by
leveldown authors as described here:
https://gist.github.com/nolanlawson/f2f4d7e2e6f426502b94

It also fixes the MemDOWN Travis tests to use
the SERVER_ADAPTER variable. (Before it was hard-coded
to run the server with --in-memory, which does the
same thing.)


daleharvey:
Good catch, cheers

https://github.com/pouchdb/pouchdb/commit/f91b6e8e18af1a8b0d4c08e5ace30ad623b3b0dc

PR #3797, opened 2015-05-05
(#3797) - consistently use utils.btoa/utils.atob

nolanlawson:
Also move them to their own file outside of utils.
This is just code cleanup and doesn't change the
functionality at all.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -6,7 +6,6 @@ exports.ajax = require('./deps/ajax');
 exports.createBlob = require('./deps/blob');
 exports.uuid = require('./deps/uuid');
 exports.getArguments = require('argsarray');
-var buffer = require('./deps/buffer');
 var errors = require('./deps/errors');
 var EventEmitter = require('events').EventEmitter;
 var collections = require('pouchdb-collections');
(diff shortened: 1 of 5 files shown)


lib/utils.js
```diff
@@ -248,31 +247,9 @@ Changes.prototype.notify = function (dbName) {
...
-    return btoa(str);
-  };
-} else {
-  exports.btoa = function (str) {
-    return new buffer(str, 'binary').toString('base64');
-  };
-}
+var base64 = require('./deps/base64');
+exports.atob = base64.atob;
+exports.btoa = base64.btoa;
```
daleharvey: These are only called in 6 places, before merging fancy getting rid of them from utils at all and requiring base64 instead?

daleharvey:
:+1: I think we can clean it up a bit further but either way can merge, cheers

lib/utils.js
nolanlawson: yeah, I like that idea. will do

nolanlawson:
Ack, well, I can't seem to get the tests to pass; we're using `utils.btoa` and `utils.atob` in tons of places in the tests. I'll keep working on it later, but opened up a separate WIP PR for now.

daleharvey:
Heh sorry I only grepped ./lib and thought it was only in a few places

nolanlawson:
Merging this PR; as I said, the other one was a bit tricker.

nolanlawson:
06e84f266f387ef351501ec777b8916e7d88fd77

Issue #3804, opened 2015-05-06
Replication: Include attachments when fetching generation one docs


jo:
As @karlwestin just found out attachments are fetched seperately, regardless if the doc is generation one.

CouchDB has [attachment support for views](https://issues.apache.org/jira/browse/COUCHDB-1923) (and therefore `_all_docs`) since 1.6.
So lets request with `attachments=true` and see if the data is returned and use it then. This should not break backwards compatibility.

jo:
Pull is here: #3805

nolanlawson:
fixed by 8f512b9149310ba06174ae9aeaf996ab787b2559

PR #3805, opened 2015-05-06
(#3804) - Replication: Include rev-1 attachments

jo:
Include attachments when fetching generation one docs during replication.
The ability to include attachments on a view result was added in CouchDB 1.6.

Add `attachments=true` to the query and examine the result wheather it has
attachment data. If yes, use it.

### DIFF
--- a/lib/replicate.js
+++ b/lib/replicate.js
@@ -262,17 +262,23 @@ function replicate(repId, src, target, opts, returnValue, result) {
     }
     return src.allDocs({
       keys: ids,
-      include_docs: true
+      include_docs: true,
+      attachments: true
     }).then(function (res) {
       if (state.cancelled) {
         completeReplication();
         throw (new Error('cancelled'));
       }
       res.rows.forEach(function (row) {
+        var hasAttachments = row.doc && row.doc._attachments &&
+          Object.keys(row.doc._attachments).length > 0;
+        var hasAttachmentData = hasAttachments &&
+          !row.doc._attachments[Object.keys(row.doc._attachments)[0]].stub;
+
         if (row.doc && !row.deleted &&
           row.value.rev.slice(0, 2) === '1-' && (
-            !row.doc._attachments ||
-            Object.keys(row.doc._attachments).length === 0
+            !hasAttachments ||
+            hasAttachmentData
           )
         ) {
           result.docs_read++;


nolanlawson:
I like the idea of this PR! But the test failures seem genuine:

```
  1) test.attachments.js- local:http Attachments replicate:
     Uncaught TypeError: Cannot read property '_attachments' of undefined
      at /home/travis/build/pouchdb/pouchdb/tests/integration/test.attachments.js:1896:25
      at /home/travis/build/pouchdb/pouchdb/lib/utils.js:347:11
      at process._tickCallback (node.js:448:13)
```

jo:
Thanks for having a look at it. I have found and fixed the error.

jo:
@nolanlawson the tests pass now :)

lib/replicate.js
```diff
@@ -262,17 +262,23 @@ function replicate(repId, src, target, opts, returnValue, result) {
...
       if (state.cancelled) {
         completeReplication();
         throw (new Error('cancelled'));
       }
       res.rows.forEach(function (row) {
+        var hasAttachments = row.doc && row.doc._attachments &&
+          Object.keys(row.doc._attachments).length > 0;
+        var hasAttachmentData = hasAttachments &&
+          !row.doc._attachments[Object.keys(row.doc._attachments)[0]].stub;
+
```
nolanlawson: ^ why would a gen-1 document have a stub in it? this test seems unnecessary

nolanlawson:
One minor question about possibly dead code, and also I would appreciate a comment explaining the optimization, but otherwise this looks good to me.

lib/replicate.js
jo: This is to support CouchDB <1.6. Support for `attachments=true` in views [was added in 1.6](https://issues.apache.org/jira/browse/COUCHDB-1923).

nolanlawson: ahhh, good one! :) might be worth adding a comment, then

nolanlawson:
Actually I can just merge with added comments. Thanks a bunch!

nolanlawson:
8f512b9149310ba06174ae9aeaf996ab787b2559

jo:
Good comment are good. :+1:

PR #3809, opened 2015-05-07
(#3797) - WIP - use btoa/atob without exporting from utils

nolanlawson:
Seems a LevelDB migration test is failing, no idea why.


nolanlawson:
@daleharvey So it turns out this is impossible to extract from `utils` without either duplicating it to `tests/integration/utils` or exposing it in `extras`... think it's best if we just merge the other one as-is and leave it be. I do agree that someday we should really work on breaking out all the `utils` stuff; it's grown too big.

PR #3813, opened 2015-05-08
remove listeners added per replicate() call on error

dholth:
PouchDB's builtin replication with retry is useless if you actually go offline because backOff() will call replicate() repeatedly adding  the EventEmitter will soon fill up, and then replication will never be able to actually continue.

This pull request avoids the problem by removing all of the per-replicate()-call listeners before calling backOff().

### DIFF
--- a/lib/replicate.js
+++ b/lib/replicate.js
@@ -130,6 +130,13 @@ function genReplicationId(src, target, opts) {
 }
 
 function replicate(repId, src, target, opts, returnValue, result) {
+  // Remember per-replicate()-call listeners for later removal
+  var listeners = [];
+  var newListener = function(event, listener) {
+    listeners.push([event, listener]);
+  };
+  returnValue.on('newListener', newListener);
+
   var batches = [];               // list of batches to be processed
   var currentBatch;               // the batch currently being processed
   var pendingBatch = {
@@ -428,6 +435,10 @@ function replicate(repId, src, target, opts, returnValue, result) {
       if (allErrors.length > 0) {
         error.other_errors = allErrors;
       }
+      // backOff() will call replicate() again; remove leaky listeners
+      for (var i = 0; i < listeners.length; i++) {
+        returnValue.removeListener.apply(returnValue, listeners[i]);
+      }
       error.result = result;
       backOff(repId, src, target, opts, returnValue, result, error);
     } else {
@@ -582,6 +593,7 @@ function replicate(repId, src, target, opts, returnValue, result) {
       throw err;
     });
   }
+  returnValue.removeListener('newListener', newListener);
 }
 
 exports.toPouch = toPouch;
--- a/lib/replicate.js
+++ b/lib/replicate.js
@@ -107,6 +107,9 @@ Replication.prototype.ready = function (src, target) {
     target.removeListener('destroyed', onDestroy);
   }
   this.then(cleanup, cleanup);
+
+  // avoid EventEmitter leaks by making second ready() call a no-op
+  this.ready = function () { };
 };
 
 


nolanlawson:
@calvinmetcalf Can you take a look at this? It's way over my head.

@dholth You can check our [existing retry replication tests](https://github.com/pouchdb/pouchdb/blob/d9782ad76827320940d25baa315f40d7684fb60b/tests/integration/test.replication.js#L3053-L3115) to see how to write a test. Basically the workflow is: write a test that fails (without your fix), add your fix and see if it fixes it, rinse, repeat.

Also I suspect that the `this.ready = function () { };` is an avoidable hack, and there's a more elegant way using the existing [EventEmitter API](https://nodejs.org/api/events.html). :)

dholth:
replicate() calls ready() every time it is called, adding event listeners on the source and target databases. (If they are destroyed, replicate wants to abort.) Since the callback functions are defined inside the ready() closure, they cannot be used to remove redundant destroyed listeners from the source and destination databases.

https://github.com/dholth/pouchdb/blob/issue-3699/lib/replicate.js#L176

A different way would be to have backOff() pass an 'is retry' flag to replicate() in which case replicate() would not call ready() at all, or to have ready() save its cleanup() function to be called either right before replicate() calls backOff, or on the second call to ready().

The second strategy might be better since if ready's cleanup is called before calling backOff then replication doesn't get aborted if the source and destination are destroyed during backOff.

Generally these bugs are caused by the Replication object originally having been designed to be used once (ready() cleanup in .then()) but PouchDB's live with retry replication wants to re-use the same Replication object by calling replicate() to re-try.

nolanlawson:
@dholth Yeah, I just think it would be cleaner if it set a flag on itself or something instead of the function setting itself to be a no-op.

I'll work on writing a test for this.

nolanlawson:
I can reproduce this, but it's subtle as hell.

nolanlawson:
I can reproduce by counting the number of listeners on the source, target, and the return value from `replicate()`. I can see that the listener count goes up over time.

There is also an issue where multiple changes feeds can be operating at the same time, causing the browser to bottleneck and halt because it only allows 8 simultaneous transactions. Retry replication is a hot mess, to be honest.

nolanlawson:
Also FWIW @dholth your fixes only remove some of the leaks, not all of them, which is exactly why it's crucial to have tests. :)

As for our own test (yes, there was only one), it was really not sufficient for this "retry" feature. I'm pretty bummed that these bugs made it out into a release.

nolanlawson:
In case anyone wants to help out: [3818-nolan](https://github.com/pouchdb/pouchdb/tree/3818-nolan). Gonna take a break for now, because this is hard.

dholth:
Yes we are still managing retry ourselves based on the offline npm
example in production.

On Sat, May 9, 2015, at 05:43 PM, Nolan Lawson wrote:
> Also FWIW @dholth[1] your fixes only remove some of the leaks, not all
> of them, which is exactly why it's crucial to have tests. :)


> As for our own test (yes, there was only one), it was really not
> sufficient for this "retry" feature. I'm pretty bummed that these bugs
> made it out into a release.


> — Reply to this email directly or view it on GitHub[2].

>



Links:

  1. https://github.com/dholth
  2. https://github.com/pouchdb/pouchdb/pull/3813#issuecomment-100551960

nolanlawson:
I have the beginnings of some fixes in https://github.com/pouchdb/pouchdb/pull/3822.

nolanlawson:
Okay, I think I have the full fix, with tests. Closing in favor of #3825. Thanks for providing such detailed info @dholth!

PR #3815, opened 2015-05-09
(#3815) - increase "retry stuff" test timeout

nolanlawson:
When running the tests, I often see this one time out.
I think the test itself is fine; the timeout is just
too low. E.g. in pouchtest.com it will often fail
because the remote database is slow.


daleharvey:
Cheers - https://github.com/pouchdb/pouchdb/commit/1deaf087ad7703ee586f1a9ce71493923e9969ba

PR #3817, opened 2015-05-09
(#3817) - increase another test timeout

nolanlawson:
We're manually setting the timeout to 30s,
when the default is 60. This causes the test
to fail pretty frequently as I test on my local network.


daleharvey:
Yeh these must have got set before we increased the global default

https://github.com/pouchdb/pouchdb/commit/3c36cfeeecad6254c7cfa9a7e039e5894111a312

Issue #3699, opened 2015-04-06
EventEmitter leak when live replication with retry goes offline


dholth:
I'm replicating twice, once for each direction: local.replicate.from(remote); local.replicate.to(remote); with {live:true, retry:true}. When I go offline (see also handy desktop Chrome mobile emulation with network throttling) each retry seems to cause the replicator to add a new 'destroy' listener to the database. Soon it goes to 11 and stops working.

Happens when the backOff() method calls replicate() from setTimeout().

dholth:
In related news, exceptions in the replication event handlers will cause the same problem. It may just be the same "needs to remove listeners" bug.

dholth:
This bug is tricky because much of the replication state and callbacks are closures from inside the replicate() function itself. One way to fix:

- Trap all per-replicate() call callbacks with returnValue.on('newListener', // append listener to list). This should work as long as replicate() doesn't add any callbacks asynchronously.
- Remove 'newListener' listener when replicate() returns.
- Remove all collected listeners right before calling backOff() (the function that eventually calls replicate() again, which re-adds new versions of all the callbacks, and is causing the bug)
- Only allow ready(), which just makes sure the replication is canceled if either the source or target databases are destroyed, to be called once. I just assigned this.ready = // no-op function;  at the end of the ready function.

While I was testing this issue I noticed the retry timeout may not be reset correctly after replication restarts. It takes it a long time to try again after going off- and back on-line the second time.

dholth:
It also leaks every time getChanges() is called from the replicator's onChangesComplete() handler, the changes feed wants to know when 'destroyed' happens on the source database, but the completed Changes doesn't stop listening.

ctlacko:
:+1: It seems like Chrome does a decent job of mitigating this leak as you mentioned, but toggling live updates on/off in Firefox makes my Macbook go into testicle melt mode. I know that's not super helpful, but I can definitely reproduce this.

dholth:
This should fix the problem, I've been using a similar fix privately to be able to use replication-with-retry. https://github.com/pouchdb/pouchdb/pull/3813

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/d8902a1825d3c60d0b0151b42569ee1bb6d283ff, thanks so much for the debugging and initial patches @dholth, we usually try to avoid taking over peoples patches however this ended up needing a fairly big fix, thanks @nolanlawson for all the work as well

PR #3819, opened 2015-05-09
(#3699) - source db doesn't leak "destroyed" event

nolanlawson:
I'm going to take it slow with fixing these memory leaks.
There are a LOT of events leaking, so this is going to
take more than one commit to fix.

This commit just ensures that the source database (during
"retry" replication) doesn't leak the "destroyed" event.

The test fails before the fix but succeeds after. I've also
started moving tests into a separate `test.retry.js` file.


lib/replicate.js
```diff
@@ -97,6 +97,11 @@ Replication.prototype.cancel = function () {
 
 Replication.prototype.ready = function (src, target) {
   var self = this;
+  if (self._readyCalled) {
+    return;
+  }
+  self._readyCalled = true;
+
```
nolanlawson: @dholth This is what I was talking about. ^ Maybe it's 6 of one, half a dozen of the other, but I find this more readable than making `ready()` a noop.

tests/integration/test.replication.js
```diff
@@ -3050,71 +3050,6 @@ adapters.forEach(function (adapters) {
...
-      rep.on('change', function () {
-        if (++numChanges === 3) {
-          rep.cancel();
-        }
-      });
-
-      remote.put({}, 'hazaa');
-    });
-
-
```
nolanlawson: ^ this test was simply copied into the new file

tests/integration/test.retry.js
```diff
@@ -0,0 +1,153 @@
...
+      rep.on('change', function () {
+        if (++numChanges === 3) {
+          rep.cancel();
+        }
+      });
+
+      remote.put({}, 'hazaa');
+    });
+
+    it('source doesn\'t leak "destroyed" event', function (done) {
```
nolanlawson: this is the test that fails before the fix but succeeds after. In fact it just hangs rather than "failing," which is one of the annoying things I find about Mocha, but c'est la vie.

nolanlawson:
closing in favor of https://github.com/pouchdb/pouchdb/pull/3822

PR #3823, opened 2015-05-10
Update create_database.html documentation to match API on authentication

mainakae:
Username and password credentials for remote http databases are not defined (anymore?) inside options.ajax, but options.auth. Updated documentation to account for that.


nolanlawson:
Good catch! Updated the PR to match our commit message format: 9a81c1855a06a3614c33fe74ab6512446c0df8c5

Also yeah, sorry about that; that was a typo when I wrote that documentation. :)

PR #3825, opened 2015-05-10
(#3699) - fix retry memory leaks

### LINKED ISSUE #3699: EventEmitter leak when live replication with retry goes offline

### LINKED ISSUE #3824: Live replication replicates 1 doc, then 1 again, then batch_size
As mentioned here: https://github.com/pouchdb/pouchdb/issues/2940#issuecomment-63135931.

This is a longstanding issue that we're all aware of (I used to see it in the NPM Browser - "replicated 1", "replicated 2", then "replicated 502", etc).

For some reason it's never had its own issue, but now it does.

nolanlawson:
Goals:
* Fixes #3699 by ensuring that listeners aren't leaked (heavily tested)
* The tests should fail instantly rather than time out

You can verify that the tests fail instantly when we have a memory
leak, by running the same tests against master:

    git checkout origin/master -- lib/

You'll see that we've been leaking the following events:

  * destroyed, cancel, complete, error

There are some events we weren't leaking, but I added tests for them
anyway:

  * paused, active, change

~~The reason the tests are slow is because we have no way to configure
the backoff (currently a random number between 0 and 2000 ms).
Once we add a way to configure the backoff, we can make these tests
faster.~~ Nevermind, realized we do have that option. Fixed.


lib/utils.js
```diff
@@ -39,6 +39,21 @@ exports.pick = function (obj, arr) {
...
+    return listener._key === key;
+  });
+  if (listeners.length) {
+    return;
+  }
+  emitter.once(event, listener);
+  listeners = emitter.listeners(event);
+  listeners[listeners.length - 1]._key = key;
+};
+
```
nolanlawson: This `oncePerKey` works, but maybe it's poor Node style, because I don't know any better. @calvinmetcalf Is there a more canonical way of saying "do this once, but only if I haven't already told you to do it once?"

lib/replicate.js
```diff
@@ -97,6 +102,11 @@ Replication.prototype.cancel = function () {
 
 Replication.prototype.ready = function (src, target) {
   var self = this;
+  if (self._readyCalled) {
+    return;
+  }
+  self._readyCalled = true;
+
```
nolanlawson: @dholth This is equivalent to your `ready()` no-op.

nolanlawson:
OK, sped up the tests by using the `back_off_function` option.

lib/utils.js
dholth: I think this will cause problems. The first set of listeners attached with oncePerKey() will reference the state from the first call to replicate()'s closure. If the second or third call to replicate() uses those callbacks referencing the original closure, then all of these var's will be different from the current replication attempt compared to the ones inside the callback: https://github.com/pouchdb/pouchdb/blob/3699-3824/lib/replicate.js#L150

A different way to solve the same leak might be to have the cleanup callbacks all live on the Replication() class so they cannot reference the replicate() function's closure. The ordinary "if existing callback === new callback" check could be used since the cleanup handlers would now be identical between replicate() calls.

And/or instead of calling replicate() multiple times, backOff() could call a private function that doesn't try to install the cleanup handlers. Example:

```javascript
function replicate() {
   install_cleanup_handlers();
   do_replication();
}

function do_replication() {
    // do it
}

function backOff() {
    // wait
    do_replication();
}
```

nolanlawson: I could also make `oncePerKey` remove the old listeners and add the new one. Should be 6 of one, half a dozen of the other.

In general, the way I would like to solve this in the future is for `retry` replication to just use `live` replication's APIs &ndash; i.e. to be implemented on top of it, rather than inside of it.

In any case, I don't think this objection is enough to block this PR. Sure the original listeners will maintain references to the original closure, but at least the listeners don't grow out of control. :)

dholth: Yes the remove - add strategy would be more resilient in case someone decides to make the proper functioning of the cleanup functions depend on the value of the closure variables.

Does the replication behave correctly if the source or destination fires 'destroy' while backOff is waiting?

nolanlawson: Probably not - that's a good test. :)

nolanlawson: Actually wait, there is no way to sub out the old listener for a new listener, unless we hold on to a reference to all listeners. Ugh.

nolanlawson:
Okay, made some changes:

* Heavy refactoring. Split up `replicate.js` into several separate files.
* Got rid of deprecated/undocumented stuff: say goodbye to `uptodate`, `outofdate`, and `default_back_off`.
* No more `utils.oncePerKey`, replaced with a simple `if returnValue._listenersAdded` check
* Added and tweaked the tests

@dholth To respond to your original comment, it doesn't matter if we keep the original listener for `'cancel'`, `'change'`, and `'complete'`, because their functionality doesn't change between retries.

Summary: I'm gonna solve these damn retry bugs, but it's going to have to be done in stages. First step is refactoring so that I can start to modularize and understand the code. Also adding tons of tests.

daleharvey:
Awesome work, this was the bug I was planning to work on next and beat me to it :)

https://github.com/pouchdb/pouchdb/commit/d8902a1825d3c60d0b0151b42569ee1bb6d283ff

Cheers

Issue #3826, opened 2015-05-10
Document backoff function


nolanlawson:
As discussed in #3825, right now it's hard-coded in `replicate.js` to 2000 milliseconds. We could speed up the tests a lot and provide a nice feature for users if that was configurable.

Suggestion:
```js
replicate({
  live: true,
  retry: true,
  backoff: 1000 // <-- defaults to 2000
});
```

nolanlawson:
Hah, I just realized we do have this option; it's just not documented.

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/f76c53a7463641b625cd04f6fe7f2f07e750c735

PR #3828, opened 2015-05-10
(#3826) - document the backoff function option

nolanlawson:
I decided not to document the `default_back_off` option because
I think users don't actually need it, and it might be more
confusing to document it. Simpler if we just say that the
value always starts at 0.


daleharvey:
This looks good, cheers

https://github.com/pouchdb/pouchdb/commit/f76c53a7463641b625cd04f6fe7f2f07e750c735

PR #3829, opened 2015-05-11
(#3768) - fix replication when id() errors

### LINKED ISSUE #3768: Retry replication emits 'error' if started while offline
This is a pretty bad bug; users aren't supposed to have to handle `'error'`s during "retry" replication unless something goes really horribly wrong. See [this SO question](http://stackoverflow.com/questions/29880471/pouchdb-database-replication-when-the-application-is-launched-offline) for context.

nolanlawson:
This doesn't fully fix #3768, but it's a step in the right direction.

If you use the Chrome dev tools to make the browser go offline while running any of these tests, then you can repro the error:

```
{"status":500,"name":"unknown_error","message":"Database encountered an unknown error"}
```

This fix just ensures that, if `id()` errors during the `genReplicationId()` function, an `'error'` event is correctly emitted during live non-retry replication, but during live retry replication we will recover.

Rebased on top of https://github.com/pouchdb/pouchdb/pull/3825 because I am lazy. The `test.retry.js` test fails before the fix but succeeds after. The other test succeeds in both; it's just a sanity check.


nolanlawson:
failures seem genuine

nolanlawson:
closing in favor of a future where I am smart enough to do this

PR #3831, opened 2015-05-11
Fix call to es3ify to work on windows

yaronyg:
We need to add in an explicit call to node for it to work on Windows.

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
   },
   "scripts": {
     "build-js": "npm run build-main-js && npm run min && npm run build-plugins",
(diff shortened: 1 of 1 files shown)


nolanlawson:
We should just run a Travis job to fail if the build fails on Windows... :P

daleharvey:
Good call that would be useful, looks like travis wouldnt be an option though :( (http://docs.travis-ci.com/user/multi-os/)

daleharvey:
In the mean time :+1: 

https://github.com/pouchdb/pouchdb/commit/b9ba2520714e3b8ec0e90e622dd7d65ed21241b4 

cheers

PR #3832, opened 2015-05-12
(#3768) - fix replication when http adapter fails

nolanlawson:
Rebased on top of #3825 because yolo

Also yeah, the `'active'` tests have been tweaked
now to be a little more forgiving. I don't think
we ever really understood exactly how many
`'active'` events there ought to be. I aim to fix
that, but not in this pull request.

Fixing retry replication is going to be a slow and
meticulous process.


tests/integration/test.replication.js
```diff
@@ -1355,7 +1355,7 @@ adapters.forEach(function (adapters) {
         db.replicate.to(dbs.remote)
         .on('complete', function (res) {
           should.exist(res);
-          db.replicate.to('http://0.0.0.0:0')
+          db.replicate.to('http://0.0.0.0:13370')
```
nolanlawson: This test was actually failing in Firefox. Who knows.

daleharvey:
:+1: for https://github.com/pouchdb/pouchdb/commit/f279655f12d308c957c1ab147fa37b9a362dd6eb, new tests look nice

daleharvey:
Merged in https://github.com/pouchdb/pouchdb/commit/2afb03d1d1042814242d4a856b9a9f5c6a4842db

PR #3835, opened 2015-05-12
add androidDatabaseImplementation option for websql

rubensayshi:
this option has been added to the cordova sqlite plugin to fallback to an older sqlite implementation; https://github.com/litehelpers/Cordova-sqlite-storage#android-sqlite-implementation

### DIFF
--- a/lib/adapters/websql/websql.js
+++ b/lib/adapters/websql/websql.js
@@ -105,7 +105,8 @@ function WebSqlPouch(opts, callback) {
     description: api._name,
     size: size,
     location: opts.location,
-    createFromLocation: opts.createFromLocation
+    createFromLocation: opts.createFromLocation,
+    androidDatabaseImplementation: opts.androidDatabaseImplementation
   });
   if (!db) {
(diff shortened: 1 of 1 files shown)


nolanlawson:
It pains me to keep adding hacks like this to support the SQLite Plugin.

That being said, this is a huge issue for people trying to use both libraries together, so I'm +1. Could you also add some documentation to [this file](https://github.com/pouchdb/pouchdb/blob/2afb03d1d1042814242d4a856b9a9f5c6a4842db/docs/_includes/api/create_database.html) to explain the new parameter to people? We should probably also explicitly recommend the "2" implementation, since "1" seems quite broken.

I would also be very happy to see a version of this PR that simply passed all the options in, or merged it into a new options object, so that we don't have to keep whitelisting options as the SQLite Plugin adds them.

nolanlawson:
Added some documentation, fixed up the commit message, and merged as 9d1b0328c2380ad6a9eb7cd524ac743860c947fd.

Whatever we have to do to get the SQLite Plugin to work again, I guess we should do that.

Issue #3838, opened 2015-05-13
Proxy configuration and node 0.12


mainakae:
Hello all,

I'm working with Ubuntu 14.04 and node v0.12.2.

I've developed several applications using PouchDB to interact with CouchDB through HTTP, and have come to notice that while on node v0.10, PouchDB (or perhaps more accurately, "request") successfully used the environment variables ($HTTP_PROXY / $http_proxy...) to route request via the proxy, in the latest node version it is not longer the case. Even worse: when trying to manually force the use of a proxy, it throws an "TypeError: Request path contains unescaped characters.".

Fortunately, I've checked with the latest version of "request" and it definitely works right with node 0.12. I've manually updated pouchdb's package.json et voila: the problem is fixed. I haven't fully tested that nothing else is broken as a result of the version bump, though.

Summary: update "request" dependency on package.json to "~2.55.0" to fix proxy configuration with node version 0.12

daleharvey:
Awesome, you have a commit up there with the fix, would you mind turning that into a PR?

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/fc48a9bff3cf04ab6a58424a0ab761b3debce341

PR #3840, opened 2015-05-13
Fixing #3838 (updating request to latest)

mainakae:
Fixing proxy issues on node v0.12, result of using old version of request lib.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "pouchdb-extend": "^0.1.2",
     "pouchdb-mapreduce": "~2.3.2",
     "pouchdb-upsert": "^1.0.2",
-    "request": "~2.28.0",
+    "request": "~2.55.0",
     "spark-md5": "0.0.5",
     "through2": "^0.4.1",
     "vuvuzela": "^1.0.0"


daleharvey:
Hrm this is failing in travis which by default tests 0.10, we should at the very least be testing .12, whether we want to drop support for .10 is a different question though

nolanlawson:
dropping 0.10 support should be a major version bump (surprised `request` didn't do that). Could be done with 4.0.0, but I suspect we would lose a lot of users because many of us are still stuck on 0.10 due to uncertainty especially with the node/io split.

Believe it or not, I myself am still mostly running 0.10, due to compatibility headaches that I don't want to deal with just yet.

nolanlawson:
Level's solution was to (temporarily?) publish a `level-11` package with 0.11 support. We could maybe do a `pouchdb-12` branch.

daleharvey:
Yeh locally I am still using 0.10 too and we should be cautious about dropping older version support, looking into at least having tests run across versions before we decide on anything

mainakae:
Seems 0.12 has broken some things, yup.

Still have a machine with 0.10.33, and now I'm trying to run the test battery with the current version of Pouch. Running into some errors, I'll try to make sense of it, probably my fault, not configuring properly the test environment or something.

daleharvey:
Ok should hopefully have multiple versions of nodejs testing working in https://github.com/pouchdb/pouchdb/pull/3844

nolanlawson:
Seems the new version of request fails our node tests. I'm getting an error in the attachments test:

```
  1) test.attachments.js- local:http Many many attachments replicate:
     Something wrong with the request
```

nolanlawson:
Apparently related to request and not Node 0.10 vs 0.12, because Travis shows the same error with 0.10: https://travis-ci.org/pouchdb/pouchdb/jobs/62392954

nolanlawson:
This seems to be failing the tests. Please reopen if you're able to get it working; else I'm happy to stay stuck on an old version.

By way of precedent, I will note that we are still on an old version of LevelDB even, due to breaking changed in `leveldown`. I don't believe code rots if left out in the sun for too long, so I am happy to stay on an older version if it's passing the tests. :)

PR #3845, opened 2015-05-14
(#3845) - add xmlns:gap to cordova config

nolanlawson:
As pointed out by @brodybits in
https://github.com/litehelpers/Cordova-sqlite-storage/issues/255.
To be honest, I don't even know how it was passing
in Blackberry when I added this. It's not valid XML without
this fix.


daleharvey:
Heh happy trusting this is the right thing till we get it automated, cheers

https://github.com/pouchdb/pouchdb/commit/46a99a35900504b1c665760e9118b2e47fdabd7e

brodybits:
:+1:

PR #3851, opened 2015-05-16
(#3768) - truly test offline

nolanlawson:
I wanted to be 100% certain that #3768 is fixed,
so I wrote a test that actually stubs XHR with
an XHR implementation that does the same thing
that would happen if the user is offline. Happily
I can report that the tests pass.

### DIFF
--- a/lib/deps/request-browser.js
+++ b/lib/deps/request-browser.js
@@ -115,7 +115,11 @@ function xhRequest(options, callback) {
     xhr.abort();
   };
 
-  if (options.xhr) {
+  if (typeof PouchXMLHttpRequest !== 'undefined') {
+    // this is only used for unit tests
+    /* global PouchXMLHttpRequest */
+    xhr = new PouchXMLHttpRequest();
+  } else if (options.xhr) {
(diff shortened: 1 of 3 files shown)


nolanlawson:
Okay good, well this seems to reveal a race condition, because the test is actually occasionally failing.

nolanlawson:
Yeah unfortunately this trick doesn't work with Selenium, so it's not nearly as clever as I thought. Can't think of another way. `options.xhr` is really unreliable.

nolanlawson:
Okay, maybe this is dumb, but my fix is to have `request-browser.js` look for a global `PouchXMLHTTPRequest` object and use that if it exists.

Hacky as hell, but frankly I kind of like it because 1) it actually allows us to test true offline state without messing with Selenium/any other HTTP requests, and 2) I think it's actually simpler than the undocumented `options.ajax.xhr` thing, which I noticed was flaky anyway because `http.js` wasn't consistently passing it in to all its `ajax` requests.

nolanlawson:
Still not working. Seems like a Travis-only failure. I give up.

Issue #3853, opened 2015-05-17
exec-sync not supported in Node 0.12 or iojs


nolanlawson:
So we should probably stop using it. We're only using it for the `build-site.js` script anyway.

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/571952fba7dec9d0e0262c7496ec18b16843dc59

PR #3854, opened 2015-05-17
(#3853) remove exec-sync to support 0.12

nolanlawson:
We don't have Travis tests to ensure that `npm run build-site` works, but I tested informally and it's fine for me (live reload, can do `BUILD=1`).

### DIFF
--- a/bin/build-site.js
+++ b/bin/build-site.js
@@ -2,51 +2,52 @@
 
 'use strict';
 
-var fs = require('fs');
-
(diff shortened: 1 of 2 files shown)


daleharvey:
Awesome, all the nodes back to green, I removed them from allowed failures

https://github.com/pouchdb/pouchdb/commit/571952fba7dec9d0e0262c7496ec18b16843dc59

PR #3856, opened 2015-05-17
(#3847) - cancel sync in an onunload event

### LINKED ISSUE #3847: Live Sync + window.location.reload throws unhandled exception
Title should say it all.

I gisted a workaround (in coffee) : https://gist.github.com/mrzor/8d0d12be413e873682b8 
And there is a jsfiddle that reproduces the error : http://jsfiddle.net/mrzor/51ps6qza/2/

Catch it with your devtools / pause on exceptions.
No need to "pause on caught exception" as this one appears uncaught despite the fact that `.on('error', ...)` is set.

nolanlawson:
This may or may not fix #3847. I can't tell, because
I can't actually reproduce it.

This commit should probably not be merged without
further discussion and an actual reproducible test.

### DIFF
--- a/lib/replicate/replicate.js
+++ b/lib/replicate/replicate.js
@@ -473,6 +473,13 @@ function replicate(src, target, opts, returnValue, result) {
         opts.complete(null, result);
       });
     }
+    if (typeof window !== 'undefined') {
+      // cancel when the page is refreshed
+      window.addEventListener('beforeunload', function unloadListener() {
+        returnValue.cancel();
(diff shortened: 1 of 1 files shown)


daleharvey:
Yeh one of the fairly fundemental assumptions is that we have a clean state between page refreshes, if this is broken it would be a fairly big deal

nolanlawson:
Closing this because we still cannot reproduce. (@mrzor? @prust? little help? :smiley)

PR #3858, opened 2015-05-17
(#3858) - fix all EventEmitter leaks in tests

nolanlawson:
This also adds [a new module I wrote](https://github.com/nolanlawson/throw-max-listeners-error) which throws an error
if the maximum number of listeners is ever exceeded. It
only runs in the Node tests, but this should be a great
way to prevent future memory leaks.

I can confirm that the module works, and that these fixes were
actually necessary to get the tests passing again.

### DIFF
--- a/lib/setup.js
+++ b/lib/setup.js
@@ -145,7 +145,7 @@ PouchDB.defaults = function (defaultOpts) {
   eventEmitterMethods.forEach(function (method) {
     PouchAlt[method] = eventEmitter[method].bind(eventEmitter);
   });
-  PouchAlt.setMaxListeners(0);
+  PouchAlt.setMaxListeners(10);
 
   PouchAlt.preferredAdapters = PouchDB.preferredAdapters.slice();
   Object.keys(PouchDB).forEach(function (key) {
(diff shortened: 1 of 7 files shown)


lib/setup.js
```diff
@@ -145,7 +145,7 @@ PouchDB.defaults = function (defaultOpts) {
   eventEmitterMethods.forEach(function (method) {
     PouchAlt[method] = eventEmitter[method].bind(eventEmitter);
   });
-  PouchAlt.setMaxListeners(0);
+  PouchAlt.setMaxListeners(10);
```
nolanlawson: I don't know what we were thinking setting this to 0. The default is 10.

lib/utils.js
```diff
@@ -150,7 +150,7 @@ function Changes() {
   var self = this;
   EventEmitter.call(this);
   this.isChrome = isChromeApp();
-  this.listeners = {};
+  this._listeners = {};
```
nolanlawson: I also assume this was a mistake, to overwrite the built-in `EventEmitter.prototype.listeners()` method.

daleharvey:
Very nice, thanks

https://github.com/pouchdb/pouchdb/commit/f83e0e3b62fc7d4e28669fc128f3228d082d2d94

Issue #3846, opened 2015-05-15
Use fakeIndexedDB to speed up our tests


nolanlawson:
We could also test it in Node.js, it seems: https://github.com/dumbmatter/fakeIndexedDB

PR #3859, opened 2015-05-17
(#3846) - don't merge - test in fakeIndexedDB

nolanlawson:
I don't actually want to merge this; I just want to
confirm that it's testing on Travis.

This thing is interesting, but I don't think it actually
buys us much. It's not much faster than regular IndexedDB,
it would be tricky for us to run it in Node, and out of all
the browsers we test, it only works in PhantomJS, because it
turns out you can't override window.indexedDB in either Firefox
or Chrome. But it is kinda neat, and it does seem to pass the tests!


nolanlawson:
Satisfied my curiosity, happy to close this now.

PR #3862, opened 2015-05-20
(#3862) - add putAttachment() test

nolanlawson:
Based on a SO question, I was worried
that we didn't support the case where `putAttachment()`
is creating a new document with a base64 string.
But it seems to work.


marten-de-vries:
Looks good. +1, merged: 6a31c99f167ca6f2701be1762cac364b531f7fa0

Issue #3642, opened 2015-03-16
PouchDB using ungodly amounts of memory with big attachments


nolanlawson:
Reported via email:

> in the end of the last year I was implementing offline functionalities for our online-campus. First I was using PouchDB for storing the offline data, until I stumbled about a problem/bug and performance issues while storing large pdf files (>40MB) as blob. While storing the pdf, the tab in Google Chrome crashes. On Firefox everything works well.
> 
> I filed an issue (https://code.google.com/p/chromium/issues/detail?id=412409) on Chromiums bug tracker and posted a [test script](https://code.google.com/p/chromium/issues/detail?id=412409#hc18) this week, which reproduces the problem. Chris Mumford from the chromium team detected, that the RAM usage exceeds 1,5GB while storing the pdf with PouchDB and his „guess“
> 
> is that there is something inefficient in pouchDB that uses more RAM than  necessary - and Chrome (my debug build) just hits it before Firefox does. Maybe references held until the test ends whereby they can be GC’d.
> 
> 
> Can you have a look on my testscript and check what leads to this high RAM usage when storing large blobs with PouchDB?
> 
> As much as I can say, I am using PouchDBs API as it should be, but maybe there is also a problem in my code (200 lines).
> Thanks in advance for your help.

[This gist](http://bl.ocks.org/nolanlawson/38e3cd6705f50b074566) says whether or not Chrome is using base64. Seems unrelated, though, since Firefox also uses a lot of memory. MD5 problem?

nolanlawson:
Duplicate of #3640, my bad.

PR #3863, opened 2015-05-20
(#3642) - update to latest leveldown/levelup

nolanlawson:
This seems to be failing due to the migration tests,
but I'm opening a PR so we can see exactly what
is failing.


nolanlawson:
So this is very red ([test run here](https://travis-ci.org/pouchdb/pouchdb/builds/63268247)), which I think indicates that we cannot update levelup/leveldown without a migration. I'm not interested in this right now, but anyone who wants to pick up the slack is welcome.

PR #3865, opened 2015-05-21
(#3865) - temporarily add uptodate event back in

nolanlawson:
I'd like to do a patch/minor release before the
major release, so let's put the `uptodate` event
back in, before we remove it for 4.0.0.

Unfortunately it appears there aren't (and weren't) any
tests for `uptodate`, but you can compare this
to an older version to see that this is where we were
emitting it:
https://github.com/pouchdb/pouchdb/blob/88336ad236926bffce41b40c7566febade6ef7d0/lib/replicate.js#L359-L360

### DIFF
--- a/lib/replicate/replicate.js
+++ b/lib/replicate/replicate.js
@@ -178,6 +178,7 @@ function replicate(src, target, opts, returnValue, result) {
         if ((continuous && changesOpts.live) || changesCompleted) {
           returnValue.state = 'pending';
           returnValue.emit('paused');
+          returnValue.emit('uptodate', result);
         }
         if (changesCompleted) {
           completeReplication();


daleharvey:
When did we actually remove this? I thought it was long before the last release

nolanlawson:
It's never been officially removed only deprecated: https://github.com/pouchdb/pouchdb/wiki/Breaking-changes

daleharvey:
But havent we already dont various releases that dont include this? remember having a conversation about whether it was breaking or not before but if we have already done the release

nolanlawson:
No, this was (accidentally) removed in my https://github.com/pouchdb/pouchdb/commit/d8902a1825d3c60d0b0151b42569ee1bb6d283ff commit, which hasn't been published yet.

daleharvey:
Ah ok, then yeh :+1: apologies for the confusion

daleharvey:
Merged https://github.com/pouchdb/pouchdb/commit/91e668e6dba811d8761995d040bfbcce53a9c010

PR #3866, opened 2015-05-21
(#3611) - foo as shorthand for foo/foo in filters

### LINKED ISSUE #3611: Should be able to filter using names like 'foo' instead of 'foo/bar'
We support this in mapreduce, so we ought to support it for filtered changes/replication as well.

E.g. if your ddoc is:

```js
{
  _id: '_design/foo',
  filters: {
    'foo': function (doc) { /* ... */}.toString()
  }
}
```

Then you should be able to filter using just `'foo'` instead of `'foo/foo'`.

thriqon:
These were the only instances I found where a name like `foo/foo` is needed.

fixes #3611

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -400,3 +400,22 @@ exports.safeJsonStringify = function safeJsonStringify(json) {
     return vuvuzela.stringify(json);
   }
 };
+
+exports.parseDesignDocFunctionName = function (s) {
+  if (!s) {
+    return null;
+  }
+  var parts = s.split('/');
+  if (parts.length === 2) {
+    return parts;
+  } else if (parts.length === 1) {
+    return [s, s];
+  } else {
+    return null;
+  }
+};
+
+exports.normalizeDesignDocFunctionName = function (s) {
+	var normalized = this.parseDesignDocFunctionName(s);
+	return normalized ? normalized.join('/') : null;
+};
--- a/docs/_includes/api/changes.html
+++ b/docs/_includes/api/changes.html
@@ -28,7 +28,7 @@ All options default to `false` unless otherwise specified.
 * `options.filter`: Reference a filter function from a design document to selectively get updates. To use a view function, pass `_view` here and provide a reference to the view function in `options.view`. See [filtered changes](#filtered-changes) for details.
 * `options.doc_ids`: Only show changes for docs with these ids (array of strings).
 * `options.query_params`: Object containing properties that are passed to the filter function, e.g. `{"foo:"bar"}`, where `"bar"` will be available in the filter function as `params.query.foo`. To access the `params`, define your filter function like `function (doc, params) {/* ... */}`.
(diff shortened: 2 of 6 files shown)


nolanlawson:
Test failures seem genuine; could you re-run the test?

Also it would be preferable to add a new test instead of rewriting the old test, so that we can be sure to support both formats. :) Other than that, this PR looks exceptional; tests and documentation all in the same commit, nice work!

thriqon:
I'm currently investigating the failure, it appears to be caused by pouchdb-server behaving slightly differently than CouchDB...

I will certainly add a new test, if that is preferred.

nolanlawson:
Yep, we never say no to new tests. :)

lib/utils.js
```diff
@@ -788,3 +788,18 @@ exports.safeJsonStringify = function safeJsonStringify(json) {
...
 };
+
+exports.parseDdocFunctionName = function (s) {
+  var parts = s.split('/');
+  if (parts.length === 2) {
+    return parts;
+  } else if (parts.length === 1) {
+    return [s, s];
+  } else {
+    throw ("Unrecognized function name");
```
daleharvey: I would much rather be returning false here, we arent being super consistent about it but generally exceptions are exceptional, this is user input and can be expected to fail

Also as a style need there no need for the `} else`'s and I think they confuse the fact that it will early return

tests/integration/test.utils.js
```diff
@@ -0,0 +1,32 @@
+
```
daleharvey: We have /tests/unit/ for unit tests

daleharvey:
Yeh I have a few little nits but this looks great, I would expect pouchdb-server and express-router will need some fixes for the routing, ill take a look at express router

nolanlawson:
There shouldn't be any changes to express-pouchdb or pouchdb-express-router, since CouchDB never accepted this `foo` format anyway. I just tested 1.6.1 to double-check:

```
{"error":"bad_request","reason":"filter parameter must be of the form `designname/filtername`"}
```

So I think instead the `http.js` will have to be modified to convert `foo` -> `foo/foo`, unless this is already passing because we're handling that somehow.

thriqon:
I updated the branch, but I'm not able to make the Travis build pass. It always fails (with seemingly unrelated stuff).

nolanlawson:
We have a lot of intermittent test failures. Sorry about that; I've restarted them.

daleharvey:
:+1: this is awesome, the only fail is in couch master which is permafail, once we get that green / ignored this looks good to land, thanks

nolanlawson:
Green in Travis, so merged as 4ce82edcb2ef8b72a3a433f455e7fc1ecc733f9b. Fixed some weird indentation issues (tabs?) that somehow JSHint didn't complain about, but other than that: great tests, great documentation, great code &ndash; basically a perfect pull request. Nice work. :+1:

thriqon:
Thanks for this nice words and the guidance :) looking forward to contribute again...

Issue #2832, opened 2014-10-05
Use multipart/mixed for attachments


nolanlawson:
`multipart/mixed` is more efficient for sending attachments around, and plus Sync Gateway only supports that format (#2676 and #2681).

jchris:
I think this is the place to discuss which implementation to test and how to test it. I think Apache CouchDB will use this if we support it so we maybe it's an easy first place to work on Sync Gateway interop.

I'd like to learn how to interact with the test automation, maybe this is a good chance to do it?

nolanlawson:
Yep, currently non-org members won't have their PRs tested in Saucelabs, but all the non-Saucelabs tests will be run. Probably the most important thing here is to verify support for PouchDB Master as well as CouchDB 1.6, so it should be fine for now. After that, we can run the Sauce tests to verify the proprietary browsers.

jchris:
Is this the change we should be testing? https://github.com/preplay/pouchdb/commit/296df7af184ef496681b517590415df7d3c200b5

nolanlawson:
Yep, that looks right. TBH can't remember why I stopped working on it.

nolanlawson:
https://github.com/preplay/pouchdb/commit/296df7af184ef496681b517590415df7d3c200b5 looks the same as 93b40b0d94b86516213b082ca4f3f6a9ac3e3e27 - maybe @preplay just rebased it?

gabrielnau:
Sorry for the late reply, I didn't see the Github notification.
Yes indeed, I just rebased it against master.

nolanlawson:
Started looking into multipart again. It's gonna be a mess.

First off, it appears CouchDB returns the content as gzip-encoded. Meaning I guess we'll need to bundle a gzip decoder with PouchDB so that we can actually decode that content?? Unless there's a good Blob-related hack we can use to make the browser do it (or something...).

Second, I'm stuck right now trying to actually decode multipart/mixed. If anybody can figure out how I'm supposed to decode a string like `'H4sIAAAAAAAAAwMAAAAAAAAAAAA=\n'` (which should be a base64-encoded, gzip-encoded string `"This is a base64 encoded text"`), then I'd be happy to hear how.

If anyone is curious, here is an example of a full multipart/mixed response from CouchDB, completely base64 encoded: https://gist.github.com/nolanlawson/5ecdce6ea62ecb7ded4c

nolanlawson:
[...]

PR #3870, opened 2015-05-24
(#2832) - avoid application/json for doc GETs

nolanlawson:
This is step 1 to get Sync Gateway support, as well as to have
more efficient attachment parsing.

My solution involves the insight that multipart/mixed is actually
really stupidly difficult to parse in the browser (gzipped
binary strings? wtf), so it makes more sense to just avoid using
`attachments: true` with `get()`s entirely, and do separate
HTTP GET requests instead.

Now, usually I'm opposed to additional HTTP requests, but in this case
I think it makes sense. By fetching attachments independently,
the browser can do the hard work of gzip-decoding, and we can still
send attachments over the wire in the most efficient binary format.
People who don't like this, and who prefer to have everything in a single
JSON object, can just avoid using attachments (as they should anyway,
for small bits of text data).

I decided to put this logic in `http.js` itself rather than the replicator,
because 1) I see no case where anybody would prefer actually passing
`attachments:true` to CouchDB and get the attachments back in base64 format,
and 2) this helps us get ALL the tests passing against CSG, even those
where we directly `get()` with `attachments:true`.

Step 2 for CSG support will be to support PUTing documents with multipart/related
type. This PR does not do that.

This PR also has some unrelated fixes that I noticed as I was writing it:

1) accidental usage of raw `btoa` in idb.js
2) ~~we weren't actually supporting `getAttachment` with the `rev` option. Whoops!~~ moved to #3871

### DIFF
--- a/lib/adapters/http/http.js
+++ b/lib/adapters/http/http.js
@@ -12,10 +12,25 @@ var MAX_URL_LENGTH = 1800;
 var readAsBinaryString = require('../../deps/binary/readAsBinaryString');
 var binaryStringToBlob = require('../../deps/binary/binaryStringToBlob');
 var utils = require('../../utils');
+var Promise = utils.Promise;
+var base64 = require('../../deps/binary/base64');
+var btoa = base64.btoa;
+var atob = base64.atob;
 var errors = require('../../deps/errors');
 var log = require('debug')('pouchdb:http');
 var isBrowser = typeof process === 'undefined' || process.browser;
 var buffer = require('../../deps/buffer');
+
+function blobToBase64(blobOrBuffer) {
+  if (!isBrowser) {
+    return Promise.resolve(blobOrBuffer.toString('base64'));
+  }
+  return new Promise(function (resolve) {
+    readAsBinaryString(blobOrBuffer, function (bin) {
+      resolve(btoa(bin));
+    });
+  });
+}
 
 function encodeDocId(id) {
   if (/^_design/.test(id)) {
@@ -29,22 +44,15 @@ function encodeDocId(id) {
 
 function preprocessAttachments(doc) {
   if (!doc._attachments || !Object.keys(doc._attachments)) {
-    return utils.Promise.resolve();
+    return Promise.resolve();
   }
 
-  return utils.Promise.all(Object.keys(doc._attachments).map(function (key) {
+  return Promise.all(Object.keys(doc._attachments).map(function (key) {
     var attachment = doc._attachments[key];
     if (attachment.data && typeof attachment.data !== 'string') {
-      if (isBrowser) {
-        return new utils.Promise(function (resolve) {
-          readAsBinaryString(attachment.data, function (binary) {
-            attachment.data = utils.btoa(binary);
-            resolve();
-          });
-        });
-      } else {
-        attachment.data = attachment.data.toString('base64');
-      }
+      return blobToBase64(attachment.data).then(function (b64) {
+        attachment.data = b64;
+      });
     }
   }));
 }
@@ -82,7 +90,7 @@ function getHost(name, opts) {
 
     if (opts.auth || uri.auth) {
       var nAuth = opts.auth || uri.auth;
-      var token = utils.btoa(nAuth.username + ':' + nAuth.password);
+      var token = btoa(nAuth.username + ':' + nAuth.password);
       uri.headers.Authorization = 'Basic ' + token;
     }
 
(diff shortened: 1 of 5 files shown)


lib/adapters/http/http.js
```diff
@@ -418,15 +447,15 @@ function HttpPouch(opts, callback) {
...
     }
-    opts = utils.clone(opts);
-    if (opts.auto_encode === undefined) {
-      opts.auto_encode = true;
-    }
-    if (opts.auto_encode) {
-      docId = encodeDocId(docId);
-    }
-    opts.auto_encode = false;
-    api.get(docId + '/' + encodeAttachmentId(attachmentId), opts, callback);
```
nolanlawson: I wasn't a big fan of overloading `get()` here for both attachment-fetching and document-fetching. In my opinion, it's cleaner to  just directly call `ajax()`.

docs/_includes/api/get_attachment.html
```diff
@@ -6,11 +6,17 @@
 
 Get attachment data.
```
nolanlawson: Improved the `getAttachment` documentation and added notes on `options.rev`.

lib/adapters/http/http.js
```diff
@@ -10,11 +10,21 @@ var CHANGES_BATCH_SIZE = 25;
 var MAX_URL_LENGTH = 1800;
 
 var utils = require('../../utils');
+var Promise = utils.Promise;
+var base64 = require('../../deps/base64');
+var btoa = base64.btoa;
+var atob = base64.atob;
```
nolanlawson: I'm trying to get in the habit of moving dependencies up top, because 1) it makes it easier to extract them from `utils` later, and 2) uglify can shorten the variable names.

lib/adapters/http/http.js
```diff
@@ -271,9 +289,6 @@ function HttpPouch(opts, callback) {
       opts = {};
     }
     opts = utils.clone(opts);
-    if (opts.auto_encode === undefined) {
-      opts.auto_encode = true;
-    }
```
nolanlawson: This `auto_encode` option was undocumented and not doing anything.

lib/adapters/http/http.js
```diff
@@ -344,30 +350,53 @@ function HttpPouch(opts, callback) {
     var getRequestAjaxOpts = opts.ajax || {};
     utils.extend(true, options, getRequestAjaxOpts);
 
-    // If the given id contains at least one '/' and the part before the '/'
-    // is NOT "_design" and is NOT "_local"
-    // OR
-    // If the given id contains at least two '/' and the part before the first
-    // '/' is "_design".
-    // TODO This second condition seems strange since if parts[0] === '_design'
-    // then we already know that parts[0] !== '_local'.
```
nolanlawson: This weird code noted in the TODO was no longer necessary once we stopped overloading `get()` for both attachments and documents.

tests/integration/test.attachments.js
```diff
@@ -1518,6 +1519,53 @@ adapters.forEach(function (adapter) {
...
+          return promise.then(function () {
+            throw new Error('expected an error');
+          }, function (err) {
+            should.exist(err);
+            err.status.should.equal(404);
+          });
+        }));
+      });
+    });
+
```
nolanlawson: We didn't previously have any tests for `rev` in `getAttachment`. Happily, this just worked&trade; for the local adapters, although it required a fix for the http adapter.

nolanlawson:
Split this PR into two commits - one to address #2832 and one for #3871.

nolanlawson:
Actually, scratch that &ndash; I broke this into two separate PRs.

nolanlawson:
This is passing in everything but `pouchdb-express-router` due to https://github.com/daleharvey/pouchdb-express-router/pull/7/files

lib/adapters/http/http.js
```diff
@@ -344,30 +351,53 @@ function HttpPouch(opts, callback) {
...
-    });
+    ajaxPromise(options).then(function (res) {
+      return Promise.resolve().then(function () {
+        if (opts.attachments) {
+          return fetchAllAttachments(res);
+        }
+      }).then(function () {
+        callback(null, res);
+      });
+    }).catch(callback);
```
nolanlawson: I probably should move this `then()` one lower. It's clearer that way. TODO, since right now I've got a lot of PRs on top of this one.

daleharvey:
I think this should probably have been implemented as an internal option, its a fairly big change that may have unknown consequences and it would make it easier to switch the behaviour, but it sounds like the sensible option for now, Cheers - https://github.com/pouchdb/pouchdb/commit/6e4aca8802c377aa93ccb95a13874338f43a8cb8

nolanlawson:
Yeah, I'm willing to make it an option on the http adapter later, but for now I just can't think of any case where a user would prefer to have attachments as base64-encoded text, even if it meant slightly fewer requests. (The individual attachments `GET`s are also parallelized, so it scales pretty well since we can do 8 requests at a time.)

Issue #3871, opened 2015-05-24
Support "rev" option in getAttachment()


nolanlawson:
I thought we supported getting an attachment for a particular revision, but apparently we don't.

daleharvey:
Merged

PR #3872, opened 2015-05-24
(#3871) - support "rev" option in getAttachment()

nolanlawson:
Originally this commit was part of #3870, but I broke
it into a separate PR to make it easier to digest.

I did rebase this on top of #3871, however, to avoid
bitrotting myself.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -20,7 +20,9 @@
   "globals": {
     "PouchDB": true,
     "should": true,
-    "Promise": true
+    "Promise": true,
+    "atob": true,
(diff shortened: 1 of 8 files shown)


nolanlawson:
Forgot that the "get an attachment at a particular rev" test should only be enabled when we're not doing auto-compaction.

daleharvey:
Looks great, cheers - https://github.com/pouchdb/pouchdb/commit/6eddeb39a83a63b899f1e5a75d3249b8507b5076

PR #3876, opened 2015-05-24
(#2832) - implement multipart/related

nolanlawson:
This *would* be the only commit necessary to upload documents
as multipart/mixed, but it appears there's a bug in CouchDB
1.6.1: https://issues.apache.org/jira/browse/COUCHDB-2700

I'm opening this pull request to confirm that the tests
pass in PouchDB Server, and also to see if they pass in
CouchDB master.

### DIFF
--- a/lib/adapters/http/http.js
+++ b/lib/adapters/http/http.js
@@ -13,6 +13,7 @@ var readAsBinaryString = require('../../deps/binary/readAsBinaryString');
 var binaryStringToBlob = require('../../deps/binary/binaryStringToBlob');
 var utils = require('../../utils');
 var Promise = utils.Promise;
+var clone = utils.clone;
 var base64 = require('../../deps/binary/base64');
 var btoa = base64.btoa;
 var atob = base64.atob;
@@ -20,6 +21,7 @@ var errors = require('../../deps/errors');
 var log = require('debug')('pouchdb:http');
 var isBrowser = typeof process === 'undefined' || process.browser;
 var buffer = require('../../deps/buffer');
+var createMultipart = require('../../deps/multipart');
 
 function blobToBase64(blobOrBuffer) {
   if (!isBrowser) {
@@ -85,7 +87,7 @@ function getHost(name, opts) {
     // except for the database name) with '/'s
     uri.path = parts.join('/');
     opts = opts || {};
-    opts = utils.clone(opts);
+    opts = clone(opts);
     uri.headers = opts.headers || (opts.ajax && opts.ajax.headers) || {};
 
     if (opts.auth || uri.auth) {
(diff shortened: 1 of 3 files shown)


nolanlawson:
@jchris Is `multipart/related` even required for CSG? I know that `multipart/mixed` is required for `GET /db/doc` with attachments, but wondering if it's even necessary for attachment uploads.

jchris:
I think if you are willing to do per attachment uploads you can get it to work but maybe the new_edits=false rev handling wouldn't be invoked... I'm not sure but if there's a downgrade option like this I think that's robust. SG can be patched, too.

nolanlawson:
@jchris Hm, I actually haven't looked into `_bulk_docs` yet, just `PUT /db/doc`. Plum forgot about that. So you're saying that CSG requires `multipart/related` for both?

And as for patching CSG: yeah, that may be easier in the short run, especially if it turns out there's no workaround for COUCHDB-2700. I'd help out, but I don't know Go. :stuck_out_tongue:

jchris:
I don't know which specific APIs are in SG off the top of my head, except
I'm fairly certain you can do a direct attachment GET PUT. The question is
can you do that in with replication-style rev handling...

Not the worst place to start reading some Go
https://github.com/couchbase/sync_gateway/blob/master/src/github.com/couchbase/sync_gateway/rest/bulk_api.go#L366

On Sun, May 24, 2015 at 4:23 PM, Nolan Lawson <notifications@github.com>
wrote:

> @jchris <https://github.com/jchris> Hm, I actually haven't looked into
> _bulk_docs yet, just PUT /db/doc. Plum forgot about that. So you're
> saying that CSG requires multipart/related for both?
>
> And as for patching CSG: yeah, that may be easier in the short run,
> especially if it turns out there's no workaround for COUCHDB-2700. I'd help
> out, but I don't know Go. [image: :stuck_out_tongue:]
>
> —
> <https://github.com/pouchdb/pouchdb/pull/3876#issuecomment-105078456>.
>



-- 
Chris Anderson
http://www.couchbase.com
http://twitter.com/jchris

nolanlawson:
Good news: the bug was on my end, and this seems to be working.

As pointed out by @kxepal in https://github.com/pouchdb/pouchdb/commit/15a1968798dcb1109a724f47dd05cab56d3544c4#commitcomment-11344522, though, a more efficient method would be to buffer the attachments as they are sent to the server. But this is still an improvement over our current "stringify everything as base64" method.

nolanlawson:
Waiting on some PRs in express-pouchdb and pouchdb-express-router in order for this to pass.

lib/adapters/http/http.js
```diff
@@ -77,12 +87,12 @@ function getHost(name, opts) {
     // except for the database name) with '/'s
     uri.path = parts.join('/');
     opts = opts || {};
-    opts = utils.clone(opts);
+    opts = clone(opts);
```
nolanlawson: If you're wondering why I changed so many of these, it's because one of them caused a tricky bug that was hard to track down. So I decided to start cloning everywhere just as a safety precaution for the future.

daleharvey:
Awesome job - https://github.com/pouchdb/pouchdb/commit/c6059d2717e1114fe723ffd4c6392344ae540f4d

PR #3877, opened 2015-05-26
(#3877) - refactor binary

nolanlawson:
Move them out of `utils.js` and into their own separate files.
None of this is a functional change, except for a small optimization
I noticed in `request-browser.js`, where we can directly use
`readAsArrayBuffer` instead of first converting to a binary string.

### DIFF
--- a/lib/deps/request-browser.js
+++ b/lib/deps/request-browser.js
@@ -2,8 +2,9 @@
 /* global Headers */
 'use strict';
 
-var createBlob = require('./blob.js');
+var createBlob = require('./binary/blob.js');
 var utils = require('../utils');
+var readAsArrayBuffer = require('./binary/readAsArrayBuffer');
 
 function wrappedFetch() {
   var wrappedPromise = {};
@@ -52,8 +53,8 @@ function fetchRequest(options, callback) {
   }
 
   if (options.body && (options.body instanceof Blob)) {
-    utils.readAsBinaryString(options.body, function(binary) {
-      fetchOptions.body = utils.fixBinary(binary);
+    readAsArrayBuffer(options.body, function (arrayBuffer) {
+      fetchOptions.body = arrayBuffer;
     });
   } else if (options.body &&
              options.processData &&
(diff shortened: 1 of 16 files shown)


lib/deps/request-browser.js
```diff
@@ -52,8 +53,8 @@ function fetchRequest(options, callback) {
   }
 
   if (options.body && (options.body instanceof Blob)) {
-    utils.readAsBinaryString(options.body, function(binary) {
-      fetchOptions.body = utils.fixBinary(binary);
```
nolanlawson: Here's where we can simplify by just directly using `readAsArrayBuffer`.

lib/deps/request-browser.js
```diff
@@ -203,8 +204,8 @@ function xhRequest(options, callback) {
   };
 
   if (options.body && (options.body instanceof Blob)) {
-    utils.readAsBinaryString(options.body, function (binary) {
-      xhr.send(utils.fixBinary(binary));
```
nolanlawson: Ditto.

nolanlawson:
As for file naming conventions, I think we should start moving away from `hyphenated-names-like-this.js`, at least internally. It looks weird when you're requiring a function, e.g. `var readAsArrayBuffer = require('read-as-array-buffer')`. I also noticed that libraries like lodash use camel case for inner modules, which is why you can do nice things like `require('lodash/array/zipObject')`.

nolanlawson:
Another note: this does increase the min+gz size from 47655 to 47725, probably due to the extra browserify boilerplate around each little function.

tests/integration/utils.js
```diff
@@ -59,6 +59,21 @@ testUtils.makeBlob = function (data, type) {
...
+};
+
+testUtils.btoa = function (arg) {
+  return PouchDB.utils.btoa(arg);
+};
+
+testUtils.atob = function (arg) {
+  return PouchDB.utils.atob(arg);
+};
+
```
nolanlawson: Trying to reduce our dependence on `PouchDB.utils` in the tests.

nolanlawson:
forgot the commit message format. we can fix on the way in

daleharvey:
all makes sense, cheers - https://github.com/pouchdb/pouchdb/commit/2376abe799d20d1ca84bdc6ed8efe3c5598a07a7

PR #3878, opened 2015-05-26
(#3878) - more efficient bit-twiddling for md5

nolanlawson:
If there's any place where we're justified in
micro-optimizations, it's gotta be md5summing
and related stuff.


daleharvey:
Heh fine be me, thanks - https://github.com/pouchdb/pouchdb/commit/03fa60890aef73b39d3f56dd7f9f5c3811388606

Issue #3881, opened 2015-05-27
Filter keys from _attachments object


nolanlawson:
As @daleharvey pointed out in https://github.com/daleharvey/pouchdb-express-router/pull/8, we should probably have some logic in PouchDB itself to filter out any keys that don't belong in `_attachments`, i.e. something other than digest, type, data, revpos, stub, etc.

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/5bad545888057436e46c66f7a71442db0340ea76

PR #3888, opened 2015-05-29
(#3881) - filter extraneous keys from _attachments

nolanlawson:
@daleharvey this is the fix that will make it so we don't
need to explicitly filter "follows" in express-pouchdb
and pouchdb-express-router. As it turns out, CouchDB
does this same thing if you include odd keys in the
_attachments object.

### DIFF
--- a/lib/adapter.js
+++ b/lib/adapter.js
@@ -34,6 +34,35 @@ function yankError(callback) {
   };
 }
 
+// clean docs given to us by the user
+function cleanDocs(docs) {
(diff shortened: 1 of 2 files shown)


daleharvey:
This is awesome, whenever there is a non http related change in express-pouchdb / router then I figure its likely something to be done in core - https://github.com/pouchdb/pouchdb/commit/5bad545888057436e46c66f7a71442db0340ea76

PR #3890, opened 2015-05-29
(#3885) - explicitly test for XHR

nolanlawson:
Proposed changed to #3885.


marten-de-vries:
This should be a good solution indeed. Merged & squashed into #3885. Thanks!

Issue #3894, opened 2015-05-29
Intermittent test failure: syncing two endpoints


nolanlawson:
I'm seeing this one fail constantly in Travis:

```
 1) test.sync.js-local-http Test syncing two endpoints (issue 838):

     Uncaught missing
```

Really annoying to have to keep restarting Travis all the time.

daleharvey:
Havent seen this in a while and tracking intermittents in https://github.com/pouchdb/pouchdb/issues/3934

PR #3895, opened 2015-05-30
(#3894) - fix race condition in sync when canceling

nolanlawson:
I managed to actually track this down and write
a test that fails much more consistently. (It's a race
condition, so the test pre-fix doesn't fail 100%
of the time, but it fails about 75% of the time, which
is decent.)

I also rewrote one of the old tests in Promise style,
just for kicks.


daleharvey:
This is awesome, cheers - https://github.com/pouchdb/pouchdb/commit/eea8d9eb3e48df62a61a3dc685d1123465498971

Issue #2035, opened 2014-04-24
Get rid of cors server


daleharvey:
its just messy, and I want to have the tests pick up COUCH_HOST

nolanlawson:
Seems iPhone is the only browser that fails when we try to remove the CORS proxy: https://github.com/pouchdb/pouchdb/pull/4576

daleharvey:
FINALLY - https://github.com/pouchdb/pouchdb/commit/2f7b57baca50e893a61c33f7de15835796f083a6

nickcolley:
Epic work 👍 🎆🎇🎉🎊

PR #3896, opened 2015-05-30
(#2035) - WIP - get rid of cors proxy

nolanlawson:
I'm really tired of this thing. It
slows down the test and causes
weird bugs occasionally. Let's try
removing it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,10 @@ before_script:
     # Lets know what CouchDB we are playing with
   - "curl -X GET http://127.0.0.1:5984/"
 
+    # set up CORS
+  - "npm install add-cors-to-couchdb"
+  - "./node_modules/.bin/add-cors-to-couchdb"
+
 script: npm run $COMMAND
 
 env:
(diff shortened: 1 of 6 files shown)


nolanlawson:
Shoot, apparently when I run couch master using `python dev/run -n 1 --with-admin-party-please`, there's no config database, so I can't enable CORS. @kxepal Do you have any idea how to fix that?

kxepal:
Sure I have (: /_config is not available on cluster port (x5984), but it still is on local one (x5986). And this means that you'll need to configure CORS on the each mode in order to sync their behavior.

nolanlawson:
I think this will be blocked for the moment on couchdb-master (@rnewson is looking into it!), but it's nice to see what fails in Travis.

nolanlawson:
Blocked by this Couch issue: https://issues.apache.org/jira/browse/COUCHDB-2705

nolanlawson:
Whoa, couch-master is actually working in Travis somehow. But unfortunately iPhone doesn't work, and i cannot repro. Trying to add some new cors headers.

nolanlawson:
Meh, I'm not going to try to debug Travis-only errors. It's not worth my time.

PR #3897, opened 2015-05-30
(#3897) - add pouchdb-express-router to allowed failures

nolanlawson:
All my recent attachment changes have been breaking
due to this thing, and TBH I don't see much reason to
maintain it anymore. Now that we have `minimumForPouchDB`
mode in express-pouchdb, that should be good enough, right?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -81,6 +81,7 @@ matrix:
 
   allow_failures:
   - env: GREP=test.replication.js CLIENT=node SERVER=sync-gateway BAIL=0 COMMAND=test
+  - env: CLIENT=node SERVER=pouchdb-express-router COMMAND=test
 
   fast_finish: true
 


daleharvey:
I wanna keep this maintained, its a few hundred lines of code, super simple and covers my use case 99% of the time, but completely fine with this being part of allowed_failures, its a fairly big barrier to contributors to have to make fixes in pouchdb-express-router to get stuff landed in pouchdb

daleharvey:
https://github.com/pouchdb/pouchdb/commit/1eb5558d90e551f35413c6ca45c1a985ca62cf8e

nolanlawson:
Appreciate it. FWIW I also fixed the bug in pouchdb-express-router (https://github.com/daleharvey/pouchdb-express-router/issues/10), but it's nice that I don't have to wait for an `npm publish`. (Also WTF is up with Travis' npm cache never updating; I had to manually change the dependency version in `package.json`.)

Issue #3898, opened 2015-05-30
Make auto_compation=true default?


daleharvey:
So users almost never care about having outdated revisions stored locally and similiarly dont want to deal with when to compact, When this was originally made it was planned to be the default, should we make the switch to having it enabled by default, turned off as an option?

nickcolley:
:+1: I think since it's important to keep used space to a minimum on user's devices this should be opt-out.

nolanlawson:
(Sorry, I commented in #3899, but then I realized it's more useful to comment here.)

I'm fairly -1 on changing the auto_compaction default to true. I think it will be a confusing change for users, even in a major version, with no benefit other than (unmeasured benefits to) reduced space, and at the expense of extra CPU cycles to delete stuff for every `bulkDocs()`. The other big cost here is the cognitive overhead for users who have to figure out why a default was switched from false to true.

In my mind, software should do the safest thing by default, and then allow users to opt-in to more dangerous options. To me, `auto_compaction` is a dangerous option, because it removes the possibility of recovering lost revisions (or of using a plugin like [pouchdb-full-sync](https://github.com/nolanlawson/pouchdb-full-sync)). Basically we would be punishing users who don't read the documentation thoroughly enough (by deleting revisions they didn't realize would be lost) instead of rewarding users who read the documentation (by giving them an option to eke out more storage gains).

One of the big benefits of PouchDB/CouchDB over something like Firebase is that we *don't* just delete stuff willy-nilly, and I love being able to tell people, "PouchDB is the database that cares about your data." In practice, I also see plenty of users adding `{auto_compaction: true}` anyway, which shows that they do read the documentation and are happy to opt-in.

daleharvey:
I very much agree that caring about users data is of the largest importance, but I dont think having storing every version of every document forever is a reasonable default. I think users that are using the archival features of pouchdb, maybe for a wiki or delta-pouch will need to know the implications of them doing so whereas the majority of users dont need to fetch outdated revs but are very worried about disk space limits and we should have defaults that make sense for the majority

ermouth:
@daleharvey –1. Autocompaction becomes serious perfomance problem, especially when you use tablets or mobiles. 

@nolanlawson CouchDB has setting `revs_limit` that limits depth of revision tree stored (not the number of revisions saved). If Pouch could limit number of stored doc revs in similar way...

nolanlawson:
[...]

PR #3899, opened 2015-05-30
WIP - (#3898) - Make auto_compaction default to true

daleharvey:
Seeing what tests may fail

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,9 +45,9 @@ env:
   - CLIENT=selenium:phantomjs ES5_SHIM=true COMMAND=test
 
   # Test auto-compaction in Node, Phantom, and Firefox
-  - AUTO_COMPACTION=true CLIENT=node COMMAND=test
-  - AUTO_COMPACTION=true CLIENT=selenium:firefox COMMAND=test
-  - AUTO_COMPACTION=true CLIENT=selenium:phantomjs ES5_SHIM=true COMMAND=test
+  - AUTO_COMPACTION=false CLIENT=node COMMAND=test
+  - AUTO_COMPACTION=false CLIENT=selenium:firefox COMMAND=test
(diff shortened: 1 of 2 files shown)


nolanlawson:
I'm fairly -1 on changing the auto_compaction default to true. Many tests will break, yes, because there are a lot of tests to ensure that old revisions are still there, but also I think it would just be a confusing change for users, even in a major version. And plus, plugins like [pouchdb-full-sync](https://github.com/nolanlawson/pouchdb-full-sync) show the value of the "keep everything forever" feature.

If you want to see which tests will break, though, just grep the tests for `auto_compaction: false`, because I explicitly have to do that, given that we are now testing auto-compaction in CI. There are also special tests for `auto_compaction: true`.

nolanlawson:
If you want to see which tests will break, though, just grep the tests for `auto_compaction: false`, because I explicitly have to do that, given that we are now testing auto-compaction in CI. There are also special tests for `auto_compaction: true`.

nolanlawson:
I would actually hold off on merging this immediately, due not only to the failures but also because I have @boennemann's "fear of incrementing the major version": https://twitter.com/janl/status/598840861015801856.

So far, there is nothing in master that would force us to increment the major version. I would like to focus on stability for the next few releases (especially due to recent calamities that make me wonder about PouchDB's robustness). Then when things settle down, I'd be happy to change `auto_compaction`, remove `PouchDB.destroy()`, remove the `updated`, `created`, `deleted` events, and all the other stuff we want to deprecate. Maybe we can make a branch that contains all these breaking changes, and then merge it in when we feel it's time for a 4.0?

daleharvey:
Ok, I want this to happen but lets leave it a short while so it doesnt overlap which possible important fixes

Issue #3900, opened 2015-05-31
Databases no longer seem to emit `destroy` events


nolanlawson:
Noticed this while looking at Istanbul coverage reports. May be a change since 3.5.0 (i.e. not affecting the last release).

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/26edf35a899946e3acf15ebf2b83d4f7f0fd9f80

PR #3901, opened 2015-05-31
(#3900) - add db.once('destroyed') tests

nolanlawson:
Well, I was surprised to see we weren't testing
this (except for PouchDB.defaults), so I added
some tests. They are passing, so maybe it's just
that the `'destroyed'` events aren't emitted
in certain tests (e.g. the changes/replicate tests).

### DIFF
--- a/tests/integration/test.events.js
+++ b/tests/integration/test.events.js
@@ -58,6 +58,27 @@ adapters.forEach(function (adapter) {
       });
     });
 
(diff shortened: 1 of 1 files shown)


daleharvey:
good catch - https://github.com/pouchdb/pouchdb/commit/2a307114643eb13c149f93b44f5a32bcd5955e9b

PR #3902, opened 2015-05-31
WIP - testing

daleharvey:


nolanlawson:
Got to 800 tests and then:

```
UnknownError - An unknown server-side error occurred while processing the command.","status":13,"cause":{"status":13,"sessionId":"44b7572386314789bef6de93c8181597","value":{"message":"The selendroid server on the device became unreachable.\nThis most likely means the app under test crashed or has been killed by the OS in a way that can't be detected using the default uncaught exception handler.\nTry to look for the reason of the
```

nolanlawson:
Maybe it's time to split our tests into two HTML files. I know that would be nice for Safari, so we could avoid the popup.

nolanlawson:
That was 5.1, btw. 4.4 failed much earlier.

daleharvey:
woot, green on 5.5, we may need to look into splitting the tests out, possibly just the replication ones

PR #3889, opened 2015-05-29
(#2858) - add "binary" option, use it in replication

nolanlawson:
Proposal: currently it is kinda awkward to get Blobs/Buffers
back from the API, because you can only use `getAttachment()`.
This is a shame, because some of our adapters can directly
return blobs (namely idb and http), meaning that we're doing a useless
base64-encoding and decoding operation during replication
of attachments.

I suggest adding a new option to `get()`: `{binary: true}`.
If included, and if `{attachments: true}`, then attachments
will be returned as Blobs/Buffers instead of base64. Also,
we can start using this option in the replicator.

In the future, I would also like to support `{binary: true}` in
`allDocs()`, `changes()`, and `query()`, but for now just
`get()` is enough for replication between http-idb to be
sped up.

I also put this commit on top of all the other ones to
avoid bitrotting myself.

Issue #2858, opened 2014-10-13
API feature request: getAttachment as URL string


ssured:
I use PouchDB to store images as attachments on documents. The regular use case is to let the browser render these images in a `<img>` tag by setting the `src` property.

There is no API method to easily figure out where to point the `src` to:
- for CouchDB you can construct the url manually: http://`server`/`dbname`/`docId`/`attachment`
- for PouchDB when getting a Blob and when there is support for http://caniuse.com/#feat=bloburls: `URL.createObjectURL`  https://developer.mozilla.org/en-US/docs/Web/API/URL.createObjectURL
- for PouchDB returning base64: use a data uri scheme, http://en.wikipedia.org/wiki/Data_URI_scheme

This can be implemented as an option in the current getAttachment code: `{asURL: true}`

Usage could look like:

```javascript
db.getAttachment('abc', 'image.png', {asURL:true}, function(err, url){
  $('img').attr('src', url);
});
```

nolanlawson:
As of 1f24d8b, it's now possible to directly get Blobs/Buffers from all read APIs. Since this is the most efficient format, and since `blob-util` can convert to all other formats such as data URIs and base 64, I'm going to consider this issue fixed. :)

ssured:
Cool stuff, nice addition and awesome you have an api for changing the default behaviour. This issue was about something a little different though. My request was for getting an url to an attachment. Say you store images as attachments, this url is what you put in the src of an img. The http adapter can, in most circumstances, give a direct url to the attachment on the server (when it is a stub), except for new unsaved attachments. Other adapters will always return blob-urls. I use this because my code does not know if it runs against a remote couchdb, or a local pouchdb (offline use case). Hope this explains my proposal a little more. Anyways, awesome work again, especially as this makes CSG compatibility better. Thank you so much!

nolanlawson:
I don't think we will add such a thing to core, but a plugin world be welcome. :)  `db.type() === 'http'` and `db.getUrl()` are the methods you will want to use. (Yes, undocumented, but pretty stable nonetheless.)

Issue #1586, opened 2014-03-06
We should only enable nonce if we are fairly sure we are in ie


daleharvey:
I hate browser sniffing, but adding nonce to everything is kinda annoying (and is causing bugs, although it shouldnt)

daleharvey:
Actually we do need this in Firefox - https://travis-ci.org/pouchdb/pouchdb/builds/72126685

nolanlawson:
Too bad - so sniffing didn't work?

daleharvey:
Nope, it started failing the tests hard in chrome and firefox in the same way the tests failed when we had the caching bugs, ie random tests failing with unexpected results from http

PR #3903, opened 2015-05-31
(#2858) - Re-integrate mapreduce back into PouchDB core

nolanlawson:
The first step to getting secondary indexes implemented is to move mapreduce back into PouchDB core.

To be honest, it never made much sense to me to have this as a separate plugin. We've always called it a "plugin," but in reality is has a lot of special access that goes above and beyond what a plugin normally uses (`dependentDbs`, `db.request()`, etc.).

Now, this doesn't preclude the possibility of us removing mapreduce at some point. As I work on this, I would like to keep the "indexy" part of mapreduce (i.e. similar to `pouchdb-abstract-mapreduce`) and leave that in, but remove the "mapreducy" part of mapreduce and put that in a separate module. So that way, `pouchdb-find` can just use the "indexy" stuff.

Then we can simply build two versions of PouchDB using normal browserify tricks. We can call the dist file `pouchdb.no-mapreduce.js`, or alternatively folks can use it from browserify with something like `var PouchDB = require('pouchdb/no-mapreduce').` No need for a separate repo or plugin.

Some interesting things about this commit:

* It only brings mapreduce back in. It does not implement native secondary indexes (yet).
* It is on top of the https://github.com/pouchdb/pouchdb/pull/3889 because I am lazy and wanted to solve the `binary: true` problem everywhere at once (i.e. in `query()` as well as `allDocs()`, `changes()`, and `get()`).
* We are finally testing mapreduce in Safari, IE, iPhone, Chrome, and PouchDB Server!
* The http temp views tests are disabled in CouchDB master because it's not supported.
* The local tests are disabled in Safari due to the dreaded popup, which comes up because we are creating lots of data with the secondary indexes. Maybe native indexes will fix this.
* The http tests in Safari actually managed to reproduce the mythical Safari POST bug 100% of the time (where Safari caches the POST request - I had reproduced it once, but then wasn't ever able to reproduce it again). So we actually need to modify `request-browser.js` to append a `_nonce` to POST requests as well as GET requests. Oh you silly Safari!


nolanlawson:
I'm also upgrading to the latest version of chai/chai-as-promised, because there were some differences with the version used by mapreduce. This required slightly rewriting some of the tests.

daleharvey:
Hey sorry, can you rebase since the binary change makes it hard to follow the changes here, cheers

daleharvey:
Yeh I think right now we have the worst of both worlds, mapreduce has internal apis, and it doesnt get the same integration testing per commit. I think the tests should be split into `/tests/mapreduce/` and we should still make an effort to reduce the internal api's (ie remove dependantdbs), but having it live in the same repo makes sense

nolanlawson:
yep, dependent dbs will be the first thing to go when I write the naive secondary indexes. As for tests/mapreduce, that makes sense.

nolanlawson:
I'm going to make it my goal to get this PR working by the end of the month. Technically it will be required for `binary` to be supported in `query()` anyway.

nolanlawson:
Moved the map/reduce tests to a separate command, only testing Node/PhantomJS/Firefox in Travis for now.

I think it will be a bigger push to get the tests passing in more than one environment. There are intermittents. Safari has more popups than when you win in Windows 95 Solitaire. PouchDB Server doesn't pass the tests (wtf?). Lots more to do here. Never a dull moment in PouchDB!

daleharvey:
Hrm, wondering if saucelabs just craps out here, the new map reduce tests are passing, all the fails are existing ones that should be green, all passing locally

daleharvey:
It seems like it is a saucelabs issue, seeing it on other pr's https://github.com/pouchdb/pouchdb/issues/3974

nolanlawson:
rebased

daleharvey:
hrm actually I was wrong, these are failing in saucelabs much more than master is, it must be the cache busting or the chai changes, looking into it now

daleharvey:
Given this is failing in saucelabs, I assume it must be the cache busting change that is breaking here, initial assumption would be that the cache busting was increasing the time taken for the tests and that would start causing timeouts, but some of these are failing in <20 minutes

daleharvey:
So https://travis-ci.org/pouchdb/pouchdb/builds/67346060 is with the cachebusting removed so it does look like that is causing a problem. I have a PR for the common failing test (https://github.com/pouchdb/pouchdb/pull/3984) so would like to see how this does rebased on that

nolanlawson:
Weird, why would adding a param to POSTs cause the tests to fail? Also irritating, because without it, Safari fails. :(

daleharvey:
I am not entirely sure, I thought it may be triggering extra cors preflights but for POST requests it really shouldnt be doing any extra work, but will see what its like based on the other fix

nolanlawson:
I would be fine with user agent-sniffing Safari and only applying POST cache buster in that case. Hell, we could start sniffing IE as well to do it for GETs.

daleharvey:
Heh yeh a fairly old and simple bug I hadnt got round to https://github.com/pouchdb/pouchdb/issues/1586

daleharvey:
Testing based on master - https://github.com/pouchdb/pouchdb/pull/3985

daleharvey:
So based on master this is passing all but the android tests which are also very flaky on master still, on top of https://github.com/pouchdb/pouchdb/pull/3988 I am hoping will be grand (expect I can get this green with another restart or 2, but would like to land after 3988)

daleharvey:
Woot green, will still need to improve the android tests but its unrelated

https://github.com/pouchdb/pouchdb/commit/3a8e0541af8a22cffd71fe9cda9800d3f1b583e4

nolanlawson:
Wow, I'm surprised we managed to get this in. Excited to start work on native indexes soon!

PR #3905, opened 2015-05-31
(#3904) - Add iojs to allowed failures

daleharvey:


nolanlawson:
Huh, that's odd. +1, but definitely worth looking into. Might have been something I changed recnetly.

nolanlawson:
SSL Error: https://registry.npmjs.org/debug does not support SSL

This is probably a Travis issue. It fails before it even `npm install`s.

daleharvey:
Yeh hoping this fixes itself

https://github.com/pouchdb/pouchdb/commit/6bd65402c1e0c434fb47cfbd8546fd94ea751df6

Issue #3908, opened 2015-06-01
Binary or base64 attachments?


nolanlawson:
It just occurred to me that #3804 and #3870 basically have two contradictory strategies for performance. One prefers fetching attachments as inlined base64-encoded strings, whereas the other prefers fetching them directly as binary in separate requests. (Not quiiiite directly, but once #3889 lands, then yes, directly as arraybuffers/blobs from XHR.)

I suppose the argument could be made that, the case of `_all_docs`, we are already fetching many documents in a single request, so the benefit of inlining attachments is high, whereas for `GET`, we are doing one-request-per-document anyway, so we might as well grab the attachments directly.

I admit I didn't do any performance comparison between the two strategies. I have a hunch that directly fetching attachments is better than inlining for both cases, but I have no data. Anyway, we would need to remove the inlining if we wanted to support CSG.

(Obviously the best strategy would be `multipart/mixed`, but I've determined we cannot support that in the browser because CouchDB sends back the data as `gzipped`, so we would lose any benefits by having to unzip stuff in JavaScript.)

@jo Do you have any thoughts on this?

jo:
Because the performance benefit of both strategies mainly depend on the size of the attachments (for many small attachments it might be the best to fetch them inline) I would suggest to do both ;)

Consider a photo app which provides small thumbnails as well as larger images. For the thumbnails it will be better to fetch them all together in one bulk request (if they're rev-1 or in a single doc), while doing binary xhrs for the large images would be definitely more performant.

The problem with this dual approach is, of course, to determine the attachment sizes in advance. For the bulk rev-1 fetch for example we do not know the size before before the response.
A simple configure option alone would not solve the abovementioned example where both small and large size attachments exist in a single database.

I don't have a solution yet but I think replication performance is important enought to go that somewhat difficult route or at least prepare for a hybrid approach.

nolanlawson:
Interesting points. I definitely agree that a hybrid approach would be best. I think, though, that it makes more sense to make this a global option  e.g. `prefer_inlining` for a few reasons:

1. If docs contain mixed large/small attachments, there's really no way to do two strategies for the two groups
2. It makes it easy to support CSG.
3. It's simpler. :) And for the thumbnail case you mention, users can simply have two separate DBs for the larger and smaller images, and use two strategies for each one.

Maybe this can be an option on the http adapter's constructor.

nolanlawson:
[...]

PR #3909, opened 2015-06-01
(#3908) - prefer non-inlined attachments for allDocs()

nolanlawson:
This should probably be configurable, because the strategy you want will depend on the size and number of your attachments.

But for the time being, we should at least be consistent about which one we're doing. Plus, this version supports CSG.

### DIFF
--- a/lib/replicate/get-docs.js
+++ b/lib/replicate/get-docs.js
@@ -59,16 +59,8 @@ function getDocs(src, diffs, state) {
     return Promise.all(diffKeys.map(processDiffDoc));
   }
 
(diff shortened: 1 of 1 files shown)


nolanlawson:
This basically reverts 8f512b9149310ba06174ae9aeaf996ab787b2559.

daleharvey:
Im good with being consistent for now, going to be working on standing up our performance tests for replication soon and we can back these changes with numbers 

https://github.com/pouchdb/pouchdb/commit/8207da70de80dad930cec6ae2e541f06c8391fac

PR #3917, opened 2015-06-02
(#3481) - Split up the android tests

daleharvey:


daleharvey:
Woot, green :) (will remember to remove from expected fails)

nickcolley:
This is awesome :D

nolanlawson:
What is `invert` and how is it working? It kinda looks like `invert` isn't doing anything, and the Android tests are just passing twice in a row... Yet I do see in Travis that 174 passed in one and 909 in the other, so I'm puzzled.

daleharvey:
Heh `INVERT` I found trying to debug why the regex I was using for grep wasnt working (https://github.com/mochajs/mocha/blob/master/mocha.js#L1641) It inverts the grep so GREP=test.replication.js will only do replication tests, GREP=test.replication.js INVERT=true will test everything but the replication ones

nolanlawson:
Oh, awesome! That is an amazingly helpful trick. +1 then.

daleharvey:
Yeh the nice thing about it is I was starting on setting up a `SUITE1` vs `SUITE2`, but its hard to guarantee that all the tests will run, with this we can just add to the `GREP=test.replication.js,test.compaction.js` if we need to split it out further and ever test will run

daleharvey:
Cheers, merged in https://github.com/pouchdb/pouchdb/commit/47944e2c507fc6904a565432e553ebc42e738cc1

bin/test-browser.js
```diff
@@ -49,6 +49,9 @@ if (client.runner === 'saucelabs') {
   qs.saucelabs = true;
 }
 if (process.env.GREP) {
```
fzzzy: Uhh, shouldn't this be INVERT?

nolanlawson: whoops, good catch!

daleharvey: heh copy and paste fail, luckily we only use INVERT when GREP is defined but good catch, thanks :)

PR #3919, opened 2015-06-02
(#XXXX) - Reduce setimeout time in tests

daleharvey:


daleharvey:
Just a minor thing I noticed, this runs like 8 times so reducing it speeds up the test suite by like a minute and 1 second should be enough to be proving any negatives, ideally we can get rid of the settimeout but this is a smaller change

nolanlawson:
Sure thing; the tests are indeed kinda slow. :+1:

daleharvey:
Cheers, landed in https://github.com/pouchdb/pouchdb/commit/a3f27979308c9ef0673f1f75484cbef00f5b0048

PR #3920, opened 2015-06-02
(#XXXX) - Test chrome stable

daleharvey:
I would like to reduce the amount of chrome tests we have running, we only need to support the latest stable of chrome (testing the beta version may be a sensible option too). I know we currently have older chromes as a proxy for older androids so not removing them yet, but hopefully we can get the rest of android tests working and explicity test the things we support and not test the things we dont

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,7 +52,7 @@ env:
   # Testing in saucelabs
   - CLIENT=saucelabs:chrome:36 COMMAND=test
   - CLIENT=saucelabs:chrome:37 COMMAND=test
-  - CLIENT=saucelabs:chrome:39 COMMAND=test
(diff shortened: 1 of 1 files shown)


nolanlawson:
Agree; even with Chrome 37/38/39, there are probably few of those floating around out there these days (citation needed). Even in Android-land, Lollipop should be keeping most folks up to date (launched with 37 but now up to 42), so the only "old" Androids we can test via a Chrome proxy would be 33 (Android 4.4.4) or 30 (Android 4.4.2).

:+1:

daleharvey:
Cheers, landed in https://github.com/pouchdb/pouchdb/commit/32ffc0677f222ea90a385603bd22d5ee494aa4b6

PR #3922, opened 2015-06-03
(#3922) - move more stuff out of utils

nolanlawson:
Feels cleaner when everything is separate files,
rather than having them all inside one big `utils.js`.


daleharvey:
Yeh I prefer this too, cheers

https://github.com/pouchdb/pouchdb/commit/1ed01f20307e3c594ddf241b348cc86a42ab0727

Issue #3923, opened 2015-06-03
Can't set default prefix for the http adapter


slaskis:
I've tried setting up a few pouch "defaults" in my node app according to [this method](http://pouchdb.com/api.html#defaults) described in the docs:

```
exports.remote = PouchDB.defaults({
  adapter: 'http',
  prefix: exports.prefix
});

exports.admin = PouchDB.defaults({
  adapter: 'http',
  prefix: exports.prefix,
  auth: exports.adminAuth
});
```

But it seems like it ignores the prefix entirely. First it didn't automatically use the http adapter (which is why I added it). Then it still just tries to connect to "\_pouch_{name}".

Looking at `lib/setup.js` it seems like the prefix is a global variable set on the PouchDB constructor, and not an option which can be overridden, or am I reading it wrong?

Besides this, thanks for a wonderful library!

nolanlawson:
I believe `prefix` was never designed to be used with the http adapter. How are you trying to use it?

slaskis:
Oh, well I understood this part of the docs like it was pretty much meant for the http adapter:

> Note the special constructor option prefix, which appends a prefix to the database name and can be helpful for URL-based or file-based LevelDOWN path names.

I'm using it like above. I'm trying to DRY up my PouchDB instance options a bit with defaults. So then I can just do this in the rest of my code:

```
var PouchDB = require('./db').admin;

var db = new PouchDB();
// do things that require admin auth here...
```

nolanlawson:
So if you have the prefix `'foo_'` and you create `new MyCustomPouchDB('http://mysite.com:5984/mydb')`, do you expect it to be transformed into `new PouchDB('http://mysite.com:5984/foo_mydb')`? Just checking to make sure I understand.

slaskis:
Sorry for being a bit vague, trying again :)

I have set the default prefix using something like:

```
var AdminPouch = PouchDB.defaults({
  prefix: "http://couchhost.com/", 
  auth: {...}
});
```

and then create using 

```
var db1 = new AdminPouch('mydb1');
var db2 = new AdminPouch('mydb2');
``` 

which I then expect it to be transformed to

```
var db1 = new PouchDB('http://couchhost.com/mydb1', {auth: {...}});
var db2 = new PouchDB('http://couchhost.com/mydb2', {auth: {...}});
```

In my case it's because I use multiple dbs on the same host and this seemed like a clean way to only define that host once (as well as other things like auth).

[...]

PR #3924, opened 2015-06-03
(#3923) Use opts.prefix if it exists

slaskis:
This is a quick workaround for prefix defaults not being respected when using the http adapter.

### DIFF
--- a/lib/setup.js
+++ b/lib/setup.js
@@ -70,7 +70,7 @@ PouchDB.parseAdapter = function (name, opts) {
       adapter.use_prefix : true;
 
   return {
-    name: usePrefix ? (PouchDB.prefix + name) : name,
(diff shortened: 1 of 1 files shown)


slaskis:
This PR still does not fix the adapter discovery. In other words in node at least it will use the leveldb adapter even though it's prefixed with a URL unless `adapter: 'http'` option is set.

nolanlawson:
Could you please add a new test to `tests/integration/test.defaults.js`? Also the test failures in Travis seems genuine; this seems to break for local LevelDB-based databases (i.e. one of the tests in `test.defaults.js`).

slaskis:
Yikes, yeah I'll just close this if that's not the intended functionality of the prefix option.

PR #3926, opened 2015-06-04
Confirm to plugin authors that bulkDocs() is the update workhorse; closes #3916

### LINKED ISSUE #3916: Relationship between .put(), .post() and .bulkDocs
Hi! Are all update methods (.put(), .post(), .remove(), etc.) guaranteed to execute bulkDocs() ultimately, or is that an implementation detail, subject to change? If the latter, then may I suggest documentation or advice to plugin authors about how to intercept updates.

(I'm happy to submit that PR once I understand the answer.)

## Longer version

Currently, put(), post(), and other methods are implemented internally with calls to bulkDocs  (Only one document is in the batch.)

* post: https://github.com/pouchdb/pouchdb/blob/3.6.0/lib/adapter.js#L170
* put: https://github.com/pouchdb/pouchdb/blob/3.6.0/lib/adapter.js#L212
* remove: https://github.com/pouchdb/pouchdb/blob/3.6.0/lib/adapter.js#L311
* removeAttachment: https://github.com/pouchdb/pouchdb/blob/3.6.0/lib/adapter.js#L276
* replicator: https://github.com/pouchdb/pouchdb/blob/3.6.0/lib/replicate/replicate.js#L61

I wrote a plugin to intercept updates, for validation. I was shocked, shocked! to read #1412. No validate_doc_update? What? What! Have you all lost your minds? But then after some work and consideration, I got religion. That is a fantastic judgment call. Plugin-based validation is both simpler and more powerful. Bravo!

Unfortunately, now I need to be sure to validate all updates. So, that's my question. I works now with a single bulkDocs interception. However, if that changes in the future (say, a 3.x release), then I expose my database to invalid data.

I could intercept every possible method (put, post, removeAttachment), at a cost of much more complexity to my plugin (and potentially double-execution of the validator for those methods which call into bulkDocs).

In my opinion, PouchDB should eschew purity and acknowledge what is happening: bulkDocs is the workhorse of PouchDB. It will remain the workhorse for, say, the 3.x series. Look here! A custom handler to deal with this very fact:

https://github.com/pouchdb/pouchdb/blob/3.6.0/lib/adapter.js#L25-L34

Thus, documentation for plugin authors should say as much: "To intercept all updates, simply implement your own bulkDocs() method." This would also tie up the loose end in #1412, where people hope for some sort of pre-commit hook. bulkDocs() *is* that pre-commit hook.

Also, to bolster my argument, I would observe that the LevelDB community has had great success with this trick. For example, level-sublevel (a package used by PouchDB) does exactly this to simulate namespaces in a Level DB. level-sublevel just monkey-patches the .put() and .batch() methods. It works great.

Thank you very much, if you've read this far down!

jhs:
My first crack at documenting the relationship with bulkDocs() and other update functions. Perhaps it is too verbose?

There is not really a dedicated place yet for plugin authors, so I sort of shoehorned in a discussion and simple example.

### DIFF
--- a/docs/_includes/api/plugins.html
+++ b/docs/_includes/api/plugins.html
@@ -21,3 +21,45 @@ PouchDB.plugin({
 });
 new PouchDB('foobar').sayHello(); // prints "Hello!"
 {% endhighlight %}
+
+#### Example Plugin: Intercept Updates
+
+A useful feature of plugins is to intercept updates before they are stored in PouchDB. In this way, a plugin might validate that the data is correct for the application, or even alter documents before they are committed to the database.
+
+The best way to intercept all updates to a PouchDB database is to **override the `bulkDocs()` method**. All changes to PouchDB documents ultimately pass through the `bulkDocs()` method. For example, a call to `put()` will become become a `bulkDocs()` call with a "batch" of one document.
+
+Because PouchDB guarantees to plugin authors that all data changes ultimate happen via `bulkDocs()`, it is the ideal place for an application or plugin to intercept updates.
+
+{% highlight js %}
+// Keep a reference to the "upstream" function.
+var pouchBulkDocs = PouchDB.prototype.bulkDocs;
+PouchDB.plugin({bulkDocs: validBulkDocs});
+
+function validBulkDocs(body, options, callback) {
+  if (typeof options == 'function') {
+    callback = options
+    options = {}
+  }
+
+  if (Array.isArray(body)) {
+    var docs = body;
+  } else {
+    var docs = body.docs;
+  }
+
+  // All documents must have a .name field.
+  for (var i = 0; i < docs.length; i++) {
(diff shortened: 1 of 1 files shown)


jhs:
re #3916

jhs:
1. I don't see how I could have broken the build. Maybe it's touchy these days?
2. On the other hand, I am not familiar with the site building process, so this PR is *not tested*. I just wanted something out there to begin the discussion at the least.

nolanlawson:
Yes, there are lots of intermittents in our tests sadly. You can add `[skip ci]` anywhere in the commit message and Travis will ignore it. Handy for documentation changes. :)

nolanlawson:
LGTM, love having detailed documentation like this.

nolanlawson:
Thanks! 1a047346ddf361600b094a1a5699a77e2eb1eef2

Issue #3932, opened 2015-06-06
3.3–3.6 versions can loose attachments data


ermouth:
Looks like data loss takes place in this scenario (Chrome 43, Mac):
* create a doc with attachments in 3.2.1
* upgrade to 3.3.1
* read the doc
* put exactly same attachment(s) in the doc as `{content_type:"whatever", data:"base64 string goes here"}`
* add some new attaches
* save the doc 
* from now replication is broken – re-saved attaches has no data / revpos and newly saved have no revpos
* upgrade to 3.6.0
* read the doc with `{attachments:true}`
* save it to another DB

Now *all* attaches of the clone are lost.

Interesting, that MD5 checksums are preserved – they all are different – but `data` returned are empty strings.

Also I experienced accidental attachments loss doing nothing with docs – just upgrading from 3.3 to 3.6. Exact cases are not yet clear.

nolanlawson:
Hm, that's weird: there was no IndexedDB migration between 3.2.1 and 3.3.1.

Have you tried upgrading directly from 3.2.1 to 3.6.0?

nolanlawson:
Also a reproducible test case would be very helpful.

ermouth:
All I can provide right now is remnants, they look like this:

```json
  "css/ptm55f-webfont.ttf": {
   "digest": "md5-SbEzACAvJcCFxzHSFvq0IQ==",
   "content_type": "application/octet-stream",
   "data": ""
  },
  "css/ptm75f-webfont.ttf": {
   "digest": "md5-ICZexyAYJD7Ox4eZxbWdOQ==",
   "content_type": "application/octet-stream",
   "data": ""
  },
  "i/tip-apps.png": {
   "digest": "md5-wtAFDfJRZZCqN3IYGyx4QQ==",
   "content_type": "image/png",
   "data": ""
  },
  "i/tip-db.png": {
   "digest": "md5-clHzg3xV1VRVkSKd4I66Bw==",
   "content_type": "image/png",
   "data": ""
  },
  "i/tip-profile.png": {
   "digest": "md5-V24SYJpSeIkEiYBteC8TuQ==",
   "content_type": "image/png",
   "data": ""
  },
  "i/tip-start.png": {
   "digest": "md5-s/bZ3Or71kFT1zCycfJIIw==",
   "content_type": "image/png",
   "data": ""
  },
  "i/tip-sync.png": {
   "digest": "md5-TGHl6/ss0gx8fhygUOFLmQ==",
[...]

PR #3935, opened 2015-06-06
(#3932) - another attachment replication test

nolanlawson:
Another test that doesn't reproduce #3932. But
it's always good to have more tests.


nolanlawson:
BTW yes, I am running these tests against the old 3.6.0 commit. It succeeds back then too.

daleharvey:
:+1: merged in https://github.com/pouchdb/pouchdb/commit/ddb4d1079291fb62379009449d33d1156c331dba, cheers

Issue #3839, opened 2015-05-13
Promises and node v0.12


mainakae:
Hello all,

I've noticed that after upgrading my node, form 0.10.x to 0.12.2, I'm no longer able to call the "finally" method on promises returned by the api; now it throws "TypeError: undefined is not a function".

var db = new PouchDB('http://localhost:5984/db");

db.allDocs().then(...).finally(...); // NO LONGER WORKING!

daleharvey:
Hrm I dont think `Promise.finally` is in the official spec? if so it shouldnt have ever been part of our API apologies

calvinmetcalf:
right and you were never able to do this in the browser, it's because we
only use bluebird for promises if native ones are not available

daleharvey:
We should either 1. Figure out how to remove `finally` from bluebird promises, 2. Use a spec compliant lib, 3. Not support node versions that dont support native promises.

It looks like all stable releases of node support Promises natively but not sure what the uptake for these releases are, may be harsh to drop support already

calvinmetcalf:
you can just switch to using lie only and not bluebird

mainakae:
I know it's not part of the specification, I've been trying out lately lie, q and bluebird among others, and though it's somehow arguable and prone to change in the future, seems that bluebird is the right choice. In fact I adopted it as the promises library for my latest app (in part in sake of homogeneity with PouchDB).

The issue is not so much about loosing that functionality (again: as things go by, you might be inclined to chose another lib, and then following tightly the specification will be a bonus) but about some shaky things going on under the hood, maybe worth checking.

daleharvey:
In node 0.12 Promises are natively supported, so it will start to use the native promises which dont have `.finally`, similiar to what happens in the browser. We have always planned to be spec compatible so we can remove any of our own Promises code when its widely enough supported. 

Removing bluebird and just using lie for older node versions sounds like a good plan

nolanlawson:
I'm fine with removing bluebird and using `global.Promise || require('lie')`. Simplifies things a lot, although technically may justify a major version release.

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/09ce6c885dece08cc722d0897a546d694759eaee

PR #3941, opened 2015-06-07
(#3839) - Use lie as promise polyfill everywhere

daleharvey:


nolanlawson:
If we're going to do this, then we should also 1) update the documentation that mentions bluebird, and 2) do a similar PR on mapreduce.

daleharvey:
ok, will take a shot at helping out with https://github.com/pouchdb/pouchdb/pull/3903 so its easier to switch mapreduce at the same time

nolanlawson:
I'm also on the fence about whether this should be a breaking change or not. We did document it, so there might be codebases out there that only run in Node and that depend on the Bluebird-specific APIs.

daleharvey:
Yeh we have only ever documented that we use bluebird and never suggested or documented using its extended functionality plus that would have never works outside node, I would vote to say its an bugfix but not fussy

nolanlawson:
After some thought, I am -1 on this change, because it doesn't simplify our code much, it might break people's code if they were depending on our (documented!) use of bluebird in Node, and it's just one more thing we have to remember moving forward. Plus bluebird has crazy good debug tools in Node (e.g. logging an uncaught error as a warning), so that may help people out when they are trying to debug PouchDB. (They get that in the browser as well, assuming they are using Chrome, since we default to Chrome native Promises.)

I would also like some input from @calvinmetcalf on this. He has a library called [pouchdb-promise](https://www.npmjs.com/package/pouchdb-promise) which is based on the assumption that we want bluebird in Node and lie everywhere else (the same assumption that a lot of plugins make).

daleharvey:
We use native promises by default everywhere and in future we will remove both `lie` and `bluebird` to just use native Promises, the issue is by using bluebird as a polyfill now we are exposing non standard promises functionality that will work in old versions of node and break in the browser and newer versions of node which is fairly confusing (https://github.com/pouchdb/pouchdb/issues/3839)

Users still have the option to override native promises and use bluebird in node when they want to but I think this gets us one step closer to using no polyfills which is the ultimate goal, so still think its worth doing so more users dont start using `finally` etc in their code and get burned when upgrading node

nolanlawson:
You make good points. I tend to be really conservative about stuff like this, though, since it could potentially break user code. This definitely needs to be a major version change IMO.

I also think @calvinmetcalf should either deprecate or change `pouchdb-promise` if we're going to do this, and the hoodies should be notified since they are probably using the `extras/promise` API. cc @janl @zoepage @gr2m.

gr2m:
I don't know all details about the upcoming Hoodie server side implementation, but so far we only use the PouchDB Promises in the front-end with browserify. But thanks a lot fro keeping us in the loop

daleharvey:
ok cool I can live with a major version change, will rerun to make sure nothing has broken since the last test run and get it landed

daleharvey:
ok, next release is a major bump :) Cheers

https://github.com/pouchdb/pouchdb/commit/09ce6c885dece08cc722d0897a546d694759eaee

nickcolley:
On top of this: Should we allow users to be able to swap in their own promise implementation? Just thinking if people are depending on bluebird's extra features I'm not sure how they'd easily migrate their code?

daleharvey:
Yup, people can easily override global.Promise and use whichever implementation they like

janl:
just to wrap this up: Backend Hoodie doesn’t care either way yet, so all good :)

PR #3943, opened 2015-06-09
Return PouchDB when loading a plugin

jhs:
This way, plugins can be applied at require() time and chained:

    var PouchDB = require('pouchdb').plugin({/*...*/});

This is a one-line change. I do not see any unit tests that explicitly test the plugin API. So I did not add any testing. The only plugin stuff I see exercised is `tests/integration/test.basics.js` in the "putting is override-able" test.


nolanlawson:
I like it; it's a nice change!

nolanlawson:
22466f422c452def2911189b18d3876129dcf9a3

Issue #3946, opened 2015-06-09
getAttachement doesn't return the type


natcohen:
getAttachment doesn't return the type in the blob response when getting an attachment (I tried application/pdf).

version 3.6.0
Browser: Chrome

natcohen:
It seems that it's not the getAttachment function but rather the update of the browser DB... when I updated to 3.6.0, the type disappeared. When I tried to rollback to 3.5.0, the DB still didn't have the types. I had to delete the database and recreate it under the 3.5.0 version.

nolanlawson:
Can you provide a test to reproduce please? Also I'm not sure what you mean by "doesn't return the type." You mean `blob.type` is not correct, or `attachment.content_type` is not correct...?

nolanlawson:
To be fair, it just occurred to me that we don't actually test the blob content types anywhere, since we don't get them in Node (only in the browser). I think it would be reasonable to add these tests to the current test suite, maybe hidden behind a flag like `if (inBrowser) {`.

natcohen:
The blob.type is empty...

Put a pdf file in your couchdb, sync pouchdb and get the file with getAttachment. The blob response won't have type. It broke from 3.5.0 to 3.6.0

nolanlawson:
Sorry, I cannot reproduce. Please see https://github.com/pouchdb/pouchdb/pull/3948

Can you please offer a live test case? An `index.html` file using a local CouchDB at `localhost:5984` would be fine.

natcohen:
Well, I simply use the following to get the file:
    
    var pouch = new PouchDB('transport', {size: 50});
    pouch.replicate.to('http://127.0.0.1:5984/transport', {live: true});
    pouch.replicate.from('http://127.0.0.1:5984/transport', {live: true});
    
    pouch.getAttachment(docid, name, function(err, res){
        console.log(res);
    })

It returns a blob with correct data and size but with type = ""

nolanlawson:
@natcohen Is there any way you could provide a live test case? I don't have access to your CouchDB, so I cannot test your particular case. :)

nolanlawson:
Please provide more details if this is an issue. Also feel free to try the nightly. :)

natcohen:
@nolanlawson 
Sorry, I was away for few days... 

I made a mistake, it doesn't occur when I add the pdf through CouchDB but when I generate it with PDFkit and add it through PouchDB. Here is the procedure to reproduce it:
[...]

Issue #3758, opened 2015-04-22
Node Buffers should return with content_type


nolanlawson:
`getAttachment()` in Node.js doesn't return the Buffer's content-type, whereas in the browser it does. This leads to awkward situations in express-pouchdb and socket-pouch where we need to do both a `get()` and a `getAttachment()` in order to grab both the buffer and the content-type ([see express-pouchdb code](https://github.com/pouchdb/express-pouchdb/blob/master/lib/routes/attachments.js#L47-L62)). This is also risky, because it's possible (although unlikely) that the database could change in the interim between the two requests.

What's a good way to do this? Can we just attach a `type` to the `Buffer `object? Or should we introduce a breaking change where Node returns an object like:

```js
{
  buffer: new Buffer(...),
  type: 'image/png'
}
```

@calvinmetcalf thoughts on this?

ermouth:
If you decide to return an object, I think you can add an option, that defines a format of returned data.

Something like `db.getAttachment("docid", "file.name", {format:"binary"})` – or `base64`, or `url`. Latter is objectURL, for browsers only.

Also I think to avoid breaking changes this method can have different name – something short, may be `db.file()`?

nolanlawson:
There's already an open issue about that (#2858). I would be hesitant to add such a feature, though, because Blobs are the most efficient method, and if you want something else, you can always use `blob-util` to transform into base64 or binary strings or object URLs or data URLs or whatever you need.

ermouth:
When I use Pouch in browser it‘s ok more or less. But if you have efficient converters right inside Pouch – why not use them as `.getAttachment` postprocessors to make end user life more sugary?

About node. When I use Pouch server-side, I need streams – and sometimes badly. So I have systems where Pouch co-exist with nano – because of streams. It‘s bit weird, dont‘t you think?

nolanlawson:
> postprocessors to make end user life more sugary?

Hmm, but since this could also be done in a plugin, I think it's safer to say it doesn't belong in core. Also reduces code bloat. :smiley:

Plus `blob-util` is already pretty sugary; I mean, it's hard to beat `imgSrcToBase64String()`.

> When I use Pouch server-side, I need streams – and sometimes badly.

Agreed, `getAttachment()` definitely needs a streaming API for Node.js. Maybe when the W3C finalizes their streams spec, we can do it in the browser too.

ermouth:
[...]

PR #3948, opened 2015-06-10
(#3946 #3758) - add buffer.type, test types

nolanlawson:
Two goals here:

1. Test `blob.type`, which we actually weren't testing before
   (#3946).
2. Add an ad-hoc `type` to buffers, for consistency across
   Node and the browser (#3758).


nolanlawson:
Everything passing except for Couch master.

daleharvey:
:+1: but would prefer to land on green (ie when couch master tests are green or allowed fails)

nolanlawson:
Seems after 3 or 4 retries, I managed to make it green. Maybe the Couch master failures are intermittents after all?

nolanlawson:
Green, so merging. I feel unsure about actually documenting this `buffer.type` behavior, since it's kinda non-standard and weird, and I'm gonna wait for @calvinmetcalf to tell me if I'm crazy and going against the grain of Node or not. But if he thinks it's okay, then we can start using it to avoid an extra method call in `express-pouchdb` (because there's no need to do a `get()` in addition to a `getAttachment()` anymore), so @marten-de-vries you may be interested in this. Else we should just modify the tests to do an `if (isBrowser)` check.

nolanlawson:
2bd898d07fd23a21f000af713e0dd63983f1b4a8

Issue #3942, opened 2015-06-08
Corrupt indexeddb data can replicate back to server


nolanlawson:
I'm opening a new issue based on https://github.com/pouchdb/pouchdb/issues/3932. The old issue devolved into a lot of personal attacks and comments on those personal attacks, so it's not helpful to continue the conversation over there.

I suspect that by not using `revpos` correctly, our replicator somehow gets CouchDB into a state where it starts replicating documents without their attachments. I have [a test case](https://gist.github.com/nolanlawson/f0dd3e35ba908df691bb) that shows PouchDB 3.6.0 saving the attachments with `revpos: undefined` and `revpos: 0`, but I cannot reproduce the case where the attachment data is actually lost.

@kxepal @rnewson This is one of the most serious bugs PouchDB has faced, but I'm kind of at a loss to figure out how to attack it. Is there any way you can think of that a malignant replicator could misuse `revpos` in order to convince CouchDB to just completely drop attachments? According to the reporter of #3932, the attachments started looking like:

```
"css/ptm55f-webfont.ttf": {
   "digest": "md5-SbEzACAvJcCFxzHSFvq0IQ==",
   "content_type": "application/octet-stream",
   "data": ""
  }
```

But I cannot reproduce.

nolanlawson:
OK, here's [a slightly different attempt](https://gist.github.com/nolanlawson/a980820bd7e75d1fcf2b). In this case, I can get CouchDB to save an attachment, and it ends up with `revpos: 0`, even though the `1-` revision *does not actually contain the attachment*. (It's contained in `2-`.) Is it possible that this could cause the attachments to get lost somehow?

nolanlawson:
Hm, even if I use CouchDB's replicator to replicate that document to another database, the attachment is still there.

nolanlawson:
Ditto for compaction. The attachment is still there, even though `revpos` is wrong.

kxepal:
There is no magic behind revpos: if you remember `_revisions` field of documents (`GET /db/doc?revs=true`) then it's the value for `start` field inside that object.

Undefined or `0` revpos are also [fine case](https://github.com/apache/couchdb-couch/blob/master/src/couch_att.erl#L385-L395) - it's not an issue.

rnewson:
`revpos` goes hand in hand with `atts_since`. The goal being to optimize replication by not having to transfer the bytes of attachments already present on the target.

[...]

PR #3957, opened 2015-06-12
(#3942) - add migration test to try to capture error

nolanlawson:
I am trying my damndest to repro #3942, but I'm still coming up short.

This test creates a PouchDB using version 3.5.0, partially replicates
to a remote server, then reopens the database using the latest
version and re-replicates. I still cannot reproduce the data loss.
The tests pass just fine.


nolanlawson:
I don't actually want to merge this

Issue #3962, opened 2015-06-14
Docs wtih many large attaches fail to replicate from remote in 3.6.0


ermouth:
Docs that have lots of large attaches fail to replicate from remote, since browser limits number of simultaneous connections.
See screenshot here http://jquerymy.com/pouchdb/shot-361-1.png

nolanlawson:
Yeah, this is a dumb mistake on my part due to the simultaneous fetches using `Promise.all` in `http/index.js`. I should limit those to 6 at a time to avoid browser limits.

Alternatively, this would be a good opportunity to introduce the feature to allow inlining (https://github.com/pouchdb/pouchdb/issues/3913), e.g. for people who have documents with many small attachments.

I still don't see how this explains #3961, though, since if a single request fails, then the entire `db.get()` should fail and thus replication should throw an error instead of inserting empty attachments.

ermouth:
It really does not explain #3961 in any way, #3961 root post was induced by my stupid automated test that did not check if doc is present before enumerating attaches from all rows of `.allDocs`request . There were no docs – and surely no attaches ) Thats why I added UPD to disregard.

nolanlawson:
UPD?

daleharvey:
update I think

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/6a2955c8a4036cce24e4475e1e52bfc897fce228

Issue #3961, opened 2015-06-13
Attachments loss


ermouth:
On https://github.com/pouchdb/pouchdb/issues/3932, https://github.com/pouchdb/pouchdb/issues/3942.

__UPD. Disregard the post__, both fix and reason seem to be invalid :( More info in comments.

Seems that issue takes place in, at least, 3.2.1 and 3.3.1 and, possibly, 3.6.0 (did not test) because:
* timeout in ajax module is too short (1e4 hardcoded, smth around 1e6 look more reasonable)
* remote DB can have several docs with large atts (tens of megabytes) and fetching them induces timeout
* this case is somehow incorrectly handled
* doc is saved with original md5, but zero-length attachment(s).

If all above is correct it seems that Pouch never checks if attachment received from remote source is consistent to its digest.

As for 3.2.1 there was enough to increase timeout. For general fix I think it‘s reasonable to re-hash received data and revisit timeout handler.

Also this case shows why browser tests with CouchDB on `localhost` better be avoided. Virtual loop in general has much wider bandwidth capacity than any real end-user network.

I can reproduce it in all browsers with remote DB that contains several docs with large attachments. Please let me know if you also can. No code example since case is very specific to network conditions and DB, sorry.

Seems that https://github.com/pouchdb/pouchdb/issues/3932 could not be reproduced, even repeating steps provided exactly, because my remote DB had several large docs and my network is nearly always very busy.

To note – general practice to avoid this kind of issues is to have a test for each numeric constant inside the code, as well as for assumption under each particular constant.

ermouth:
In 3.6.0 situation seem to be more complex, still investigating.

ermouth:
All below is for Chrome.

I went down to IDB level and what I see is that Blobs can be fetched well with direct transactions, they have correct type, correct size, but I can not read these Blobs using FileReader. Any available method (readAsArrayBuffer, asBinaryString, asDataUrl, asText) – all fail with error "NotFound".

Code, that fetches attach, looks like this (think it‘s correct):

``` javascript
var res, req = indexedDB.open('_pouch_dbname');
req.onsuccess = function(e) {
  var db = e.target.result;
  db.transaction(["attach-store"], "readonly")
  .objectStore("attach-store")
  .get("md5-SBwxEsOQ1mwFXc3DLe4YSQ==")
  .onsuccess = function(e) {
  	console.log(res = e.target.result)	
  }
};
```
`res.blob` now must be valid Blob.

[...]

Issue #3913, opened 2015-06-01
inline_attachments option for attachments in the http adapter


nolanlawson:
Proposal (per me and @jo in https://github.com/pouchdb/pouchdb/issues/3911):

``` js
new PouchDB('http://example.com:5984/foo', {inline_attachments: true});
```

`inline_attachments` defaults to false. When false, the adapter will prefer to fetch attachments in separate requests, which can be more efficient for larger images because we avoid base64. Also, it supports Couchbase Sync Gateway properly.

When `inline_attachments` is true, then the adapter will prefer to use the `attachments=true` option and fetch the attachments as base64 strings, which can be useful for many small attachments (such as thumbnails).

nolanlawson:
I would also be partial to calling the option `inline`, just to avoid a camelCase vs snake_case dilemma. :)

jo:
I'm also in favour of `inline` :+1:

ermouth:
To notice: inlining can significantly reduce replication costs when using Cloudant or Smileupps. They both charge for number of requests, not for amounts of data transfer.

jo:
Good point

daleharvey:
From @ermouth 
> 
> 
> Since it‘s not an issue, but a question, I‘ve discarded a template, sorry.
> 
> Upgrading from 3.3 to 5.4, I‘ve discovered, that now Pouch fetches attaches using separate requests. It‘s nice, since it reduces possibility of timeout on docs with lot of large attaches, also it reduces traffic.
> 
> Although, since Pouch does not check integrity of an attach received, any unreported error during transmission can have destructive effect.
> 
> If we, for example, have misconfigured proxy or rewrite in front of Couch, direct request to an attach may cause any answer but attach itself. This response will be stored in Pouch and then replicated back to Couch (if user changes doc in Pouch, surely).
> 
> Initial attach is totally destroyed in this case. Example isn‘t taken from thin air – I‘ve discovered it testing real configuration with rewrites. Also I know proxies, configured to add cache-control header to all images, and it may also break intended behavior.
> 
> So is there any way to fetch docs during sync in prev versions style? I mean an option or something to enforce Pouch 5.4+ make only application/json requests and fetch docs with attaches inlined?
>

daleharvey:
Good point about the rewrites, I wonder if there is something we can do to prevent that, it isnt obvious off the top of my head.

However yeh having an option to use inline attachments is something we should probably support, it should be a fairly minor change

[...]

PR #3964, opened 2015-06-14
(#3962) - stop fetching attachments in separate requests

nolanlawson:
This essentially reverts 6e4aca8802c377aa93ccb95a13874338f43a8cb8 and 8207da70de80dad930cec6ae2e541f06c8391fac.

This was a dumb mistake on my part, to try to eke out additional performance improvements and improve CSG support by removing base64 inlining of attachments and fetching the attachments in a separate request instead.

CSG needs to add `application/json` support if they want to be able to replicate attachments to Pouch. I'm tired of implementing crazy workarounds.

As for Couch, it seems like `multipart/mixed` is never going to be a viable solution for Pouch, because 1) since they send the attachments gzipped, we would have to manually gunzip using some slow JS library, and 2) even if they made gzipping optional, then we're sending attachments unzipped, which might mean a larger payload than gzipped json (citation needed).

If we want to bring this feature back, then as @ermouth says, we would probably need tests that better simulate the browser dropping many concurrent connections.

### DIFF
--- a/lib/adapters/http/index.js
+++ b/lib/adapters/http/index.js
@@ -36,8 +36,8 @@ function blobToBase64(blobOrBuffer) {
   });
 }
 
-function readAttachmentsAsBlobOrBuffer(row) {
-  var atts = row.doc && row.doc._attachments;
+function readDocAttachmentsAsBlobOrBuffer(doc) {
+  var atts = doc._attachments;
   if (!atts) {
     return;
   }
@@ -45,6 +45,13 @@ function readAttachmentsAsBlobOrBuffer(row) {
     var att = atts[filename];
     att.data = b64StringToBluffer(att.data, att.content_type);
   });
+}
+
+function readRowAttachmentsAsBlobOrBuffer(row) {
+  if (!row.doc) {
+    return;
+  }
+  return readDocAttachmentsAsBlobOrBuffer(row.doc);
 }
 
 function encodeDocId(id) {
@@ -361,6 +368,11 @@ function HttpPouch(opts, callback) {
       params.push('conflicts=' + opts.conflicts);
     }
 
+    // yay attachments
+    if (opts.attachments) {
+      params.push('attachments=true');
+    }
+
     // Format the list of parameters into a valid URI query string
     params = params.join('&');
     params = params === '' ? '' : '?' + params;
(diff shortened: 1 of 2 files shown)


nolanlawson:
This means we might as well also un-revert @jo's commit: https://github.com/pouchdb/pouchdb/commit/8f512b9149310ba06174ae9aeaf996ab787b2559

I'm willing to eat humble pie and say I was wrong about this. Trying to do many concurrent requests in a browser is a recipe for disaster.

daleharvey:
So I think ultimately we do want attachments in seperate requests, but this PR is probably needed however I do not want to revert it until we can reproduce the dataloss bug, I am certain this will fix it however somewhere we are writing empty data and I want to know why. The general process I think is something like:

   1. Possibly revert to 3.5.1 on the website to make sure nobody else runs into this
   2. Reproduce the data loss and find the root cause
   3. Do this revert and fix it
   4. Implement seperate attachment fetching better

Dont take the blame I reviewed the code as well and merged it even with having hesitations.

nolanlawson:
Fair enough. Until we have a reproducible test, I am indeed just taking shots in the dark.

I'd be fine with rolling back to 3.5.0 and unpublishing 3.6.0.

nolanlawson:
Well this is green, but I guess we would need @ermouth to test and see if it fixes the issue.

daleharvey:
:+1:  when green

nolanlawson:
Well, do we still want this, given that the zero-length issue turned out to be something else?

If not, we can still go for the `inline` option as planned in https://github.com/pouchdb/pouchdb/issues/3913.

jo:
Where is @ermouth's issue described?

daleharvey:
His issue is in https://github.com/pouchdb/pouchdb/issues/3961

We still need to fix the 'ton of parallel requests' problem that the current implementation of attachment fetching has, ill take a look to see if I can get a variation of https://github.com/pouchdb/pouchdb/pull/3966 to fail

nolanlawson:
Done! See #3970.

As for the `inline` feature, I'll try to get that done soonish. I think we can close this PR.

daleharvey:
Hrm, my worry is that the test it is based (https://github.com/pouchdb/pouchdb/pull/3966) on is green which means we havent hit the case where the browser starts killing requests? 

But yeh good with skipping this and going in for fixes

nolanlawson:
We could try upping that test to ~10000 attachments and see what happens.

Issue #3963, opened 2015-06-14
Inconsistent .length property for attaches in viewindexer, 3.6.0


ermouth:
Can prevent correct calculation of attachments sizes for a doc.
Repro is here http://jquerymy.com/pouchdb/badviews2.html

nolanlawson:
Nice find. This is indeed a bug introduced in 3.6.0. I have a fix.

nolanlawson:
fixed by f6720b64b901921f4dec7d7aa9be604d6e9ad8ae

ermouth:
@nolanlawson seems that it won‘t work for docs, saved by at least 3.3.1

nolanlawson:
Yep, there's no migration. I didn't think this was important enough to justify a migration.

Keep in mind a migration would mean re-reading in all the attachments and re-measuring their blobs. An expensive operation.

You can do it yourself with allDocs and bulkDocs with new_edits:false. I can provide a script when the new release comes out.

ermouth:
@nolanlawson this approach seems inapplicable to saved conflicts. Although since they are unreacheable from views, it‘s not so important. 

It seems that saved conflics with attaches are destined to have no length property for their files attached – or solution becomes too low-level.

PR #3965, opened 2015-06-14
(#3963) - fix attachment length for stubs

nolanlawson:

In short, we weren't being careful enough to
test the `.length` property after a `put()` with
stubs. The new test should handle that.


daleharvey:
:+1: this looks good, cheers

nolanlawson:
f6720b64b901921f4dec7d7aa9be604d6e9ad8ae

PR #3966, opened 2015-06-15
(#3961) - replication of many attachments

nolanlawson:
Still not reproducing the attachment loss bug. But at least we have another test.

### DIFF
--- a/tests/integration/test.attachments.js
+++ b/tests/integration/test.attachments.js
@@ -3476,6 +3476,53 @@ repl_adapters.forEach(function (adapters) {
       });
     });
 
+      it('#3961 Many attachments on same doc', function () {
+        var doc = {_id: 'foo', _attachments: {}};
+
+        var db = new PouchDB(dbs.name);
+        var remote = new PouchDB(dbs.remote);
+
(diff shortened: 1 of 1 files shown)


daleharvey:
As mentioned in the other thread, worried that this is green which means we arent seeing the case where the browser is killing concurrent requests, taking a look

daleharvey:
So yeh, if we bump this up to 1000 it starts to fail, I also expect with additional latency and larger files it will be much more sensitive. I think we can expand this to include more and larger files until we see it fail then we have a good basis for the upcoming fixes

nolanlawson:
OK sweet, glad to know we have a test for it. And yeah, I see where you're coming from now. Makes sense.

nolanlawson:
@daleharvey Any reason not to fix this? Seems like a reasonable test and may catch something in the future, even as-is.

daleharvey:
Yeh I was meaning to rebase this and merge once green :+1:

nolanlawson:
rebased

daleharvey:
Merged - https://github.com/pouchdb/pouchdb/commit/70998d157c056a345f00fa9776ca971dc4de9565

PR #3970, opened 2015-06-16
(#3961) - fetch http attachments in batches

nolanlawson:
Rebased on https://github.com/pouchdb/pouchdb/pull/3966 to
demonstrate that this passes the tests even if you have
many attachments on the same doc.

### DIFF
--- a/lib/adapters/http/index.js
+++ b/lib/adapters/http/index.js
@@ -8,6 +8,7 @@ var CHANGES_BATCH_SIZE = 25;
 // URL into account, we fudge it a bit.
 // TODO: we could measure the full URL to enforce exactly 2000 chars
 var MAX_URL_LENGTH = 1800;
+var ATTACHMENT_BATCH_SIZE = 6; // browsers cap simultaneous requests at ~6-8
 
 var readAsBinaryString = require('../../deps/binary/readAsBinaryString');
 var binStringToBluffer =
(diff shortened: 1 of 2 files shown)


daleharvey:
I think a better solution for handling concurrent requests is to have all http requests go through a proxy function that manages how many concurrent requests are active (globally) so clients like the replicator can just make http requests without worrying about how many they are making but we manage it across the whole instance

daleharvey:
So we can have multiple replications going or various other clients triggering requests but still be confident that we arent just throwing tons of them up

nolanlawson:
Remind me again - what's the issue with opening up many simultaneous requests? I'm worried we're going to end up reimplementing a lot of what the browser itself is doing - i.e. capping the number of concurrent requests to 6-8.

daleharvey:
We had bugs where when you make a whole bunch of requests the browser would start killing any that were idle a short period of time

daleharvey:
> Remind me again - what's the issue with opening up many simultaneous requests? I'm 
> worried we're going to end up reimplementing a lot of what the browser itself is doing - 
> i.e. capping the number of concurrent requests to 6-8.

My main issue is that here we are doing the capping in the wrong place, this is implementing a batch size of 6 in the the http adapter, for node we may have a huge or practically unlimited simultaneous requests, in the different browsers there are different amounts and this doesnt cap the total limit of requests pouchdb itself is making, just in this particular api call.

I agree we should avoid doing any work the browser can handle itself, preferably we do no capping, but if the browser does not do it well enough (as we are seeing in the test for this put up to 1000) I think we should do it in one place that handles all requests. Will hopefully get  apatch for the weekend

daleharvey:
Closing this, so with the comments I think we should either

  1. Not limit requests at all if there is no evidence to suggest that multiple connections are a problem as the browser handles this for us (I think theres already evidence and will work on the other test to make it testable)

  2. If we do limit the browser connections, do it in a global way, at the least per instance so we dont need to handle request queing everywhere we may want to make multiple requests

Issue #3974, opened 2015-06-16
Saucelabs is faily


daleharvey:
Seems like an issue on their end, all of a sudden getting a large but not total failure on saucelabs. Will ping someone

daleharvey:
pinged some people I know from saucelabs, hoping its a known issue, gonna try upgrading sauce-connect-launcher to see if that makes a difference since we are on a fairly old version

nolanlawson:
fixed by https://github.com/pouchdb/pouchdb/pull/3978

daleharvey:
This isnt completely fixed, gonna keep it open till we can get back to around the stability we were having before

daleharvey:
So I think between the sauceconnect upgrade and https://github.com/pouchdb/pouchdb/pull/3984 this should be back to being more stable

PR #3978, opened 2015-06-17
(#3974) - Upgrade sauce connect launcher

daleharvey:


daleharvey:
This needed a few restarts so certainly not fixing the problem completely, but is a lot more stable that it was

nolanlawson:
cool, looks good

Issue #3986, opened 2015-06-19
Split out android tests more


daleharvey:
Seeing them fail and they keep coming close to the 35 minute mark that I suspect is our limit

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/1d98f4b8283beebff615abaad1ce1d7174a18245

PR #3988, opened 2015-06-19
(#3986) - Split out android tests more

daleharvey:


.travis.yml
```diff
@@ -58,8 +58,8 @@ env:
   - CLIENT="saucelabs:internet explorer:10:Windows 8" COMMAND=test
 
   # split up the android tests as it goes over time
-  - GREP=test.replication.js CLIENT="saucelabs:Android:5.1:Linux" COMMAND=test
-  - GREP=test.replication.js INVERT=true CLIENT="saucelabs:Android:5.1:Linux" COMMAND=test
+  - GREP=test.replication.js,test.attachments.js CLIENT="saucelabs:Android:5.1:Linux" COMMAND=test
+  - GREP=test.replication.js,test.attachments.js INVERT=true CLIENT="saucelabs:Android:5.1:Linux" COMMAND=test
```
nolanlawson: whoa, I didn't know you could put a comma in there. that is rad

nolanlawson:
+1 when green, should help with android flakiness

.travis.yml
daleharvey: Heh I was sure I had something like this working, but it isnt now

daleharvey:
ah, found the issue, | not ,

daleharvey:
Not sure this is working correctly at all, seeing the basics tests run when we are grepping attachments + replication

nolanlawson:
If it isn't, let's just add a string like "suite2" to the attachments and replication tests.

daleharvey:
Dumb mistake, need to quote special characters that are gonna be run through bash

daleharvey:
Ugh, that was a but now its fixed, there must be a bug in webview passing in parameters, the url we generate is working fine in the browser

daleharvey:
Going with your suggestion @nolanlawson, debugging the param issue into webview didnt seem worth it

daleharvey:
No restarts :) landed - https://github.com/pouchdb/pouchdb/commit/1d98f4b8283beebff615abaad1ce1d7174a18245

Issue #3990, opened 2015-06-20
Switch to fetch as primary xhr api 


daleharvey:

daleharvey:
Actually not sure this is the best idea until at least the `abort()` issue gets sorted - https://twitter.com/daleharvey/status/612406302350921729

daleharvey:
Closing for now

PR #3992, opened 2015-06-21
(#3990) - Create new global ajax wrapper

daleharvey:
So we may not want to switch to fetch() api, but this is still worth doing I think, we should try to have clean reusable polyfills in `/deps/` and the PouchDB logic like nonce should be kept outside of it, if we want to limit the number of requests pouchdb makes in the browser than it can be done here or in a similiar way

### DIFF
--- a/lib/adapters/idb/blobSupport.js
+++ b/lib/adapters/idb/blobSupport.js
@@ -1,6 +1,7 @@
 'use strict';
 
 var utils = require('../../utils');
+
 var idbConstants = require('./constants');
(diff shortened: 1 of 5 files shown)


daleharvey:
@nickcolley @jo @karlwestin Yo, Nolan and Calvin are busy these days, anyone fancy taking a look at / +1'ing this or other PRs I have open, most of them are minor, cheers and apologies for the bother

nickcolley:
This looks good to me +1

daleharvey:
Sweet, cheers - https://github.com/pouchdb/pouchdb/commit/07e32b307c4622dcfd5a7bf16e10691c5c60fcbf

Issue #3994, opened 2015-06-21
pouchdb.js is 734KB


daleharvey:
Assuming the browserify upgrade https://github.com/pouchdb/pouchdb/commit/39e5367e71228884e0115ffde63664d2c910c415 broke some of the hacks we have been using to avoid bundling a bunch of things

levelup.js and buffer are both in there

daleharvey:
actually 838K, not sure if it may be including bluebird as well (original test was on the remove bluebird branch). The base size for the last release is 367KB

daleharvey:
hrm not looking like https://github.com/pouchdb/pouchdb/commit/39e5367e71228884e0115ffde63664d2c910c415

daleharvey:
https://github.com/pouchdb/pouchdb/commit/39da6013d74685189fcdf5bc2c617db5b76e88bd is the problem

daleharvey:
fixed in https://github.com/pouchdb/pouchdb/commit/df596be6d48929524d36094f21a8cc24297b3e0a

PR #3995, opened 2015-06-21
(#3994) - Fix path to leveldb exclusion

daleharvey:


daleharvey:
Easily missed (I review :))

Fixed in https://github.com/pouchdb/pouchdb/commit/df596be6d48929524d36094f21a8cc24297b3e0a

re: twitter it probably is worth a test that fails the build when it gets obviously too large, I didnt want to pick an arbitrary number (500KB seemed best) but thats probably perfect getting in the way of being good

Issue #3626, opened 2015-03-10
Force appcache update


nickcolley:
I'm wondering if there's a better solution that current to update the appcache when a user visits the site.
Currently if you came from Twitter and had an older version cached you'd see no changes until you refreshed again.

I'd go as far as saying most people will not refresh.

nolanlawson:
I think that's a pretty poor solution. Isn't the recommended solution to show a non-modal popup asking the user if they'd like to refresh?

Alternatively, we could just get rid of AppCache... :stuck_out_tongue:

nickcolley:
I'm not suggesting to force a refresh, was hoping there was another way. The update indicator is a nice idea but just seems like a faff when we just want don't really want the user to control the version they're seeing..

(Could get rid of it yeah..)

nolanlawson:
Adding Dash support (https://github.com/pouchdb/pouchdb/issues/3630) would be another way to make it work offline. ;)

nickcolley:
Not interested in closed platform doc apps, so don't really consider that an alternative. :P

nolanlawson:
Yeah, it's not a true replacement. Also we recently got this compliment, so we must be doing something right: https://twitter.com/sujayt/status/575398863151833089

nickcolley:
I think we should go with the ol' "There's new stuff press me" notification.

nolanlawson:
1c4f1f14e5b5c5fb5e3bf0e3794047157a84b4c9

PR #4005, opened 2015-06-26
(#3626) - Offer to reload when update is avaliable

nickcolley:
With this patch when a user hits the page and downloads the new appcache goodness a little button comes down (complete with transition) from the top of the site to prompt them to update

image

### DIFF
--- a/docs/_layouts/default.html
+++ b/docs/_layouts/default.html
@@ -181,13 +181,25 @@
           giveIntro();
         }
       }
-      
       function giveIntro() {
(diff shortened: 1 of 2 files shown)


nickcolley:
If anyone has something more exciting to use instead of 'Content update, reload now?' feel free to share (keep it small though since it has to fit on smaller screens..)

nolanlawson:
Nice work! I think the wording is perfect. :)

In the future we may want to consider ServiceWorkers, but I like that this works on all browsers.

1c4f1f14e5b5c5fb5e3bf0e3794047157a84b4c9

PR #4009, opened 2015-06-28
Improve replication ID generation

KlausTrainer:
This improves replication ID generation in the following ways:

* it ignores the order of array elements in the `doc_ids` option
* it ignores the `query_params` option if there's no `filter` option
* it ignores the order of object properties in the `query_params` option
* it ignores the `view` option unless the `filter` option's value is
  `_view`.

### DIFF
--- a/lib/replicate/genReplicationId.js
+++ b/lib/replicate/genReplicationId.js
@@ -1,15 +1,38 @@
 'use strict';
 
 var md5 = require('../deps/md5');
+var collate = require('pouchdb-collate').collate;
+
+function sortObjectPropertiesByKey(queryParams) {
+  return Object.keys(queryParams).sort(collate).reduce(
+    function (result, key) {
+      result[key] = queryParams[key];
+      return result;
+    },
+    {}
+  );
+}
 
 // TODO: check CouchDB's replication id generation
 // Generate a unique id particular to this replication
 function genReplicationId(src, target, opts) {
+  var docIds = opts.doc_ids ? opts.doc_ids.sort(collate) : '';
   var filterFun = opts.filter ? opts.filter.toString() : '';
+  var queryParams = '';
+  var filterViewName =  '';
+
+  if (opts.filter && opts.query_params) {
+    queryParams = JSON.stringify(sortObjectPropertiesByKey(opts.query_params));
+  }
+
+  if (opts.filter && opts.filter === '_view') {
+    filterViewName = opts.view.toString();
+  }
+
   return src.id().then(function (src_id) {
     return target.id().then(function (target_id) {
-      var queryData = src_id + target_id + filterFun +
-        JSON.stringify(opts.query_params) + opts.doc_ids;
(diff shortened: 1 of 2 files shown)


janl:
So this is “improve” as in “get closer to CouchDB?”

KlausTrainer:
> So this is “improve” as in “get closer to CouchDB?”

Without having looked into CouchDB's replication id generation too closely, I believe that this will make it closer to it :)

Note that this addresses the following issues:

* If I want to trigger a new filtered replication that is different to an existing replication only with respect to the `view` option, the new replication won't be triggered, as there's already a replication with the same ID.
* It is possible to trigger a new replication that is equivalent to an existing one. That is, the new replication has options that are equivalent to an existing replication's options.

An example of an equivalent, but different set of replication options, is having the same document IDs in the `doc_ids` array in a different order. Another example would be a set of replication options with different `query_params` that's missing a `filter` option. As the replicator ignores `query_params` if it doesn't come along with a `filter` option, `query_params` should only be used for calculating the replication ID if it comes along with a `filter` option. If I specify replication options that are equivalent to an active replication (i.e., they would lead to a replication that is exactly the same), I expect to get the same replication ID, even if e.g. I use a different order when specifying the same set of document IDs in the `doc_ids` array.

Generally, I believe that it's good for replication ID generation to be as predictable as possible, also with regard to use cases outside of PouchDB proper. An example would be a PouchDB plugin that allows for checking the progress of active replications by looking at their respective replication document's `last_seq` value.

janl:
> Generally, I believe that it's good for replication ID generation to be as predictable as possible

+1, that’s what the CouchDB algorithm is designed to do.

janl:
Might be worth taking a peek at the Erlang implementation :)

daleharvey:
Actually taking a look, at least inside couch replicator it looks like this could also be an improvement for couchdb, it looks like it just picks up DocID's directly, unless they get sorted somewhere else

https://github.com/apache/couchdb-couch-replicator/blob/master/src/couch_replicator_utils.erl#L125

daleharvey:
So I had one very tiny nit, if possible I think 

```function sortObjectPropertiesByKey(queryParams) {
  return Object.keys(queryParams).sort(collate).reduce(function (result, key) {
    result[key] = queryParams[key];
    return result;
  }, {});
}```

is cleaner than 

```function sortObjectPropertiesByKey(queryParams) {
  return Object.keys(queryParams).sort(collate).reduce(
    function (result, key) {
      result[key] = queryParams[key];
      return result;
    },
    {}
  );
}```

I switched it on the way in, its a nice and clean fix and the tests look great, thanks a lot :)

daleharvey:
Merged in https://github.com/pouchdb/pouchdb/commit/7aa5ed1fd8c6f8f93c770b4c2d4101e5a38042d3

nolanlawson:
Related to https://github.com/nolanlawson/pouchdb-load/issues/29, sadly this will break pouchdb-load when we publish it in PouchDB, but that's due to my own laziness, so I will fix it.

PR #4021, opened 2015-07-03
Fix a bug where an aborted fetch on a binary attachment fails silently

karlwestin:
If the request that is breaking is for example an image, there's a 'parse error' thrown, which causes the replication to halt. The replication is not able to detect the error and retry, or quit correctly.  
pouchdb-attachment-unexpected-end

### DIFF
--- a/lib/deps/ajax.js
+++ b/lib/deps/ajax.js
@@ -129,9 +129,12 @@ function ajax(options, adapterCallback) {
     if (response.statusCode >= 200 && response.statusCode < 300) {
       onSuccess(data, response, callback);
     } else {
-      if (options.binary) {
+      try {
+        // See if we received an error message from the server
+        // If the connection was interrupted, the data might be nothing
+        // that's why we have the try/catch
         data = JSON.parse(data.toString());
-      }
+      } catch(e) {}
       error = errors.generateErrorFromResponse(data);
       error.status = response.statusCode;
       callback(error);
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
     "less": "~1.7.0",
     "mkdirp": "~0.4.2",
     "mocha": "^1.18.2",
+    "mockery": "^1.4.0",
     "ncp": "~0.5.0",
     "phantomjs": "^1.9.7-5",
     "pouchdb-express-router": "^0.0.6",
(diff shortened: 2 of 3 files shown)


lib/deps/ajax.js
```diff
@@ -129,7 +129,7 @@ function ajax(options, adapterCallback) {
     if (response.statusCode >= 200 && response.statusCode < 300) {
       onSuccess(data, response, callback);
     } else {
-      if (options.binary) {
+      if (options.binary && data.length) {
```
daleharvey: Sorry if I am being dumb, why are we `JSON` parsing when we are expecting binary in the first place?

daleharvey: Oh, because this is an error, not entirely sure the `options.binary` check should be in there at all, it seems like

    try { 
      data = JSON.parse(data.toString());
    } catch(e) {}
    error = errors.generateErrorFromResponse(data);
    error.status = response.statusCode;
    callback(error);

Would what we should do?

karlwestin:
good point, i'll have a look tomorrow!

daleharvey:
Cool thanks, if it turns out we need the `if (binary)` check then we should likely still have the try catch in case of invalid json, otherwise this looks good, cheers :)

karlwestin:
It fixes the bug for me without the binary check, i just don't understand this test failure. 
I can't reproduce it when doing `SERVER_ADAPTER=memdown LEVEL_ADAPTER=memdown SERVER=pouchdb-server COMMAND=test npm run test` locally.

daleharvey:
Its very likely that this is an intermittent failure, will try restarting (travis is looking a little flakey) 

Otherwise this is looking great, +100 for the new tests

daleharvey:
Travis is being annoying so opened a squashed PR to try this - https://github.com/pouchdb/pouchdb/pull/4029

daleharvey:
@karlwestin Also I have invited you to the PouchDB owners group, we try to do this fairly early once someone is participating in the project, it mostly means you have full write access to the repostory as well as a few other things, keep in mind the guidelines @ https://github.com/pouchdb/pouchdb/blob/master/CONTRIBUTING.md#committers if you want to start pushing code etc, thanks for the patches :)

daleharvey:
Between the 2 PR's we have all green now, travis is just acting up today, merged in https://github.com/pouchdb/pouchdb/commit/931df967ef211ea12362b3afc6ee7f1570e89c8f, cheers

karlwestin:
awesome, thanks!

PR #4022, opened 2015-07-03
Feature: Allow user to raise changes/attachments timeout

karlwestin:
This is a feature-request, since we're seeing the changes call taking longer than 30 seconds sometimes. 
Would be nice to be able to up it a notch!

Cheers!

### DIFF
--- a/tests/integration/test.http.js
+++ b/tests/integration/test.http.js
@@ -86,4 +86,31 @@ describe('test.http.js', function () {
     uri.host.should.equal('foo.com');
   });
 
+
+  it('Allows the "ajax timeout" to extend "changes timeout"', function() {
+    var timeout = 120000;
+    var db = new PouchDB(dbs.name, {
+      skipSetup: true,
+      ajax: {
+        timeout: timeout
+      }
+    });
+
+    var ajax = PouchDB.utils.ajax;
+    var ajaxOpts;
+    PouchDB.utils.ajax = function(opts) {
+      if(/changes/.test(opts.url)) {
+        ajaxOpts = opts;
+      }
+    };
+
+    db.changes({
+      onChange: function(change) {}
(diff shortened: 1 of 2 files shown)


daleharvey:
Yeh this has been one of those things that has needed cleaned up for a while, I think however if the user specified a timeout it should be applied whether or not it is below or above the default time (ie the user can optionall reduce the timeout), I think the Math.max should be reduce, but otherwise this looks great

(a side benefit of taking the explicit user timeout is that we can reduce the time the test takes :)

@karlwestin how does that sound?

karlwestin:
Thanks for having a look!
Sounds good, i'm just not sure i understand 'i think math.max should be reduce' ?
if we go for either opts or default, wouldn't it just be like this?
`opts.timeout = opts.timeout || ajaxOpts.timeout || DEFAULT;`

daleharvey:
Sorry, "should be able to reduce" I mean, so as you posted

tests/integration/test.http.js
```diff
@@ -86,4 +86,31 @@ describe('test.http.js', function () {
...
+    var ajaxOpts;
+    PouchDB.utils.ajax = function(opts) {
+      if(/changes/.test(opts.url)) {
+        ajaxOpts = opts;
+      }
+    };
+
+    db.changes({
+      onChange: function(change) {}
+    });
```
daleharvey: `onChange` is deprecated, so this should be 

    db.changes().on('change', function() {});

I can fix on the way in though

daleharvey:
Awesome thanks a lot, merged in https://github.com/pouchdb/pouchdb/commit/4da908d9db11d4facd58aa8edbc5c1a420fbc78e

karlwestin:
:+1: cheers!

Issue #3873, opened 2015-05-24
Update leveldown/levelup to latest


nolanlawson:
See https://github.com/pouchdb/pouchdb/issues/3462 and https://github.com/pouchdb/pouchdb/pull/3863 for past discussion. This will require a migration script because there are breaking changes in levelup/leveldown.

dukex:
The levelup API will change in v2, see https://github.com/Level/levelup/pull/350

nolanlawson:
I took at a stab at just updating to leveldown 1.4.0 (e9b20d7bae1433033a9c24b93f6c2a2d936f323c
), but I was not successful. There seems to be an issue with `destroy()`; I asked in the leveldb room in IRC but I honestly don't have much interest in this right now:

```
5:16 PM <nolanlawson> hi all, I am trying to upgrade pouchdb from leveldown 0.10.5 to 1.4.0
5:16 PM <nolanlawson> I seem to be running into this issue where if you create a level, then destroy it, then reopen it, you get:
5:17 PM <nolanlawson> OpenError: IO error: lock LOCK: already held by process
5:18 PM <nolanlawson> I looked at the changelog, but nothing jumps out at me about what changed in 1.0.0 to cause this to happen
5:18 PM <nolanlawson> also is there a list of breaking changes somewhere?
```

nolanlawson:
^ issue filed on leveldown

cshum:
The `.destroy()` aren't that reliable. It's not even part of the [*DOWN](https://github.com/Level/abstract-leveldown/) spec. 
Would it be ok to simply clear up the folder [rimraf](https://github.com/isaacs/rimraf)?

nolanlawson:
If rimraf works more reliably, then a strong +1, yeah

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/8fd74079cb799d47a2a7679a7ed74d9c0c1aeacc

PR #4025, opened 2015-07-05
leveldown (>= 1.2.2) now supports prebuilds

broerse:
`leveldown` (>= 1.2.2) now supports prebuilds so you shouldn't have to use `leveldown-prebuilt` anymore.

### DIFF
--- a/docs/errors.md
+++ b/docs/errors.md
@@ -208,16 +208,16 @@ var db = new PouchDB('database', { db: require('sqldown') });
 
 ```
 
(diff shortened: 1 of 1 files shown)


daleharvey:
Awesome thanks a lot, we also want to upgrade level @ https://github.com/pouchdb/pouchdb/issues/3873 however a few issues in the way

daleharvey:
Merged the doc updates in https://github.com/pouchdb/pouchdb/commit/dc02a682137eebead727b86bb75764c612f74172 though, thanks

broerse:
Thanks

Issue #3632, opened 2015-03-11
High CPU usage if db.changes({live:true}) called more than once


mgjv:
I stumbled on this when I accidentally included the same javascript twice. The following will result in high CPU usage

var db = new PouchDB('tester')
db.changes({ since: 'now', live: true })
db.changes({ since: 'now', live: true })

environment: Google Chrome Version 41.0.2272.76 (64-bit) on Mac OS X 10.10.2
out of 8 cores, 4 run up to 50%

mgjv:
Problem also occurs with Firefox16.0.1 and Safari 8.04 (10600.4.10.7) on Mac OS X

Problem does NOT occur on Firefox Developer Edition 38.0a2

oliverdm:
Similar problem with high CPU and disk utilization (> 80%) when using live sync and live changes in Internet Explorer 11 simultaneously.

Resource utilization stays high up to several minutes, goes down at some point, but also goes up again with new changes. Also happens with an empty database.

The following HTML should reproduce the issue when running on a localhost with CouchDB:

```
<!DOCTYPE html>
<html>
<head>
    <title>HIGH CPU AND DISK UTILIZATION IN IE 11</title>
    <script src="path/to/pouchdb.js"></script>
</head>
<body>
<script>
var localDb = new PouchDB("test");
var remoteDb = new PouchDB("http://localhost:5984/test");

var syncLive = PouchDB.sync(localDb, remoteDb, {
    live: true
});
var changesLive = localDb.changes({
    live: true,
    since: 'now'
});
</script>
</body>
</html>
```

According to IE profiler hot spots are:
 - Anonymous function in pouchdb.js line 8441
 - EventEmitter.prototype.emit in pouchdb.js 9118

(PouchDB version 3.3.1)

6f:
I ran into the same issue, but I think I fixed mine with the following commit:
6f/pouchdb@123d41b89a6b6b3d499102119fb75a3d6f5205ca

I'd appreciate if you could test if the patch solves the issue that you're having.

mgjv:
Well found.

I checked out branch 3632-issue-non-deterministic-loop which contains the above commit, and can confirm that the problem no longer occurs. It still occurs on master.

tested on Chrome, Firefox and Safari, as in initial report/

nolanlawson:
@6f could you open up a PR? I'd be happy to approve the fix, assuming it passes the tests. Seems like a good find to me.

nolanlawson:
[...]

PR #4028, opened 2015-07-06
fix for high CPU utilization

michalchochol:
Fix resolves the problem of high CPU utilization, caused by by infinite loop of change listeners. 

Here is the code to reproduce problem:

<!doctype html>
<title>PouchDB issue #3632</title>
<script type="text/javascript" src="https://github.com/pouchdb/pouchdb/releases/download/3.6.0/pouchdb-3.6.0.min.js"></script>
<script type="text/javascript">                                                                                                                 
var testDb = new PouchDB("bug3632");                                                                                                      

for(var i=1; i<10; ++i) {
  (function(i) {                                                                                                                       
    testDb.changes({live:true}).on("change", function(change) {                                                                                   
      console.log('listener' + i + ' got a change event');                                                                                        
    });
  })(i);
}

setTimeout(function() {                                                                                                                         
  testDb.put({'_id': ''+new Date().getTime(), 'msg': 'Hello PouchDB'});                                                                         
}, 200);                                                                                                                                        
</script>

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -181,9 +181,9 @@ Changes.prototype.addListener = function (dbName, id, db, opts) {
       }
     }).on('complete', function () {
       if (inprogress === 'waiting') {
-        process.nextTick(function () {
-          self.notify(dbName);
-        });
+        setTimeout(function(){
+          eventFunction();
+        },0);
       }
       inprogress = false;
     }).on('error', function () {
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -159,16 +159,10 @@ Changes.prototype.addListener = function (dbName, id, db, opts) {
     return;
   }
   var self = this;
-  var inprogress = false;
   function eventFunction() {
     if (!self._listeners[id]) {
       return;
     }
-    if (inprogress) {
-      inprogress = 'waiting';
-      return;
-    }
-    inprogress = true;
     var changesOpts = pick(opts, [
       'style', 'include_docs', 'attachments', 'conflicts', 'filter',
       'doc_ids', 'view', 'since', 'query_params', 'binary'
@@ -180,14 +174,7 @@ Changes.prototype.addListener = function (dbName, id, db, opts) {
         exports.call(opts.onChange, c);
       }
     }).on('complete', function () {
-      if (inprogress === 'waiting') {
-        setTimeout(function(){
-          eventFunction();
-        },0);
-      }
-      inprogress = false;
     }).on('error', function () {
-      inprogress = false;
     });
   }
   this._listeners[id] = eventFunction;


daleharvey:
hrm @calvinmetcalf any opinions on this? Happy to merge as a workaround but I would really like to understand whats causing this in the first place and see if we can figure out a test to ensure it doesnt happen again

daleharvey:
So taking a look into this, it does seem to reduce CPU in chrome but the CPU is still very high and in firefox it seems pretty constant either way. Given the test we should go back to being idle at some point so something is very wrong

michalchochol:
@daleharvey 
I tested also in Firefox (ver. 38.0 under Linux), according to test described in my first post, and it solves the problem. 

@nickcolley 
eventFunction is defined in line 163 of utils.js :)

daleharvey:
@michalchochol You are right, I was making local changes but didnt change the line that fetched pouchdb from the cdn

So this does fix the CPU issue for me, however despite writing the first version of this code I am having an extremely hard time following what exactly is happening and putting us into an infinite loop, @michalchochol what led you to this change? going to dig into this more and see if we can get a test figured out

michalchochol:
@daleharvey 
In my opinion infinite loop is caused by "inprogress" variable usage in Changes.prototype.addListener.
As you can see, it is declared and initiated in scope of Changes.prototype.addListener, but its reference is used by eventFunction. So each one eventFunction listener uses the same instance of inprogress variable (from eventFunction closure) and changes its state. In result listeners influence on each other and finally causes infinite loop of listeners execution.

Calling eventFunction in this way:
        setTimeout(function(){
          eventFunction();
        },0);
probably separates listeners in some way and solves the problem. But of course it is only workaround.
After deeper analysis, I concluded, that inprogress variable is not needed at all. I prepared new version of Changes.prototype.addListener - it looks like this (it is available in repository too):

Changes.prototype.addListener = function (dbName, id, db, opts) {
  if (this._listeners[id]) {
    return;
  }
  var self = this;
  function eventFunction() {
    if (!self._listeners[id]) {
      return;
    }
    var changesOpts = pick(opts, [
      'style', 'include_docs', 'attachments', 'conflicts', 'filter',
      'doc_ids', 'view', 'since', 'query_params', 'binary'
    ]);

    db.changes(changesOpts).on('change', function (c) {
      if (c.seq > opts.since && !opts.cancelled) {
        opts.since = c.seq;
        exports.call(opts.onChange, c);
      }
    }).on('complete', function () {
    }).on('error', function () {
    });
  }
  this._listeners[id] = eventFunction;
  this.on(dbName, eventFunction);
};

Finally everything works fine - it passes all pouchdb tests, that I had done few times to be sure. I have tested modified version in our project too, and everything work like a charm.

daleharvey:
@michalchochol Thanks for looking into this more, sadly looks like removing inProgress fails the tests that were added when that property was added, I think it is intentional that it works across instances. 

I am gonna do a full run of your previous patch @ https://github.com/pouchdb/pouchdb/pull/4042, if that passes all tests then I think we should merge as I have confirmed that it does fix the CPU across browsers, then I would like to dig deeper and try to follow up with more tests to make sure we dont hit infinite loops again

michalchochol:
@daleharvey Which tests are failed after inprogres removing?

daleharvey:
https://travis-ci.org/pouchdb/pouchdb/builds/70385750 has the failures, https://travis-ci.org/pouchdb/pouchdb/jobs/70385758 in running in plain firefox and fails with 

```{ failures: 

   [ { message: 'expected [ Array(5) ] to have a length below 5 but got 5',

       title: 'it handles a bunch of individual changes in live replication'
```

daleharvey:
https://github.com/pouchdb/pouchdb/commit/801e07a531edc568d6cbcc8f822f8a26d470f03c is the commit that added the tests

daleharvey:
Ok, I merged the original commit, sorry for taking so long but thanks for following up. I am going to carry on looking into this and see if we can make it clearer what its supposed to do and less liable to break in future, but for now the issue is fixed - https://github.com/pouchdb/pouchdb/commit/52225348190e89b20e9a10109bf2fc56ee753728

michalchochol:
ok, thanks :)

PR #4029, opened 2015-07-06
(#4021) - Aborted fetch on a binary attachment causes the replication to halt.

daleharvey:
implement other fix for attach error
Added a test for interrupted binary file


daleharvey:
https://github.com/pouchdb/pouchdb/commit/931df967ef211ea12362b3afc6ee7f1570e89c8f

Issue #3999, opened 2015-06-24
Mismatching checkpoints causes replication to start over from change 0


karlwestin:
Trying to give a little bit of background to explain the issue:

We're replicating from a 'central', relatively well connected CouchDB, to a PouchDB embeddded in a Cordova app, that is running on android tablets. Those tablets are sometimes offline, sometimes running on very bad connections (think 50 kbps, 0.5-1s latency, some packet loss). Because of the bad connections, we're using very small `batch_size` (less than the number of outstanding requests the browser can have) to avoid timeouts. 

The scenario i describe below represents a replication of data from the central server, to the mobile device. Central = Source, Mobile = Target

In the screenshot here, you see PouchDB reading the replication last_sequence number from central, 3945, right after, it starts reading changes from last_seq 0. 
![http://i.imgur.com/n04z260.png](http://i.imgur.com/n04z260.png)

What happens is that the [comparison of replication state on Central/Mobile disagrees](https://github.com/pouchdb/pouchdb/blob/269e236faf302a4065267f2d0f5d17bc67b38c10/lib/replicate/checkpointer.js#L71), and therefore, returns to 0.

```
if (collate(targetDoc.last_seq, sourceDoc.last_seq) === 0) {
  return sourceDoc.last_seq;
}
return 0;
```

### Where does this difference come from?

When a replication batch is done, we PUT a checkpoint, first on Mobile (Target) then on Central (Source). If the PUT on Central fails, Mobile and Central will have different opinions on the state of replications. We have tried a solution where the PUT on Central would be retried 5x before failing, and if it fails, Mobile would roll back to its previous check point. The problem with this seems to be that when a PUT fails, Pouch can't know if the write failed, or if the write succeeded and we didnt get the response. So it seems half the time, a rollback is correct, and half the time the checkpoint shouldnt be rolled back. 

### Why is this a problem?

Unfortunately 'catching up with the changes', like pouch does now can be pretty slow for us. The problem is worse by the catch-up being non-deterministic. A device can spend an hour or two catching up on checkpoints, just to slip and go back to 0 just before it was time to read docs. (The central DB in our case has a sequence number around 20k, with batch_size 5 that takes a while). I understand that the main problem here is that our connection sucks, but it would be helpful to reach a solution where we can leave a device syncing over night, and know that it will not fall back to 0 again.

### Proposed solution

[...]

Issue #4002, opened 2015-06-26
CouchDB master is failing


daleharvey:
There was a mention about an overhaul of the build system recently so assuming it broke how we launch

https://github.com/pouchdb/pouchdb/blob/master/bin/run-couch-master-on-travis.sh#L34

@janl any pointers on what the equivalent to startup couch like above is now? cheers

janl:
This should still work. Can you point me to instructions on how to run the the test suite and what errors you are seeing? I’ll have a look then.

daleharvey:
To run the test suite against couch master you start it with the above line and

    COUCH_HOST='http://127.0.0.1:15984' npm test

my git clone of couch needed to `rm -rf src` to start building again, otherwise https://gist.github.com/daleharvey/620fab43a66b92a5f61d happened, but got a checkout building now and should be able to see whats up

Added couch master to allowed failures also

janl:
that looks like get-deps failing during configure

daleharvey:
Yeh, although that was just a local checkout, was just mentioning in case that was something you noticed would want to fix on couch side, a clean checkout worked

It looks for the pouchdb tests that it may just be a timing issue, we do a sleep to wait for couch to start and it looks like it may take a little longer to start now

daleharvey:
https://travis-ci.org/pouchdb/pouchdb/jobs/68684560 is an example of a failing run, building works fine, we dont see the start logs with the & but it cant connect

daleharvey:
hrm, the exact same thing is working fine locally

janl:
yay! :D

daleharvey:
So the issue is a change in either couchdb or travis made failing to build the docs a full failure as instead of just a warning as it was previously

daleharvey:
So I merged https://github.com/pouchdb/pouchdb/commit/9a0354fced16787a625d25255889905a0f71c56b with r=me since it was just infrastructure and it uncovered the fact that couchdb-master is actually failing now

daleharvey:
https://travis-ci.org/pouchdb/pouchdb/jobs/70516090

```
test.changes.js-http #2569 Big non-live doc_ids filter:

     Something wrong with the request
```

daleharvey:
ok, the _doc_ids bug should get fixed by https://github.com/apache/couchdb-chttpd/pull/42

daleharvey:
Tested the above locally and all fixed, should hopefully get merged soon

rnewson:
[...]

PR #4035, opened 2015-07-07
(#3999) Implement CouchDB-like checkpoints

karlwestin:
Sending the PR for review, I'll be happy to squash commits but i expect some comments on this first :)

This PR implements CouchDB-like checkpoints, to avoid losing the entire replication history when 2 checkpoints mismatch. Every replication writes a `session_id` that is compared, and if they match, the last_seq from the source db checkpoint is used. If the session_id does not match, the checkpoint docs has a `history` array, where session_id from previous replications are compared to find an older common point. Fir doc size reasons, I did not include all the keys that CouchDB has, and the history is limited to 5 items. 

Comparison:
```
// Checkpoint format used in this PR:
{
     "_id": "_local/edd1d4be175dd4c88b6ec1f07731d98e",
     "_rev": "0-6",
     "session_id": "6030857ab3a97de2a59238c2943da145",
     "last_seq": 8308,
     "replication_id_version": 2,
     "history": [{
         "session_id": "6030857ab3a97de2a59238c2943da145",
         "last_seq": 8308
     }]
}

// CouchDB checkpoint style:
{
     "_id": "_local/edd1d4be175dd4c88b6ec1f07731d98e",
     "_rev": "0-6",
     "session_id": "6030857ab3a97de2a59238c2943da145",
     "source_last_seq": 8308,
     "replication_id_version": 3,
     "history": [{
         "session_id": "6030857ab3a97de2a59238c2943da145",
         "start_time": "Tue, 30 Jun 2015 14:27:08 GMT",
         "end_time": "Tue, 30 Jun 2015 14:27:28 GMT",
         "start_last_seq": 2304,
         "end_last_seq": 8308,
         "recorded_seq": 8308,
         "missing_checked": 6004,
         "missing_found": 6004,
         "docs_read": 6004,
         "docs_written": 0,
         "doc_write_failures": 6004
     }],
     "_revisions": {
         "start": 0,
         "ids": ["6"]
     }
 } 
// old PouchDB checkpoints look like this:
{
     "_id": "_local/edd1d4be175dd4c88b6ec1f07731d98e",
     "_rev": "0-6",
     "last_seq": 8308
}
```

This means that new PouchDB checkpoints are backwards compatible, which they wouldn't be if we used the CouchDB format. 

I've been testing this in the app where i found the original issue, and it has been solving the problem well there. 

Looking forward to any feedback on this!

Issue #3999

### DIFF
--- a/lib/replicate/checkpointer.js
+++ b/lib/replicate/checkpointer.js
@@ -50,7 +50,7 @@ function updateCheckpoint(db, id, checkpoint, session, returnValue) {
     return db.put(doc).catch(function (err) {
       if (err.status === 409) {
         // retry; someone is trying to write a checkpoint simultaneously
-        return updateCheckpoint(db, id, checkpoint, returnValue);
+        return updateCheckpoint(db, id, checkpoint, session, returnValue);
       }
       throw err;
     });
--- a/lib/replicate/checkpointer.js
+++ b/lib/replicate/checkpointer.js
@@ -6,6 +6,15 @@ var pouchCollate = require('pouchdb-collate');
 var collate = pouchCollate.collate;
 
 var REPLICATION_ID_VERSION = 2;
+// This is an arbitrary number to limit the
+// amount of replication history we save in the checkpoint.
+// If we save too much, the checkpoing docs will become very big,
+// if we save fewer, we'll run a greater risk of having to
+// read all the changes from 0 when checkpoint PUTs fail
+// CouchDB 2.0 has a more involved history pruning,
+// but let's go for the simple version for now.
+var CHECKPOINT_HISTORY_SIZE = 5;
+var LOWEST_SEQ = 0;
 
 function updateCheckpoint(db, id, checkpoint, session, returnValue) {
   return db.get(id).catch(function (err) {
@@ -38,9 +47,8 @@ function updateCheckpoint(db, id, checkpoint, session, returnValue) {
 
     // Just take the last pieces in history, to
     // avoid really big checkpoint docs.
-    // CouchDB 2.0 has a more involved history pruning,
-    // but let's go for the simple version for now.
-    doc.history = doc.history.slice(0, 5);
+    // see comment on history size above
+    doc.history = doc.history.slice(0, CHECKPOINT_HISTORY_SIZE);
 
     doc.replication_id_version = REPLICATION_ID_VERSION;
 
@@ -113,11 +121,11 @@ Checkpointer.prototype.getCheckpoint = function () {
   var self = this;
   return self.target.get(self.id).then(function (targetDoc) {
     return self.src.get(self.id).then(function (sourceDoc) {
-      // TODO: Discuss the 'return 0's here
-      // can we do something better?
+      // Since we can't migrate an old version doc to a new one
+      // (no session id), we just go with the lowest seq in this case
       if(targetDoc.replication_id_version !==
          sourceDoc.replication_id_version) {
-        return 0;
+        return LOWEST_SEQ;
       }
 
       var version;
@@ -127,24 +135,24 @@ Checkpointer.prototype.getCheckpoint = function () {
         version = "undefined";
       }
 
-      if(comparisons[version]) {
+      if(version in comparisons) {
         return comparisons[version](targetDoc, sourceDoc);
       }
 
-      return 0;
+      return LOWEST_SEQ;
     }, function (err) {
       if (err.status === 404 && targetDoc.last_seq) {
         return self.src.put({
           _id: self.id,
-          last_seq: 0
+          last_seq: LOWEST_SEQ
         }).then(function () {
-          return 0;
+          return LOWEST_SEQ;
         }, function (err) {
           if (err.status === 401) {
             self.readOnlySource = true;
             return targetDoc.last_seq;
           }
-          return 0;
+          return LOWEST_SEQ;
         });
       }
       throw err;
@@ -153,14 +161,12 @@ Checkpointer.prototype.getCheckpoint = function () {
     if (err.status !== 404) {
       throw err;
     }
-    return 0;
+    return LOWEST_SEQ;
   });
 };
 // This checkpoint comparison is ported from CouchDBs source
 // they come from here:
 // https://github.com/apache/couchdb-couch-replicator/blob/master/src/couch_replicator.erl#L863-L906
-
-var LOWEST_SEQ = 0;
 
 function compareReplicationLogs(srcDoc, tgtDoc) {
   if(srcDoc.session_id === tgtDoc.session_id) {
--- a/lib/replicate/checkpointer.js
+++ b/lib/replicate/checkpointer.js
@@ -35,7 +35,7 @@ function updateCheckpoint(db, id, checkpoint, session, returnValue) {
       return;
     }
     // Filter out current entry for this replication
-    doc.history = (doc.history || []).filter(function(item) {
+    doc.history = (doc.history || []).filter(function (item) {
       return item.session_id !== session;
     });
 
@@ -123,19 +123,19 @@ Checkpointer.prototype.getCheckpoint = function () {
     return self.src.get(self.id).then(function (sourceDoc) {
       // Since we can't migrate an old version doc to a new one
       // (no session id), we just go with the lowest seq in this case
-      if(targetDoc.replication_id_version !==
+      if (targetDoc.replication_id_version !==
          sourceDoc.replication_id_version) {
         return LOWEST_SEQ;
       }
 
       var version;
-      if(targetDoc.replication_id_version) {
+      if (targetDoc.replication_id_version) {
         version = targetDoc.replication_id_version.toString();
       } else {
         version = "undefined";
       }
 
-      if(version in comparisons) {
+      if (version in comparisons) {
         return comparisons[version](targetDoc, sourceDoc);
       }
 
@@ -168,8 +168,8 @@ Checkpointer.prototype.getCheckpoint = function () {
 // they come from here:
 // https://github.com/apache/couchdb-couch-replicator/blob/master/src/couch_replicator.erl#L863-L906
 
-function compareReplicationLogs(srcDoc, tgtDoc) {
-  if(srcDoc.session_id === tgtDoc.session_id) {
+function compareReplicationLogs (srcDoc, tgtDoc) {
+  if (srcDoc.session_id === tgtDoc.session_id) {
     return {
       last_seq: srcDoc.last_seq,
       history: srcDoc.history || []
@@ -181,7 +181,7 @@ function compareReplicationLogs(srcDoc, tgtDoc) {
   return compareReplicationHistory(sourceHistory, targetHistory);
 }
 
-function compareReplicationHistory(sourceHistory, targetHistory) {
+function compareReplicationHistory (sourceHistory, targetHistory) {
   // the erlang loop via function arguments is not so easy to repeat in JS
   // therefore, doing this as recursion
   var S = sourceHistory[0];
@@ -189,7 +189,7 @@ function compareReplicationHistory(sourceHistory, targetHistory) {
   var T = targetHistory[0];
   var targetRest = targetHistory.slice(1);
 
-  if(!S || targetHistory.length === 0) {
+  if (!S || targetHistory.length === 0) {
     return {
       last_seq: LOWEST_SEQ,
       history: []
@@ -197,7 +197,7 @@ function compareReplicationHistory(sourceHistory, targetHistory) {
   }
 
   var sourceId = S.session_id;
-  if(hasSessionId(sourceId, targetHistory)) {
+  if (hasSessionId(sourceId, targetHistory)) {
     return {
       last_seq: S.last_seq,
       history: sourceHistory
@@ -205,7 +205,7 @@ function compareReplicationHistory(sourceHistory, targetHistory) {
   }
 
   var targetId = T.session_id;
-  if(hasSessionId(targetId, sourceRest)) {
+  if (hasSessionId(targetId, sourceRest)) {
     return {
       last_seq: T.last_seq,
       history: targetRest
@@ -215,15 +215,15 @@ function compareReplicationHistory(sourceHistory, targetHistory) {
   return compareReplicationHistory(sourceRest, targetRest);
 }
 
-function hasSessionId(sessionId, history) {
+function hasSessionId (sessionId, history) {
   var props = history[0];
   var rest = history.slice(1);
 
-  if(!sessionId || history.length === 0) {
+  if (!sessionId || history.length === 0) {
     return false;
   }
 
-  if(sessionId === props.session_id) {
+  if (sessionId === props.session_id) {
     return true;
   }
 
--- a/lib/replicate/checkpointer.js
+++ b/lib/replicate/checkpointer.js
@@ -107,21 +107,23 @@ Checkpointer.prototype.getCheckpoint = function () {
   var self = this;
   return self.target.get(self.id).then(function (targetDoc) {
     return self.src.get(self.id).then(function (sourceDoc) {
-      debugger;
-      console.log('Got CheckPoints', targetDoc, sourceDoc);
       // TODO: Discuss the 'return 0's here
       // can we do something better?
       if(targetDoc.replication_id_version !== sourceDoc.replication_id_version) {
-        console.log('diff versions');
         return 0;
       }
 
-      if(comparisons[targetDoc.replication_id_version.toString()]) {
-        console.log('chosing comparinson function', targetDoc.replication_id_version.toString());
-        return comparisons[targetDoc.replication_id_version.toString()](targetDoc, sourceDoc);
+      var version;
+      if(targetDoc.replication_id_version) {
+        version = targetDoc.replication_id_version.toString();
+      } else {
+        version = "undefined";
       }
 
-      console.log('same version, no comparison');
+      if(comparisons[version]) {
+        return comparisons[version](targetDoc, sourceDoc);
+      }
+
       return 0;
     }, function (err) {
       console.log('caught err in sourceDoc', err);
--- a/lib/replicate/checkpointer.js
+++ b/lib/replicate/checkpointer.js
@@ -5,7 +5,8 @@ var explain404 = require('./../deps/explain404');
 var pouchCollate = require('pouchdb-collate');
 var collate = pouchCollate.collate;
 
-var REPLICATION_ID_VERSION = 2;
+var CHECKPOINT_VERSION = 1;
+var REPLICATOR = "pouchdb";
 // This is an arbitrary number to limit the
 // amount of replication history we save in the checkpoint.
 // If we save too much, the checkpoing docs will become very big,
@@ -26,7 +27,8 @@ function updateCheckpoint(db, id, checkpoint, session, returnValue) {
         session_id: session,
         _id: id,
         history: [],
-        replication_id_version: REPLICATION_ID_VERSION
+        replicator: REPLICATOR,
+        version: CHECKPOINT_VERSION
       };
     }
     throw err;
@@ -50,7 +52,8 @@ function updateCheckpoint(db, id, checkpoint, session, returnValue) {
     // see comment on history size above
     doc.history = doc.history.slice(0, CHECKPOINT_HISTORY_SIZE);
 
-    doc.replication_id_version = REPLICATION_ID_VERSION;
+    doc.version = CHECKPOINT_VERSION;
+    doc.replicator = REPLICATOR;
 
     doc.session_id = session;
     doc.last_seq = checkpoint;
@@ -111,7 +114,7 @@ var comparisons = {
 
     return 0;
   },
-  "2": function(targetDoc, sourceDoc) {
+  "1": function(targetDoc, sourceDoc) {
     // This is the comparison function ported from CouchDB
     return compareReplicationLogs(sourceDoc, targetDoc).last_seq;
   }
@@ -123,14 +126,14 @@ Checkpointer.prototype.getCheckpoint = function () {
     return self.src.get(self.id).then(function (sourceDoc) {
       // Since we can't migrate an old version doc to a new one
       // (no session id), we just go with the lowest seq in this case
-      if (targetDoc.replication_id_version !==
-         sourceDoc.replication_id_version) {
+      if (targetDoc.version !==
+         sourceDoc.version) {
         return LOWEST_SEQ;
       }
 
       var version;
-      if (targetDoc.replication_id_version) {
-        version = targetDoc.replication_id_version.toString();
+      if (targetDoc.version) {
+        version = targetDoc.version.toString();
       } else {
         version = "undefined";
       }
--- a/lib/replicate/checkpointer.js
+++ b/lib/replicate/checkpointer.js
@@ -5,20 +5,44 @@ var explain404 = require('./../deps/explain404');
 var pouchCollate = require('pouchdb-collate');
 var collate = pouchCollate.collate;
 
-function updateCheckpoint(db, id, checkpoint, returnValue) {
+function updateCheckpoint(db, id, checkpoint, session, returnValue) {
   return db.get(id).catch(function (err) {
     if (err.status === 404) {
       if (db.type() === 'http') {
         explain404('PouchDB is just checking if a remote checkpoint exists.');
       }
-      return {_id: id};
+      return {
+        session_id: session,
+        _id: id,
+        history: [],
+        replication_id_version: 2
+      };
     }
     throw err;
   }).then(function (doc) {
     if (returnValue.cancelled) {
       return;
     }
+    // Filter out current entry for this replication
+    doc.history = (doc.history || []).filter(function(item) {
+      return item.session_id !== session;
+    });
+
+    // Add the latest checkpoint to history
+    doc.history.unshift({
+      last_seq: checkpoint,
+      session_id: session
+    });
+
+    // Just take the last pieces in history, to
+    // avoid really big checkpoint docs.
+    // CouchDB 2.0 has a more involved history pruning,
+    // but let's go for the simple version for now.
+    doc.history = doc.history.slice(0, 5);
+
+    doc.session_id = session;
     doc.last_seq = checkpoint;
+
     return db.put(doc).catch(function (err) {
       if (err.status === 409) {
         // retry; someone is trying to write a checkpoint simultaneously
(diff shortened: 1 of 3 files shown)


karlwestin:
seems like i need to look into those test failures first!

lib/replicate/checkpointer.js
```diff
@@ -5,24 +5,60 @@ var explain404 = require('./../deps/explain404');
 var pouchCollate = require('pouchdb-collate');
 var collate = pouchCollate.collate;
 
-function updateCheckpoint(db, id, checkpoint, returnValue) {
+var REPLICATION_ID_VERSION = 2;
```
kxepal: Does PouchDB has any analogue of CouchDB instance UUID? If it has, then better raise version to 3 which is current one.

lib/replicate/checkpointer.js
```diff
@@ -5,24 +5,60 @@ var explain404 = require('./../deps/explain404');
...
+// This is an arbitrary number to limit the
+// amount of replication history we save in the checkpoint.
+// If we save too much, the checkpoing docs will become very big,
+// if we save fewer, we'll run a greater risk of having to
+// read all the changes from 0 when checkpoint PUTs fail
+// CouchDB 2.0 has a more involved history pruning,
+// but let's go for the simple version for now.
+var CHECKPOINT_HISTORY_SIZE = 5;
```
kxepal: CouchDB keeps history size up to [50 records](https://github.com/apache/couchdb-couch-replicator/blob/master/src/couch_replicator.erl#L762-L766). 5 is too small value especially assuming that replication ID is based on remote host URL (if it's really true IDv2) which has higher collision risks.

lib/replicate/checkpointer.js
```diff
@@ -93,8 +161,74 @@ Checkpointer.prototype.getCheckpoint = function () {
...
+    return {
+      last_seq: LOWEST_SEQ,
+      history: []
+    };
+  }
+
+  var sourceId = S.session_id;
+  if(hasSessionId(sourceId, targetHistory)) {
+    return {
+      last_seq: S.last_seq,
```
kxepal: Here and below [`recorded_seq`](https://github.com/apache/couchdb-couch-replicator/blob/master/src/couch_replicator.erl#L891) should be used instead of `last_seq`.

lib/replicate/checkpointer.js
```diff
@@ -93,8 +161,74 @@ Checkpointer.prototype.getCheckpoint = function () {
...
   });
 };
+// This checkpoint comparison is ported from CouchDBs source
+// they come from here:
+// https://github.com/apache/couchdb-couch-replicator/blob/master/src/couch_replicator.erl#L863-L906
+
+function compareReplicationLogs(srcDoc, tgtDoc) {
+  if(srcDoc.session_id === tgtDoc.session_id) {
+    return {
+      last_seq: srcDoc.last_seq,
```
kxepal: In checkpoint document root [`source_last_seq`](https://github.com/apache/couchdb-couch-replicator/blob/master/src/couch_replicator.erl#L871) field is used, not `last_seq`.

daleharvey:
Apologies, I wanted to get https://github.com/pouchdb/pouchdb/issues/4002 fixed before this landed so we make sure the tests pass in master too, the fix for that should land upstream shortly and will take a look at this tomorrow

lib/replicate/checkpointer.js
```diff
@@ -5,24 +5,60 @@ var explain404 = require('./../deps/explain404');
...
+        replication_id_version: REPLICATION_ID_VERSION
+      };
     }
     throw err;
   }).then(function (doc) {
     if (returnValue.cancelled) {
       return;
     }
+    // Filter out current entry for this replication
+    doc.history = (doc.history || []).filter(function(item) {
```
daleharvey: Small nit, we generally put spaces here (need to get a linter to start checking for that)

     doc.history = (doc.history || []).filter(function (item) {

lib/replicate/checkpointer.js
```diff
@@ -64,27 +102,57 @@ Checkpointer.prototype.updateSource = function (checkpoint) {
...
+
 Checkpointer.prototype.getCheckpoint = function () {
   var self = this;
   return self.target.get(self.id).then(function (targetDoc) {
     return self.src.get(self.id).then(function (sourceDoc) {
-      if (collate(targetDoc.last_seq, sourceDoc.last_seq) === 0) {
-        return sourceDoc.last_seq;
+      // Since we can't migrate an old version doc to a new one
+      // (no session id), we just go with the lowest seq in this case
+      if(targetDoc.replication_id_version !==
```
daleharvey: and same here

    if (targetDoc.replication_id_version !==

daleharvey:
@karlwestin Where is the `session_id` generated? may be being dumb but cant find it here, does the `session_id` refer to a single replication instance?

The main reason for the check that both the remote and local databases agree on the checkpoint is for the case that the source database is deleted (or the target in a push replication) in that case we do want to start from 0.

This does seem like it would work, I am wondering if we need to the couch algorithm in its entirety though, it seems like loosening the restriction on our comparison would be far simpler and just as effective

```
// Happy path, both wrote same seq
if (source.seq === target.seq) {
   return seq

// In these cases we can lose track of a dropped database, but not much way around that
if (!targetCheckpointExists && targetReadOnly) 
  return source.seq
if (!sourceCheckpointExists && sourceReadOnly) 
  return target.seq

// If either database has not got a checkpoint, then it was likely dropped, its possible replication 
// failed to write a checkpoint on the first batch, but even if so we arent losing a lot
if (!sourceCheckpointExists || !targetCheckpointExists) 
  return 0

// If we try to write a checkpoint it is because we have confirmed writing up until that seq, so we 
// should be able to safely pick the largest existing seq between the databases
return Math.max(source.seq, target.seq);
```

Can anyone bring up any drawbacks on this approach? (ping @kxepal, @kocolosk et al)

kxepal:
@daleharvey session_id is just [random UUID4](https://github.com/apache/couchdb-couch-replicator/blob/master/src/couch_replicator.erl#L629) which is used for a single replication session.

```source.seq === target.seq```  are rarely would be in sync since database update seq also increments even for non-replicable changes (like security or local docs updates). 

```
// In these cases we can lose track of a dropped database, but not much way around that
if (!targetCheckpointExists && targetReadOnly) 
  return source.seq
if (!sourceCheckpointExists && sourceReadOnly) 
  return target.seq
```

This assumes that evil user may update/remove checkpoint only on the one side to let some documents be not replicated.

```
// If we try to write a checkpoint it is because we have confirmed writing up 
// until that seq, so we should be able to safely pick the largest existing seq 
// between the databases
return Math.max(source.seq, target.seq);
```

seq may be not numerical for such comparison: for CouchDB 2.0 it's a string or array, depending on the case.

**Comment by daleharvey** [comment]
@kxepal I was referring to `session_id` in this PR, we dont have any expectations of interoperation between couchdb replication and pouchdb replication, they are seperate replicators (ie if you replicate A <> B with CouchDB, replicating A <> B with PouchDB will run entirely seperately), we shouldnt see or understand the replication checkpoint format from couchdb's replicator at all in code, just a useful reference for discussing this

> This assumes that evil user may update/remove checkpoint only on the one side to 
> let some documents be not replicated.

An evil user that has access to the database has access to the database, not entirely sure what we can prevent in that case.

> seq may be not numerical for such comparison: for CouchDB 2.0 it's a string 
> or array, depending on the case.

Sorry yup that was just psuedo code, it would be something along the lines of

    return collate(source.seq, target.seq) === 1 ? source.seq  : target.seq;

File: lib/replicate/replicate.js [hunk 7]
```diff
@@ -30,6 +30,8 @@ function replicate(src, target, opts, returnValue, result) {
   var checkpointer;
   var allErrors = [];
   var changedDocs = [];
+  // Like couchdb, every replication gets a unique session id
+  var session = utils.uuid();
```
kxepal: cc @daleharvey

daleharvey: Doh dunno how I missed that, thanks

kxepal:
@daleharvey 
While I understand and agree with interoperation point, I don't see any reasons to avoid reuse CouchDB checkpoint structure for PouchDB, even since it requires no cost. Benefits of such simple solution are great: in CouchDB 2.0 you may browse all _local documents as like regular ones and administrators may finally have some interface to inspect and manager recorded on server checkpoints. With single format there would be no worries about which fields each implementation uses.

>> This assumes that evil user may update/remove checkpoint only on the one side to let some documents be not replicated.
>
> An evil user that has access to the database has access to the database, not entirely sure what we can prevent in that case, what threat model are you thinking about?

I was not about threat, but about the case when replication will go one for inconsistent state of source and target. CouchDB replication prevents such cases and also monitors if source or target peers were restarted (this moment reflects in instance_start_time field) and if it happens it restarts replication to ensure that instances are still the same and replication may be continued.

Otherwise it's not a replication, but just a data moving from point A to point B.

daleharvey:
> source.seq === target.seq are rarely would be in sync since database update 
> seq also increments even for non-replicable changes 
> (like security or local docs updates). 

The `seq` in this case refers to the `last_seq` we fetched in changes during replication, ie this check is just checking whether the source and target agree on the last checkpoint the replicator managed to write, it will be the default case aside from 1. failures to write the checkpoint, 2. dropped databases

>  I don't see any reasons to avoid reuse CouchDB checkpoint structure for PouchDB

Mostly for complexity purposes, if we can store less, if we can learn from the evolution CouchDB had to go through and get closer to the simpler end solution then its a bonus for us. It also means less maintenance to keep compatibility with CouchDB where it isnt required and there is the issue with `uncanny valley` where things may look very similiar but be used differently

> I was not about threat, but about the case when replication will go one for inconsistent 
> state of source and target. CouchDB replication prevents such cases and also monitors 
> if source or target peers were restarted (this moment reflects in instance_start_time field) 
> and if it happens it restarts replication to ensure that instances are still the same and 
> replication may be continued.

So those 2 clauses are only for the case where we are replicating to / from a read only database, its never had a checkpoint written to it, it should do the right thing in every case except when a database gets dropped, how does couchdb guard against that?

kxepal:
>> source.seq === target.seq are rarely would be in sync since database update
>> seq also increments even for non-replicable changes
>> (like security or local docs updates).
>
> The seq in this case refers to the last_seq we fetched in changes during replication, ie this check is just checking whether the source and target agree on the last checkpoint the replicator managed to write, it will be the default case aside from 1. failures to write the checkpoint, 2. dropped databases

This case assumes that:
1) Replication goes only from source to target
2) There are no any updates made on target ever
3) No other replications are ever run with target

So, it may works, but it's very limited use case and easy to break with the next situation:
1) Source replicated to Target. last_seq on both is 10
2) Target db updated it security - it has now last_seq 11
3) Source updated some doc - it has now last_seq 11
4) Replication won't replicate anything because both source and target are agree on their seq.

>> I don't see any reasons to avoid reuse CouchDB checkpoint structure for PouchDB
>
> Mostly for complexity purposes, if we can store less, if we can learn from the evolution CouchDB had to go through and get closer to the simpler end solution then its a bonus for us. It also means less maintenance to keep compatibility with CouchDB where it isnt required and there is the issue with "uncanny valley" where things may look very similiar but be used differently

Well, complexity isn't a case here. If we speak about this PR changes, then it's about rename two fields which are used for the same purpose. Storing replication stats in checkpoints is nice, but optional, though useful, feature indeed - they aren't used by replicator anywhere.

> So those 2 clauses are only for the case where we are replicating to / from a read only database, its never had a checkpoint written to it, it should do the right thing in every case except when a database gets dropped, how does couchdb guard against that?

Nohow: without checkpoints (`use_checkpoints: false` replication option) it will always run full replication all the time. It's hard to say for sure if you can run incremental replication for such situation, but false assumption may leads to have some data not replicated.

daleharvey:
> So, it may works, but it's very limited use case and easy to break with the next situation

Not sure you are understanding what this comparison does, will try to clarify

    1. User writes docs x,y to A
    2. We replicate from A to B
    3. During replication we get a changes feed with `{results: [{seq: 1, _id: x}, {seq: 2, _id: y}], last_seq: 2}`
    4. We successfully write those 2 documents, the replicator considered 2 as the last_seq that was successfully written during replication
    5. We tell both A and B that 2 is the last seq that was written

When the replication starts again, it fetches the checkpoint information from A and B, it does not check A or B.info or what their current update_seq is, only what the replicator has stored as the last written seq. The only way those checkpoints should not match is if the database is dropped, one of the databases is read only or if we failed to write one of the checkpoints

kxepal:
@daleharvey Sorry, I was thought you're about read-only source/target and `source` and `target` in code referenced to their db info.

Actually, then, this is how CouchDB works and this PR [implements](https://github.com/pouchdb/pouchdb/pull/4035/files#diff-bdbe5c3aac3d6d14af4b502a1b035595R172) that case, though via session_id comparison (which also ensures that the last checkpoint was made for the same replication). Otherwise it fallbacks to history comparison where it seeks the match of the recorded_seq.

daleharvey:
So I am still not certain why we need history tracking and why picking the largest of the 2 checkpoints would not be sufficient.

if 2 checkpoints exist then we know that:

  1. It was the same replication, since thats the point of the replication id generation
  2. Neither databases have been dropped
  3. The replicator has written up until at least the highest seq, we could pick the smallest and that would also fix the problem, but it seems like it would be safe to pick the largest.

It is possible / likely I am missing something and there is a reason that couchdb has suck extensive history tracking, but I havent seen any reasoning for it yet, anyone able to explain why this wont work?

kxepal:
@daleharvey Assume one of the peers is restored from backup. It has the checkpoint, but it's outdated from comparison with the other peer. 

Also Replication ID has a room for collisions. It's not an unique hash for all time (especially because of md5 nature). Older than 3 versions hit it with better chances. Even now with v3 you can hit it now with using view as a filter.

karlwestin:
eHealthAfrica/sense-id-command#73

kxepal:
@karlwestin private repo? 404 there.

karlwestin:
yeah sorry that was a link to a private repo, i've been out of the loop for a few days, trying to catch up with your conversation right now!

daleharvey:
@kxepal Could you explain the issue with restoring from backup? it seems like that should work in this case too

As for collisions, do you mean 2 seperate replications that use the same id? if so then it seems like a `session_id` mechanism doesnt do anything to prevent that since its randomly generated, if a replication picks up this checkpoint document (via its id) then its going to use the history to find a checksum, if we can get different replications using the same id then that seems like a bug that needs to be fixed in the generation of the replication id

daleharvey:
@karlwestin Will try and get a PR up if I get time, but wondering if 
   
    return collate(targetDoc.last_seq, sourceDoc.last_seq) ? targetDoc.last_seq : sourceDoc.last_seq;

would fix the problem, pass your tests

kxepal:
@daleharvey Assume Source and Target shares the following checkpoint:

```
{
    "_id": "_local/b3e44b920ee2951cb2e123b63044427a",
    "_rev": "0-8",
    "history": [
        {
            "doc_write_failures": 0,
            "docs_read": 2,
            "docs_written": 2,
            "end_last_seq": 5,
            "end_time": "Thu, 10 Oct 2013 05:56:38 GMT",
            "missing_checked": 2,
            "missing_found": 2,
            "recorded_seq": 5,
            "session_id": "d5a34cbbdafa70e0db5cb57d02a6b955",
            "start_last_seq": 3,
            "start_time": "Thu, 10 Oct 2013 05:56:38 GMT"
        },
        {
            "doc_write_failures": 0,
            "docs_read": 1,
            "docs_written": 1,
            "end_last_seq": 3,
            "end_time": "Thu, 10 Oct 2013 05:56:12 GMT",
            "missing_checked": 1,
            "missing_found": 1,
            "recorded_seq": 3,
            "session_id": "11a79cdae1719c362e9857cd1ddff09d",
            "start_last_seq": 2,
            "start_time": "Thu, 10 Oct 2013 05:56:12 GMT"
        },
        {
            "doc_write_failures": 0,
            "docs_read": 2,
            "docs_written": 2,
            "end_last_seq": 2,
            "end_time": "Thu, 10 Oct 2013 05:56:04 GMT",
            "missing_checked": 2,
            "missing_found": 2,
            "recorded_seq": 2,
            "session_id": "77cdf93cde05f15fcb710f320c37c155",
            "start_last_seq": 0,
            "start_time": "Thu, 10 Oct 2013 05:56:04 GMT"
        }
    ],
    "replication_id_version": 3,
    "session_id": "d5a34cbbdafa70e0db5cb57d02a6b955",
    "source_last_seq": 5
}
```

Then Target becomes restores from backup. Now it has the same checkpoint in the following state:

```
{
    "_id": "_local/b3e44b920ee2951cb2e123b63044427a",
    "_rev": "0-2",
    "history": [
        {
            "doc_write_failures": 0,
            "docs_read": 2,
            "docs_written": 2,
            "end_last_seq": 2,
            "end_time": "Thu, 10 Oct 2013 05:56:04 GMT",
            "missing_checked": 2,
            "missing_found": 2,
            "recorded_seq": 2,
            "session_id": "77cdf93cde05f15fcb710f320c37c155",
            "start_last_seq": 0,
            "start_time": "Thu, 10 Oct 2013 05:56:04 GMT"
        }
    ],
    "replication_id_version": 3,
    "session_id": "77cdf93cde05f15fcb710f320c37c155",
    "source_last_seq": 2
}
```

What will CouchDB do here:
1. Check session_id match at the document root - MISMATCH
2. Pick first element of Source history and seek session_id match in the Target history log - MISMATCH
3. Now try pick first element of Target history log and seek for session_id match in Source log - MATCH
4. Start replication from update_seq=2

What will happens if there would be no history as you propose:
1. Fetch Source and Target checkpoints and [_pick the largest_ seq_](https://github.com/pouchdb/pouchdb/pull/4035#issuecomment-120758451)
2. Start replication from update_seq=5 - WRONG

Collision is another sort of matter: replication id is a hash and every hash has a collisions - what you or not. It's implementation responsibility to reduce their chances as much as possible. That's why there were existed versions of replication ids: 1, 2, now 3. Old versions was used generation algorithms that have higher collision rates (like they didn't distinguish multiple replicators on the same host). But even with UUID there is a collision chance if source or target was replaced by some other, completely different, instance - Replication ID would be the same, but it's completely unrelated to what Source has. To solve that problem, random session_id is used: it's guarantee to be unique globally enough to negotiate peer replacement edge case issue.

So, in fact, all this "complexity" is about solving edge cases which eventually may happens and you'll be not happy if there would be no mechanism to handle them.

daleharvey:
Ok thats a great description, thanks a lot

@karlwestin apologies for the sidetrack, will take a look at this PR as it is now (and see whats up with pouchdb-server etc)

daleharvey:
Nice all green :)

Weirdly the error with pouchdb-express-router seems to pass if I test GREP=3999 but if I GREP=test.replication.js it fails, it sounds like something is leaking out of the test.

Looking over the tests, do you think it would be simpler if instead of faking the checkpoint return, we just refused to write the remote checkpoint during the first replication? In general we have found the tests to be more reliable and less likely to break if we test closest to the actual condition as we can

daleharvey:
ok, so the main issue with tests intermittently failing is that we are calling `done()` inside the mocked `get` that is happening during replication

We need to make sure all operations / replications are completely finished before calling done / finishing the test, will add a patch here for that

daleharvey:
@karlwestin ok I added a commit @ https://github.com/pouchdb/pouchdb/pull/4048 that should hopefully make the tests more reliable, they are passing locally in pouchdb-express-router for me.

The first test I refactored a bit more towards our general style, space after `function` / `if` and brackets, try promises over nested callbacks, early returns and avoid try catching. How do those changes look to you?

karlwestin:
@daleharvey thank you so much for fixing those changes! I actually didn't know i could return the promise for the test, i really like this! I rewrote the other tests to the same style as well. Its getting late here now, but looking forward to see the test matrix tomorrow, i tried pouchdb-server, pouchdb-express-router and a few others locallay and they all looked good

cheers!

daleharvey:
So the failure in couchdb-master is genuine, but the rest is looking nice and green :) ill try taking a look at couchdb master soon

kxepal:
@karlwestin could you please check my commentaries on PR code?

kxepal:
@daleharvey Is it possible to throw last 100 lines of CouchDB log to stderr in case of test failure for this target?  Should help to guess what's wrong if it was on CouchDB side.

karlwestin:
@kxepal hi!
It seems from your comments that you want the checkpoint to be exactly like the couchdb checkpoint format. I agree right now they're in the 'uncanny valley' and i would rather make it look less like a couchdb checkpoint than more. The reason for this is to support downgrading from this format to the old checkpoint format. I don't see so much value in having the exact same format as couchdb, since it seems to me those checkpoints will not be used interchangingly.

My suggested solutions is to change the replication_id_version property to something else, so we do not imply CouchDB similarity, and to increase the history size, maybe not all the way to 50 though (i believe theres a value in keeping checkpoint docs small). 

Does this sound completely off to you?

kxepal:
@karlwestin Hi (:

My main concern here is keep things in order: if PouchDB is going to implement CouchDB like checkpoints, then it shouldn't leave this work half-baked. I don't much follow 'uncanny valley' reason because there is no underwater stones: structure is pretty clear and PouchDB may reuse it with full power.

If PouchDB would like to have own checkpoint format - fine, but lets somehow make this moment explicitly. For instance, by adding few more fields to the document like:
```
{
  "replicator": "PouchDB",
  "version": 1
}
```
- simple change, that will simplify life a lot in future for everyone. I think, it's good to have similar change for CouchDB as well if we're going to change checkpoints format.

The main problem here is not about checkpoint interchangeability, but about introspection and processing checkpoints data. In CouchDB 2.0 it is possible to view all _local documents and very first idea I came was: "Oh, finally I can make a page to browse and manage all checkpoints on my server". But to fulfil such ambition, these checkpoints have to share same format or provide a way to distinguish one schema from the other. In the light of these changes, PouchDB checkpoints starts to looks as like as CouchDB ones, but not exactly like.

Changing replication_id is not a good idea since it's used not about document schema versioning, but about algorithm to use for replication id generation.

Or my worries are groundless? What do you think? cc @daleharvey

daleharvey:
```
{
  "replicator_id": "pouchdb",
  "version": 1
}
```

Looks like a nice idea, we can add a check to ensure we have the expected replicator_id and it makes it fairly clear what the distinction is, any checkpoint viewer thing can detect what replicator it was generated with etc. We definitely shouldnt use `replication_id_version: 3` since that indicates a level of compatibility that doesnt exist. 

5 seems like a totally reasonable number for history, I would worry about raising it at all, we already generate a lot of traffic with checkpoints in pouchdb and pouchdb users care about the traffic they generate more than couchdb in general, I think we leave it at 5 and then if it becomes an issue we can make it configurable

karlwestin:
@kxepal so then the suggested format would be like this?
```
// Checkpoint format used in this PR:
{
     "_id": "_local/edd1d4be175dd4c88b6ec1f07731d98e",
     "_rev": "0-6",
     "session_id": "6030857ab3a97de2a59238c2943da145",
     "last_seq": 8308,
     "replicator_id": "PouchDB",
     "version": 1,
     "history": [{
         "session_id": "6030857ab3a97de2a59238c2943da145",
         "last_seq": 8308
     }]
}
```

I would be very happy to change it into that! I joined the IRC channel too if you'd rather discuss there!

daleharvey:
The problem with couchdb-master is http://127.0.0.1:15984/test_repl_remote/_changes?timeout=25000&style=all_docs&since=2&limit=100 fires 

```
2015-07-15 15:28:44.865 [debug] node1@127.0.0.1 <0.2204.0> httpd 400 error response:
 {"error":"bad_request","reason":"Malformed sequence supplied in 'since' parameter."}
```

kxepal:
@karlwestin Yes, something like that. Though assigning software name to `replicator_id` looks strange for me, but may be why not? 

@daleharvey quite reasonable failure since sequence format had changed.

daleharvey:
So yup, looked into this a little, `3999-1` changes the last_seq in the `.get` to `_local/etc`, the format it returns confuses couchdb-master, as I mentioned before I think these types of bugs are best fixed if instead of malforming the checkpoint as we read it, we have a test that does not write one of the checkpoints

karlwestin:
Ok i m not really sure how that test would work then:

This test would then set a batch size of 1, and write 2 docs, then replicate.
Then we'd stub out remote.put instead of remote.get, and make that fail on the 2nd checkpoint write.
The replication would then fail, and we'd have different state on the different sides.
When we pick up the next replication, we'd test through the changes feed that we still not start at 0

I could take a shot at doing that, just need a little more time :)

daleharvey:
Yeh its a little more hassle, but its closer to the error condition we are accounting for, I would have been ok with merged as is but if we already need to fix it for couch-master, may as well fix it properly. If you dont get to it tonight I may give it a shot, thanks loads for all these changes, its not an easy PR

kxepal:
I wonder how could you confuse couch-master with last_seq if last_seq always references to source and eventually have to satisfy it format...

karlwestin:
I confuse couchdb by making up my own 'last_seq' in that test https://github.com/karlwestin/pouchdb/blob/fix-3999/tests/integration/test.replication.js#L2549
gonna try 'the failed checkpoint put' instead as a method to create a mismatch

kxepal:
Ah, I see. How about to request database info, pick the update_seq and fake the number? Should be quite generic solution.

daleharvey:
Heh couchdb-master is always the pickest when it comes to changes assertions, but the new tests looks great :) ill take a look at how much needs changed to get couchdb master working, this is looking just about landable

tests/integration/test.replication.js
```diff
@@ -2519,6 +2519,326 @@ adapters.forEach(function (adapters) {
...
+        });
+      };
+
+      // 2. We measure that the replication starts in the expected
+      // place in the 'changes' function
+      var changes = source.changes;
+      source.changes = function (opts) {
+
+        if (mismatch) {
+          opts.since.should.be.at.least(1);
```
daleharvey: We cant use numerical comparisons for couchdb-master checkpoints since they are arrays / objects, we can use `collate` to compare them, although in this case I think 

    opts.since.should.not.equal(0);

is fine

tests/integration/test.replication.js
```diff
@@ -2519,6 +2519,326 @@ adapters.forEach(function (adapters) {
...
+         // Verify that we have old checkpoints:
+         should.exist(checkpoint);
+         var target = new PouchDB(dbs.name);
+         return PouchDB.utils.Promise.all([
+           target.get(checkpoint),
+           source.get(checkpoint)
+         ]);
+       }).then(function (res) {
+        // [0] = target checkpoint, [1] = source checkpoint
+        res[0].last_seq.should.equal(res[1].last_seq);
```
karlwestin: @daleharvey the `opts.since.should.not.equal(0)` trick works fine when i do it locally! thanks for that!
i'd like to do a `collate(res[0].last_seq, res[1].last_seq)` here, but i don;t know how to access pouchdb-collate so it works both on node and browser... do you have any trick for that?

karlwestin: thinking again about it, maybe its overkill to check last_seq there, should be enough with the check in changes that the replication doesn't start at 0 :)

karlwestin:
Yeah that's for sure the most challenging one! But now it seems we got greeeeeeeen!

daleharvey:
Woot, thanks so much for all the hard work @karlwestin, this is awesome. Thanks loads for the helpful comments @kxepal as well, made this much better

This is a really nice feature to get in, https://github.com/pouchdb/pouchdb/commit/9de0faa21378812e89b9c82dc1b6950f9868bcd2

karlwestin:
Yay! thanks a lot @daleharvey and @kxepal for being so helpful to a new contributor! You rock!

kxepal:
Hooray! (:

nickcolley:
This is epic :tada: :+1:

Issue #1526, opened 2014-02-26
make sure all adapters use AbstractPouch.prototype.put/post


calvinmetcalf:
mainly for the http pouch, whats currently in AbstractPouch.prototype.put/post should got into AbstractPouch.prototype._put/_post and AbstractPouch.prototype.put/post should just call those, this would allow us to put all the option parsing in the same place and allow us to do stuff like pre update hooks, etc.  we'd need to do the same thing for httpPouch.get as well

nolanlawson:
fixed by ef0944f72429915c7248689cc6950895de78982e

PR #4039, opened 2015-07-09
(#1526) - Reuse generic adapter code in http adapter

daleharvey:


daleharvey:
This is one for Nolan since this will remove the optimisations of using multipart requests for attachments, however one of the main drivers for using multipart requests was sync with couchbase gateway and we do not go through the `.put` during replication.

The only people who would have got benefit from this were people using pouchdb as a http client writing documents with a lot of attachments manually. Given we have to fix performance for attachments during replication and this reduces a whole ton of duplicate code, I think we should merge this then deal with optimising attachments via replication (possibly copy the behaviour over to bulkDocs)

daleharvey:
Would need fixes for pouchdb-server etc, but will wait for consensus before digging into those

nolanlawson:
Why are we removing multipart support? It's not just for CSG; it's also just a good optimization all around.

daleharvey:
The main driver was to reduce duplicated code across the adapters, I would be very surprised if the argument handling didnt have bugs or slightly different behaviours around https://github.com/pouchdb/pouchdb/pull/4039/files#diff-bb99ab1c6caa56f1860b6016240dbabaL570, every other place we use bulkDocs alone to deal with write operations on the database and it would be nice to be consistent here.

I cant remember if it came up during review, but I dont think we should have vastly different code paths for `db.put(docWithAttachments)` vs `db.bulkDocs([docWithAttachment])`, I cant see how this can affect csg / replication since we only ever call `bulkDocs` in the replicator, this only affects people using pouchdb as a http client doing manual `put`s right? as I mentioned earlier I think we should fix up this PR and then look at implementing the optimisation for bulkdocs.

nolanlawson:
Oh I see, so the issue is that I only ever used `multipart/related` for `put`, whereas it's just as appropriate for `bulkDocs` and `post`. Plus, the HTTP adapter can just use `_bulk_docs` everywhere instead of differentiating between put/post/bulkDocs. That makes sense.

Can we just go ahead right now and use multipart for bulkDocs, then? Otherwise that `multipart.js` file is just dead code anyway, since nothing is using it.

nolanlawson:
BTW this is failing in the pouchdb-server tests because apparently we never tried to handle the case of local docs inside of `_bulk_docs`:

    TypeError: Cannot read property 'should' of undefined[0m[90m
      at /home/travis/build/pouchdb/pouchdb/tests/integration/test.local_docs.js:22:15

daleharvey:
Yeh I think multipart should be implemented bulkDocs, I think the optimisation is worth it however I am worried about its risks, it changes the error handling semantics we have had up until now, right now if we fail to write a document we fail to write a document, but with this method we can partially fail to write which is very different, we should be very sure at the least that this plays nicely with replication

nolanlawson:
I'm not sure that's true that we can partially write if we use `multipart/related`. It's still a single HTTP request.

Can someone from CouchDB confirm? @kxepal?

kxepal:
@daleharvey Multipart for /db/_bulk_docs is indeed desirable, but it raises question of content-lenght-less implementation which causes chain of another issues. Though, it still the right direction.
Another idea is to receive stream of JSON docs line by line ala reversed changes feed.

@nolanlawson Yes, partially write isn't possible since there is no intermediate storage for such requests. But that's also the right direction as it will helps a lot with half-replicated document with a lot of attachments - we can resume such cases from the middle.

PR #4044, opened 2015-07-11
(#4002) - Disable large doc_ids test

daleharvey:


daleharvey:
Thisll get fixed upstream

Issue #2971, opened 2014-11-11
make default test run against pouchdb-server


daleharvey:
Just to make it easier for contributors, if they arent dealing with bugs that involve couch vs pouchdb behaviour, then having git clone && npm test work out the box would be nice

daleharvey:
Dropping although I may revisit

PR #4050, opened 2015-07-17
(#2971) - Default test run against pouchdb-express-router

daleharvey:


daleharvey:
After watching various people try to get the test suite up and running, I think this may vastly increase how easy it is to get started with contributions, we no longer need a caveat for installing couchdb etc and the new flow to get the test suite running is:

    $ npm install
    $ npm test

The downsides may possibly be, when developing features that are aimed very much towards couchdb compatibility that people may assume they have a working patch and then see that it fails against couchdb in CI, we already have that between couchdb and couchdb-master, plus the vast majority of the work done on pouch isnt directly working on couchdb compatibility. The other issue is when developing feature that arent currently supported by pouchdb-express-router and require patches for them, in that case I am happy to take responsibility to get them working.

daleharvey:
Basically I would really like us to get to 

    $ git clone https://github.com/pouchdb/pouchdb.git
    $ cd pouchdb
    $ npm install 
    $ npm test

To get the test suite running, this seems like the lowest easiest way

nickcolley:
I like this a lot :+1:

nolanlawson:
Sorry to be that guy, but -1. I vastly prefer testing against CouchDB as a default, since otherwise we are potentially chasing our own tail. Yes it's a barrier to entry for new contributors, but `brew install couchdb` or `sudo apt-get install couchdb` is really not a lot to ask for most developers.

Additionally I have to admit that I've always been on the fence about pouchdb-express-router. It's a side project that I think only you (Dale) have really been using, we never managed to integrate it into express-pouchdb because @marten-de-vries came up with the `minimum-for-pouchdb` solution instead (which I think is fine), and up to now every time there's a bug in express-pouchdb, I end up having to fix it in two places, which is sub-optimal.

If anything, I would rather we made pouchdb-server the default (since people are actually using that), but then again we have the tail-chasing problem.

daleharvey:
In terms of chasing our own tail, we are testing CouchDB in travis by default, I dont believe it will change anything in terms of merging things without couchdb compatibility.

As for making it easier, `apt-get install couchdb` installs couchdb 1.4, I have had a lot of troubles with `brew install couchdb` previously, I opened this bug originally after watching 10 people try to get the test suite running and almost all having troubles at the couch install step (as I have done in the past too), The vast majority of bugs we have open dont require running the tests against a CouchDB instance and so I think the tradeoff is easier for pouchdb devs (we already have couch installed etc) vs easier for new contributors

And with `pouchdb-server` vs `pouchdb-express-router`, I think pouchdb-server has been built the wrong way and its complexity is already showing, pouchdb-express-router is around 500 lines and conceptually extremely simple, it is a http router for pouchdb function calls whereas pouchdb-server is a couchdb-like full app platform. (Which I and many others already fundamentally disagree with, hence the couchapp discussions)

However pouchdb-server can work for now, how about if we fix pouchdb-server to not spit out a whole ton of logs to stdio by default (so tests dont look like http://pastebin.com/VVL6MCWM) and then use pouchdb-server, if it proves to be problematic and those problems are easier fixed by pouchdb-express-router then we can do that later

Hows that sound?

nolanlawson:
Good points; I can agree with this approach. Let's fix the pouchdb-server logs and then use that as the default.

As for pouchdb-express-router: if pouchdb-server is too complicated (which may well be true) and you want to continue to maintain pouchdb-express-router, could you please at least move it to the pouchdb org and give me, calvin, and marten npm owner permissions? I'm still not really a big fan, especially given that we don't advertise pouchdb-express-router anywhere on the site (which is why nobody knows about it), but I can see the attraction of a lite node server. If we're going to own it, though, then we should own it.

Also this PR would be good to fit into the 4.0 release, because I may have to update the [PouchDB for *down authors](https://gist.github.com/nolanlawson/f2f4d7e2e6f426502b94) gist, although TBH I don't know if anybody is even using the PouchDB test suite in that way.

nolanlawson:
So honestly since this is just changing some defaults, I'm +1 on this. But I really really really wish we could find a way to deprecate pouchdb-express-router or merge it with express-pouchdb somehow.

daleharvey:
Ok pouchdb-express-router is transferred (https://github.com/pouchdb/pouchdb-express-router)

As for deprecating it, `express-pouchdb` was supposed to be exactly `pouchdb-express-router` with `pouchdb-server` doing all the extra couch stuff, this was very purposeful, couchdb has tried and failed to be an app server for over 5 years now and almost every person that has spent a long time building apps that way has decided it is a bad architecture which is why couchdb is having discussions about deemphasizing the app'y parts of CouchDB. So while copying the current couchdb monolith app server is useful to current couchdb users, I am hoping we can start building more scalable ecosystem and provide better answers for "how to we do authentication" etc by building on top of well designed single use modules.

It would be awesome if we kept the original plan and express-pouchdb only did the routing required for replication to work, but at this point it would take a lot of work to split it back out into those parts and I dont think anyone is particularly motivated to do so, I am however very motivated to make sure `pouchdb-express-router` has great compatibility.

daleharvey:
Filed https://github.com/pouchdb/pouchdb-server/issues/110 to get the logs fixed, once thats done ill merge this pre 4.0, cheers

nolanlawson:
CouchDB definitely tried to be too many things to too many people, but in my opinion where it failed was in not going far enough. Creating an authentication system where you can't even serve the 99% use case of "everybody can only read and write their own data" is a pity.

pouchdb-server is definitely going down that same route, but to its credit, it is 1) simple to set up and 2) doesn't favor any particular server design. Whereas pouchdb-express-router is only of interest to 1) adept programmers who prefer small modules and 2) those adept programmers who are also using Express. I think the latter contingent is going to be shrinking over the years, especially with the popularity of Hapi.

That said, it's a big tent, and we have room for lots of server implementations. So I'm happy to start documenting pouchdb-express-router as long as we have a sane way of maintaining it. Just don't discount pouchdb-server - it may be bloated, but it sure is easy to use and batteries-included for a lot of people. :wink:

I also agree that express-pouchdb is not really serving the needs of anybody at this point. It is way too deeply tied to pouchdb-server.

daleharvey:
pouchdb-server isnt suitable for this yet since it spits out a bunch of logs both in stdio and the filesystem, requests to change that behaviour have been given push back so im gonna drop this for now, though if I see it still remains a troublesome step for contributors will revisit it

marten-de-vries:
https://github.com/pouchdb/pouchdb-server/issues/110 for stdout.

As for the file system, the only required file there is a config file. [More info here](https://github.com/pouchdb/pouchdb-server/issues/98). It's not that different from it generating database files to persist databases.

PR #4055, opened 2015-07-18
(nolanlawson/pouchdb-load#29) - export generateReplicationId

nolanlawson:
Renamed from `genReplicationId` because if we're going to
export it, we should use the full name.


daleharvey:
Looks good - https://github.com/pouchdb/pouchdb/commit/9879f3f6c0a4970a49d4067fdb78476bfbc591a7

PR #4070, opened 2015-07-24
localstorage availability in safari private

mattbailey:
Safari private mode still exposes the localStorage object but sets the DOM quota to 0, making it non-functional.

I'd like to still be able to use the pouchdb memory adapter in safari private mode, but this test seems to make it fail, since it attempts to use localStorage thinking it will be functional.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -114,8 +114,8 @@ exports.hasLocalStorage = function () {
     return false;
   }
   try {
-    localStorage.setItem('pouchdblstest', 1);
-    return !!localStorage.getItem('pouchdblstest');
+    localStorage.setItem('_pouch_check_localstorage', 1);
+    return !!localStorage.getItem('_pouch_check_localstorage');
   } catch (e) {
     return false;
   }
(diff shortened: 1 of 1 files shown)


lib/utils.js
```diff
@@ -114,7 +114,8 @@ exports.hasLocalStorage = function () {
     return false;
   }
   try {
-    return localStorage;
+    localStorage.setItem('pouchdblstest', 1)
+    return localStorage.getItem('pouchdblstest') ? true : false;
```
nolanlawson: it's shorter to do `return !!localStorage.getItem('pouchdblstest`)`

nolanlawson:
+1 when green. I'd also like to have the key be something like `'_pouch_check_localstorage'` because all of our localstorage keys are prefixed with `_pouch`.

nolanlawson:
This is also failing JSHint because of a missing semicolon, hence why Travis is failing.

daleharvey:
Nice, good catches in the review nolan and cheers for the patch + the updates, :+1: 

merged in https://github.com/pouchdb/pouchdb/commit/348ef3c3d0ad0ba35c0abadc452192161bd00714

PR #4072, opened 2015-07-25
(#4701) - extract out browser-only code

nolanlawson:
This improves our code coverage from 65.4% (lines) to 86.91% (lines).

The basic idea is to remove places where we use `if (process.browser)` and instead rely on Browserify switches in package.json. This has a few benefits:

1) Smaller bundle size when browserifying, since we don't ship any code that's supposed to run in Node
2) Conversely, Istanbul auto-ignores browser-only modules like `idb` and `websql`
3) If somebody messes this up, we'll know immediately because browserify will flat-out fail if it can't find a file

Downsides:

1) It's kind of ugly. I can't think of a better convention than `file.js` and `file-browser.js`.
2) It's arguably harder to read than `if (process.browser)`

I think this is the way to go to reach 100% code coverage, but I'm only doing part of the work and then opening a PR so I can get feedback.

### DIFF
--- a/lib/adapters-browser.js
+++ b/lib/adapters-browser.js
@@ -1,0 +1,6 @@
+'use strict';
+
+module.exports = {
+  idb: require('./adapters/idb'),
+  websql: require('./adapters/websql')
+};
--- a/lib/adapters.js
+++ b/lib/adapters.js
@@ -1,0 +1,5 @@
+'use strict';
+
+module.exports = {
+  leveldb: require('./adapters/leveldb')
+};
--- a/lib/deps/binary/base64StringToBlobOrBuffer-browser.js
+++ b/lib/deps/binary/base64StringToBlobOrBuffer-browser.js
@@ -1,0 +1,8 @@
+'use strict';
+
+var atob = require('./base64').atob;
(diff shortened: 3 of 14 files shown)


nolanlawson:
And right on schedule - browserify failed. :P Should be fixed now.

daleharvey:
`file-browser` seems like a good convention to me, I dont think it is particularly ugly and I much prefer this to the `if (process.browser)` checks. The smaller file size is a pretty big bonus.

One issue I noticed although not this commits fault, is our duplicate `md5` between mapreduce and core, is there a reason for that or something we can fix up?

But :+1: on this commit, I like it

nickcolley:
Semi-related but if we used https://github.com/hughsk/envify wouldn't uglify remove the dead code automatically?

nolanlawson:
@daleharvey Yeah in one case it's md5 on a string, whereas in the other case it's md5 on an ArrayBuffer/Blob. I could rename it to avoid confusion.

@nickcolley I considered this, but I couldn't figure out how I could make it work with Istanbul. I also find this PR's method more explicit, which is nice.

nolanlawson:
228c580127a5a6b6d69927e37853cf81c31fb4bd

nickcolley:
:+1: fair enough

Issue #4073, opened 2015-07-25
Check the pouchdb.min.js size in Travis


nolanlawson:
To make sure we don't accidentally browserify something we don't mean to.

nolanlawson:
fixed by 5a3602cf35d05f98326f5d88ebb8cbeabd5fa3b7

PR #4074, opened 2015-07-25
(#4073) - check pouchdb.min.js size < 50000

nolanlawson:


daleharvey:
:+1: , though if you wanted to switch it to 200KB when you land I would be down for that, cheers I should have done this a while ago

nolanlawson:
Yeah, 50KB is a bit strict, but I'd be fine with upping it as necessary. I'm hoping we can figure out some tricks to continue keeping it under 50.

nolanlawson:
5a3602cf35d05f98326f5d88ebb8cbeabd5fa3b7

daleharvey:
Ah I missed your were testing grip size, I thought it was 500KB and too high. My mistake

nolanlawson:
Yeah I'm using `gzip -c pouchdb.min.js | wc -c` which is the min+gz size in bytes.

PR #4075, opened 2015-07-25
(#4075) - rimraf dist before building

nolanlawson:
I know this has bitten both me and @daleharvey
before, that we accidentally published some
extra files in dist/ that we didn't mean to.
This should fix that.


daleharvey:
Yup nice fix, cheers - https://github.com/pouchdb/pouchdb/commit/a280f3e2ff49d2bc953fb3996667e0eebafbc162

Issue #3112, opened 2014-12-04
Internationalization


nolanlawson:
I've seen a lot of articles in other languages about PouchDB ([Spanish](http://www.kabytes.com/programacion/pouchdb-base-de-datos-de-bolsillo-para-crear-aplicaciones-offline/), [French](http://rougetube.fr/2014/12/4/pouchdb-une-base-donnee-pour-les-navigateurs-web-amp-nbsp_87), [Japanese](http://www.moongift.jp/2014/12/pouchdb-%E3%82%B5%E3%83%BC%E3%83%90%E3%82%B5%E3%82%A4%E3%83%89%E3%81%A8%E3%81%AE%E5%90%8C%E6%9C%9F%E3%81%AB%E3%82%82%E5%AF%BE%E5%BF%9C%E3%81%97%E3%81%9F%E3%82%AF%E3%83%A9%E3%82%A4%E3%82%A2%E3%83%B3)). What they all have in common is that they just provide a very basic "hello world" without delving too deeply.

This shows that non-native English speakers have a real need for documentation in their own language. And I think the API/guides are getting stable enough that we could reasonably start thinking about doing it ourselves on the website.

My French is decent enough that I could translate the guide and would just need a native speaker to grammar-check. Anybody else interested in this project? Which parts do you think should be translated? Or is it a crazy idea?

7ictor:
Is someone working on this? I can help with Spanish (my native language), I run my personal blog in Jekyll and enjoying using Pouch in production for a whole year :blue_heart:

nickcolley:
@7ictor the pull request is ready to have a translation as far as I can remember. I couldn't really go any further because I'm only a native English speaker, so if you want to pair up and get some Spanish going that'd be ace.

mnewmedia:
I could do german, but most germans can or should be able to read english. And since almost all good IT-resources are english, every IT-guy should be able to understand english. And if not they will have to learn it, and when they are forced to learn english it's maybe a bit bad in the beginning but better at the end.
Also translated docs would getting outdated and produce unnecessary issues. I know PouchDB guys are just too lovely and want to help everyone =) but still they really should learn english even I should improve my english skills.
Also my English is only beginner Level and it was sometimes hard to understand the CouchDB Docs, but the PouchDB Docs are mostly written in easy to understand english.

nickcolley:
Hey @mnewmedia thanks for comment.

[...]

PR #4076, opened 2015-07-25
(#3112) - Docs i18n

nickcolley:
Ok got something working here.

Also needs the following:
- [ ] Nice language selector
- [ ] Contribution README doc update
- [ ] At least one page translated.

Think if we can get one page translated and make sure the site is working good with just that one we can gradually do the rest later?

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -12,3 +12,5 @@ github:
 collections:
   guides:
     output: true
+languages: ["en", "es"]
+exclude_from_localizations: ["static"]
--- a/docs/_guides/index.md
+++ b/docs/_guides/index.md
@@ -4,59 +4,7 @@ layout: guide
 nav: Intro
 title: Introduction to PouchDB
 sidebar: guides_nav.html
+translated: true
 ---
 
-Welcome to the PouchDB guide! Consider this your starting point for anything and everything related to the world of PouchDB and CouchDB.
-
-For a quicker TodoMVC-based tutorial, you can also check out the ["Getting Started" guide](/getting-started.html).
-
-Feel free to skip ahead using the sidebar at any time.
-
-What is PouchDB?
-------
-
-**PouchDB** is a JavaScript implementation of [CouchDB](http://couchdb.com). Its goal is to emulate the CouchDB API with near-perfect fidelity, while running in the browser or in Node.js.
-
-What is CouchDB?
----------
-
-**CouchDB** is a NoSQL database created in 2005 by Damien Katz, and now maintained by the Apache Software Foundation.  If you are a JavaScript developer, you probably use CouchDB every day, because it's the core technology that powers [npm](https://www.npmjs.org/).
-
-Couchbase, CouchDB, Couch-wha?
-----------
-
-Today there are two major database companies that
-can trace their lineage back to CouchDB: [**Couchbase**](http://couchbase.com) and [**Cloudant**](http://cloudant.com). Both of them are separate products compared to CouchDB.
-
-However, all three of these databases share the same **CouchDB sync protocol**. This means that PouchDB can sync with either one of them, and you can always swap out one database for another. You're never locked in.
-
-In a sense, these databases are like competing phone companies, and the CouchDB sync protocol is the underlying telephony infrastructure.
-
-CouchDB's one-two punch: HTTP and sync
-----------
-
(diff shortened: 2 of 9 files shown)


daleharvey:
> Think if we can get one page translated and make sure the site is working good with just that one we can gradually do the rest later?

I think ideally we can put up 2 pages, maybe like the home page in french and spanish using a google translate version for testing purposes, then invite people to come along and start contributing proper translations.

I would really like to see how to fallback, so if a translated page isnt available, maybe it shows the english with a link to a google translate?

This is an amazing change, cheers for picking it up :)

nickcolley:
With this plugin if there's no translation it just shows the english content.

nickcolley:
@daleharvey @nolanlawson do you think I could (or someone else) put a tweet out to see if anyone would be up for doing some translation?

nolanlawson:
I have to admit that I have my reservations about this PR.

Having documentation in multiple languages is a worthy goal, but I don't think PouchDB has the resources to do this right now, and if we try to start down this path, then we will likely bite off more than we can chew.

Imagine when we change something and then need to go back and fix the docs. Is the committer going to fix the docs in Spanish, French, and German as well as English? No, what'll happen instead is that the foreign-language docs will rot and speakers of those languages will just go to the English version because they can trust it.

I'm all for inclusion, and as a fluent speaker of French who has worked in multilingual environments, I understand the pain that non-native English speakers have to go through just to participate in the global programming community. But I don't think we at PouchDB are in a reasonable position to help solve this.

daleharvey:
Alternatively people contributing to translations is one of the most popular ways to get started within a project, I know for Mozilla our translation community is huge and we are still seeing a bunch of non english activity - http://frontinbrazil.com.br/pouchdb/, https://vpease.wordpress.com/2015/08/11/ionic-framework-pouchdb-y-cloudant-la-combinacion-perfecta-totalmente-expuesta/ etc

We will never require maintainers or contributors to fix things across all languages, but I think providing the option for people to do so would be awesome.

I do think the example @nolanlawson brings up is important to think about though, we would really need a mechanism for determining what translations are out of date and what changed. I think ideally we would want to show a warning on pages that have out of date translations but at the least someone translating locally needs a simple way to figure out what translations needed to do.

I think we should figure that out before asking people to start to contribute translations

marten-de-vries:
Having something gettext-like would fix it: there wouldn't be multiple pages: just an English one in which all pieces of text that would need to be translatable are marked in some way. The translations would then be in separate (.po) files, and would only replace the English text if the piece of text didn't change in the meantime. If it really used gettext (.po) files, an additional benefit would be that we could use a dedicated translation site, making understanding git not a requirement for contributing.

The problem is implementing it. I haven't seen a ready made solution for jekyll yet.

marten-de-vries:
As for implementation, I have suggested a node preprocessor for jekyll before: https://gist.github.com/marten-de-vries/d734b71fde92617e717c

A preprocessor because I don't know Ruby, regardless of language this does explain the concept in code though.

nolanlawson:
Yeah, as I've said, I think this is more of a social problem than a technical problem for us. Mozilla might have the resources to do wide-scale translation, but we're small-time, so I don't think we do. :/

Also to Dale's point, I think the needs of foreign-language speakers is largely being served by the kind of blogposts he cites. I imagine there is a correlation between 1) people with low English comprehension, and 2) people who rely more on tutorials/blogposts/videos than official docs. Which indicates to me that translating the official docs would not be very high-value (although maybe the guide is an exception).

That said, if we do push forward on this, I could try my hand at translating the guide to French. :) I'd need a native-language speaker to go through and proofread, though. (My French is correct but lacks pizzazz.)

zoepage:
> Alternatively people contributing to translations is one of the most popular ways to get started within a project...
100% yes on that.

It's always nice to announce one person per language to at least check if the documentation is up to date. No need they have to take care of it anytime. Just make clear if they don't have the time, they need help and get the word out.

A warning on the page is a good idea. Also providing a diff, if the main resource (here I'd suggest the english version of the docs) changes, so people who translate the docs in their language know what to change without a huge hassle.

This is how we do it with hoodie:
https://github.com/hoodiehq/documentation#translations

// folder structure
We have one folder per language. This is double content and not the best practice, but we figured the most simple solution for people to contribute to translations.

nickcolley:
So looks like we need to decide a good way to automatically show a message for:
- A page that is in English because there's no translation, and how to contribute
- A page that is translated but out of date, how to contribute and where to find the original.

daleharvey:
So I think this is awesome, the branch will be around however closing out PR's that arent active at the moment. People will still be able to find it when looking from the issue

Issue #4078, opened 2015-07-26
Nasty ETag bug in http-proxy


nolanlawson:
This is a bug that took me a long time to track down, and I don't have a good solution, so I'm going to document it thoroughly.

The `Testing successive new_edits to the same doc, different content` can sometimes fail with a bizarre error like this:

![screenshot 2015-07-26 17 06 11](https://cloud.githubusercontent.com/assets/283842/8895907/abe85b06-33b8-11e5-9b3a-e698c781c358.png)

The root problem is that one test creates a document with the name `foo` and the rev `1-x` and the next test does the same thing, but with different fields in the document. The second test ends up getting the document from the first test rather than its own.

How could this be possible? The database is deleted between every test, so that's not it. It turns out it's a proxy issue. When I set a breakpoint, I get this (correct) document back from `localhost:5984`:

```js
{"_id":"foo","_rev":"1-x","bar":"baz"}
```

Whereas I get this (incorrect) document back from `localhost:2020`:

```js
{"_id":"foo","_rev":"1-x","bar":"huzzah"}
```

What's happening is that `pouchdb-http-proxy` is caching the GET requests based on the ETag. Since the ETag is the same for both documents (`1-x`), it returns the old data.

I dug into the guts of `pouchdb-http-proxy`, but it seems like this is the behavior of a Node core module (`http`). I don't think it's reasonable for me to open a bug on Node (since it's doing what it's supposed to do - ETag is being used correctly). However I think this highlights the many problems we have with the proxy, and shows why I'm so eager to get rid of it.

daleharvey:
I filed a bug on couchdb instead (https://issues.apache.org/jira/browse/COUCHDB-2764) Etags are supposed to reflect the content that is being served, we shouldnt get different documents with the same etag. Im down for removing the proxy since it has shown to be buggy, but I dont think its at fault here.

Merged the fix in https://gist.github.com/daleharvey/2010c3d3f873bc03fc5e

nolanlawson:
Fair enough. I guess I assumed the ETag behavior from CouchDB was correct. (Makes sense to me that it would be the rev, anyway.)

PR #4079, opened 2015-07-26
(#4078) - simple fix for ETag bug

nolanlawson:
Simple solution to the problem: don't put ourselves
in a situation where the ETag bug can bite us.

Changes the document IDs to be more unique, and also
adds `bar` values to the documents so we get more
meaningful errors than "couldn't read undefined.should".

### DIFF
--- a/tests/integration/test.bulk_docs.js
+++ b/tests/integration/test.bulk_docs.js
@@ -355,8 +355,9 @@ adapters.forEach(function (adapter) {
 
       var db = new PouchDB(dbs.name);
       var docs = [{
-        '_id': 'foo',
+        '_id': 'foobar123',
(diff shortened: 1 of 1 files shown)


nolanlawson:
Heh, it's funny. I just realized the only reason I noticed this bug was because I accidentally removed the cache-busting in another PR.

daleharvey:
I filed a bug on couch for this, https://issues.apache.org/jira/browse/COUCHDB-2764, :+1: for the workaround for now

daleharvey:
Merged in https://github.com/pouchdb/pouchdb/commit/7c166045194775bcfd8468b31bf8fd6f70ccaaa1

Issue #4080, opened 2015-07-26
Remove local_seq


nolanlawson:
Might as well get this in for 4.0.0.

nolanlawson:
f0639ea6bd0a9041592f890f463e905aa05d2158

sansato:
The ?local_seq=true option is still active in couchdb, and I don't see how else to get the sequence number matching the current rev of a document.  I'm tasked with providing a promise for a particular replication stream being up-to-date with regard to a particular doc@rev, which is easy if I can get its seq.  Is there a recommended pouchdb alternative (that works with a remote db) that returns such seq?

https://docs.couchdb.org/en/2.3.1/api/document/common.html

karlwestin:
This depends a lot about what you're trying to do, it's a bit tricky

Unfortunately PouchDB sequence numbers do not match CouchDB sequence numbers since 2.0, and if CouchDB runs in multi-node mode CouchDB sequence numbers might not match each other either.

What you can do is to store the result of your push replication (i assume this is about whether you uploaded all your changes to the server) in a local doc, then use that sequence to look for changes after that number

```javascript
localDB.changes({ since: last_push_update_seq, doc_ids: 'myDocID' })
   .then(changes => {
      // if there's no changes to this doc since last sync, you're good
      return changes.results.length === 0
   })
```

If you are online, you can compare the revision on your local doc against the revision on the server.

Storing the last update seq of a replication is a bit involved and i'm sorry i can't provide that to support your project, but in general:
- if you're replicating with `live: false` you get the last synced sequence from the promise
- if you are doing `live: false` you'll have listen for the 'paused' event and call localDB.info() when that happens

sansato:
Thanks for this.  It opens some options I hadn't seen.  Especially appreciating the heads-up about server seq numbers when using a couchdb cluster.

PR #4081, opened 2015-07-26
(#4080) - remove local_seq

nolanlawson:
For the record, this will also improve code coverage,
because we are not testing it.


lib/utils.js
```diff
@@ -353,12 +353,6 @@ exports.cancellableFun = function (fun, self, opts) {
 
 exports.explain404 = require('./deps/explain404');
 
-exports.info = function (str) {
-  if (typeof console !== 'undefined' && 'info' in console) {
-    console.info(str);
-  }
-};
-
```
nolanlawson: I can verify that this is no longer being used anywhere.

daleharvey:
:+1: I seem to need a rebase though

nolanlawson:
And `local_seq` is gone! f0639ea6bd0a9041592f890f463e905aa05d2158

PR #4083, opened 2015-07-27
(#4083) - fix typo in docs

nolanlawson:


nolanlawson:
Besides fixing the typo, I also decided to come right out and un-recommend the SQLite Plugin.

Although I'm technically a maintainer of the SQLite Plugin, I think it is too complex/untested/buggy at this point, and I don't have much motivation to work on it. Also, its only real reason for existence is the 50MB limit on iOS, which will go away once Apple fixes IndexedDB. (Preloading/native access is nice, but I don't think it's a dealbreaker for us since we have pouchdb-load.)

In any case, I'm very aware of the problems with the SQLite Plugin, so I'd be doing our users a disservice if I wasn't upfront about them.

nickcolley:
:+1: much more clear when you'd use the plugin.

daleharvey:
:+1: looks good to me

nolanlawson:
c55db55d7edbf0821538e027399536c260af5876

PR #4085, opened 2015-07-27
corrected Python http server modules information

karolklp:


nickcolley:
Since this is a beginners guide it might be confusing to have two options, especially as this isn't a python project. Maybe we should use https://www.npmjs.com/package/http-server or something?

karolklp:
So let's leave only Python 3 option. It was released in 2008, so we it's almost certain, that beginner will have freshly-downloaded Python >= 3

nickcolley:
My new install of Fedora ships with v2.X.

I'd say since python is fragmented in that respect we can't assume what version people will have.

nolanlawson:
I think it's harmless to have instructions for both, although I do agree that Python is fragmented and the current situation is not ideal.

daleharvey:
Yeh python 2.7.5 on my mac, I think both is simple enough

nolanlawson:
canceled the travis build because it's not necessary

karolklp:
Thanks guys, I wasn't aware of Python's fragmentation. Should we keep it that way or change it to https://www.npmjs.com/package/http-server like @nickcolley suggested? I like Python solution because it's nice trick for easy local server that may be useful..

nolanlawson:
In my experience, new developers have an even harder time getting Node/npm up and running, because it's not installed by default on most systems.

nickcolley:
I'm cool with having the both methods then I guess. :+1:

nolanlawson:
Merged as 043498fa9dd54b6aec61b40619f6d139115f2f1a. I amended the commit on the way in to match our commit message format and to use a `#` comment instead of parens (just in case they copy-paste it). Thanks for the contribution! :)

nickcolley:
Thanks @karolklp ! :+1: :tada:

karolklp:
Thanks guys for great OSS experience :)

Issue #4086, opened 2015-07-28
Should be able to remove _local docs in bulkDocs()


nolanlawson:
This is something I realized we were doing wrong, by looking at the Istanbul coverage results. I have a fix.

daleharvey:
https://github.com/pouchdb/pouchdb/commit/43d50f675d8d447eb3ce6fa75c409ce1a0c22216

PR #4082, opened 2015-07-26
(#4071) - more coverage fixes

nolanlawson:
Yet more code coverage fixes, on top of #4077.
Coverage is now up to 93.73%.

Major changes:

1. `COVERAGE=1 npm t` now runs integration, unit, and mapreduce tests
2. `Changes` moved from `utils.js` to its own file, `changesHandler.js`
3. Dead code removed
4. `/* istanbul ignore */` added where it makes sense to ignore code (e.g. `isChromeApp()`)
5. `parseUri.js` simplified
6. General refactoring to move stuff out of `utils.js`.

PR #4087, opened 2015-07-28
(#4086) - removeLocal() during bulkDocs()

nolanlawson:
This increases our code coverage very slightly
by allowing the user to remove a `_local` doc
during a `bulkDocs()`.

Incidentally this test led me to find a bug in
CouchDB 1.6.1. I'll report it in their bugtracker.

### DIFF
--- a/tests/integration/test.bulk_docs.js
+++ b/tests/integration/test.bulk_docs.js
@@ -320,6 +320,89 @@ adapters.forEach(function (adapter) {
       });
     });
 
+    it('Deleting _local docs with bulkDocs' , function () {
+      var db = new PouchDB(dbs.name);
+
+      var rev1;
+      var rev2;
+      var rev3;
+      return db.put({_id: '_local/godzilla'}).then(function (info) {
+        rev1 = info.rev;
+        return db.put({_id: 'mothra'});
+      }).then(function (info) {
+        rev2 = info.rev;
+        return db.put({_id: 'rodan'});
+      }).then(function (info) {
+        rev3 = info.rev;
+        return db.bulkDocs([
+          {_id: 'mothra', _rev: rev2, _deleted: true},
+          {_id: '_local/godzilla', _rev: rev1, _deleted: true},
+          {_id: 'rodan', _rev: rev3, _deleted: true}
+        ]);
+      }).then(function () {
+        return db.allDocs();
+      }).then(function (info) {
+        info.rows.should.have.length(0);
+        return db.get('_local/godzilla').then(function () {
+          throw new Error('expected 404');
+        }, function (err) {
+          should.exist(err);
+        });
+      });
+    });
+
+    if (adapter === 'local') {
+      // these tests crash CouchDB with a 500, neat
+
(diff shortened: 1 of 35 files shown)


nolanlawson:
I rebased this on #4082 to avoid bitrotting myself.

nolanlawson:
I'll note that since removing a local doc inside of bulkDocs is something that our users might never do (and even CouchDB doesn't accept it), then another valid solution to the coverage problem would be to remove the `removeLocal` code from `processDocs.js`.

daleharvey:
Does couchdb-master let you delete a local doc in bulkDocs? it sounds like a reasonable thing to do so this is probably good, but would like to make sure we arent breaking compatibility with Couch on purpose here, @rnewson ?

tests/integration/test.bulk_docs.js
```diff
@@ -320,6 +320,89 @@ adapters.forEach(function (adapter) {
...
+        return db.allDocs();
+      }).then(function (info) {
+        info.rows.should.have.length(0);
+        return db.get('_local/godzilla').then(function () {
+          throw new Error('expected 404');
+        }, function (err) {
+          should.exist(err);
+        });
+      });
+    });
```
nolanlawson: ^ @daleharvey This test passes on CouchDB 1.6, so yes, it lets you delete a _local doc within _bulk_docs.

nolanlawson:
Weird, GitHub is still showing 2 commits instead of only 1, even though this branch is up-to-date. Sorry for the noise.

daleharvey:
Yeh I have had that problem with github before, but yeh if its a plain bug thats failing on couch then :+1: to this, looks great, Did you file the bug? I think if would be very useful to have a reference next to the test so we can enable it when couch gets fixed

nolanlawson:
Yup I did: https://issues.apache.org/jira/browse/COUCHDB-2758

I can add a reference to the bug in the test comments when I merge.

rnewson:
curl foo:bar@localhost:15984/db1/_bulk_docs -Hcontent-type:application/json -d '{"docs":[ {"_id":"_local/foo"}   ]}'
[{"ok":true,"id":"_local/foo","rev":"0-1"}]
➜  ~  curl foo:bar@localhost:15984/db1/_local/foo
{"_id":"_local/foo","_rev":"0-1"}

rnewson:
curl foo:bar@localhost:15984/db1/_bulk_docs -Hcontent-type:application/json -d '{"docs":[ {"_id":"_local/foo","_deleted":true}   ]}'
[{"ok":true,"id":"_local/foo","rev":"0-0"}]
➜  ~  curl foo:bar@localhost:15984/db1/_local/foo
{"error":"not_found","reason":"missing"}

nolanlawson:
@rnewson That's weird; I do think I found a test case that's failing. Did you look at that jira?

daleharvey:
I added a reference to the CouchDB issue and merged, cheers

https://github.com/pouchdb/pouchdb/commit/43d50f675d8d447eb3ce6fa75c409ce1a0c22216

PR #4095, opened 2015-08-01
(#4095) - 'destroyed' event emits on all DBs

nolanlawson:
Here's the story: the coverage reports showed me that there was an unused function in mapreduce's `create-view.js`:

screenshot 2015-08-01 16 07 50

Delving in, I realized that the problem is that destruction events aren't emitted on a db object if an identically-named db object is the one that was destroyed.

So I fixed that (plus a test to catch it), but adding a new event listener on `PouchDB` caused some tests to fail due to exceeding the max listeners of 10.

So I also added some temporary `PouchDB.setMaxListeners()` to those tests that were failing. In the process, I also pulled out the persisted-only tests from mapreduce's test suites into a separate file (`test.persisted.js`), so I could isolate exactly which tests happen to go over the limit of 10. (Since the limit goes right back to 10 after the tests are done, we'll still be able to catch any true memory leaks).

With this fix, coverage improves from 93.88% (4051/4315 lines) to 93.94% (4058/4320 lines).


nolanlawson:
As for the question of why we needed that unused function anyway: I couldn't find a test to make it fail without that code, but I would feel uncomfortable if database objects remain cached even after they're destroyed, so I would prefer to keep it in there.

lib/setup.js
```diff
@@ -10,23 +10,18 @@ PouchDB.preferredAdapters = [];
...
+function bindEventEmitterMethods(Pouch) {
+  Object.keys(EE.prototype).forEach(function (key) {
+    if (typeof EE.prototype[key] === 'function') {
+      Pouch[key] = eventEmitter[key].bind(eventEmitter);
+    }
+  });
+}
+
+bindEventEmitterMethods(PouchDB);
+
```
nolanlawson: Explicitly listing out all the EventEmitter prototype methods is not necessary because we can determine it programmatically. Hopefully this should minify a bit smaller.

tests/integration/test.events.js
```diff
@@ -58,6 +58,23 @@ adapters.forEach(function (adapter) {
...
+          if (++called === 2) {
+            resolve();
+          }
+        }
+        db1.once('destroyed', checkDone);
+        db2.once('destroyed', checkDone);
+        db1.destroy();
+      });
+    });
+
```
nolanlawson: This test fails before the fix.

tests/mapreduce/test.persisted.js
```diff
@@ -0,0 +1,687 @@
...
+          res.rows.map(function (row) {
+            return [row.id, row.key];
+          }).should.deep.equal([
+              ['baz', 'bar'],
+              ['bar', 'baz'],
+              ['foo', 'foo']
+            ]);
+        });
+      });
+    });
```
nolanlawson: This is a new test that I used to check the destroyed event on the dependent db. It succeeds before and after the fix, but I think it's always good to have more tests.

tests/integration/test.replication.js
```diff
@@ -19,6 +19,16 @@ adapters.forEach(function (adapters) {
...
+    before(function () {
+      // briefly increase the limit from 10 because we momentarily
+      // go over during these tests
+      PouchDB.setMaxListeners(20);
+    });
+
+    after(function () {
+      PouchDB.setMaxListeners(10);
+    });
+
```
nolanlawson: I probably should have broken `test.replication.js` into two separate tests and only had this `setMaxListeners` on one of them, but I got a little lazy here.

nolanlawson:
Error in PouchDB Server, looking into it...

nolanlawson:
OK, should be fixed now.

daleharvey:
:+1: merged in https://github.com/pouchdb/pouchdb/commit/26edf35a899946e3acf15ebf2b83d4f7f0fd9f80

Issue #4071, opened 2015-07-25
Improve code coverage


nolanlawson:
Now that mapreduce is folded into pouchdb core, one of the things I really miss is the confidence that code coverage is at 100%.

I think we can get there, but it will take some discipline to go through the codebase, add `istanbul ignore` where it makes sense, extract out browser-only code (since we can't test coverage in the browser using istanbul), and of course add tests as well.

I'm willing to do the work to try to get us up to 100%, or at least as close as possible.

nickcolley:
What percentage are we at the moment?

nolanlawson:
BTW @daleharvey where is code coverage reported to? Can't seem to find it.

daleharvey:
http://pouchdb.cloudant.com/coverage_results + http://pouchdb.cloudant.com/performance_results

nolanlawson:
Hm, it doesn't seem to be storing the actual coverage number anywhere. I wish there was some way we could make Travis surface that information for every pull request.

nickcolley:
Could try out something like https://coveralls.io/ which gives those kinds of stats.

daleharvey:
@nolanlawson What do you do to figure out what exact lines are not being covered? I would like to get this to 100% so its easier to not regress it, even if we add some ignores for where there is browser specific stuff we can remove it when we start testing the browser implementation (and I would prefer to add ignores than tests that only exist for coverage purposes)

daleharvey:
Also filed a bug to see why we dont see the source files in coveralls - https://github.com/lemurheavy/coveralls-public/issues/633

daleharvey:
And we are at 100% - https://github.com/pouchdb/pouchdb/commit/79785ee234b77af36137969b73f5d80dd7f0b0e4

PR #4096, opened 2015-08-01
(#4071) - simplify ajax and changes, increase coverage

nolanlawson:
This increases coverage from 93.88% (4051/4315 lines) to 94.1% (4036 / 4289).

### DIFF
--- a/tests/integration/test.changes.js
+++ b/tests/integration/test.changes.js
@@ -568,6 +568,90 @@ adapters.forEach(function (adapter) {
       }).on('error', done);
     });
 
+    it('Immediately cancel changes', function () {
+      // fixes code coverage by ensuring the changes() listener
+      // emits 'complete' even if the db's task queue isn't
+      // ready yet
+      return new PouchDB.utils.Promise(function (resolve, reject) {
+        var db = new PouchDB(dbs.name);
+        var changes = db.changes({live: true});
+        changes.on('error', reject);
+        changes.on('complete', resolve);
+        changes.cancel();
+      });
+    });
+
+    it('Changes with invalid ddoc view name', function () {
+      return new PouchDB.utils.Promise(function (resolve, reject) {
+        var db = new PouchDB(dbs.name);
+        db.post({});
+        var changes = db.changes({live: true, filter: '_view', view: ''});
+        changes.on('error', resolve);
+        changes.on('change', reject);
+      });
+    });
+
+    it('Changes with invalid ddoc view name 2', function () {
+      return new PouchDB.utils.Promise(function (resolve, reject) {
+        var db = new PouchDB(dbs.name);
+        db.post({});
+        var changes = db.changes({live: true, filter: '_view', view: 'a/b/c'});
+        changes.on('error', resolve);
+        changes.on('change', reject);
+      });
+    });
+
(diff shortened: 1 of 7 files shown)


daleharvey:
hrm, do you think we can fallback from the aborted connection couch gives us? I can take a look at this later, I would prefer to avoid adding more pouchdb only tests for this type of functionality

nolanlawson:
Sorry, what do you mean about aborted connections?

daleharvey:
as discussed in https://issues.apache.org/jira/browse/COUCHDB-2765, the tests we add mostly check that pouchdb emitted the error event, so it seems we should be able to do that consistently with cocuhdb if its aborting the changes request we are making

tests/integration/test.changes.js
```diff
@@ -746,6 +746,86 @@ adapters.forEach(function (adapter) {
...
+              live: true,
+              filter: '_view',
+              view: 'name/name'
+            });
+            changes.on('error', resolve);
+            changes.on('change', reject);
+          });
+        });
+      });
+    }
```
nolanlawson: @daleharvey  You're talking about this thing? Yeah, the only reason I added it was because it improves the code coverage, since there's a line in `changes.js` where we say `var mapFun = ddoc && ddoc.views && ddoc.views[viewName[1]] && ddoc.views[viewName[1]].map`. I don't think there's any way to improve coverage without making a Pouch-only test, since CouchDB breaks for the same test.

daleharvey: Yup, but its a bug that we do not handle this error consistently between pouchdb / couchdb, any time we have `if (adapter === 'local') {` in our tests it means we behave differently and one of the main parts of the pouchdb api is to be consistent between backends. We occasionally make an explicit choice to differ, but otherwise it should only really be plain bugs in couchdb that we expect to be fixed soon that get this. The discussion in https://issues.apache.org/jira/browse/COUCHDB-2765 suggests this is an API choice for couchdb and we should be handling it.

Ill take a look to see if we can get it consistent here given the current couch behaviour, if it takes a while them I am fine landing this and filing a follow up to remove the inconsistency but any time we are adding pouch only tests its kinda like settimeout in that its a big flag we should be thinking about how to remove

daleharvey:
Actually since this bug is documenting the bug and adding tests for it, not introducing it, I think its better to land straight away and file a follow up for fixing the bug :+1:

tests/integration/test.changes.js
nolanlawson: Fair enough. If they don't plan on fixing it, then the fix on our side would be to put a `try/catch` around the `JSON.parse`, but I didn't want to do that since I thought they were going to be fixing it soon. I can add a TODO in a comment to let us know that we need to fix the bug on our side.

Also I'll note that this is a super rare bug that I doubt any users will actually run into, and either way if they do, they will either see an error from Couch or from Pouch.

nolanlawson:
There are minor merge conflicts; I will fix on the way in.

nolanlawson:
Rebased and waiting for a green from Travis just to be 100% sure.

daleharvey:
All green, merged - https://github.com/pouchdb/pouchdb/commit/7725eccd25a126b500d555d2e28233476701008e

PR #4097, opened 2015-08-01
(#4071) - remove dead code from TaskQueue

nolanlawson:
According to Istanbul, this was all dead code. Since TaskQueue is
internal, I feel safe removing it.


daleharvey:
Very happy to get rid of this type of code - https://github.com/pouchdb/pouchdb/commit/324c087b77c8dd2b06b7ccc942b370d46cc7a0db

Issue #4098, opened 2015-08-01
Remove deprecated onChange and complete options from changes()


nolanlawson:
These have been deprecated for a long time, but it will be a lot of effort to remove them because many of the tests use this format.

This would really help with the code coverage, though.

daleharvey:
Removed in https://github.com/pouchdb/pouchdb/commit/a0a8ff31938d010eaf9b919323a3ce97bad3a04b

PR #4099, opened 2015-08-02
(#4098) remove deprecated onChange and complete

nolanlawson:
Removes these options both from `changes()` and from `replicate()`.
Only the new style is supported.

### DIFF
--- a/lib/adapters/http/index.js
+++ b/lib/adapters/http/index.js
@@ -950,7 +950,7 @@ function HttpPouch(opts, callback) {
             if (returnDocs) {
               results.results.push(c);
             }
-            utils.call(opts.onChange, c);
+            opts.onChange(c);
           }
           return ret;
         });
--- a/lib/changes.js
+++ b/lib/changes.js
@@ -14,7 +14,6 @@ function Changes(db, opts, callback) {
   var self = this;
   this.db = db;
   opts = opts ? utils.clone(opts) : {};
-  var oldComplete = callback || opts.complete || function () {};
   var complete = opts.complete = utils.once(function (err, resp) {
     if (err) {
       self.emit('error', err);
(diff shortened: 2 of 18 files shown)


nolanlawson:
Coverage increased from 93.88% (4051/4315 lines) to 93.9% (4032/4294).

lib/adapters/http/index.js
```diff
@@ -950,7 +950,7 @@ function HttpPouch(opts, callback) {
             if (returnDocs) {
               results.results.push(c);
             }
-            utils.call(opts.onChange, c);
+            opts.onChange(c);
```
nolanlawson: whenever we know that the function is not undefined, I'd prefer to just call it directly than use `utils.call()`, which is a weird function anyway.

daleharvey: Yeh the only point of `utils.call` is when we dont know whether the function is undefined, otherwise definitely isnt needed, a lot of places we use it then we ended up wrapping the function we call it from at a higher level and its no longer needed

daleharvey:
Sweet lets do it - https://github.com/pouchdb/pouchdb/commit/a0a8ff31938d010eaf9b919323a3ce97bad3a04b

lib/adapters/http/index.js
nolanlawson: That makes sense; good to know!

daleharvey: Heh taking a look at the code (https://github.com/pouchdb/pouchdb/blob/master/lib/deps/call.js) Im gonna assume it has ended up not being made redundant and we have just missed it and forgot to take it out

nolanlawson: Oh yeah, that was me who removed a bunch of dead code. But I guess after removing the dead code, I didn't notice that the whole thing was dead. XD

nickcolley:
https://github.com/pouchdb/pouchdb/blob/master/lib/changes.js#L43 is this meant to be there?

nolanlawson:
@nickcolley that's just an internal vestigial thing; users never see it. :)

Issue #4103, opened 2015-08-02
Remove undocumented/untested SSE code


nolanlawson:
This code is not being used right now and is undocumented. Can we just get rid of it? cc @calvinmetcalf 

In the meantime, there's been lots of progress in things like replication over streams and over sockets (Thali project, pouchdb-replication-stream, express-pouchdb-replication-stream, pouchdb-dump, pouchdb-load, socket-pouch), so I think the need for SSE might have diminished somewhat.

Happy to keep https://github.com/pouchdb/pouchdb/issues/2823 open while we work on re-integrating it, though.

daleharvey:
https://github.com/pouchdb/pouchdb/commit/c36333a378aed91ee05813c486669a5b6c66a756

PR #4104, opened 2015-08-02
(#4103) - remove SSE code for now

nolanlawson:
All of this was dead code per Istanbul.

### DIFF
--- a/lib/adapters/http/index.js
+++ b/lib/adapters/http/index.js
@@ -871,9 +871,6 @@ function HttpPouch(opts, callback) {
       }
     }
 
-    if (opts.continuous && api._useSSE) {
-      return  api.sse(opts, params, returnDocs);
-    }
     var xhr;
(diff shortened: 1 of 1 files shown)


daleharvey:
So I am :+1: it does seem like the best option for streaming continuous changes, but with compatibility issues / work to get it enabled / multiple code paths and no improvement on replication. I would rather see the work going in to getting real replication streaming working and then base whatever changes we make to the `changes` api on top of that.

But will let @calvinmetcalf speak for this since he implemented it, entirely possible I am missing something

calvinmetcalf:
:+1:

daleharvey:
Ok lets do it, getting the streaming replication up and running is one of my next goals, likely then we will wanna revive this or similiar.

https://github.com/pouchdb/pouchdb/commit/c36333a378aed91ee05813c486669a5b6c66a756

PR #4105, opened 2015-08-02
(#4071) - fix code coverage of sync()

nolanlawson:
`sync()` is now at 100% code coverage.

### DIFF
--- a/lib/sync.js
+++ b/lib/sync.js
@@ -30,13 +30,6 @@ function Sync(src, target, opts, callback) {
   this.push = replicate(src, target, opts);
   this.pull = replicate(target, src, opts);
 
-  var emittedCancel = false;
-  function onCancel(data) {
-    if (!emittedCancel) {
-      emittedCancel = true;
-      self.emit('cancel', data);
-    }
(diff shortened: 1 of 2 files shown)


daleharvey:
So this looks mostly good, but fairly worried about the need for `setTimeout`s in these tests, so taking a look at whether they are absolutely required

nolanlawson:
Yeah I just wanted to make sure that the `complete` callbacks get called after the `change` callbacks, since the order is unspecified. I believe in practice, though, `change` comes before `complete`, so they may be safe to remove.

Note that they are setTimeouts with 0, though! So I'm not trying to do any janky sleeps here. :stuck_out_tongue:

daleharvey:
ok cool, yeh they arent the 'just wait for x to happen, they are all yeilding on something sync happening so I 'think' its all fine, so :+1: but will keep an eye on failures

daleharvey:
Merged in https://github.com/pouchdb/pouchdb/commit/83d7e10d139897ac3c6c26e38e13c5ca05d14c28

PR #4106, opened 2015-08-02
(#4071) - fix coverage of constructor and merge

nolanlawson:
`merge.js` is now at 100%, `constructor.js` is close but there's one unused code path
that I would like to handle by actually testing rather than ignoring.

### DIFF
--- a/lib/merge.js
+++ b/lib/merge.js
@@ -135,9 +135,6 @@ function doMerge(tree, path, dontExpand) {
           }
           continue;
         }
-        if (!item.ids) {
-          continue;
-        }
         /*jshint loopfunc:true */
         item.ids[2].forEach(function (el, idx) {
           trees.push(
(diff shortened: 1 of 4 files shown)


lib/merge.js
```diff
@@ -135,9 +135,6 @@ function doMerge(tree, path, dontExpand) {
           }
           continue;
         }
-        if (!item.ids) {
-          continue;
-        }
```
nolanlawson: ^ This dead code removal does make me a little nervous, since it was copied from the Erlang code and may cover a case we're not aware of. @daleharvey could you take a look?

nolanlawson:
Should be fixed now. Turned out we weren't handling the `new`-less constructor for `PouchAlt`.

lib/merge.js
daleharvey: Heh all this code is fairly hairy but its well tested (https://github.com/pouchdb/pouchdb/blob/master/tests/unit/test.merge.js), if we arent hitting this then I think we are safe.

daleharvey:
looks good to me, cheers - https://github.com/pouchdb/pouchdb/commit/35dcc5aa17078568bc126cd918e6271c4f1441d9

Issue #4107, opened 2015-08-02
Show code coverage numbers in Github PRs


nolanlawson:
So right now we are running Istanbul coverage checks, but it's really hard to surface them because you need to dig deep into the Travis UI to find them. I would prefer to have the coverage numbers front-and-center.

I [asked around on Twitter](https://twitter.com/nolanlawson/status/627829374809542656), and apparently the best resource is indeed Coveralls (thx @nickcolley!) combined with [istanbul-coveralls](https://github.com/shinnn/istanbul-coveralls). Unfortunately Coveralls is down today for maintenance, but it seems really easy to set up; I've got some sample repos here:

* https://github.com/noopkat/avrgirl-stk500v2/blob/master/package.json
* https://github.com/hoodiehq/pouchdb-hoodie-sync/blob/master/package.json

After my recent PRs, our code coverage should be somewhere around 95% or 96%. Not amazing, but not bad either! I hope to make it even better by adding [fakeIndexedDB](https://github.com/dumbmatter/fakeIndexedDB) tests in Node so that we can actually cover our IndexedDB adapter.

(Follow-up to https://github.com/pouchdb/pouchdb/issues/4071)

daleharvey:
I also followed up to add the badge to the repo, this is awesome :)

nickcolley:
@daleharvey if you remove the space between them they'll line up. (lil nit)

nolanlawson:
+1 for @nickcolley's suggestion. Looks great though!

daleharvey:
Yeh ill sneak that change into the release blog post

nickcolley:
Woop it's working and green :tada:

nickcolley:
Just noticed https://coveralls.io/builds/3206189/source?filename=..%2Flib%2Fsetup.js the files are not accessible, wonder if we need to config that?

nolanlawson:
@nickcolley Not sure what the fix is: https://github.com/lemurheavy/coveralls-public/issues/263

PR #4109, opened 2015-08-02
(#4107) - add coveralls coverage reports

nolanlawson:
Seems this should be all we need to add to get Coveralls support.
I already signed us up for the pouchdb/pouchdb org.

Here's an example of what a PR will look like:
* https://github.com/noopkat/avrgirl-stk500v2/pull/2


daleharvey:
Merged immediately because this is awesome :) https://github.com/pouchdb/pouchdb/commit/439174031fd9254365fc1195084c138e3bfe4c6c

Issue #3921, opened 2015-06-02
PouchDB takes 10-15 seconds to update 5,000 documents


nolanlawson:
Prompted by [a SO question](http://stackoverflow.com/questions/30536760/pouchdb-about-3x-slower-inserting-4800-documents-over-websql/30560249).

Fastest combo seems to be WebSQL without auto-compaction (in Chrome anyway). No combo can update these docs faster than 5-10 seconds. I don't know if we've reached the theoretical limit of IndexedDB/WebSQL or the CouchDB revision tree model, but I hope not. I have not profiled yet.

[Live test](https://gist.github.com/nolanlawson/84028dcfcdc7a0e340b7).

Results for IndexedDB, no auto-compaction (Chrome 43):

```
initial bulkDocs: 13147.961ms
(index):58 allDocs: 4159.247ms
(index):65 bulkDocs (update): 16072.757ms
(index):58 allDocs: 3765.193ms
(index):65 bulkDocs (update): 14294.470ms
(index):58 allDocs: 3508.268ms
(index):65 bulkDocs (update): 18408.058ms
(index):58 allDocs: 4030.973ms
(index):65 bulkDocs (update): 15931.974ms
(index):58 allDocs: 4018.197ms
(index):65 bulkDocs (update): 17534.399ms
```

Results for WebSQL, no auto-compaction (Chrome 43):

```
initial bulkDocs: 9348.997ms
(index):58 allDocs: 119.222ms
(index):65 bulkDocs (update): 11301.562ms
(index):58 allDocs: 125.094ms
(index):65 bulkDocs (update): 13029.120ms
(index):58 allDocs: 188.741ms
(index):65 bulkDocs (update): 11270.408ms
(index):58 allDocs: 159.362ms
(index):65 bulkDocs (update): 11863.117ms
(index):58 allDocs: 161.994ms
(index):65 bulkDocs (update): 12099.767ms
```

nolanlawson:
Profiled. `processDocs`, `updateDoc`, and `merge` seem to be the main culprits, on the JS side anyway. Chrome Dev Tools can be a little deceptive because they don't really show the time taken by IndexedDB/WebSQL.

nolanlawson:
Also: wow, WebSQL allDocs is 40 times faster than IndexedDB allDocs.

nolanlawson:
[...]

PR #4110, opened 2015-08-02
(#3921) - speed up updateDoc()

nolanlawson:
For now, I just want to see if the tests are passing.

This is a slightly dangerous optimization and may require a migration.
The fact that it seems to be working without migrations kinda worries
me; I think I might need to add a new migration test to catch it.

Also I have not measured the performance boost yet, but this is
pre-caching a lot of values as well as speeding up `revExists()`.


nolanlawson:
Also I don't plan on getting this into 4.0.0, so no rush.

nolanlawson:
I'm stunned this actually passed. This weekend I'll get some numbers on the perf boost and try to confirm whether or not we need a migration.

nolanlawson:
I've added a migration test, so I feel confident now that this PR is good to go.

Speedup is slight, but any speedup is good, right? Numbers are from a low-spec Acer Aspire laptop running Linux Mint for the `bulk-inserts` test:

Chrome 11 (websql): 33002ms -> 32277ms
Chrome 44 (idb): 15690ms -> 14988ms
iojs 2.5 (leveldb): 4878ms -> 4862md

daleharvey:
The failure is just a coveralls issue that seems to happen on branches that arent completely up to date
+1 on extra migration tests - https://github.com/pouchdb/pouchdb/commit/cd1357a6a63a048a71336a336e516a3d02afad6e

Issue #4094, opened 2015-07-31
genReplicationId fails if the root / is html


olafura:
PouchDB starts trying again and again to get the uuid from the root, and since the ajax code tries to parse it as JSON it fails,

http adaptors api.id is the problem it needs a try catch to use genDBurl if it fails.

chttp also doesn't seem to have a uuid in the json but that can probably be added.

I can provide a patch for this if you want.

daleharvey:
A patch for this would definitely be welcome, thanks, I think the first thing that would need to be fixed is a try / catch around the JSON parse in ajax.js, we have had an issue open for a while on that (https://github.com/pouchdb/pouchdb/issues/2670)

The http api.id function could then ignore any errors from the callback and use the uuid if available, if not just the host

kxepal:
Why PouchDB ever tries to get remote UUID?

nolanlawson:
@kxepal Seems it was ported from Erlang (see [these lines](https://github.com/pouchdb/pouchdb/blob/324c087b77c8dd2b06b7ccc942b370d46cc7a0db/lib/replicate/checkpointer.js#L187-L234)). But I'm unfamiliar with this part of the codebase; @daleharvey?

daleharvey:
So I cant tell if the couchdb implementation was changed significantly or this was just a mistake, however I am going to guess it was written that way to prevent failure in the case of the remote database hostname switching itself when we only stored the checkpoint on one node. 

It fairly certainly isnt needed now, but changing it isnt untrivial, need to make sure peoples existing checkpoints remain stable so a patch to just fallback correctly would be fine, then another to take the remote uuid fetch out, looks fairly certain that isnt what the current couch implementation looks like - https://github.com/apache/couchdb-couch-replicator/blob/master/src/couch_replicator_utils.erl#L86


https://github.com/apache/couchdb-couch-replicator/blob/master/src/couch_replicator_utils.erl#L86

kxepal:
@nolanlawson referenced part is about checkpoint docs comparison. They don't have any UUID nor require additional requests to do.

@daleharvey CouchDB implementation is about that replicator uses local UUID for replication ID generation: not source nor target ones (imagine server A runs remote-remote replication). So, for PouchDB that would be all the same: generate own UUID at once and store it in the db or somewhere else.

olafura:
@kxepal It will be interesting reading through how CouchDB does this.

I'm wondering if my patch is good enough for now.

[...]

PR #4114, opened 2015-08-03
Simpler id that is less likely to break #4094

olafura:
I've tried to find something more unique in the headers but it's hard to find anything that a proxy will not mess with.

### DIFF
--- a/lib/adapters/http/index.js
+++ b/lib/adapters/http/index.js
@@ -220,18 +220,7 @@ function HttpPouch(opts, callback) {
   };
 
   api.id = utils.adapterFun('id', function (callback) {
-    ajax({
-      headers: clone(host.headers),
-      method: 'GET',
-      url: genUrl(host, '')
-    }, function (err, result) {
-      /* istanbul ignore next */
-      if (err) {
-        return callback(err);
-      }
-      var uuid = result.uuid + host.db;
-      callback(null, uuid);
-    });
+    callback(null, genDBUrl(host, ''));
   });
 
   api.request = utils.adapterFun('request', function (options, callback) {


olafura:
#4094

nolanlawson:
Hm, it fails with `DASHBOARD_HOST is required`. Is this one of those things where only owners can get their PRs checked for code coverage? @zoepage did you have a workaround for this by any chance? Looking at [this code](https://github.com/hoodiehq/pouchdb-hoodie-sync/blob/7be8ce852e796c7d51e07ee81072b3f1407f07c6/.travis.yml#L22-L23) and it seems promising.

zoepage:
@nolanlawson If you add the `GH_repo` and coveralls key to travis, it should check the repo by default with every PR you get in, independent from the ownership / rights. (If you invite me to coveralls and travis, I can take a look).

Running coveralls in `after_success ` helps us the build won't fail, because it doesn't have any influence on the build. So this is what you'd need :)

daleharvey:
the `DASHBOARD_HOST` is our issue, we upload the coverage results to our host so we could build all that fancy graph stuff ourselves - https://github.com/pouchdb/pouchdb/blob/master/bin/run-coverage.js#L4

Can just stop calling + delete that file now, just call `COVERAGE=1 npm test`

daleharvey:
@olafura Sorry for the side track

For this commit, it is important we dont change the way id's are generated for existing clients, if they are using the uuid now they should continue to do so otherwise peoples replications will all start from 0 (reminding myself we should add a test to ensure that nobody breaks that)

I think the best short term fix is a try/catch around the JSON.parse and remove the `if (err)` from here

nolanlawson:
The Coveralls thing was a temporary bug in Coveralls itself. Should be fixed now.

I tend to agree with @daleharvey though that we need to maintain backwards compatibility. Else everybody who upgrades will have their replications start from scratch. Handling the error sounds like a good compromise.

daleharvey:
Ill write up tests for this shortly

daleharvey:
@olafura Hey, cheers for the patch, I came up with a pretty similiar solution but handles the backwards compatibility issues with session_id here - https://github.com/pouchdb/pouchdb/pull/4217, Apologies for not explaining it fully, will try to next time, hope you can get to in another contribution :)

olafura:
@daleharvey Thanks for that, I'm sorry about not being available to respond. Been really busy planning my move to Canada this week so hopefully I can help you out in the future.

PR #4125, opened 2015-08-05
(#4098) fix test.http.js against couchdb master

willholley:
Tests should not depend upon numeric sequence numbers else they fail against couchdb 2.X


daleharvey:
Awesome, thanks a lot for keeping on top of the couch-master suite, once it goes red then its easy for some of these to sneak in.

Merged in https://github.com/pouchdb/pouchdb/commit/77724e4de8745573e3cab89ff70b194d446d4039

PR #4129, opened 2015-08-05
(#3838) - Update request

daleharvey:


nolanlawson:
+1 when green

nolanlawson:
Failure seems genuine:

> Uncaught Error: too many listeners of type "error" added to EventEmitter. Max is 0 and we've added 2.

^ This is designed to catch memory leaks; not sure what it has to do with request, though.

daleharvey:
Yeh genuine failure, will take a look at this later

Issue #2239, opened 2014-05-17
TypeError in "npm run shell" while replicating


nolanlawson:
Do `npm run shell` and then paste this in:

```js
pouch = new PouchDB('foo');
pouch.put({_id: 'whatever'}).then(function() {
  pouch.replicate.to('http://localhost:5984/my_test_db2').on('error', function(err) {
    console.log(err);
    console.log(err.stack);
  }).on('complete', function (resp) {
    console.log(resp);
  });
});
```

It prints a TypeError:

```
TypeError: batch processing terminated with error
    at /Users/nolan/workspace/pouchdb/lib/replicate.js:207:12
    at tryCatch1 (/Users/nolan/workspace/pouchdb/node_modules/bluebird/js/main/util.js:64:19)
    at Promise$_callHandler [as _callHandler] (/Users/nolan/workspace/pouchdb/node_modules/bluebird/js/main/promise.js:708:13)
    at Promise$_settlePromiseFromHandler [as _settlePromiseFromHandler] (/Users/nolan/workspace/pouchdb/node_modules/bluebird/js/main/promise.js:724:18)
    at Promise$_settlePromiseAt [as _settlePromiseAt] (/Users/nolan/workspace/pouchdb/node_modules/bluebird/js/main/promise.js:896:14)
    at Promise$_fulfillPromises [as _fulfillPromises] (/Users/nolan/workspace/pouchdb/node_modules/bluebird/js/main/promise.js:1041:14)
    at Async$_consumeFunctionBuffer [as _consumeFunctionBuffer] (/Users/nolan/workspace/pouchdb/node_modules/bluebird/js/main/async.js:64:12)
    at Async$consumeFunctionBuffer (/Users/nolan/workspace/pouchdb/node_modules/bluebird/js/main/async.js:37:14)
    at Promise$_Scheduler (/Users/nolan/workspace/pouchdb/node_modules/bluebird/js/main/schedule.js:40:58)
    at process._tickDomainCallback (node.js:463:13)
```

daleharvey:
Removed in https://github.com/pouchdb/pouchdb/commit/b70b99300a6dc05042dc13be721159d56ffd2403

gr2m:
damnit, and I just wanted to add a shell to https://github.com/hoodiehq/pouchdb-hoodie-store for simpler debugging. Do you think this is a bad idea in general? Or was it just the implementation that you removed that was not good?

PR #4133, opened 2015-08-06
(#2239) - Remove npm run shell

daleharvey:


daleharvey:
This had a whole raft of bugs and hasnt proved to be very useful, I would like to do something along the same idea, but figure out a way to do it that isnt so restrictive / error prone

nolanlawson:
I agree; I've never used it. Let's also remove the description from TESTING.md.

daleharvey:
Good catch, forgot about that

daleharvey:
Merged

https://github.com/pouchdb/pouchdb/commit/b70b99300a6dc05042dc13be721159d56ffd2403

Issue #4117, opened 2015-08-04
API page takes a long time to load


nolanlawson:
This doesn't seem to be an appcache issue, because it happens every time you load http://pouchdb.com/api.html

It seems there's a lot of synchronous JavaScript that's running, because this page can get quite slow for the initial load. On a ~2008 era Linux-running Toshiba laptop I'm playing around with, it takes up to 15 seconds to load. Any way to speed this up?

nickcolley:
Shouldn't be taking 15seconds that's awful.

The page loads and then promise and callback examples get merged (incase you don't have javascript).

![image](https://cloud.githubusercontent.com/assets/2445413/9058591/424bf070-3a9a-11e5-85a0-211c8a6b4542.png)

I'll take a look at this, looks like we should batch the updates rather than doing them one at a time.

daleharvey:
clientHeight is going to force a reflow on every access, if this is happening predender then its probably what is being costly. Its weird that it is blocking the render since we have everything in the footer, but yeh its very noticeable for me but hopefully shouldnt be too hard to fix

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/253609994e99911d4be6f3fa8971826335de1c0a

I think it can likely be faster, but this is a huge improvement

PR #4136, opened 2015-08-08
(#4117) - Api page takes a long time to load

nickcolley:
So indeed hitting clientHeight constantly was what was so slow. I've split it out so it only happens when you use the tabs, and then it gets cached.

I experimented with viewport only focus but it was getting more complicated than it's worth. This gets the first render of the example boxes around 100ms so hopefully that improvement translates to the devices @nolanlawson was using.


daleharvey:
Awesome job :) this loads far faster and it also fixed the fragment links going to the right location

Its pretty funny that this was so obviously slower since we have appcache and every other page loads immediately, generally I would have just put it down to page load time

Merged in https://github.com/pouchdb/pouchdb/commit/253609994e99911d4be6f3fa8971826335de1c0a

nolanlawson:
Just published; much faster than before! Also I love the little "content updated, click to reload?" button. :heart:

Issue #2204, opened 2014-05-09
'options.doc_ids' matches wrong ids


aldelbal:
I am using replicate method with 'options.doc_ids' to match only certain documents.

*TEST 1*

db.replicate.from('https://mycouchdb/mydatabase',{
live : false,
doc_ids : 'config'
})

I have following docs in couchdb:
* doc 1: id = 'config'
* doc 2: id = 'fi'
* doc 3 : id = 'en'

Replicate get doc 1 ('config') and doc 2 ('fi'). It means that doc ids partially matching doc_ids are replicated as well.

TEST 2
db.replicate.from('https://mycouchdb/mydatabase',{
live : false,
doc_ids : 'configen'
})

I have following docs in couchdb:
* doc 1: id = 'config'
* doc 2: id = 'fi'
* doc 3 : id = 'en'

Replicate gets doc 2 ('fi') and also doc3 ('en'). It means that doc ids partially matching doc_ids are replicated as well.

According to coucndb documentation:
'doc_ids (array) – Array of document IDs to be synchronized'

So this has to be a faulty implementation either in PouchDB or in CouchDB ?

Note: this can be work-arounded by making sure your doc ids are not partial match of any other but this is not really sustainable approach when having many rows.

nolanlawson:
fixed by 07207d4ca83a2a6795ee3dd046815c306932eb9c

PR #4137, opened 2015-08-09
(#2204) - Fix error for invalid doc_ids

daleharvey:
I was going to do this as a guided bug fix thing, but the error mechanism (throwing) in replicate.js confused me so I figured it wasnt particularly friendly for new contributors, so just fixed it


nolanlawson:
Nice!

Issue #1974, opened 2014-04-18
Return helpful error if user provides both key and keys


nolanlawson:
CouchDB is adding it in 1.6.0; should as well: https://issues.apache.org/jira/browse/COUCHDB-2053.

daleharvey:
Closing out, our message isnt word for word identical here, ours is actually slightly more informative imo and not sure we need that level of consistency yet

PR #4138, opened 2015-08-09
(#1974) - Consistent error when providing keys + key

daleharvey:


daleharvey:
Same story, thought it would be a good first bug, wasnt great, I think we were doing ok with the error before but being consistent and having more tests isnt terrible

nolanlawson:
Seems to be consistently failing in the PouchDB Server tests:

```
 AssertionError: expected 'Some query parameter is invalid' to equal '`keys` is incompatible with `key`, `start_key` and `end_key`'
      + expected - actual
      +"`keys` is incompatible with `key`, `start_key` and `end_key`"
      -"Some query parameter is invalid"
```

Probably needs a fix in express-pouchdb to not just overwrite the error message.

daleharvey:
> Probably needs a fix in express-pouchdb to not just overwrite the error message.

The problem is that pouchdb-server is using its own copy of pouchdb, I believe it used to use some fancy trick to switch out its pouchdb for the current master, but looks to be broken

daleharvey:
actually that part seems to be working, screwed up my local test I think

daleharvey:
Yup your original theory was right - https://github.com/pouchdb/express-pouchdb/blob/master/lib/utils.js#L159

When I start fixing it it starts to cause other failures, thinking these should be moved to expected failures similiar to express-pouchdb-router. Contributors shouldnt need to be worrying about them and you shouldnt have to be fixing 2 additional servers for every patch that isnt compatible. We can land fixes on pouchdb without worrying and motivated persons can get them compatible + green again.

Hows that sound?

nolanlawson:
I agree that we shouldn't have to modify code in two places for every new commit, but I'd like to take this opportunity to remove the offending code from express-pouchdb rather than letting the master branch become red.

nolanlawson:
I actually still don't see where that message is coming from, either. Apparently it's from `exports.BAD_ARG` in `lib/deps/errors.js`, but I can't see where express-pouchdb is using that. This will require more debugging.

nolanlawson:
How about we modify the test so that it only runs for local databases, not remote? I would be surprised if CouchDB master or CSG have exactly the same error message anyway, and it feels flakey to check for the precise wording.

daleharvey:
https://github.com/pouchdb/express-pouchdb/blob/master/lib/utils.js#L159 modifies the error object that is actually returned, if we change that code it ends up breaking other tests

nolanlawson:
I'll take the time to clean up express-pouchdb next weekend. I don't think it should be trying to modify the errors that way; that smells to me like a problem that should be fixed in PouchDB itself instead of via monkey-patching. In any case, are you okay with not merging this right away until express-pouchdb is fixed?

daleharvey:
For running the tests locally, the only point of this change is to have consistency of the wording with CouchDB, we already behaved perfectly fine just with different wording in the error so if we arent going to test / enforce the same wording then really this can just be closed which I actually think is the best option I think.

I do still believe pouchdb-server should be moved to allowed failures and not block contributors patches if they were to come in and pass our core suite, I will bring that up if / when it happens. 

Will close this out, Im still not totally certain its a good idea to be testing word for word error messages with couchdb yet

nolanlawson:
It bothers me too that pouchdb-server has become a blocker for contributors. I think it either needs to be moved into the same repo, or we need to seriously clean it up so that it's not so fragile.

PR #4140, opened 2015-08-09
(#3921) - small optimizations to merge.js

nolanlawson:
According to performance profiling I've been doing, the big boost here should come from not calling `extend()`, which it turned out was unnecessary anyway. I also went through and removed as many functions-within-functions as I could, and migrated everything from `forEach()` to regular `for` loops.

When this is green, I'll provide some perf numbers to justify it. Looking at the Chrome profiler, though, this should help a lot.

### DIFF
--- a/lib/merge.js
+++ b/lib/merge.js
@@ -1,6 +1,4 @@
 'use strict';
-var extend = require('pouchdb-extend');
-
 
 // for a better overview of what this is doing, read:
 // https://github.com/apache/couchdb/blob/master/src/couchdb/couch_key_tree.erl
@@ -27,6 +25,20 @@ function binarySearch(arr, item, comparator) {
     }
   }
   return low;
+}
+
+function sortByPos(a, b) {
+  return a.pos - b.pos;
+}
+
+function sortByDeletedThenPosThenId(a, b) {
+  if (a.deleted !== b.deleted) {
+    return a.deleted > b.deleted ? 1 : -1;
+  }
+  if (a.pos !== b.pos) {
+    return b.pos - a.pos;
+  }
+  return a.id < b.id ? 1 : -1;
 }
 
 // assuming the arr is sorted, insert the item in the proper place
@@ -105,7 +117,8 @@ function doMerge(tree, path, dontExpand) {
     return {tree: [path], conflicts: 'new_leaf'};
   }
 
-  tree.forEach(function (branch) {
+  for (var i = 0, len = tree.length; i < len; i++) {
+    var branch = tree[i];
     if (branch.pos === path.pos && branch.ids[0] === path.ids[0]) {
       // Paths start at the same position and have the same root, so they need
       // merged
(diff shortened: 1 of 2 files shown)


lib/merge.js
```diff
@@ -286,11 +291,11 @@ PouchMerge.rootToLeaf = function (tree) {
     history.push({id: id, opts: opts});
     if (isLeaf) {
       var rootPos = pos + 1 - history.length;
-      paths.unshift({pos: rootPos, ids: history});
```
nolanlawson: `shift()` and `unshift()` are slow, `push()` and `pop()` are fast (-er).

nolanlawson:
true failures in the unit tests. I will fix

nolanlawson:
Okay, so AFAICT, our `merge()` function now actually mutates the object it is given as input. But it doesn't seem to matter, because whenever we do a merge, it's because we're going to update the single `metadata` object in the database anyway. So mutating the input tree is actually just fine, and in fact much better for performance than calling `extend()` all over the place.

However, this would normally break the `test.merge.js` unit test, so I modified it to redefine the constants every time. This is kinda risky, but since our integration tests are passing, I feel confident doing this.

lib/merge.js
```diff
@@ -194,9 +209,6 @@ function stem(tree, depth) {
 var PouchMerge = {};
 
 PouchMerge.merge = function (tree, path, depth) {
-  // Ugh, nicer way to not modify arguments in place?
-  tree = extend(true, [], tree);
-  path = extend(true, {}, path);
```
nolanlawson: ^ this is where we avoid `extend()`. The big perf boost should be here.

nolanlawson:
Performance boost for the `bulk-inserts` test (10 iterations in browser, 100 in Node for better granularity)

* iojs 2.5 (leveldb): 36855ms -> 36822ms
* Chrome 44 (idb): 15179ms -> 14471ms
* Chrome 11 (websql): 32476ms -> 30788ms

I admit, I wish the boost were bigger in Node. :( It's an improvement, but not a huge one.

daleharvey:
Following up on the coverage failure in https://github.com/pouchdb/pouchdb/issues/4142

Do you still think its worth it? modifying args in place is the default behaviour of javascript, so its arguably not confusing, I do think its terrible confusing behaviour and it turns this from a decoupled functional api to one that could be potentially confusing, however we are the only consumers. 

Having mixed feelings really

nolanlawson:
I have mixed feelings as well. I think I will split this into the micro-optimizations (which are fairly uncontroversial) and the big optimization (removing `extend`). Then I want some better measurements about how much exactly this is improving.

When I profile I consistently see `extendInner` very high on the list. It might not figure highly into the raw speed numbers of our perf test, but in terms of what is taking up the most *JavaScript* time it is quite high, which means it's the kind of thing that could be blocking the DOM. So yeah, I think it's imprortant, but we should be cautious about adding such stuff.

nolanlawson:
Closing in favor of two new PRs

Issue #4049, opened 2015-07-14
Retry doesnt work when started offline


daleharvey:

daleharvey:
Hrm, should have been fixed in https://github.com/pouchdb/pouchdb/issues/3768

daleharvey:
ok confirmed this is still a bug, for some reason the checkpointing is failing looking into it now, not sre why previous test isnt catching it

nolanlawson:
I tested manually using the Chrome dev tools to toggle offline and online, and it seemed to work. I may have missed something though.

daleharvey:
So the issue here is that we cant construct a http pouchdb instance while we are offline, it errors out and `isReady` never gets set to true, it doesnt look like we reconstruct the database on retry

daleharvey:
The test doesnt cover that because it overrides the databases methods after construction

daleharvey:
This isnt particularly easy to fix, we cant simply reconstruct the database because the replication may have been started with

```
var db = new PouchDB('...');
var remote = new PouchDB('....');
db.sync(remote);
```

I think this is mostly a bad decision we made early on with construction + taskqueue which we have seem come up a lot with errors during construction etc.

I think we want to move to a stateless constructor that can never fail, it means all of our error handling goes via a promise api, we can get rid of taskqueue and the various bugs it causes. In Firefox OS we generally do this with something like

```
function Constructor(opts) { 
  this.opts = opts;
}

function openDb (opts) { 
  if (dbs[opts.name]) { 
    return Promise.resolve(dbs[opts.name]);
  } 
  // else do construction etc
}

function post(doc) { 
  openDb(this.opts).then(function(idb) { 
     idb.createTxnWhatever
  });
}
```

triniwiz:
Anything i could do till there's a fix  ??

daleharvey:
`skipSetup` technically should do the same as my proposal in #4057, but in my testing it didnt so a little confused

nolanlawson:
fixed by f8c06ac76d321e84bdcab25f9cf02f5d648e06af

baversjo:
thanks for your hard work!

PR #4146, opened 2015-08-11
(#4049) - Stateless constructor for http adapter

daleharvey:


daleharvey:
Some how managed to get the constructor in some type of infinite loop, but this should mostly be how it can work

daleharvey:
So this mostly looks to be working, one small problem where I am likely doing something dumb

@nolanlawson it seems like your area of expertise, https://github.com/pouchdb/pouchdb/blob/4049/lib/adapters/http/index.js#L176 if the promise is in progress or fulfilled, I want to return it, if it is rejected then I want to try again, `setupPromise` never seems to get set to false like it should do https://github.com/pouchdb/pouchdb/blob/4049/lib/adapters/http/index.js#L228 though

nolanlawson:
Chrome Canary has a dev tools experiment where you can see a list of resolved and unresolved Promises. I'll try that out and give it a go.

daleharvey:
Ok woot, I think this is all good for review aside from pouchdb-server seems to be failing in the same way again

daleharvey:
And to explain, the old test for 'retry working offline', was that in the test `var db = new PouchDB` happened successfully before we started failing requests, but if you do var db = new PouchDB('http://url') while offline that construction would fail and it would be impossible to pass it to the replicator.

This changes it so `var db = new PouchDB` returns immediately without doing any setup on the server, it can never fail, when you start doing requests against that db it then does the setup, if we start if offline + retry, the subsequent requests will retry the full db setup until you are online and it succeeds

nolanlawson:
This causes a breakage in pouchdb-server, but everything else looks great. I'll file the pouchdb-server problem as a separate issue.

nolanlawson:
f8c06ac76d321e84bdcab25f9cf02f5d648e06af

PR #4149, opened 2015-08-12
(#3921) - micro-optimizations to merge.js

nolanlawson:
Same as #4140, but containing just the micro-optimizations and none of the in-place mangling of trees/paths.

I haven't profiled this change, but `merge.js` is called a lot, so if we're going to micro-optimize anywhere, this is the place to do it.

### DIFF
--- a/lib/merge.js
+++ b/lib/merge.js
@@ -29,6 +29,20 @@ function binarySearch(arr, item, comparator) {
   return low;
 }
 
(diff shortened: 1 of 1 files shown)


daleharvey:
Yeh these ones dont have any downsides so +1 - https://github.com/pouchdb/pouchdb/commit/7611524c0377213de456d736d04d22bf89c1105e

PR #4151, opened 2015-08-12
(#3921) - avoid extend() in merge.js

nolanlawson:
Same as #4140, but containing just the (risky) in-place mangling of trees/paths.

This is a harmless commit per our unit tests, but since it introduces slightly surprising behavior (`merge()` modifies the input objects), it needs a lot of justification from profiling. When I get a chance (or if someone else wants to hop on this!) I think I can demonstrate that this offers a big boost to our speed (especially in `bulkDocs()`), so it's well worth it.

### DIFF
--- a/lib/merge.js
+++ b/lib/merge.js
@@ -1,6 +1,4 @@
 'use strict';
-var extend = require('pouchdb-extend');
-
 
 // for a better overview of what this is doing, read:
 // https://github.com/apache/couchdb/blob/master/src/couchdb/couch_key_tree.erl
@@ -211,9 +209,6 @@ function stem(tree, depth) {
 var PouchMerge = {};
 
 PouchMerge.merge = function (tree, path, depth) {
-  // Ugh, nicer way to not modify arguments in place?
-  tree = extend(true, [], tree);
-  path = extend(true, {}, path);
   var newTree = doMerge(tree, path);
   return {
     tree: stem(newTree.tree, depth),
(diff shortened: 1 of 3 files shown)


nolanlawson:
OK, I added a new performance test, because none of our current ones were capturing what I wanted. This new one is called `basic-updates`, and it just inserts 100 documents and then updates them each 100 times.

This proves that this PR is an enormous performance boost. I was actually really surprised by how big it turned out to be. All tests are on my crappy Acer Aspire Linux laptop.

| Environment | Before | After |
| ---- | ---- | --- |
| iojs 2.5 (leveldb) | 40365ms | 15350ms |
| Chrome 44 (idb) | 70521ms | 40288ms |
| Firefox 41 (idb) | 65415ms | 34425ms |
| Chrome 11 (websql) | 57045ms | 41881ms |

I would like to test in Safari to see how JSCore is affected, but I don't have an Apple device handy.

nickcolley:
Wow this looks great! Nice work.

daleharvey:
Heh, ok with that type of speedup then big :+1: 

Same comment about the test failures as in the other PR

nolanlawson:
Yeah, the idb-alt and localstorage-down tests are permafail; must be some module that they both depend on. Gonna wait it out and just cross my fingers for now.

nolanlawson:
6764777f240a48d9063900090ba964111293c895

Issue #3883, opened 2015-05-28
allDocs doesnt support start_key as alias for startkey


marten-de-vries:
CouchDB supports ``start_key`` as an alias for ``startkey`` for views. At the very least that should be supported in ``express-pouchdb``, but it would be nice if plain PouchDB supported it as well. (To make migrating from CouchDB -> PouchDB easier, that's where I encountered this).

Same for a few others, both url endpoints and parameters (like ``end_key``):
- http://docs.couchdb.org/en/latest/api/ddoc/views.html
- http://docs.couchdb.org/en/latest/api/database/bulk-api.html

nolanlawson:
whoa, I've never seen those before. Yeah, I think PouchDB core should support these.

daleharvey:
So given some [inspiration](https://medium.com/@kentcdodds/first-timers-only-78281ea47455) I decided to stop fixing this after writing the test and let a new contributor do the fix. Hopefully the following steps are reasonably clear. If you run in to any issues then feel free to [get in touch](https://github.com/pouchdb/pouchdb#getting-help). 

 * Clone the PouchDB repo and setup locally, there is a [guide](https://github.com/pouchdb/pouchdb/blob/master/CONTRIBUTING.md#building-pouchdb) to help out.
 * Ensure your local setup is working with 'npm test', it will take around 2 minutes and should pass
 * Go to `inserturlwhenmerged` and remove the `.skip` from the test
 * Run `GREP=3883 npm test`, this should now be failing
 * The first part that needs fixed is [/lib/adapters/http/index.js](https://github.com/pouchdb/pouchdb/blob/master/lib/adapters/http/index.js#L703), open that file and fix it to ensure if `start_key` is defined instead of `startkey` we use `start_keys` value
 * Once thats done and you run the test, there should be 1 passing and one failing, you will now need to add the same logic to [/lib/adapter.js](https://github.com/pouchdb/pouchdb/blob/master/lib/adapter.js#L652) to fix it for the other adapters aside from http (websql, leveldb, indexeddb)
 * Once the tests are passing, that is it done, commit and push your changes and celebrate :)

Charlotteis:
So I tried to start this:
1) Cloned, setup and tested the pouchdb repo.
2) Copy and pasted the test you wrote to line 283 of `test.all_docs.js`
3) Found the mention of `startkey` in `adapters/http/index.js`.
4) Added new code to recognise `start_key` and get it's value.

[...]

Issue #3833, opened 2015-05-12
Remove change 'update/create/delete' events


daleharvey:

nolanlawson:
yes please :)

nolanlawson:
fixed by f8629e3be86e619411598848b368ff9cd7846e56

nolanlawson:
Oh whoops looks like the wrong issue got tagged in the commit message. No biggie though.

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/83a4a0b84658159fef440028707c05615b10ced7

PR #4155, opened 2015-08-12
(#3883) - Tests for start_key alias

daleharvey:
Want to commit a failing test then write up how to fix it

PR #4156, opened 2015-08-12
(#3833) - Support start_key and end_key alias for allDocs

Charlotteis:
**Issue:** #3883 

This PR implements the `start_key` alias for `allDocs`.

I've written things quite verbose in some places. I'm new to pouchdb (this is the first time I've interacted with it!) but I wrote the solution in a way that I think would be easier to read than some alternatives.

I hope I've done a good job and welcome feedback if I haven't!
Thanks @daleharvey for the great breakdown in what to do.

:panda_face:

### DIFF
--- a/lib/adapters/http/index.js
+++ b/lib/adapters/http/index.js
@@ -695,6 +695,10 @@ function HttpPouch(opts, callback) {
     if (opts.key) {
       params.push('key=' + encodeURIComponent(JSON.stringify(opts.key)));
     }
+    
+    if (opts.start_key) {
+      opts.startkey = opts.start_key;
+    }
 
     // If opts.startkey exists, add the startkey value to the list of
     // parameters.
@@ -703,6 +707,10 @@ function HttpPouch(opts, callback) {
     if (opts.startkey) {
       params.push('startkey=' +
         encodeURIComponent(JSON.stringify(opts.startkey)));
+    }
+
+    if (opts.end_key) {
+      opts.endkey = opts.end_key;
     }
 
     // If opts.endkey exists, add the endkey value to the list of parameters.
--- a/lib/adapter.js
+++ b/lib/adapter.js
@@ -650,6 +650,12 @@ AbstractPouchDB.prototype.allDocs =
   }
   opts = utils.clone(opts);
   opts.skip = typeof opts.skip !== 'undefined' ? opts.skip : 0;
+  if (opts.start_key) {
+    opts.startkey = opts.start_key;
+  }
+  if (opts.end_key) {
+    opts.endkey = opts.end_key;
+  }
   if ('keys' in opts) {
     if (!Array.isArray(opts.keys)) {
       return callback(new TypeError('options.keys must be an array'));
--- a/tests/integration/test.all_docs.js
+++ b/tests/integration/test.all_docs.js
@@ -280,6 +280,16 @@ adapters.forEach(function (adapter) {
       });
     });
 
+    it('3883 start_key end_key aliases', function () {
+      var db = new PouchDB(dbs.name);
+      var docs = [{_id: 'a', foo: 'a'}, {_id: 'z', foo: 'z'}];
+      return db.bulkDocs(docs).then(function() {
+        return db.allDocs({start_key: 'z', end_key: 'z'});
+      }).then(function (result) {
+        result.rows.should.have.length(1, 'Exclude a result');
+      });
+    });
+
     it('test total_rows with a variety of criteria', function (done) {
       this.timeout(20000);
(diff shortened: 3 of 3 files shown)


Charlotteis:
I think that the work was already started here: https://github.com/pouchdb/pouchdb/pull/4155.

lib/adapters/http/index.js
```diff
@@ -705,6 +705,15 @@ function HttpPouch(opts, callback) {
...
+    // If opts.start_key exists, add the start_key value to the list of
+    // parameters.
+    // If start_key is given then the returned list of documents will
+    // start with the document whose id is start_key.
+    if (opts.start_key) {
+      params.push('start_key=' +
+        encodeURIComponent(JSON.stringify(opts.start_key)));
+    }
+
     // If opts.endkey exists, add the endkey value to the list of parameters.
```
daleharvey: So this will work but I think it would be better to avoid a little duplication of the logic here. If we did

   if (opts.start_key) { 
     opts.startkey = opts.start_key;
   }

Then it means we leave all the encoding / JSON work in one place and makes it clear/explicit that if you pass in `start_key` then all we (or couchdb) does is pretend you said `startkey`. I just realised I wrote those steps but didnt actually explain the bug properly, sorry.

Another note is we need to do the same logic for `end_key` => `endkey`

lib/adapters/idb/allDocs.js
```diff
@@ -59,7 +59,7 @@ function handleKeyRangeError(api, opts, err, callback) {
 function idbAllDocs(opts, api, idb, callback) {
 
   function allDocsQuery(opts, callback) {
-    var start = 'startkey' in opts ? opts.startkey : false;
+    var start = 'startkey' || 'start_key' in opts ? opts.startkey : false;
```
daleharvey: So theres 2 issues here, the first one is mostly because as mentioned above I didnt explain the bug properly. 

first is when the use gives us `start_key`, we want to treat it exactly as if they gave us `startkey`. The logic here wont quite do that but the code will be exactly the same as last time, ie `if (opts.start_key) { opts.startkey = opts.start_key; }`

The second is due to how our code is setup which is pretty confusing, you have fixed this in 4 places (http, leveldb, websql and idb), http is special for now but for the rest you can fix them all in the same place. 

We call the function @ https://github.com/pouchdb/pouchdb/blob/master/lib/adapter.js#L645 which will in turn call these implementations, so any common logic especially with the parameters can go in there once and be fixed for all 3 adapters (like how we set a default value for skip @ https://github.com/pouchdb/pouchdb/blob/master/lib/adapter.js#L652)

So if you do what you did for the http adapter and put in that function just around where we do the `opts.skip` stuff, it will only need fixed in the one place.

daleharvey:
Hey, thanks so much for this. It was a funny coincidence I noticed you asking right  for a good starter bug right after I started working on this. Its the first time I tried something writing a bug up like this and realised I didnt do a good job explaining the actual bug so there are things still to have fixed. But hopefully the comments clear up any confusion.

Charlotteis:
@daleharvey The comments helped loads! I can now say I know how to add alias keys to this specific thing (`allDocs`). Also it helped me to think about not being too verbose in the future and reduce the computation done. I think I have done what was asked now, the tests certainly pass :)

Thank you for being so helpful, I hope I can help out more in the future.

Charlotteis:
Ah, looks like I left some stray code in. :(

nolanlawson:
Wow, can't believe how fast this got implemented. :)

Thanks @Charlotteis, looks great to me. There's still a bit of verbosity, but that comes from the fact that `http.js` doesn't extend everything from `adapter.js`, so there are actually quite a few places where we have to duplicate code between the two, unfortunately.

But that being said, LGTM, +1 when green.

nickcolley:
Woop thanks for picking this one up @Charlotteis :tada: :balloon: :D

daleharvey:
That was quick and from the looks of it this looks perfect, :+1: although will wait for the tests to be certain (our tests can be very particular even when your certain they cant break anything). Thanks :)

Charlotteis:
Yahoo! :tada:
Thank you all for being so encouraging :100:

Charlotteis:
The build is still running so I'll wait for it to finish, but looks like we've had a Firefox failure and I have no idea why.

daleharvey:
Those test failures are almost certainly false negatives, our test suite isnt 100% reliable yet, just need to restart them

Charlotteis:
Phew, good to know :)

nolanlawson:
Green, ready to go. And coverage increased by 0.03%. :)

Charlotteis:
Wahoo!

PR #4159, opened 2015-08-14
(#4159) - avoid extra http call for destroy()

nolanlawson:
There is no reason to be doing the extra http call
to the `_local/dependentDbs` document for http
databases. This is something that slows down our
tests a lot, because we are constantly destroying
databases, so hopefully this fix should speed up
our test suite.

### DIFF
--- a/lib/adapter.js
+++ b/lib/adapter.js
@@ -808,6 +808,12 @@ AbstractPouchDB.prototype.destroy =
       callback(null, resp || { 'ok': true });
     });
   }
+
+  if (self.type() === 'http') {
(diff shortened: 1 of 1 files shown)


daleharvey:
So the 2 failing PR's are green on the last commit and failing in both your PR's, assuming theres a dep update failing them somewhere. 

I would :+1: and move those to allowed_failures until they get fixed, no point on blocking work while we fix them, ill try to later today

nolanlawson:
fb43b5fe5e3515406c5bc9251c62b7740dca3289

Issue #3444, opened 2015-01-24
Inconsistent date deserialisation


mtford90:
Hi - I'm running PouchDB 3.2.1.

The following test succeeds:

```js
it('get date', function (done) {
    db.post({date: new Date()}).then(function (resp) {
        db.get(resp.id).then(function (doc) {
            assert.ok(doc.date instanceof Date);
            done();
        });
    }).catch(done);
});
```

However this one fails:

```js
it('query date', function (done) {
    db.post({date: new Date()}).then(function () {
        db.query({
            map: function (doc) { emit(doc._id, doc) }
        }).then(function (objs) {
            assert.equal(objs.rows.length, 1);
            assert.ok(objs.rows[0].date instanceof Date);
            done();
        }).catch(done);
    }).catch(done);
})
```

In the second case, the date field is a `string` rather than a `Date`. This should probably be consistent - i.e. both being dates?

nolanlawson:
Yeah, you need to turn all your dates into strings/numbers before storing: https://github.com/pouchdb/pouchdb/issues/2351

mtford90:
So saving down a `Date` object is essentially undefined behaviour then? Might it be a good idea to have a warning/error for this?

nolanlawson:
Yes, we should probably add this to the [common errors](http://pouchdb.com/errors.html). A pull request would be very welcome! :)

mtford90:
Good idea - on my todo list :)

Issue #3968, opened 2015-06-15
Doc props like .constructor or .watch are lost in 3.2.1–3.6.0


ermouth:
Quite artificial example, but issue already took place ~year ago and was marked as fixed, see #2080

Repro: http://jquerymy.com/pouchdb/proploss2.html

Note, that result of the test is incorrect in all browsers, moreover result is different in FF and others.

nolanlawson:
Yeah, the reason we use that jQuery 1.9.0 version of extends is because other libraries that we tried have bugs with Ember due to their manipulation of the global Object prototype (https://github.com/pouchdb/pouchdb/issues/2158). I admit this is a bad bug, though, and needs to be fixed.

nolanlawson:
@calvinmetcalf @daleharvey What is the solution here? Should we hack `pouchdb-extend` to keep these properties in, or should we just replace `pouchdb-extend` with `JSON.parse(JSON.stringify(obj))`? Personally I wouldn't mind the latter, because it would also fix those errors where people are surprised by IndexedDB's handling of Dates and other special types compared to the other adapters.

calvinmetcalf:
JSON.parse(JSON.stringify(obj)) is probably the easiest in all honesty

nolanlawson:
^ I added a failing test, so anybody who would like to fix it is welcome to. :)

ermouth:
JSON.parse(JSON.stringify(obj)) solution is very fast, also this approach solves race issue #3955, but it‘s always a deep clone, so in cases of large docs there can be significant GC-induced stalls.

nolanlawson:
fixed by 870248e7b97a3a1a9c737abad8ebebf9b07ead3f

PR #4162, opened 2015-08-15
(#3444 #3921) - use custom clone, fix Dates

nolanlawson:
I have a few goals with this:

* improve speed of `extend()` and `clone()`, which is still a perf drain per the profiler
* remove `pouchdb-extend`, which is a weird jquery `extend()` fork and we can do better
* fix the consistency of Date parsing, because I realized we can kill two birds with one stone, and since we're always cloning `docs` when we do `bulkDocs()`, we might as well stringify the `Date`s while we're doing it, so people can stop being confused by the inconsistency between IndexedDB and other backends (with no effect on performance).

BTW to justify my `clone()` implementation compared to `JSON.parse(JSON.stringify(obj))`, check out http://jsperf.com/deep-copy-vs-json-stringify-json-parse/15

When this is green, I will provide performance numbers.

### DIFF
--- a/lib/adapter.js
+++ b/lib/adapter.js
@@ -106,7 +106,7 @@ function allDocsKeysQuery(api, opts, callback) {
     offset: opts.skip
   };
   return Promise.all(keys.map(function (key) {
-    var subOpts = utils.extend(true, {key: key, deleted: 'ok'}, opts);
+    var subOpts = utils.extend({key: key, deleted: 'ok'}, opts);
     ['limit', 'skip', 'keys'].forEach(function (optKey) {
       delete subOpts[optKey];
     });
@@ -722,7 +722,7 @@ AbstractPouchDB.prototype.bulkDocs =
     opts = {};
   }
 
-  opts = utils.clone(opts);
+  opts = utils.clone(opts || {});
 
   if (Array.isArray(req)) {
     req = {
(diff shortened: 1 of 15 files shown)


nickcolley:
+1 for the consistent date formatting.

nolanlawson:
OK, this doesn't seem to actually impact performance much, so instead I'm going to argue for it because it makes the code smaller and simpler (due to removing `pouchdb-extend`). The `pouchdb.min.js` gzip size went from 48406 to 47955.

nolanlawson:
Green!

nolanlawson:
Dang, I just did a more detailed profile and I see that the new `isPlainObject()` method is actually taking up a lot of time, probably due to that for-loop.

nolanlawson:
On second review, that was a false alarm. When I run the `basic-updates` test over and over, I see pretty much no difference between the old version and the new version. I just got worried because I did an ad-hoc profile with node-inspector and I thought I saw a lot of usage in `isPlainObject()`.

This is :+1: imo.

daleharvey:
Yup this is nice, we should probably take a look at https://github.com/pouchdb/pouchdb/issues/3968 again off the back of this patch

Merged in https://github.com/pouchdb/pouchdb/commit/f17d58a95941e7f2bb8d2505800decbe0435cf0f

Issue #4164, opened 2015-08-15
adb-alt and localstorage backends permafailing


daleharvey:

nolanlawson:
I figured it out: the culprit is `buffer` at v3.4.1.

What's happening is just that the `big png` test is now timing out, because there's [a de-optimized try/catch](https://github.com/feross/buffer/blob/a2dbe7afb66675139697de2758d46656d3f24538/index.js#L1408-L1414) that is taking up all the self time:

![screenshot from 2015-08-15 10 06 26](https://cloud.githubusercontent.com/assets/283842/9289194/7070692e-4335-11e5-818d-9ca81b55c4ae.png)

nolanlawson:
Hah, 3.4.2 fails for a totally different reason: `Maximum call stack size exceeded`. The offending line is [this one](https://github.com/feross/buffer/blob/4c2977b9a1410bae33041a14f0fd5555b8de4c7c/index.js#L703), which is being called on an Array of size 652572 leading to the call stack exceeded error in Chrome 44.

nolanlawson:
fix proposed: https://github.com/feross/buffer/pull/70

nolanlawson:
Looks like this will be fixed with buffer 3.4.3! :)

daleharvey:
Fixed - https://github.com/pouchdb/pouchdb/commit/fa72fb72c9dfb776ee8261e987b1da20f909c079

PR #4165, opened 2015-08-15
(#4164) - increase timeout for slow test

nolanlawson:
Easy way to fix this problem, at least temporarily. In the meantime
I'll look into `buffer` and see if 3.4.2 fixes the slowness
or if we need to report it as a bug.


nolanlawson:
This would fix buffer @3.4.1, but @3.4.2 now has a totally different failure, so there's no point in merging this.

PR #4166, opened 2015-08-16
(#4166) - simpler Function.bind shim for phantomjs

nolanlawson:
I always disliked the complexity of managing `ES5_SHIM=true`.
As it turns out, the only thing we really need to shim for phantom is
`Function.prototype.bind`, so it's much simpler IMO to just
include a basic shim (provided by the `pmind` module) and not
have to explicitly include all of `es5-shim`.


daleharvey:
I have no opinions on `pmind` vs `es5-shim`, eithers fine for me. However I dont want to remove the fact that we are explicitly adding it, seems like a small issue but adding things globally to our test environment makes it extremely easy for subtle bugs to get introduced that would have pouchdb fail entirely when used, but passing in our suite.

Alternatively we can just explicitly not support ie8, we dont test in it, have other reported problems with it and its very much at its end of life

nolanlawson:
pbind is very simple, and it doesn't modify the environment in any way unless `!Function.prototype.bind`. Here's [the code](https://github.com/evanlucas/pbind/blob/master/index.js).

My main complaint is just that it's so awkward to have to remember to put `ES5_SHIM=true`, and in fact the es5-shim is modifying the environment a lot more than we really need to for PhantomJS. `Function.prototype.bind` is required to even get Mocha to run, let alone PouchDB, but other than that we need zero shims in order to pass the tests. (I've given up on IE 8/9.)

I still stand by this commit in terms of simplifying the test process; I can't imagine what subtle bugs it could introduce, and if anything we're being more explicit about only supporting ES5-enabled browsers.

daleharvey:
Its a test cleanliness thing, we start testing in react-native or some other exotic environment which may or may not have `bind` and we dont know whether its needed or not, its an extreme example but if a browser comes out and breaks `bind` we do not know about it.

It seems like its a small polyfill thats easy to keep track of, but eventually we update phantom and dont need it, its just sitting in our global test environment without us noticing, we do the same with a few other polyfills for older envs and suddenly we dont know what envs require what polyfills. This isnt a slippery slope thing, in early pouch we had quite a few times where pouchdb was entirely broken in the browser but passing all test suites fine due to polluting the global scope.

Switching to pmind is absolutely fine, or removing <es5 support entirely would probably be ok, even just adding the polyfill to pouchdb itself. But I really dont want to add globals (which pouchdb requires to function in some cases), unconditionally to the global test environment.

nolanlawson:
Fair enough; I would prefer to just close this issue then. Let's keep the es5-shim around. Maybe we'll need it anyway if we can ever get iE9 passing. :)

PR #4168, opened 2015-08-16
(#4168) - increase EventEmitter max to 20

nolanlawson:
So I started up PouchDB Server and created a single database,
and I saw that I had already hit the EventEmitter warning.

This tells me that we are exceeding the limit of 10 through normal
usage, which explains why so many people have complained about this
starting with 4.0.0.

I don't think it should be the responsibility of users to increase
this limit unless it really, really indicates a memory leak. We have
lots of tests in place to ensure we don't leak memory, and we are the
ones adding all these event listeners to the PouchDB object, so
I would rather we just increase the limit and handle it automatically
for users.

### DIFF
--- a/docs/errors.md
+++ b/docs/errors.md
@@ -77,12 +77,12 @@ There is a limit of one database per app in some versions of the Android WebView
 
 If you see this warning:
 
-    (node) warning: possible EventEmitter memory leak detected. 11 listeners added.
+    (node) warning: possible EventEmitter memory leak detected. 21 listeners added.
     Use emitter.setMaxListeners() to increase limit.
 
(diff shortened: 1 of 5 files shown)


nickcolley:
I mentioned this on IRC but I don't hang around long enough I guess, I think this is a good idea as long as we're sure it's from normal use.

nolanlawson:
I just ran the `pouchdb-find` tests and the warning came up again. Seems to be happening a lot.

daleharvey:
I would like to take the time in the future to see if we can reduce the number of event emitters we create, it seems like we shouldnt need to be pushing the default config here, but in the meantime I agree this will be better for users.

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/691df0d6b7fac08d8a281f169c2381f6424510d6

nolanlawson:
Hold up, I just realized this is a true memory leak. Will fix.

nolanlawson:
Oops, didn't mean to reopen this.

Issue #3424, opened 2015-01-20
Implement _bulk_get aka _bulk_revs


nolanlawson:
Counterpart to https://github.com/pouchdb/express-pouchdb/issues/196.

nolanlawson:
It's debatable, though, whether we actually want to implement a `db.bulkGet()` or `db.bulkRevs()`. For local-local replication it doesn't offer any benefit, and may actually confuse users.

As a first pass, though, we can simply update `replicate.js` to use a server-side bulk endpoint if it exists.

dholth:
See also https://github.com/dholth/pouchdb-bulk-get

robertkeizer:
+1

nolanlawson:
@daleharvey On the CouchDB side ([COUCHDB-2310](https://issues.apache.org/jira/browse/COUCHDB-2310)), there definitely seems to be an "intent to ship" in both 1.x and 2.x.

So how do you feel about adding `db.bulkRevs()`? I'm leaning towards implementing it at the PouchDB API level, i.e. not simply an endpoint on PouchDB Server that wraps `Promise.all` and `db.get`. Although most endusers will never use it themselves, it's nice for consistency's sake.

daleharvey:
I am not so sure about this, if couchdb do add it then we probably have to for compat, but I think we would be able to spec and implement stream before this bulkrevs is ever released and replication-stream is a much more desirable end goal, going from what we have now to adding bulkDocs / bulkRevs (+detection) to adding Streaming + detection doesnt seem ideal

nolanlawson:
So you think we should just jump directly to the replication-stream stuff?

robertkeizer:
I'm finding that a middleware layer that performs the replication-stream stuff works fairly well. Have some hacky code working for express-pouchdb but need to clean it up before committing.

Personally I'd like to see Pouch implement this without replication-stream functionality as soon as couch has it.

dholth:
_bulk_revs (a more efficient API that can replace the 1-at-a-time
GET API) is useful for replication, but it's not necessarily just
about replication. You might just want to get a lot of things with a
single call.


On Mon, Feb 9, 2015, at 08:50 AM, Robert Keizer wrote:
> I'm finding that a middleware layer that performs the
> replication-stream stuff works fairly well. Have some hacky code
> working for express-pouchdb but need to clean it up before committing.


> Personally I'd like to see Pouch implement this without
> replication-stream functionality as soon as couch has it.


[...]

PR #4169, opened 2015-08-16
WIP - wait for +1 from Couch community - (#3424) - implement bulkGet()

nolanlawson:
This is published now in pouchdb-server and express-pouchdb,
so I would like to start testing it.


nolanlawson:
If we would like to disable this before we publish the next version, we can easily do so by changing a flag in `http/index.js`. Else if CouchDB folks are ready to sign off on the implementation, then we can just ship it and give PouchDB Server users a much faster replication. :)

nolanlawson:
Cool, it's green. :) And based on the [Travis logs for the pouchdb-server test](https://s3.amazonaws.com/archive.travis-ci.org/jobs/75845326/log.txt), I can see that the `_bulk_get` endpoint is indeed being hit.

Related: https://github.com/pouchdb/express-pouchdb/pull/208 and https://github.com/apache/couchdb-chttpd/pull/33

nolanlawson:
Note that the version of pouchdb-server/express-pouchdb out there in the wild is not actually using `_bulk_get`, because it relies on `db.bulkGet()` existing; otherwise the endpoint just throws a 400 (and the client ignores it). So we would still need to do another pouchdb-server release in order to really get this out to users.

However, since our tests tie pouchdb-server's pouchdb dependency to the current commit, then this test is indeed hitting `_bulk_get` and it is working.

If you would like to test pouchdb-server yourself with the `_bulk_get` endpoint, just do `npm install -g pouchdb-server@pouchdb/pouchdb-server#experimental-bulk-get`.

lib/deps/bulkGetShim.js
```diff
@@ -0,0 +1,77 @@
...
+    // just use the first request as the "template"
+    // TODO: The _bulk_get API allows for more subtle use cases than this,
+    // but for now it is unlikely that there will be a mix of different
+    // "atts_since" or "attachments" in the same request, since it's just
+    // replicate.js that is using this for the moment.
+    // Also, atts_since is aspirational, since we don't support it yet.
+    var docOpts = pick(docRequests[0], ['atts_since', 'attachments']);
+    docOpts.open_revs = docRequests.map(function (request) {
+      // rev is required, open_revs disallowed
+      return request.rev;
```
kxepal: `rev` field is optional and if it missed. `open_revs=all` semantic should be used.

nolanlawson: Ah okay, thanks!

nolanlawson: ^ @daleharvey let's leave this in and make it a "first timer" bug.

daleharvey:
I am really not a fan of bulk_get, between all_docs, _changes and get and bulk_get it seems we have a lot of options that all support slightly different parameters for slightly different use cases without a clear distinction of which should be used when.

However if couch are on track to implement it then it makes the decision for us, and good job on getting it implemented with backwards compatible support, I think until the couchdb version has landed this should be considered experimental in case there are any api changes. 

So happy to get this merged, but there are no tests being added here?

nolanlawson:
Yup, that's the point; the test is the fact that it uses the new APIs for replication.

I guess I could add explicit tests for `db.bulkGet()`, but like you, I'm wary of actually elevating this to a user-facing API since it's confusing. Note that I also didn't document it.

I'm going to test CouchDB master w/ @kxepal's branch this weekend and make sure that my tests pass. Then I would like to hold off until after 4.0.1, because I want to do an "intent to ship"-style email to inform everyone else in the Couch community that we intend to ship this features for PouchDB Server and PouchDB, so they'd better speak now or hold their peace. :)

nolanlawson:
I would like to merge this into 5.0.0. Should we do some kind of "intent to ship" email to the Couch mailing list or something? @daleharvey what's the right etiquette here?

daleharvey:
I think a message on the couchdb mailing list with an "intent to ship" seems like a good idea, see if there is any final discussion about compatibility, I would like to see what the tests look like when run against a couchdb-master implementation

willholley:
Following my comment in https://github.com/apache/couchdb-chttpd/pull/33, I'd suggest adding some validation to the responses we get from bulk_get. The effect of the current CouchDB implementation not including _revisions, for example, is to introduce conflicts in PouchDB, rather than fail the replication.

willholley:
so I think the CouchDB implementation is correct and PouchDB is currently wrong. We need to be sending revs=true and attachments=true as query parameters rather than fields in the request body. After that we should be good to go...

nolanlawson:
@willholley It looks like pouchdb-server actually accepts them either in the body or as query params. Is that acceptable? https://github.com/pouchdb/express-pouchdb/blob/1808abd579675aac8dfc4def703683ff5640a8bb/lib/routes/bulk-get.js#L33-L38

nolanlawson:
superseded by https://github.com/pouchdb/pouchdb/pull/4335

PR #4182, opened 2015-08-22
(#4168) - fix constructor memory leak

nolanlawson:
I apologize for my previous "fix" (https://github.com/pouchdb/pouchdb/pull/4168),
which just papered over the problem. I found a test that
fails (calling `new PouchDB()` multiple times), and I've implemented
a fix. As a bonus, we no longer have to increase the max listeners
from 10, either in PouchDB itself or in our tests.

### DIFF
--- a/docs/errors.md
+++ b/docs/errors.md
@@ -77,26 +77,22 @@ There is a limit of one database per app in some versions of the Android WebView
 
 If you see this warning:
 
-    (node) warning: possible EventEmitter memory leak detected. 21 listeners added.
+    (node) warning: possible EventEmitter memory leak detected. 11 listeners added.
     Use emitter.setMaxListeners() to increase limit.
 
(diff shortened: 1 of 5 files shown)


docs/errors.md
```diff
@@ -77,27 +77,23 @@ There is a limit of one database per app in some versions of the Android WebView
...
 
 If you're sure it's not a memory leak, though, you can increase the limit by doing:
 
 {% highlight javascript %}
-PouchDB.setMaxListeners(20); // or 30 or 40 or however many you need
-{% endhighlight %}
-
-Sometimes the EventEmitter is the `PouchDB` object itself, though, and in that case you need to do:
-
-{% highlight javascript %}
```
nolanlawson: It's no longer possible for the `PouchDB` constructor to have too many listeners, because we only ever add one. However, `db` objects can still exceed the limit.

nickcolley:
Great catch!

daleharvey:
Yup very nice, cheers - https://github.com/pouchdb/pouchdb/commit/f1a7cbfe5b4070f693f777eb038396fd1ef2238c

PR #4183, opened 2015-08-22
(#3921) - more optimizations to merge.js

nolanlawson:
I've been motivated to work on this recently, because I noticed that `pouchdb-server --in-memory` is actually *slower than CouchDB*. When I run the `basic-updates` perf test, I get ~17.5s for CouchDB and ~22s for PouchDB Server. This to me is pretty unreasonable, since one is in-memory and the other one writes to disk.

After profiling, I saw that `isDeleted` and `winningRev` were still the biggest performance-killers. After this fix, not so much. Now the time is down to about 21s.

Before:

screenshot from 2015-08-22 17 59 31

After:

screenshot from 2015-08-22 18 00 41

So TLDR we are still slower than CouchDB even when we run in-memory, but this is an improvement. I think next I am going to have to upgrade my tools and use some better profilers than the node-inspector profiler, so that I can find V8 deoptimizations and poor memory usage (garbage collection seems to be taking up a lot of time).

### DIFF
--- a/lib/adapters/leveldb/index.js
+++ b/lib/adapters/leveldb/index.js
@@ -61,6 +61,18 @@ var safeJsonEncoding = {
   type: 'cheap-json'
 };
 
+// winningRev and deleted are performance-killers, but
+// in newer versions of PouchDB, they are cached on the metadata
+function getWinningRev(metadata) {
+  return 'winningRev' in metadata ?
+    metadata.winningRev : merge.winningRev(metadata);
+}
+
+function getIsDeleted(metadata, winningRev) {
+  return 'deleted' in metadata ?
+    metadata.deleted : isDeleted(metadata, winningRev);
+}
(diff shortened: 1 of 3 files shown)


nolanlawson:
All tests were run on an Acer Aspire with Linux Mint. You can load up Chrome Dev tools, then Profile, then Load, then load the [before file](https://s3.amazonaws.com/nolanlawson/pouchdb/perf/express_pouchdb_memdown_before_20150822T160148.cpuprofile) and [after file](https://s3.amazonaws.com/nolanlawson/pouchdb/perf/express_pouchdb_memdown_after_20150822T174830.cpuprofile).

daleharvey:
https://github.com/pouchdb/pouchdb/commit/70761f2c6003e91f8a886a1d5968c34cb27dc047

PR #4185, opened 2015-08-22
(#3921) - optimize rootToLeaf()

nolanlawson:
Perf tests showed this was taking up a significant amount of
time as well. Basically everything in `merge.js` needs to be
optimized to the extreme, as far as I can tell.

The big detriment of these kinds of changes is that the code
becomes less readable, but I think `merge.js` is pretty opaque
and CS-y as it is, so I don't mind so much. Also this removes
all external dependencies from that particular function.

### DIFF
--- a/lib/merge.js
+++ b/lib/merge.js
@@ -304,17 +304,28 @@ PouchMerge.collectConflicts = function (metadata) {
   return conflicts;
 };
 
-PouchMerge.rootToLeaf = function (tree) {
+// build up a list of all the paths to the leafs in this revision tree
+PouchMerge.rootToLeaf = function (revs) {
   var paths = [];
-  PouchMerge.traverseRevTree(tree, function (isLeaf, pos, id, history, opts) {
-    history = history ? history.slice(0) : [];
+  var toVisit = revs.slice();
+  var node;
+  while ((node = toVisit.pop())) {
+    var pos = node.pos;
+    var tree = node.ids;
+    var id = tree[0];
+    var opts = tree[1];
+    var branches = tree[2];
+    var isLeaf = branches.length === 0;
+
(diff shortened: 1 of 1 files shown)


lib/merge.js
```diff
@@ -304,17 +304,28 @@ PouchMerge.collectConflicts = function (metadata) {
   return conflicts;
 };
 
-PouchMerge.rootToLeaf = function (tree) {
+// build up a list of all the paths to the leafs in this revision tree
+PouchMerge.rootToLeaf = function (revs) {
   var paths = [];
-  PouchMerge.traverseRevTree(tree, function (isLeaf, pos, id, history, opts) {
-    history = history ? history.slice(0) : [];
```
nolanlawson: There is a consistent trend here: anything using `traverseRevTree` tends to be slow, because 1) it creates inner functions that can't be inlined as efficiently by V8, and 2) it tends to traverse the whole rev tree even when that's not strictly necessary (in the case of this particular function, it is).

nolanlawson:
Bumping up the iterations of the `basic-updates` test to 200, we get 47109ms before this fix and 45307ms after, so it indeed has a big impact.

I also generated a flamegraph recently using MemDOWN and the `basic-updates` test to show that `merge.js` really is the big offender in most performance problems (JSON parsing also ranks highly). Here's a screenshot:

screenshot from 2015-08-23 10 49 59

And [here's an HTML file you can explore yourself](https://s3.amazonaws.com/nolanlawson/pouchdb/perf/flame2.htm).

daleharvey:
Same, cheers - https://github.com/pouchdb/pouchdb/commit/91793af1b38650c3b5e55124332fa3344f772ae0

PR #4186, opened 2015-08-23
DON'T MERGE - test removing vuvuzela

nolanlawson:
We have a test that supposedly should demonstrate why
vuvuzela is actually necessary (it's this one:
https://github.com/pouchdb/pouchdb/blob/466f5ea516c629136f81ca5201f98166ed274cde/tests/test.replication.js#L112-L182), but I want to make
sure that Travis is actually running browsers that can
catch it, because oddly my local Chrome/Firefox cannot.

### DIFF
--- a/lib/deps/json.js
+++ b/lib/deps/json.js
@@ -1,0 +1,9 @@
+'use strict';
+
+exports.safeJsonParse = function safeJsonParse(str) {
+  return JSON.parse(str);
+};
+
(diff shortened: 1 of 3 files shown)


nolanlawson:
OK, so I think we have confirmation that removing vuvuzela would break that test: https://travis-ci.org/pouchdb/pouchdb/builds/76812489

Seeing failures in Node, Firefox, Chrome, and Android. For iOS and Safari, that test is disabled due to the WebSQL quota.

PR #4187, opened 2015-08-23
(#3921) - conditional JSON.parse

nolanlawson:
Only runs `vuvuzela.parse` if the string is very large.
This function seemed to be a big offender in the profiler,
probably due to the deoptimization of try/catch, so it can
only be helpful to remove it where possible.

If this passes, I'll try to get some perf numbers.

### DIFF
--- a/lib/deps/json.js
+++ b/lib/deps/json.js
@@ -1,0 +1,32 @@
+'use strict';
+
+var RECURSIVE_JSON_LIMIT = 10000;
+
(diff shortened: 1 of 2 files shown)


nolanlawson:
This offers zero performance benefit according to my tests.

nolanlawson:
I also tried removing the try/catch and just using vuvuzela everywhere, but that was quite a bit worse: the `basic-updates` test with 200 iterations went from 46925ms to 163709ms.

PR #4188, opened 2015-08-23
(#3921) - optimize merge.stem()

nolanlawson:
Another offender per the performance tests, 
optimized as much as possible.

Using the `basic-updates` test in Node with 
leveldown and iterations increased to 200, the score is:
47173ms before, 45004ms after.

### DIFF
--- a/lib/merge.js
+++ b/lib/merge.js
@@ -47,18 +47,21 @@ function insertSorted(arr, item, comparator) {
   arr.splice(idx, 0, item);
 }
 
-// Turn a path as a flat array into a tree with a single branch
-function pathToTree(path) {
-  var doc = path.shift();
-  var root = [doc.id, doc.opts, []];
-  var leaf = root;
-  var nleaf;
-
-  while (path.length) {
-    doc = path.shift();
-    nleaf = [doc.id, doc.opts, []];
-    leaf[2].push(nleaf);
-    leaf = nleaf;
+// Turn a path as a flat array into a tree with a single branch.
+// If any should be stemmed from the beginning of the array, that's passed
+// in as the second argument
+function pathToTree(path, numStemmed) {
+  var root;
+  var leaf;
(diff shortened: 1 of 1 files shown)


lib/merge.js
```diff
@@ -47,18 +47,21 @@ function insertSorted(arr, item, comparator) {
...
+  var leaf;
+  for (var i = numStemmed, len = path.length; i < len; i++) {
+    var node = path[i];
+    var currentLeaf = [node.id, node.opts, []];
+    if (leaf) {
+      leaf[2].push(currentLeaf);
+      leaf = currentLeaf;
+    } else {
+      root = leaf = currentLeaf;
+    }
```
nolanlawson: It turned out there was no reason to destroy the `path` as we traversed it, and plus `shift()` is slow.

lib/merge.js
```diff
@@ -190,20 +193,28 @@ function doMerge(tree, path, dontExpand) {
...
-  }, [stemmedPaths.shift()]);
+    // Then we remerge all those flat trees together, ensuring that we dont
+    // connect trees that would go beyond the depth limit
+    if (result) {
+      result = doMerge(result, stemmedNode, true).tree;
+    } else {
+      result = [stemmedNode];
+    }
+  }
+  return result;
```
nolanlawson: Once again, I'm preferring for-loops to `map()` and `reduce()`. Easy trick.

A more subtle optimization is to avoid `slice()` because it turned out we don't actually need to copy the array. Since the `stemmed` array is just going to get transformed into a new array using `pathToTree()` anyway, we can just tell `pathToTree()` how many got stemmed (`numStemmed`), and it can start at that number instead of at 0 when it iterates through.

daleharvey:
Looks good, cheers - https://github.com/pouchdb/pouchdb/commit/0f83b35ae996695e9c2d398d47eed91322c71dbe

Issue #4189, opened 2015-08-23
Better error message when CouchDB doesn't have CORS configured


nolanlawson:
Currently it's very hard to figure out when you forgot to enable cors, because the error message you get is something like `unknown error 500`.

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/ac56d71cac049fd026ee1e8918bcb82d31192018

anaran:
I don't think this issue fixes the scenario I describe in #4256 where the firefox os app is to blame and the error message sends people off to investigate their couchdb server CORS settings instead.

PR #4190, opened 2015-08-23
(#4189) - better CORS error message

nolanlawson:
This adds a console error for browser users when we detect CORS is unavailable, and it also uses a custom 405 error instead of "UnknownError".

Here's what it looks like in Chrome:

screenshot from 2015-08-23 15 15 03

And in Firefox:

screenshot from 2015-08-23 15 19 22

### DIFF
--- a/lib/deps/ajax/ajax-core.js
+++ b/lib/deps/ajax/ajax-core.js
@@ -6,6 +6,7 @@ var errors = require('./../errors');
 var utils = require('../../utils');
 var applyTypeToBuffer = require('./applyTypeToBuffer');
 var defaultBody = require('./defaultBody');
+var explainCors = require('./explainCors');
 
 function ajax(options, callback) {
 
@@ -87,7 +88,14 @@ function ajax(options, callback) {
 
   return request(options, function (err, response, body) {
     if (err) {
-      err.status = response ? response.statusCode : 400;
+      if (response) {
+        if (response.statusCode === 0) {
+          explainCors(); // statusCode 0 indicates a CORS error (405)
+        }
+        err.status = response.statusCode;
+      } else {
+        err.status = 400;
+      }
       return onError(err, callback);
     }
 
--- a/lib/deps/ajax/explainCors-browser.js
+++ b/lib/deps/ajax/explainCors-browser.js
@@ -1,0 +1,9 @@
+'use strict';
+
+module.exports = function () {
+  if ('console' in global && 'error' in console) {
+    console.error('PouchDB error: the remote database does not seem to have ' +
+      'CORS enabled. To fix this, please enable CORS: ' +
+      'http://pouchdb.com/errors.html#no_access_control_allow_origin_header');
+  }
+};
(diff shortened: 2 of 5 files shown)


nickcolley:
This is such a good change, thanks a lot @nolanlawson . It's going to help so many people. :dancer:

nolanlawson:
Cripes, it would have helped *me*. :) I was scratching my head over this bug for a good 20 minutes recently, before I realized I forgot to enable CORS. :stuck_out_tongue:

nolanlawson:
This PR breaks Firefox and I cannot reproduce.

nolanlawson:
Looks like Travis is using an older version of Firefox that still has that blob+ajax bug. I think this is a good time to just simplify `blobSupport.js` with a user agent sniff. Only Firefox has the blob+ajax bug, and only Chrome has the blob+idb bug, so we might as well do the shameful, shameful, user agent sniff and not have to futz around with this stuff.

nolanlawson:
Actually, it just occurred to me that that can't possibly be the issue.

I think I'm just going to scale back the ambition of this PR and make it only log a warning, not try to refactor the error code.

nolanlawson:
Green! :)

daleharvey:
Nice, https://github.com/pouchdb/pouchdb/commit/ac56d71cac049fd026ee1e8918bcb82d31192018

darobin:
There's a problem with this change: `statusCode` can be set to 0 in several ways. Most notably, aborting the request (for instance because of a timeout) will also cause it to be set to 0. In fact, *any* network error will look like this.

CORS issues are indistinguishable from other network problems *by design*. You simply can't really check for this. I'm afraid this change needs to be backed out.

For context, I just spent 20min looking at the code and trying to figure out why I was getting a CORS error for a request that wasn't even cross-domain :)

daleharvey:
The vast majority of our users come up against statusCode 0 due to cors, the wording could be a little bit clearer with some bikeshedding, but this doesnt need backed out

darobin:
Sure, that works too. Note that in the specific instance that I've bumped into the `abort()` came from https://github.com/pouchdb/pouchdb/blob/master/lib/deps/ajax/request-browser.js#L113 which as far as I can tell can only happen in a timeout. A specific error there might be useful (it seems to be the case if fetch is used).

nolanlawson:
Agree we should change the error message to say something like

> do you see something about *Access-Control-Allow-Origin* in your console? If so, you have a CORS error

nickcolley:
Is it possible to detect that header and only show the message if it isn't there?

nolanlawson:
Apparently no. I think it makes more sense to soften the warning message. Or just remove it.

For the record, to everybody saying they wasted time because of this message: I apologize, but keep in mind we've been hearing for *years* that new users forget to enable CORS, and spend a lot of time hunting down the error.

daleharvey:
Being that this is a PR with code against it, fairly confusing to have it open, lets close this and reopen the issue filed - https://github.com/pouchdb/pouchdb/issues/4256

darobin:
@nolanlawson FWIW no need to apologise, you tried to solve a real problem, nothing wrong with that. It's unfortunate that the Web's security model relies on hiding this sort of information.

Issue #4191, opened 2015-08-23
Endless loop while syncing


nolanlawson:
I can reproduce this (https://github.com/hoodiehq/hapi-couchdb-store/issues/1), and it seems to be a bug in PouchDB or PouchDB Server, because CouchDB does not reproduce it.

nolanlawson:
This turned out to be a bug in pouchdb-hoodie-api: https://github.com/hoodiehq/hapi-couchdb-store/issues/1#issuecomment-133971319

PR #4192, opened 2015-08-24
(#4191) - test correct changes "since" behavior

nolanlawson:
but it's always good to have more tests, right? :)


daleharvey:
Removed the console.logs on the way in, cheers - https://github.com/pouchdb/pouchdb/commit/e91f220dd458e7b01719e5adf6618fb345bcb559

Issue #4193, opened 2015-08-24
Migrate to container-based Travis infrastructure


nolanlawson:
Travis keeps bugging me to do this: http://docs.travis-ci.com/user/migrating-from-legacy/

nolanlawson:
Woohoo, already 1min shaved off each individual build, now let's try with caching...

nolanlawson:
fixed by 9d6b611c6467944c5de2cad668642e35fa4109d3

PR #4194, opened 2015-08-24
(#4193) - migrate to container-based Travis

nolanlawson:
CSG will still be broken in this PR because we
don't have sudo anymore, but I'm just testing
to see if this will work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,12 +9,30 @@ services:
 git:
   depth: 30
 
+sudo: false
+
+cache:
+  directories:
+    - node_modules
+    - /tmp/phantomjs
+
+addons:
+  apt:
+    packages:
+      - build-essential
+      - ca-certificates
+      - curl
+      - erlang-dev
+      - erlang-nox
+      - git
+      - libicu-dev
+      - libmozjs185-dev
+      - python
+      - g++
+
 before_script:
   - "export DISPLAY=:99.0"
   - "sh -e /etc/init.d/xvfb start"
-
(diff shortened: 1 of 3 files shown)


nolanlawson:
So this actually seems very close to passing, but I'm not sure what benefit it's giving us. It doesn't seem much faster than the old system.

nolanlawson:
OK, actually there do seem to be big benefits [per this doc](http://docs.travis-ci.com/user/migrating-from-legacy/) (e.g. additional cores/RAM), and especially if we start [caching node_modules](http://blog.travis-ci.com/2013-12-05-speed-up-your-builds-cache-your-dependencies/)

nolanlawson:
Weird, looks like SpiderMonkey isn't being compiled correctly:

    /home/travis/couchdb/src/couch/priv/couch_js/http.c:18:19: fatal error: jsapi.h: No such file or directory

I thought this would be fixed by adding `g++` to the apt sources list, but apparent. I know I've run into this like a million times with Couch, but I can never remember what the fix is. :/

nolanlawson:
Since this improves the build time so much, and since Couch master is in the allowed failures anyway, I would suggest we just merge and then open another issue to fix Couch master. Same for CSG; it will also be permafail until we fix the sudo issues.

daleharvey:
I can live with that, hopefully someone else is working on couch / travis integration as well so maybe ping them in the new issue, :+1:

nolanlawson:
9d6b611c6467944c5de2cad668642e35fa4109d3

Issue #4195, opened 2015-08-24
Remove vuvuzela


daleharvey:

daleharvey:
@nolanlawson you had a test PR open for this, its fairly guaranteed to save a good amount of speed by removing it entirely, the issue is that we generate far too much metadata

https://github.com/pouchdb/pouchdb/blob/master/lib/deps/docs/updateDoc.js#L30 is currently defaulted to 1000, which is tons, but should be small enough to safely json.stringify, when I run https://github.com/pouchdb/pouchdb/blob/master/tests/integration/test.replication.js#L192 I can see we still have a 5000 long array in the metadata which means something is broken.

if rev_limits is like 10 then we should have a very small amount of metadata that would be very safe to write, if we dont then it is another bug, taking a look into it now

nolanlawson:
So we now have the revs_limit correctly set to 1000, and I ran another PR trying to get rid of it, and it seems like sadly it's here to stay, unless we want to decrease the max any further: https://github.com/pouchdb/pouchdb/pull/4336

I think we should close this issue for now; running `disc` on PouchDB i can see that vuvuzela takes up <1% of the total browserify size, and although it's a performance drain, we can still experiment with using string length heuristics to avoid the try/catch deoptimization.

Issue #2839, opened 2014-10-08
Implement _revs_limit


nolanlawson:
ala Couch

anaran:
Ah, so this might be the reason I cannot restore a previous revision after `db.remove(id);`
I can get `_revisions.ids` but all the revs are `missing` except for the final one, which is `ok` but has `"_deleted": true` and no doc fields available.

anaran:
Using `doc._deleted = true; db.put(doc);` keeps all doc fields intact in the tombstone revision,
just like you said in #pouchdb.

Using `db.remove(id);` does not!

In both cases the previous revision becomes `missing` immediately, even without replication or further db activity.

nolanlawson:
> Using `doc._deleted = true; db.put(doc);` keeps all doc fields intact in the tombstone revision,
just like you said in #pouchdb.
> 
> Using `db.remove(id);` does not!

This is intended behavior, and will be documented when the docs are updated soon.

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/5832d03e2796b090ae2a844278577fe2d994d982

PR #4196, opened 2015-08-24
(#4195) - Remove vuvuzela

daleharvey:


daleharvey:
So I believe vuvuzela is papering over bugs in our API, basically we use it for writing metadata, metadata is entirely under our control so we should bound it to a limit that we can comfortable store as native JSON. If we get to a situation where it is too large to be written it means there is a bug in how we bound the limit.

One method we use is `revs_limit` which is the parameter to https://github.com/pouchdb/pouchdb/blob/master/lib/merge.js#L212, we currently set that to 1000 which is huge but should be ok still to store. One bug this PR fixes is that new_edits are not subject to that `revs_limit`, we could just increase the number in the test and it would still eventually break even with vuvuzela.

We can very safely reduce that number, I am just getting this up to test but will likely implement https://github.com/pouchdb/pouchdb/issues/2839 at the same time as this then maybe reduce to 100

This would be a very nice time if @rnewson or @kxepal wanted to explain the dangers of reducing `revs_limit`, 1000 has always seemed to be a huge value for me and I dont really see the value in it being much more than 1, but would be good to hear some background before making a decision

kxepal:
@daleharvey The size of value is depended on rate of your document's update and probability of conflict causing. Having it as 1 is not good since you'll loose access to previous revisions from the same branch. 1000 is good default to not fall into situations when you should care about. Much likely, you can find lower value that suites PouchDB common workflow and doens't causes any need to tweak it all the time.

daleharvey:
I just cant think of many / any use cases where we need more than the leafs parents, I havent seen a documented case of conflict resolution that would actually use it, generally users either just delete conflicts, or they fetch all conflicts, merge them and delete the rest, none of this requires revision tree history

kxepal:
Well, quite simple: I accidentally deleted document and want to restore it. 
For CouchDB iirc, revision tree history is used to locate common ancestors by given revision on /_revs_diff call.

daleharvey:
Yup we do the same, but yeh having backup abilities is not a very typical use case, we see couchdb devs constantly telling people not to use couchdb as version control, while I think this is a perfectly valid use case I think its one users should be specifically opting into, its not a particularly sensible default.

In this issue will likely just bump it down to 100/500 depending on what the performance tests say, not do anything drastic like change it to one but something to think about for future pouch changes

kxepal:
It's indeed not version control system, just a side effect which you can use with certain limitations.

I don't see any issues with having rev_limit 100 for PouchDB unless someone will update docs with higher than 100 times between each syncs. Reducing rev_limit is also makes improvements on disk size.

nolanlawson:
Removing vuvuzela would definitely reduce both the bundle size and speed up read/write operations. (Since try/catch is a deopt.) That said, I tried recently to remove vuvuzela, and it failed the Travis tests: https://github.com/pouchdb/pouchdb/pull/4186

The [test that reproduces the error](https://github.com/pouchdb/pouchdb/blob/466f5ea516c629136f81ca5201f98166ed274cde/tests/test.replication.js#L112-L182) only adds 5000 revisions, so either as you say we have a bug and are accidentally storing >=1000, or 1000 is enough to break some browsers.

And yeah, the object is entirely within our control, so it kinda makes me wonder if a better solution wouldn't be to have a custom stringify/parse function? vuvuzela is a bit overkill since we know exactly what the object will look like.

nolanlawson:
Ah okay, I just read https://github.com/pouchdb/pouchdb/issues/4195#issuecomment-134105551, so we definitely have a bug if that array reaches 5000.

daleharvey:
Yup, the issue was that writing a brand new document with `new_edits` didnt trigger the stemming we do on revisions, they were written as is, this was passing in node and firefox locally

nolanlawson:
We should probably add an explicit test to check that the rev_limit is being respected.

nolanlawson:
Failures seem genuine, definitely browser-specific. (Safari, Chrome, iPhone, Android)

From what @calvinmetcalf has told me, there's no iron-clad rule about where the stack limit is on a given browser; it varies based on how much memory is available at any given time. We could maybe reasonably reduce `revs_limit` to 500 (who's modifying a document 1000 times?) and thereby remove vuvuzela, but that would be a breaking change IMO.

I'd say the most important thing is to have a test to demonstrate that stemming to 1000 is actually happening, then worry about vuvuzela after that.

daleharvey:
Yeh these failures are genuine, I dont think they are caused by the actual removal of vuvuzela but the extra work we are doing on new_edits writes causing a slowdown which has knock on effects.

Ill isolate the test / fix for that bug without the rest of the vuvuzela changes

Issue #4144, opened 2015-08-10
Move pouchdb-upsert into pouchdb itself


nolanlawson:
There's no reason to keep it as a separate module anymore, now that mapreduce is no longer a separate module, and it's causing problems because it's trying to both 1) be a good plugin for average users, and 2) be a good component inside of PouchDB itself. E.g. see discussion in https://github.com/pouchdb/upsert/issues/7.

I would like to just copy-paste `upsert.js` into PouchDB's core code somewhere and maybe remove stuff we don't need (e.g. promise vs callback API - just support one or the other, whichever one is being used in PouchDB itself).

This may be a good first bug.

nolanlawson:
95b1771348f7cbdbe6e665b6891af8599f00be18

nolanlawson:
For everyone wondering what this issue means, it *does not mean that pouchdb-upsert is bundled with PouchDB itself*. That's never gonna happen, due to our desire to keep the API surface small. Please use the [pouchdb-upsert](https://github.com/pouchdb/pouchdb-upsert) plugin.

tohagan:
Since the `upsert()` function is already included and used in PouchDB code under `lib/deps/upsert.js` and wiring up the plugin is ugly (`require()` is fine in NodeJS but not much use in Angular) can't we just add it as `PouchDb.prototype.upsert()`.  The fact that you needed it inside PouchDb is surely evidence enough that everyone else needs it too! 

I think we're talking 1-2 lines of code to integrate it and we move some unit tests and docs into core and a whole load of API coding pain goes away :).

nolanlawson:
Even just exposing the "extras" API (e.g. ajax, checkpointer), has already caused some headaches for us; it made it difficult in the last release to juggle the new Rollup build system. @daleharvey and I are already considering removing it.

So yeah, I don't think the fact that it's in PouchDB itself is a strong justification for exposing it. It increases the overall API surface, which increases the stuff we have to maintain and makes it harder for new users to grok the codebase. Plugins aren't seamless, but it's the best we have right now.

As for `require()` in Angular/Ionic, Ionic 2 is moving to a system more like Angular 2's (i.e. with npm, TypeScript, transpilers, etc.), meaning it will be very easy to `require()` Node-style.

tohagan:
Well I guess I just hack the plugin code into an Angular 1.x service of some kind.

PR #4206, opened 2015-08-26
(#4144) - Removes pouchdb-upsert dependency.

nicolasbrugneaux:
Basically copies the file from the npm package. Removes support for
callback style (can still be used with callbackify).

### DIFF
--- a/lib/adapter.js
+++ b/lib/adapter.js
@@ -459,7 +459,7 @@ AbstractPouchDB.prototype._compact = function (opts, callback) {
     .on('complete', onComplete)
     .on('error', callback);
 };
-/* Begin api wrappers. Specific functionality to storage belongs in the 
+/* Begin api wrappers. Specific functionality to storage belongs in the
    _[method] */
 AbstractPouchDB.prototype.get =
   utils.adapterFun('get', function (id, opts, callback) {
@@ -784,12 +784,9 @@ AbstractPouchDB.prototype.registerDependentDatabase =
     doc.dependentDbs[dependentDb] = true;
     return doc;
   }
-  upsert(this, '_local/_pouch_dependentDbs', diffFun, function (err) {
-    if (err) {
-      return callback(err);
-    }
-    return callback(null, {db: depDB});
-  });
+  upsert(this, '_local/_pouch_dependentDbs', diffFun)
+    .then(callback.bind(this, null, {db: depDB}))
+    .catch(callback);
 });
 
 AbstractPouchDB.prototype.destroy =
--- a/lib/deps/upsert.js
+++ b/lib/deps/upsert.js
@@ -1,7 +1,55 @@
 'use strict';
 
-var upsert = require('pouchdb-upsert').upsert;
+var Promise = require('lie');
 
-module.exports = function (db, doc, diffFun, cb) {
-  return upsert.call(db, doc, diffFun, cb);
+// this is essentially the "update sugar" function from daleharvey/pouchdb#1388
+// the diffFun tells us what delta to apply to the doc.  it either returns
+// the doc, or false if it doesn't need to do an update after all
+var upsert = module.exports = function upsert(db, docId, diffFun) {
+  return new Promise(function (fulfill, reject) {
+    if (typeof docId !== 'string') {
+      return reject(new Error('doc id is required'));
+    }
+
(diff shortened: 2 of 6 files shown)


lib/deps/upsert.js
```diff
@@ -1,7 +1,67 @@
 'use strict';
 
-var upsert = require('pouchdb-upsert').upsert;
+var PouchPromise;
```
marten-de-vries: This can probably be replaced by a ``require()`` that directly gives you the promise object ('./promise').

lib/deps/upsert.js
```diff
@@ -1,7 +1,67 @@
...
+  }, function (err) {
+    /* istanbul ignore next */
+    if (err.status !== 409) {
+      throw err;
+    }
+    return upsert(db, doc._id, diffFun);
+  });
+}
+
+/* istanbul ignore next */
```
marten-de-vries: No longer necessary, this isn't a PouchDB plugin (anymore).

marten-de-vries:
Thanks for working on this. I left a few inline comments.

nicolasbrugneaux:
thanks for the quick feedback, I updated accordingly :)

nolanlawson:
Travis failures seem genuine:

```
  1) test.design_docs.js-local Basic views:
     Error: timeout of 50000ms exceeded
      at null.<anonymous> (/home/travis/build/pouchdb/pouchdb/node_modules/mocha/lib/runnable.js:158:19)
      at Timer.listOnTimeout [as ontimeout] (timers.js:121:15)
```

Since it doesn't fail until it gets to `test.design_docs.js`, it looks like it's because `create-view.js` in mapreduce is using `upsert` slightly differently from where it's used elsewhere? It's timing out, which means maybe a promise is not being resolved somewhere. (If you test in Chrome, it will warn about uncaught errors in promises.)

nicolasbrugneaux:
Right, I didn't test in browser, I ran it in cmd line and kept timing out so I figured it was unrelated, I'll investigate further.

nicolasbrugneaux:
I can't seem to figure out why this line `return sourceDB.registerDependentDatabase(depDbName).then` in `mapreduce/create-views.js` doesn't resolve...

nolanlawson:
Hm, try ticking "Async" and "break on uncaught exceptions" in the Dev Tools? Also Chrome Canary has an experimental Promises view you can see, which shows unresolved promises: chrome://flags -> Enable Developer Tools experiments -> then in the developer tools, you should see in Dev Tools -> Settings -> Experiments -> "Promise inspector"  (if you don't, hit the shift key 6 times )

nicolasbrugneaux:
I think I got it, thanks for the tips, I would have never found it otherwise, one upsert in `adapter.js` was still with callback style.

lib/adapter.js
```diff
@@ -784,12 +784,9 @@ AbstractPouchDB.prototype.registerDependentDatabase =
...
     return doc;
   }
-  upsert(this, '_local/_pouch_dependentDbs', diffFun, function (err) {
-    if (err) {
-      return callback(err);
-    }
-    return callback(null, {db: depDB});
-  });
+  upsert(this, '_local/_pouch_dependentDbs', diffFun)
+    .then(callback.bind(this, null, {db: depDB}))
```
nolanlawson: `.bind` tends to be bad for performance, the callback doesn't need `this`, and I believe it's shorter to just have a regular function here

lib/deps/upsert.js
```diff
@@ -1,7 +1,55 @@
 'use strict';
 
-var upsert = require('pouchdb-upsert').upsert;
+var Promise = require('lie');
```
nolanlawson: We should pull from deps/promise so that we use the native `Promise` if it's available.

nolanlawson:
This all looks fabulous, and seems like it should pass in Travis (I restarted some spurious failures). Big +1 after it's green, and no need to fix my nitpicks if you don't want; I can just amend the commit on the way in.

nolanlawson:
Green! :)

nolanlawson:
95b1771348f7cbdbe6e665b6891af8599f00be18

nicolasbrugneaux:
Awesome! Thank you :dancer:

PR #4217, opened 2015-08-29
(#4094) - Dont fail replication if fetching uuid fails

daleharvey:


lib/adapters/http/index.js
```diff
@@ -225,11 +225,8 @@ function HttpPouch(opts, callback) {
       method: 'GET',
       url: genUrl(host, '')
     }, function (err, result) {
-      /* istanbul ignore next */
-      if (err) {
-        return callback(err);
-      }
-      var uuid = result.uuid + host.db;
+      var uuid = (result && result.uuid) ?
+        result.uuid + host.db : genDBUrl(host, '');
```
nolanlawson: ^ would prefer to wrap `result.uuid + host.db` in parens so it's clear what's going on

lib/deps/ajax/ajax-core.js
```diff
@@ -101,18 +101,14 @@ function ajax(options, callback) {
...
 
     if (response.statusCode >= 200 && response.statusCode < 300) {
       onSuccess(data, response, callback);
     } else {
-      try {
-        // See if we received an error message from the server
-        // If the connection was interrupted, the data might be nothing
-        // that's why we have the try/catch
-        data = JSON.parse(data.toString());
-      } catch(e) {}
```
nolanlawson: nice

tests/integration/test.replication.js
```diff
@@ -4053,6 +4053,41 @@ adapters.forEach(function (adapters) {
...
+        .on('complete', complete);
+
+      var changes = db.changes({live: true}).on('change', function (change) {
+        rep.cancel();
+        changes.cancel();
+      }).on('complete', complete);
+
+      remote.post({a: 'doc'});
+    });
+
```
nolanlawson: excellent test

nolanlawson:
LGTM, +1 when green

daleharvey:
Hrm, pouchdb-server seems to be failing genuinely but I have no idea how this test would effect that

daleharvey:
I dont know how this could fail pouchdb-server, and pouchdb-server is permafailing for me locally. Spent some time debugging but its fairly confusing.

So as discussed before, I dont think the burden should be on contributors (whether its me, Nolan or someone for their first time) to be supporting every server that we test, as we have done with express-pouchdb-router, I have moved pouchdb-server to allowed failures in this patch and would suggest that we merge genuine fixes even if its causing test failures on these server and people maintaining those servers can get them green again.

Sound good?

nolanlawson:
I think that's fair, but let me take a crack at it first. I think I can fix the pouchdb-server failure.

marten-de-vries:
@daleharvey Sounds good.

nolanlawson:
I'm still working on solving this; no luck reproducing yet. In any case, can we agree that this fix can wait until after 4.0.1? I don't think it's worth breaking PouchDB Server right before a monthly release.

daleharvey:
It isnt particularly ideal, this is a bug that breaks replication for a fairly large number of users. The point of moving to allowed failures is that there isnt pressure on you or anyone else to get pouchdb-server green before pouchdb bugs get fixed

nolanlawson:
The bug seems to be that Travis' npm is installing pouchdb-server 0.6.6 for some reason. The latest is 0.9.0, but it's not picking it up (hence why the logs are also so noisy as well).

I agree that contributors shouldn't have to fix every server in existence, but I do think we should try to make a good-faith effort to fix these issues and catch them early. Also pouchdb-server is a bit special IMO because 1) it is in heavy usage, especially now that Hoodie has picked it up, and 2) it is directly affected by any change we make here.

nolanlawson:
Nope, that's not the bug. And I can't reproduce locally at all. I have no idea wtf is going on.

nolanlawson:
Okay, I can repro, but only if I run the entire test suite all the way through and use Firefox.

nolanlawson:
Ugh, I have been trying to figure this out all day, but it can only be repro'd in a very specific set of conditions. It seems to be related to Selenium as well, since what's actually timing out is the `window.eval()` command.

nolanlawson:
Interesting - I can repro in Firefox, but not Phantom or Chrome.

nolanlawson:
I'm not 100% happy that pouchdb-server is failing, but this seems to be Selenium+Firefox related (or maybe proxy-related). Something subtle is going on here, and this fix is important enough to make it into 4.0.1 imo. Plus we can always fix the build by using Phantomjs instead of Firefox.

nolanlawson:
9c03e0f212b3a23e4d72962be2545382caaa3090

PR #4218, opened 2015-08-29
(#4217) - Defer jslint until after test

daleharvey:


daleharvey:
When I am quickly iterating on a test it bugs me waiting the minute or 2 it takes to lint, this just runs the lint after the test

nolanlawson:
You can just run `npm run test-node` instead of `npm t` to skip the linting. We could document this and/or rename it to `test-without-lint`.

From the perspective of a new contributor, I think it would be kind of annoying to run ~30 minutes of tests and then only find out at the end that you forgot a semicolon. :P Ditto for running all tests in Travis for 9.5 hours.

In general though this is a minor issue so I am -0.

daleharvey:
I usually comment out `npm run jshint` when I am doing a bunch of node tests, but either that or exposing another command for it instead of doing the right thing by default is fairly annoying.

I picked up the `posttest` trick from a post specifically talking about helping new contributors not get pointless errors about linting when they are trying to figure some new code, its something that gets in my way a bunch as well, comment out a variable in the test and you need to comment out its definition etc - http://blog.npmjs.org/post/127671403050/testing-and-deploying-with-ordered-npm-run-scripts

nolanlawson:
Ah yeah that is a really good point. I've been hit by that too. Okay, let's make it a pre-travis script, and then +1.

daleharvey:
Switched it up, added it to before_scripts in travis so it will still fail early, added a lint error just to make sure it does the right thing

daleharvey:
Cheers, merged - https://github.com/pouchdb/pouchdb/commit/3cf579bfb44638c3d479d4081e72c54af3a73a17

Issue #4216, opened 2015-08-29
Function.name doesnt work in IE10+


nickcolley:
https://github.com/pouchdb/pouchdb/blob/70761f2c6003e91f8a886a1d5968c34cb27dc047/lib/adapters/leveldb/index.js#L150

leveldown.name is set only in browsers that support .name on a function

http://stackoverflow.com/questions/6903762/function-name-not-supported-in-ie

nickcolley:
I wacked this polyfill in:
```
if (!(function f() {}).name) {
    Object.defineProperty(Function.prototype, 'name', {
        get: function() {
            var name = this.toString().match(/^\s*function\s*(\S*)\s*\(/)[1];
            // For better performance only parse once, and then cache the
            // result through a new accessor for repeated access.
            Object.defineProperty(this, 'name', { value: name });
            return name;
        }
    });
}
```

Memdown passes in IE! :D

Might be a nicer way to fix without polyfilling.

nickcolley:
This is again ES6... https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name

nickcolley:
This is required for pouchdb.memory to work in IE by the way, didn't make that clear.

nolanlawson:
So the thing is that `Function.name` is really a hack, and we are exploiting the fact that most of the *down adapters have names like `MemDOWN` and `LocalStorageDOWN` that make them easy to identify. But in fact there's nothing stopping them from exporting anonymous functions if they feel like it.

I'd suggest that instead of providing a `Function.name` shim, we just rewrite the part of the code that checks for it.

nolanlawson:
fixed by 2b656465bd578de1180c5b50d73664935a1f37fe

nolanlawson:
finally actually for-realseys fixed by 5c0c421081472a785b5bee58860420c57aae8885

PR #4220, opened 2015-08-29
(#4216) - remove Function.name checking

nolanlawson:
This `Function.name` checking only makes sense in Node,
and it only matters when the user creates two databases
with the same name but different `db`s. We can simplify
the code and just document the weird edge case.


marten-de-vries:
One case where this might (didn't test yet) bite us, is in pouchdb-server. It supports changing the storage engine/storage locations on the fly via ``/_config``, and there are almost always identical database names involved (``_users`` & ``_replicator``).

nickcolley:
Yeah I'd prefer if we could fix this without any gotchas. Not sure the best approach though I was thinking like a UUID or md5sum but not sure.

nolanlawson:
Hm, maybe the shim for Function.prototype.name is the best bet after all. We don't need to shim it globally; we can just shim it in this one case where we want to get the function's name.

PR #4221, opened 2015-08-29
(#4216) - test memory adapter in IE10

nolanlawson:
This seems like this may be useful, since IE10/phantomjs have
some unexpected issues with the leveldown build.


nolanlawson:
this PR will fail until #4216 is actually fixed btw

daleharvey:
Cool lets add it in https://github.com/pouchdb/pouchdb/pull/4246 so we can keep it green

PR #4222, opened 2015-08-29
BREAKING, merge after 4.0.1 - (#4222) - remove idb-alt plugin

nolanlawson:
I am fairly certain nobody is using this. I think it was
a nice test case for a long time, and especially served
to convince us we weren't crazy when MemDOWN and LocalStorageDOWN
kept failing the tests, but level.js passed them just fine.

However, I think it's pretty clear by now that we're not going
to replace the `idb` adapter with this; it browserifies too large,
there are too many PouchDB-specific optimizations in the current
idb adapter, and we don't have a reasonable migration plan. So
I think we should just drop support.

This will require a major version release, so we can wait until after 4.0.1
to merge.


nolanlawson:
Oh another thing: this increases the build time both in Travis and when you run `npm run build`.

daleharvey:
So tentative + .5 I have always seen the levelalt adapters as experimental, I dont think they are useful alternatives at the moment but its nice to see that they work. I would be happier maybe emphasizing them in the docs and keeping them in travis. But not hugely fussed either way

marten-de-vries:
If a moment comes where the idb-alt adapter is considered actually useful, re-enabling it shouldn't be that hard. Worst case scenario, a few bug fixes in either of the projects. The interface isn't going away. But spending time on it in the meantime? Naah. I think removing it is the right decision at the moment.

nolanlawson:
> I dont think they are useful alternatives at the moment but its nice to see that they work

Surprisingly, I'm seeing the Hoodie folks using the `memory` adapter everywhere; in fact they are finding bugs in IE10 that I didn't even know existed.

As for `localstorage`, I haven't heard much fuss, except from people who are trying to get PouchDB to run in IE9, which doesn't work because of ES3 issues. So its value is limited apparently.

For `idb-alt`, I have never heard a single question from a single user about it, and also it has never broken on us, most likely because it [hasn't changed much](https://github.com/maxogden/level.js/graphs/contributors) over the past year.

nolanlawson:
forgot to remove `level-js` from `package.json`

daleharvey:
Ok lets do this - https://github.com/pouchdb/pouchdb/commit/427efbcdf71e965a3aa34fc68641f948a66ae8b6

PR #4224, opened 2015-08-29
BREAKING, merge after 4.0.1 - (#4224) - remove CRUD events

nolanlawson:
Yet another deprecation I would like for 5.0.0. Again, let's
please do a 4.0.1 first though.

This one will require some explanation in the release notes
to explain how to upgrade, although I wonder how many
people are actually using it, since the docs on it were
kind of sparse.


daleharvey:
Lets do this - https://github.com/pouchdb/pouchdb/commit/83a4a0b84658159fef440028707c05615b10ced7

Issue #4209, opened 2015-08-28
Couch master failing in Travis again


nolanlawson:
The error I see is:

```
 1) suite2 test.replication.js-local-http #3999-1 should not start over if last_seq mismatches:
      AssertionError: expected '2-g1AAAAG7eJzLYWBg4MhgTmHgz8tPSTV0MDQy1zMAQsMcoARTIkOS_P___7MygKxcoAC7mbmJWZpFIk4NSQpAMskeRU9SmnGSpbkFbj0OID3xKHpMTJLMzBPx2JMA0lMP1sOcyAjWY55oYm5pmYypA6cpeSxAkqEBSAENmo9ku1mKeZqlEQF9CyD69mclMhBQeQCi8j5hlQ8gKpF8ZWBobJJsmYKpKwsA9CRvdg' to not equal '2-g1AAAAG7eJzLYWBg4MhgTmHgz8tPSTV0MDQy1zMAQsMcoARTIkOS_P___7MygKxcoAC7mbmJWZpFIk4NSQpAMskeRU9SmnGSpbkFbj0OID3xKHpMTJLMzBPx2JMA0lMP1sOcyAjWY55oYm5pmYypA6cpeSxAkqEBSAENmo9ku1mKeZqlEQF9CyD69mclMhBQeQCi8j5hlQ8gKpF8ZWBobJJsmYKpKwsA9CRvdg'
      + expected - actual
     
      at source.put.then.then.catch.then.then.then.mismatch (/home/travis/build/pouchdb/pouchdb/tests/integration/test.replication.js:2501:36)
      at Array.0 (/home/travis/build/pouchdb/pouchdb/node_modules/lie/lib/index.js:88:21)
      at nextTick (/home/travis/build/pouchdb/pouchdb/node_modules/lie/node_modules/immediate/lib/index.js:61:18)
      at process._tickCallback (node.js:448:13)
```

nolanlawson:
[Example log](https://travis-ci.org/pouchdb/pouchdb/jobs/76944090)

kxepal:
mmm...
```
[...]

PR #4229, opened 2015-08-30
(#4209) - remove couch master from allowed failures

nolanlawson:
Should be passing again.


daleharvey:
Seeing as couchdb master is not a release and commits to it are fairly regularly breaking our test suite, I think its best kept in allowed failures until we are testing against a stable release, that ok?

nolanlawson:
Yeah, probably the prudent move.

PR #4232, opened 2015-08-30
(pouchdb/pouchdb-server#110) - actually silence logs

nolanlawson:
Due to https://github.com/pouchdb/pouchdb-server/issues/115 we should be using the -n option
in order to actually silence the logs. Seems to be a bug/quirk in `yargs`.


daleharvey:
report-coverage isnt due to this, so merging, cheers - https://github.com/pouchdb/pouchdb/commit/c8857fd1a3450ea5ae9ed179642fe8e18411b606

Issue #4240, opened 2015-09-01
PouchDB Server is permafail in Firefox


nolanlawson:
https://github.com/pouchdb/pouchdb/commit/9c03e0f212b3a23e4d72962be2545382caaa3090 put PouchDB Server into permafail. Details are scarce (https://github.com/pouchdb/pouchdb/pull/4217#issuecomment-136177286), but basically you should be seeing an ESOCKETTIMEDOUT for `window.eval`. You can repro by running:

    SERVER=pouchdb-server CLIENT=selenium:firefox npm t

Oddly this cannot be reproduced in Chrome or PhantomJS in my testing.

nolanlawson:
Error:

```
=>  { lastPassed: 'issue #2393 update_seq after new_edits + replication',
  failures: [],
  passed: 610,
  failed: 0,
  browser: 'Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:40.0) Gecko/20100101 Firefox/40.0' }
{ [Error: [eval("window.results")] ESOCKETTIMEDOUT] code: 'ESOCKETTIMEDOUT' }
Doh, tests failed
```

Nothing interesting in Firefox itself except timeouts.

nolanlawson:
More data points:
* it happens with and without the proxy
* it happens with a stock pouchdb-server (i.e. using pouchdb 4.0.0 on the server side)

nolanlawson:
This is definitely a Selenium bug, because I can't reproduce if I just use `npm run dev` and open the tests in Firefox.

daleharvey:
The issue is pouchdb-server isnt sending the correct Content-type on JSON responses
```
$ curl -v -X PUT http://127.0.0.1:6984/testing
* About to connect() to 127.0.0.1 port 6984 (#0)
*   Trying 127.0.0.1...
* Adding handle: conn: 0x7fc9d4004400
* Adding handle: send: 0
* Adding handle: recv: 0
* Curl_addHandleToPipeline: length: 1
* - Conn 0 (0x7fc9d4004400) send_pipe: 1, recv_pipe: 0
* Connected to 127.0.0.1 (127.0.0.1) port 6984 (#0)
> PUT /testing HTTP/1.1
> User-Agent: curl/7.30.0
> Host: 127.0.0.1:6984
> Accept: */*
>
< HTTP/1.1 201 Created
< X-Powered-By: Express
< Location: http://127.0.0.1:6984/testing
[...]

PR #4241, opened 2015-09-01
(#4240) - switch to phantom for pouchdb-server test

nolanlawson:
So I suspect this issue is related to the proxy using
too many connections (or something), and that it's
an issue the average user will never run into and so we
might as well just do the easy fix and switch to PhantomJS
instead of Firefox. PhantomJS is faster anyway.

Obviously eventually I want to track this down, but it's
more important to catch real bugs in PouchDB Server.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,7 +55,7 @@ env:
 
   # Test against pouchdb-server
   - CLIENT=node SERVER=pouchdb-server COMMAND=test
-  - CLIENT=selenium:firefox SERVER=pouchdb-server COMMAND=test
+  - CLIENT=selenium:phantomjs ES5_SHIM=true SERVER=pouchdb-server COMMAND=test
   - SERVER_ADAPTER=memdown LEVEL_ADAPTER=memdown SERVER=pouchdb-server COMMAND=test
 
   # Test against pouchdb-express-router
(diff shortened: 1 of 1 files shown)


nolanlawson:
Passed! :) The only thing failing now is `report-coverage`, which we really need to add to allowed failures.

daleharvey:
As I have mentioned, ever since we have been able to test websql in saucelabs I dont find any value in testing phantomjs, it isnt a browser, if there are bugs with connection limits etc in real browsers this isnt going to uncover anything. I would be +0 on adding phantom js but not having it replacing firefox

nolanlawson:
I still think PhantomJS is useful to test, because I see a lot of developers actually using it in practice, mostly because it is fast. As for Sauce, according to our test runs, the fastest Sauce test is about 20 minutes slower than either Firefox or PhantomJS - hence my preference for keeping the tests in Travis where possible. (Also now that I take another look, I see Firefox is actually a bit faster than PhantomJS, but both are still way faster than anything in Sauce.)

I definitely do want to solve this problem, but this PR is mostly a stopgap. The bug is irritating because it only occurs in one browser and you have to run the full test suite, with Selenium, in order to repro it (which takes like 30 minutes on my current machine). My goal is to get pouchdb-server back to green and then put the Firefox+Selenium fix on the back burner.

daleharvey:
Sorry I forgot to mention, we already have the node pouchdb-server tests which are green and we can use them to make sure we dont further break pouchdb-server so I am happy with that. the firefox tests are there to smoke test that they dont break in a real browser so they need to be kept

nolanlawson:
In that case, it makes more sense for me to just focus on fixing the Firefox test. I'll keep the issue open and do that this weekend.

Issue #4142, opened 2015-08-10
coveralls coverage reporting failing: XXX is not an object


daleharvey:
Our coverage reporting fails fairly regularly with this error, see https://travis-ci.org/pouchdb/pouchdb/jobs/74838103. Sounds like a bug in istanbul reporter

nolanlawson:
yeah, it is a bug in coveralls as far as I can tell (not istanbul). Maybe best to just ignore it and wait for a fix from coveralls

daleharvey:
Tracking it @ https://github.com/nickmerwin/node-coveralls/issues/80

nolanlawson:
So since we upgraded to Travis containers, the new error is:

    FATAL ERROR: JS Allocation failed - process out of memory
    Aborted (core dumped)

My hunch is that Istanbul just uses up too much memory. Sadly we may need to switch back to not using containers.

nolanlawson:
Supposedly Travis containers have more memory than the old non-containers, so surprised this is happening:

> The new containers have 2 dedicated cores and 4GB of memory, vs 1.5 cores and 3GB on our legacy infrastructure. CPU resources are now guaranteed, which means less impact from ‘noisy neighbors’ on the same host machine and more consistent build times throughout the day.

daleharvey:
Ugh going to have to look into patching this myself, now we are at 100% would like PR's to start reporting this so we dont regress

daleharvey:
So havent seen this in a long time

PR #4243, opened 2015-09-01
(#4243) - add report-coverage to allowed failures

nolanlawson:
It's been permafail for the past couple of days. Not
sure if the problem is coveralls, the script, etc...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -124,6 +124,7 @@ matrix:
     env: CLIENT=node COMMAND=test
   - env: CLIENT=node SERVER=pouchdb-server COMMAND=test
   - env: CLIENT=selenium:firefox SERVER=pouchdb-server COMMAND=test
+  - env: COMMAND=report-coverage
 
 
   fast_finish: true


daleharvey:
This was broken by the upgrade to travis, it occasionally failed on PR's previous so I ignored the failure, but it only started repeatedly failing on master when we upgraded travis. :+1: on moving to allowed failures and debugging, its not super critical but would be nice to get fixed

nolanlawson:
Hm, I could have sworn this was still failing before the Travis upgrade. But yeah, having a green master is more important than incremental Coveralls reports. I'll open an issue to fix Coveralls as well.

daleharvey:
It was failing on PR's before intermittently, that was being tracked @ https://github.com/pouchdb/pouchdb/issues/4142, however master was green up until travis (https://travis-ci.org/pouchdb/pouchdb/builds), now its failing to complete the test suite. I think we can just use this existing issue to get it fixed

https://github.com/pouchdb/pouchdb/commit/ef75880d4c32a6acb11a445867f2d25bb0927f79

PR #4250, opened 2015-09-02
DON'T MERGE - (#4142) - simply test coverage

nolanlawson:
Trying to confirm my suspicion that Istanbul is just
running out of memory.


nolanlawson:
Yup okay, so just running `COVERAGE=1 npm test` is also causing the out-of-memory: https://travis-ci.org/pouchdb/pouchdb/builds/78384247

Issue #4147, opened 2015-08-12
Request support for new CSRF feature (COUCHDB-2762) added to CouchDb 2.0


tohagan:
As you're probably aware, CouchDb 2.0 [just announced](https://issues.apache.org/jira/browse/COUCHDB-2762) support for CSRF Protection. Would be nice to be able to enable this as a PouchDb feature. In my web wanderings I also notice this issue for the Drupal PouchDb module:
  - https://www.drupal.org/node/2470691

CSRF protection is a seriously needed security feature for Couch and I think it should be added to the roadmap for Couch 1.x as well so makes sense if Pouch/Couch both plan to support it.

I'm also unsure if you support CouchDB 2.0 yet (perhaps you might clarify this on the Guide Introduction?).  Thanks guys.

tohagan:
Hmmm ... Maybe it's already in Couch 1.x  ... Just found this 

https://people.apache.org/~dch/snapshots/couchdb/20140715/new-docs-build/html/cve/2010-2234.html

tohagan:
Nope ... https://issues.apache.org/jira/browse/COUCHDB-2762 ... This is BAD!

tohagan:
And looks like it won't apply for READ operations: https://github.com/apache/couchdb-couch/pull/84

tohagan:
Sorry for my ramblings ... Just trying to get my head around security best practice for using Pouch/Couch.  

I *think* the best option is to use Cookie Auth (not include user/password in URL) by using pouchdb-authentication. This should be [ok for mobile apps that use Pouch inside a webview](http://security.stackexchange.com/questions/90994/are-phone-apps-vulnerable-to-xss-or-csrf-webview-phonegap-chrometab)  and hence don't share cookies with other browsers.  For desktop web apps using Couch 1.x however, CSRF is still a major issue that IMHO still needs to be resolved.

kxepal:
@tohagan backport to 1.x isn't a problem, should be easy to.

garrensmith:
Since we just added support for the new CSRF work in Fauxton, I'm happy to take a stab at getting this working in PouchDB.

daleharvey:
So I am not certain there is anything we need to do inside pouchdb? if you are using cookie authentication against couchdb shouldnt everything 'just work'?

garrensmith:
[...]

PR #4252, opened 2015-09-02
[WIP] Add csrf protection

garrensmith:
This adds support for the csrf protection implemented in CouchDB 2.
Currently this will not work with CORS as PouchDB does not have access
to the required cookies.

Adding this as a starting point for #4147. Currently all tests will fail as this won't work with CORS. Ideally if we could only enable this for requests that are not CORS requests that would be great start. I'm not quite sure of the best way of doing that.

### DIFF
--- a/lib/deps/ajax/prequest-browser.js
+++ b/lib/deps/ajax/prequest-browser.js
@@ -2,6 +2,22 @@
 
 var ajax = require('./ajax-core');
 
+function parseCookies (cookieString) {
+
+}
+
+function parseCookies (cookies) {
+  if (!cookies) {
+    return {};
+  }
+
+  return cookies.split(';').reduce(function (list, cookie) {
+    var parts = cookie.split('=');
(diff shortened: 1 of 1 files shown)


daleharvey:
So CouchDB has switched to only do CSRF validation on form submission, pouchdb only deals with application/json so this is no longer needed, apologies but thanks so much for figuring out the details @garrensmith

tohagan:
Your cookies are probably being blocked because both client and server need to allow credentials.
Access-Control-Allow-Credentials true

See this discussion ..

http://stackoverflow.com/questions/19770359/xmlhttprequest-cors-post-sent-without-cookies

garrensmith:
@tohagan the allow-credentials is set. Its an expected behaviour. When doing a CORS request the browser will receive and send the cookies. But we do not have access to them via document.cookie. See this [link](http://www.html5rocks.com/en/tutorials/cors/#toc-withcredentials) for more.

tohagan:
Thanks  @garrensmith ... will check it out.

Issue #4251, opened 2015-09-02
'paused' event not firing during sync


willholley:
As [reported on Slack](https://pouchdb.slack.com/archives/general/p1441187392000733), calling db.sync does not generate a 'paused' event.

This appears to be a regression in 4.0.1 compared to 4.0.0.

ng-:
Also, db.sync().once('paused') does not appear to be working in 4.0.0.  For now I am just using sync.on and counting manually but it is a bit annoying.

daleharvey:
So I am very confused, There is no code https://github.com/pouchdb/pouchdb/blob/master/tests/integration/test.sync.js at handling the `paused` / `active`` events, however I have definitely used them previously - https://github.com/daleharvey/done/blob/gh-pages/js/done.js#L120

daleharvey:
So testing, apparently this worked between 3.3.0 and 3.6.0, but not before or after

daleharvey:
In so far as I can tell, it looks like this worked completely accidently, there are not tests for it or code that directly handles the situation, it worked between 3.3 and 3.6 and there arent any particular commits around 3.6 that would indicate it breaking, frankly confused.

So for now I just think we implement it, I dont think `active` and `paused` can just forward the replication events, sync is only paused when both replications are paused, I will probably look at fixing the other bugs related to active / paused before I do this one (https://github.com/pouchdb/pouchdb/issues/3852)

colinskow:
I think it is also important to add a test for `active` event during sync.

daleharvey:
our events are undertested which is why this regressed, I will be copying across https://github.com/pouchdb/pouchdb/commit/a17be0693f325449ed7716d8dc49a7c3a0069b78 as the tests for sync and add a few more

colinskow:
Here's a simple Node test, which is successfully isolating the issue:

```js
var PouchDB = require('pouchdb');
var expect = require('chai').expect;

describe('PouchDB Sync Events', function() {

  var remoteDB = new PouchDB('http://localhost:5984/pouchdb_event_test');
  var localDB = new PouchDB('pouchdb_event_test');

  var doc1 = {
    _id: 'doc1'
  };

  var doc2 = {
    _id: 'doc2'
  };

  var previous = Promise.resolve();

  var sync, rev1, rev2;

  var activeDeferred = Promise.defer();
  var pauseDeferred = Promise.defer();

  it('should fire active and paused events on sync', function(done) {
    previous
      .then(function() {
[...]

PR #4254, opened 2015-09-02
(#4251) add test for paused event during sync

willholley:
The paused event should fire when there are no morechanges to process. This appears to have regressed in PouchDB 4.0.1.

This PR just adds a failing test.

### DIFF
--- a/tests/integration/test.retry.js
+++ b/tests/integration/test.retry.js
@@ -27,6 +27,21 @@ adapters.forEach(function (adapters) {
       testUtils.cleanup([dbs.name, dbs.remote], done);
     });
 
+    it('#4251 paused fires during sync', function (done) {
+      var db = new PouchDB(dbs.name);
+
+      var sync = db.sync(dbs.remote, {
+        live: true,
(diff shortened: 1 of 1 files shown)


daleharvey:
This is great cheers, only commenting on the failing test, but this will leak the replication once it starts to pass, you need to ensure the sync is completed before starting the next test. So 

```
var sync = db.sync....
sync.on('paused', function() { sync.cancel(); });
sync.on('complete', function() { done(); });
```

It doesnt look like we have these events tested with `sync` at all, which is fairly worrying, I am not entirely certain how they should work, we usually just forward replication events to the sync emitted, however `paused` and `active` have different meanings when it comes to sync

colinskow:
Fixed issue and added test for both `paused` and `active` events. https://github.com/pouchdb/pouchdb/pull/4360

daleharvey:
Thanks @willholley but theres are being followed up in https://github.com/pouchdb/pouchdb/pull/4360 with a bunch of tests, apologies for not merging

Issue #4256, opened 2015-09-03
explainCors-browser message is misleading in Firefox OS App


anaran:
https://github.com/pouchdb/pouchdb/blob/ac56d71cac049fd026ee1e8918bcb82d31192018/lib/deps/ajax/explainCors-browser.js#L5
suggests the couchdb server is not properly set up, but it is in my case.

I had to use the following instead to get couchdb _session cookie to work in my privileged Firefox OS App, which has systemXHR permission:

```javascript
    var myXHR = function () {
      var request;
      if (window.location.protocol == "app:") {
        request = new XMLHttpRequest({ mozSystem: true, mozAnon: true });
      }
      else {
        request = new XMLHttpRequest({ mozSystem: false, mozAnon: false });
      }
      return request;
    }
    var opts = {
      ajax: {
        xhr: myXHR,
        headers: { 'Cookie': cookie }
      }
    };
    var remoteDB = new PouchDB(destination + db._db_name, opts);
```

mnewmedia:
I also get this error, and i have enabled cors. I get this message **only when i stop the live Synchronisation**.

nickcolley:
Hey folks, sorry about the confusion, you can track this issue over here: https://github.com/pouchdb/pouchdb/pull/4190

anaran:
@nickcolley At first I thought this isssue is a duplicate of issue #4190, but that's a pull request, not an issue report.

Therefor, shouldn't this issue remain open, until 4190 comes up with a better solution?

Else people will deem this issue closed, unless they read all the fine print.

daleharvey:
We use issues and PR's pretty much interchangably, would have probably been easier to open a new issue once it got reported on the PR, but the discussion is already there and its easier to have the discussion in a single place.

daleharvey:
And that discussion should really be here in fact you were right, confusing to be reviewing PR's that arent in progress, but yeh we can soften the error message and not show it in situations where it certainly isnt a cors issue (we can test whether the url matches document.host I think)

nolanlawson:
According to https://github.com/pouchdb/pouchdb/pull/4190#issuecomment-137144008 there's no way to detect a CORS error from some other error. I'm thinking we might want to just remove the warning.

anaran:
@daleharvey `document.host` ?
How about

```
>> location.protocol
<- "app:"
```

daleharvey:
[...]

PR #4258, opened 2015-09-03
(#4256) - remove CORS warning message

nolanlawson:
If we can't consistently detect CORS problems, then I'd rather just not print anything out at all. This was an effort to be helpful, but frankly false positives can be more dangerous than false negatives.

I know that a lot of newbie PouchDB users have run into this error and been hopelessly confused about what to do, and it's even bitten me as recently as a few weeks ago (not sure how I missed it, stupid me). Hopefully people will google the mysterious words they see in their console (`No 'Access-Control-Allow-Origin' header`) and come upon our error page: http://pouchdb.com/errors.html#no_access_control_allow_origin_header

As a side note, I really dislike CORS.

### DIFF
--- a/lib/deps/ajax/ajax-core.js
+++ b/lib/deps/ajax/ajax-core.js
@@ -6,7 +6,6 @@ var errors = require('./../errors');
 var utils = require('../../utils');
 var applyTypeToBuffer = require('./applyTypeToBuffer');
 var defaultBody = require('./defaultBody');
-var explainCors = require('./explainCors');
 
 function ajax(options, callback) {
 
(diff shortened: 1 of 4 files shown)


nickcolley:
I'm +1 on this since it's a great effort but clearly not ideal.

Also I think people searching that error will have good success by the looks: https://www.google.co.uk/search?q=No+%27Access-Control-Allow-Origin%27+header&oq=No+%27Access-Control-Allow-Origin%27+header&aqs=chrome..69i57&sourceid=chrome&es_sm=93&ie=UTF-8#q=No+%27Access-Control-Allow-Origin%27+header+pouchdb

daleharvey:
I also hate CORS, however I would much rather improve this message than get rid of it, its continually bitten new users and I think we can vastly reduce the confusion it caused recently by changing the wording and also ensuring it only gets shown when we know the request is actually a cross origin request

I can get a PR up to do that shortly

daleharvey:
Closing with the +1 @ https://github.com/pouchdb/pouchdb/pull/4259

PR #4259, opened 2015-09-03
(#4265) - More checks and changed wording of CORS warning

daleharvey:


daleharvey:
So while it is true we cant detect for certain that something is a CORS error, it is something that users have come up against a lot and I think the original commit still had value. I think this will help reduce the confusion, the error wont show on same origin requests and the wording is a little clearer

lib/deps/ajax/explainCors-browser.js
```diff
@@ -1,9 +1,9 @@
 'use strict';
 
 module.exports = function () {
-  if ('console' in global && 'error' in console) {
-    console.error('PouchDB error: the remote database does not seem to have ' +
-      'CORS enabled. To fix this, please enable CORS: ' +
+  if ('console' in global && 'warn' in console) {
+    console.warn('PouchDB: the remote database may not have CORS enabled.' +
+      'If not please enable CORS: ' +
```
nolanlawson: comma after "If not"

nolanlawson:
+1, didn't realize you could do this kind of check. Very nice!

daleharvey:
Have no idea whats up with the performance test, just started failing on master too https://travis-ci.org/pouchdb/pouchdb/builds/78580886

So landing and will look into the performance tests now - https://github.com/pouchdb/pouchdb/commit/f05e41206e61824a85461f24a654998f79d99342

Cheers

PR #4260, opened 2015-09-03
Update request

omsmith:
~~Test failure due to https://github.com/nolanlawson/throw-max-listeners-error/issues/1~~

Such exciting things as working keep-alive in node > 0.10 https://github.com/request/request/pull/1715

### DIFF
--- a/lib/adapters/http/index.js
+++ b/lib/adapters/http/index.js
@@ -749,7 +749,7 @@ function HttpPouch(opts, callback) {
         // query string limits
         // see http://wiki.apache.org/couchdb/HTTP_view_API#Querying_Options
         method = 'POST';
-        body = JSON.stringify({keys: opts.keys});
+        body = {keys: opts.keys};
       }
     }
 
@@ -1009,7 +1009,7 @@ function HttpPouch(opts, callback) {
       headers: clone(host.headers),
       method: 'POST',
       url: genDBUrl(host, '_revs_diff'),
-      body: JSON.stringify(req)
+      body: req
     }, callback);
   });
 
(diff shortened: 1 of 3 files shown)


daleharvey:
@nolanlawson is this something you can fix in throw-listeners? looks like if you handle 0 this should be good

lib/adapters/http/index.js
```diff
@@ -1009,7 +1009,7 @@ function HttpPouch(opts, callback) {
       headers: clone(host.headers),
       method: 'POST',
       url: genDBUrl(host, '_revs_diff'),
-      body: JSON.stringify(req)
```
nickcolley: Is this done by request now?

nolanlawson:
Yeah I can publish an update to that lib when there's a fix.

lib/adapters/http/index.js
omsmith: Seems that way. In most cases strings _weren't_ being passed, so these ones were out of the norm (as far as I could tell).

CouchDB was returning errors ("body must be valid JSON"), because request was seeming stringifying _again_ (so an escaped string was being sent).

omsmith:
@nolanlawson sent https://github.com/nolanlawson/throw-max-listeners-error/pull/2 your way

omsmith:
Should I give an explicit bump to max-listeners?

daleharvey:
Looks like one last issue with the map reduce tests

omsmith:
Hmm alright. Different tests than I ran locally?

About to have people over, but I'll take a look afterward.

daleharvey:
The map reduce tests are seperate, `TYPE=mapreduce` npm test

daleharvey:
This is awesome, thanks :) - https://github.com/pouchdb/pouchdb/commit/fc48a9bff3cf04ab6a58424a0ab761b3debce341

omsmith:
@daleharvey Great - thanks very much.

Don't mean to be pushy or anything, wondering if there's an expected ETA for next release (v5.0.0 at this point?)?

Issue #4204, opened 2015-08-26
Ensure we're actually respecting revs_limit, stemming to 1000


nolanlawson:
Follow-up to https://github.com/pouchdb/pouchdb/issues/4195 and https://github.com/pouchdb/pouchdb/pull/4196. We have `revs_limit` hard-coded to 1000, but it seems that in the case of `new_edits=false`, we're not actually applying it, so it's possible to get revision histories > 1000.

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/13e9c18dce44d90366d902a2e5ba2f34184d5887

PR #4271, opened 2015-09-04
(#4204) - Ensure we respect revs_limit on new_edits writes

daleharvey:


daleharvey:
So @nolanlawson you are probably not going to love this, all new_writes edits have been hitting the database directly so they have all been avoiding `merge` which makes the writes faster but it can lead to us having unbounded revisions stored. This is likely going to cause a bit of a slowdown however its the correct behaviour now.

lib/deps/docs/processDocs.js
```diff
@@ -84,6 +84,9 @@ function processDocs(docInfos, api, fetchedDocs, tx, results, writeDoc, opts,
         updateDoc(fetchedDocs.get(id), currentDoc, results,
           resultsIdx, docWritten, writeDoc, newEdits);
       } else {
+        // Ensure stemming applies to new writes as well
+        var merged =  merge.merge([], currentDoc.metadata.rev_tree[0], 1000);
```
nolanlawson: ^ should probably extract that 1000 to a shared constant somewhere.

nolanlawson:
+1 when green. Also I'll look to see if there's a way to speed this up by traversing the tree to detect whether or not we're at <1000 revs.

lib/deps/docs/processDocs.js
daleharvey: Yeh I was planning to do https://github.com/pouchdb/pouchdb/issues/2839 at some point soon

daleharvey:
Green, merged cheers - https://github.com/pouchdb/pouchdb/commit/13e9c18dce44d90366d902a2e5ba2f34184d5887

Issue #3357, opened 2015-01-07
attachment names with a leading underscore should throw an error


marten-de-vries:
Currently PouchDB doesn't, which could give problems when syncing with CouchDB. A test inspired by one of couchdb-harness's:
```javascript
var db = new require('pouchdb')('test')
db.put({
  _id: "bin_doc2",
  _attachments:{
    "_foo.txt": {
      content_type:"text/plain",
      data: "VGhpcyBpcyBhIGJhc2U2NCBlbmNvZGVkIHRleHQ="
    }
  }
}).then(console.log.bind(console)); // no error
```
The error that should be thrown is:
{name: 'bad_request', message: 'Attachment name can't start with '_''}

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/57bc143a3920ecf1925ffa346d2d83e80520683d

PR #4277, opened 2015-09-05
(#3357) - Validate attachment names dont start with _

daleharvey:


nolanlawson:
+1 when green.

lib/adapter.js
```diff
@@ -153,6 +153,13 @@ function doNextCompaction(self) {
   });
 }
 
+function attachmentNameError(name) {
+  if (name.charAt(0) === '_') {
+    return 'Attachment name can\'t start with \'_\'';
```
kxepal: Why not to include bad attachment name into error? This will save time to figure out what exactly caused the issue.

daleharvey: Yeh will probably will do that on the way in, I was doing compatibility with Couch but them we avoid testing for exact error message so itll be fine

kxepal: Let's improve Couch errors as well! (:

daleharvey:
Cheers, merged with a more descriptive error - https://github.com/pouchdb/pouchdb/commit/57bc143a3920ecf1925ffa346d2d83e80520683d

Issue #4278, opened 2015-09-05
Gender Specific Language


Charlotteis:
Running [alexjs](https://github.com/wooorm/alex) over pouchdb I found some gender specific language we could get rid of. I will do that now!

daleharvey:
So replying to https://github.com/pouchdb/pouchdb/pull/4279#issue-105024186

The documentation ones were pretty clear but the blog post instances less so, in those cases we referring to someone who fairly sure identifies themselves as cis male, is it still better to use gender neutral by default there? seems sensible

Not entirely sure what the suitable replacement for crazy in this context is? 

And the last issue is one of those is a quote, assuming that part should be reworded a little to not be a quote to remove it

Charlotteis:
For the `pouchdb-levels-up` post, if Nick identifies as Male I see no real need to change it. If unsure, 'they' never hurts.

```diff
- What would have sounded like a crazy dream
+ What would have sounded like a dream
```

```diff
- Or whatever other crazy combination you can cook up!
+ Or whatever other combination you can cook up!
```
---
For the `secondary-indexes` post:
```diff
- the doc ID will be a crazy string
+ the doc ID will be a string
```

```diff
- he described it as "crazy slow."
+ they described it as "crazy slow."
```

Not sure what to do with that last one there with regards to the 'crazy'. Perhaps we leave it in as it is a quote? I'd rather get rid of it but I'm not sure on a good alternative.

nolanlawson:
Sounds like you've made a lot of headway here; good to close the issue @Charlotteis ?

Charlotteis:
Yup! Thanks @nolanlawson :)

PR #4280, opened 2015-09-05
(#4278) - Remove use of 'crazy' in documentation and tests

Charlotteis:
Seems alexjs isn't as robust as I would like it to be (it skips a lot of files) so I found these with a manual lookup in my files.

:sunny:

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -220,7 +220,7 @@ The `adapters` list is a comma-separated list that will be used for `PouchDB.pre
 
     http://localhost:8000/tests/index.html?adapters=websql
 
-Or even make the `preferredAdapters` list any crazy thing you want:
+Or even make the `preferredAdapters` list anything you want:
 
     # loads websql, then memory, then idb, then localstorage
     http://localhost:8000/tests/index.html?adapters=websql,memory,idb,localstorage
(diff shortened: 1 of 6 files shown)


nolanlawson:
Wow, surprised how often I use the word "crazy." +1

nolanlawson:
So this is unfortunately failing because of some changes to the mapreduce tests; I was testing the sorting of various strings, so `strange` does not serve as a substitute for `crazy` because it sorts differently. :)

>  AssertionError: expected 'multiple values!' to equal 'strange!'

Since these strings can be anything, how about `crayon` or something that will sort the same but without the offensive connotations?

Charlotteis:
Weirdly the tests passed for me when I did `npm test`, unless this is something different.
Will update in the morning :)

daleharvey:
The map reduce tests are seperate from the integration tests, needs to be `TYPE=mapreduce npm test` for them

Charlotteis:
@nolanlawson @daleharvey thanks for the help! Have run the mapreduce tests and all seems fine :)

nolanlawson:
Cool, thanks!

nolanlawson:
1e93a0df050171ae80fc067ffc8fa896f3641af4

PR #4281, opened 2015-09-05
(#4204) - avoid revs_limit stemming unless necessary

nolanlawson:
This is a tiny performance optimization, but should
apply to the vast majority of users, whose revs are
never exceeding 1000.

### DIFF
--- a/lib/deps/docs/processDocs.js
+++ b/lib/deps/docs/processDocs.js
@@ -85,8 +85,11 @@ function processDocs(docInfos, api, fetchedDocs, tx, results, writeDoc, opts,
           resultsIdx, docWritten, writeDoc, newEdits);
       } else {
         // Ensure stemming applies to new writes as well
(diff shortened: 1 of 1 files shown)


daleharvey:
This looks reasonable, slightly worried about it reintroducing unbounded revisions in very weird cases (a long branch who's head is not the current winning revision). That seems fairly unlikely and we have lived with unbounded revisions so far and its all green. I +.9 although I think we should only merge if the performance test do indicate that it is worth it.

It also may be worth putting into the merge function so it applies to all writes, not just new_edits:false

nolanlawson:
> a long branch who's head is not the current winning revision

I'd be happy to demonstrate that a long non-winner still ends up stemming it. I'm pretty sure I'm checking the *new revision*, not the *winning revision*, because of exactly the case you describe.

daleharvey:
Yup you are right, :+1: cheers

nolanlawson:
Did a perf test and don't see any benefit (in fact, a tiny detriment for some reason), so closing.

PR #4282, opened 2015-09-05
(#4282) - refactor merge.js, remove hyphens

nolanlawson:
Fixes two things that have been bugging me:

1. `merge.js` was too big, split up into separate files now
2. removed last filenames that still use `snake-case` instead of
   `camelCase`, with the exception of the `*-browser.js` files,
   which are our own convention that I rather like.


nolanlawson:
So unfortunately this browserifies a bit bigger due to the extra `function`s (while losing some weight due to function names being minified), 47430 -> 47454 min+gz. 24 extra bytes is acceptable to me.

daleharvey:
This looks good, cheers - https://github.com/pouchdb/pouchdb/commit/4ed6704f78fb75d80475390e40e2d86bf3e6d7f5

PR #4285, opened 2015-09-06
(#3941) - Add ability to set options per request

daleharvey:


daleharvey:
Based on https://github.com/pouchdb/pouchdb/pull/4146 because I dont want to keep bitrotting myself

In very very early days we supported `new PouchDB('name', {headers`, this removes support for that which cleans up a large amount of code, we now support passing the `{ajax: ` option to any HTTP request now (aside from `.id` and `.info` which have never supported opts), should probably add some documentation however it doesnt seem best to add it to every single api, may make a new section for it

nolanlawson:
Rebased for you since  #4146 is merged now. I'll note this PR already had a green run.

nolanlawson:
Very nice fix. 3c964adbfb9a187af79d1685d63497bdc6b9ecc6

nolanlawson:
#3941 has nothing to do with this PR, not sure which one you meant, forgot to check before merging. :P

nolanlawson:
Wait, is this a breaking change or not? Looks like `opts.headers` is still supported in there.

Issue #3356, opened 2015-01-07
filter function can throw uncatchable errors


marten-de-vries:
```javascript
var PouchDB = require('pouchdb');
var db = new PouchDB('test', {db: require('memdown')});
db.put({
  _id: "_design/test",
  filters: {
    test: function (doc, req) {
      throw new Error(); // syntaxerrors can't be caught either.
    }.toString()
  }
})
db.changes({filter: 'test/test'}).catch(function (err) {
  console.log("this statement is never reached although an error is thrown.")
});
```

Why is this bad? Because it crashes express-pouchdb. For bonus points, also known as skipping less of couchdb-harness, making the second argument of a filter function ``req`` specifiable as an option to ``changes()`` would be nice.

nolanlawson:
Ouch. Same thing happens with map/reduce functions btw.

nolanlawson:
fixed by 12cfc598402f0aebd3df8c669daf601ab9ec5086

PR #4292, opened 2015-09-08
(#3356) - Catch throws in filter functions

daleharvey:


lib/adapters/idb/index.js
```diff
@@ -480,7 +480,16 @@ function init(api, opts, callback) {
...
+
+        var filtered;
+        try {
+          filtered = filter(change);
+        } catch(err) {
+          var msg = 'Filter function threw: ' + err.toString();
+          return opts.complete(errors.error(errors.BAD_REQUEST, msg));
+        }
+
+        if (filtered) {
```
nolanlawson: Normally `try/catch`es should always go into their own functions to avoid deoptimization of the entire function. Normally I wouldn't nitpick, but I do know that `onGetWinningDoc()` is called a lot.

daleharvey: So I was thinking of putting this in filterChange but the return type seemed awkward, maybe if it succeeds we coorce the result into a `boolean` and if not return the error object?

nolanlawson:
Very nice.

Issue #4295, opened 2015-09-09
nodejs stable failing


daleharvey:

daleharvey:
Updating g++ fixes this, will PR shortly - https://github.com/pouchdb/pouchdb/pull/4296

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/f046750b0b9189c65a0620c2ee98b475b7e58285

PR #4297, opened 2015-09-09
(#4295) - Use g++4.8 for stable nodejs

daleharvey:


daleharvey:
Taking an IRC +1 since all builds will fail without this applied, cheers @Charlotteis - https://github.com/pouchdb/pouchdb/commit/f046750b0b9189c65a0620c2ee98b475b7e58285

nolanlawson:
LGTM, but hope we don't have to do this kind of thing often.

Issue #4298, opened 2015-09-09
Update express-js et al


daleharvey:

daleharvey:
For dependencies that are tightly tied to pouchdb and that should be tested against pouchdb master before upgrading, I think we should use *

daleharvey:
We are all green now https://github.com/pouchdb/pouchdb/commit/984ff0332c6d5c6d04a85d5b17f8e98b3cb7ad7a

PR #4299, opened 2015-09-09
(#4298) - Update pouchdb-server etc

daleharvey:


nickcolley:
This makes sense to me, is there a risk of getting failing builds more frequently if these deps fail?

daleharvey:
pouchdb-express-router and pouchdb-server are both in allowed_failures, if these deps update and break the tests it just means they will know earlier, it wont fail the pouch test suite and since they are upstream projects they can never actually break pouchdb

However I assumed this would fix pouchdb-server at least in firefox, it hasnt so ill take another look

nolanlawson:
Weird, I did like 4 Firefox runs and it didn't show up. Guess the intermittent is still there.

PR #4303, opened 2015-09-10
(#3356) - put try/catch inside own function

nolanlawson:
This ensures that the deoptimized try/catch only affects
users who are actually filtering.


daleharvey:
Nice cheers - https://github.com/pouchdb/pouchdb/commit/29ca4deab87bcc074b019d684c4f3ac800cdf27d

PR #4304, opened 2015-09-10
try not caching node modules

nolanlawson:
I kinda wonder if our fancy `ln -s` linking of
node modules is what's breaking the pouchdb-server
tests.


nolanlawson:
Nope; I found the cause. It's a `request` thing: https://github.com/pouchdb/express-pouchdb/issues/255#issuecomment-139425431

Issue #4305, opened 2015-09-10
Coverage is permafail


daleharvey:

daleharvey:
fixed in https://github.com/pouchdb/pouchdb/commit/8f606e16ce57f2224a615a797d08d404317004fc

PR #4306, opened 2015-09-10
(#4305) - Temporarily revert to old travis infra

daleharvey:


daleharvey:
We obviously want to get on the new infrastructure, but we are losing the coverage reporting and it looks like it may be causing other version issue, so lets revert it temporarily, get the tree nice and green again then see about turning it back on

nolanlawson:
Yes absolutely; I was too hasty about migrating to the new system. Our tree is so red right now it ain't funny.

.travis.yml
```diff
@@ -9,12 +9,12 @@ services:
...
+# sudo: false
 
-cache:
-  directories:
-    - node_modules
-    - /tmp/phantomjs
+# cache:
+#   directories:
+#     - node_modules
+#     - /tmp/phantomjs
```
nolanlawson: cache might still be good regardless, but yeah, let's turn it on after we get everything green

nolanlawson:
I believe this will break couchdb-master (already broken) unless the apt-get works in sudoless mode, but +1 when green regardless.

nolanlawson:
lol, coverage is still failing with `Not a valid object name`. We can't win, Dale.

nolanlawson:
Green, merged.

PR #4308, opened 2015-09-11
(pouchdb/express-pouchdb#255) - use node 0.12 as default

nolanlawson:
I'm not really motivated to fix a PouchDB Server bug that only
occurs in Node 0.10. Let's just make 0.12 the default and
then test CouchDB using 0.10.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 
 node_js:
-  - "0.10"
+  - "0.12"
 
 services:
   - couchdb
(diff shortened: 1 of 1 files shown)


daleharvey:
My god it is so green

daleharvey:
merged - https://github.com/pouchdb/pouchdb/commit/984ff0332c6d5c6d04a85d5b17f8e98b3cb7ad7a

This is awesome, thanks

PR #4312, opened 2015-09-11
Revert "(#4305) - Temporarily revert to old travis infra"

nolanlawson:
This reverts commit 8f606e16ce57f2224a615a797d08d404317004fc.

Maybe if we're lucky, we'll no longer get an OOM because
we're on Node 0.12? If this is red for `report-coverage`, though,
then I don't want to merge it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,12 +9,12 @@ services:
 git:
   depth: 30
 
-# sudo: false
+sudo: false
 
-# cache:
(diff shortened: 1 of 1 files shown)


nolanlawson:
Istanbul is still failing; we need to report this to the Travis folks: https://travis-ci.org/pouchdb/pouchdb/jobs/79838632

PR #4313, opened 2015-09-11
(#4313) - fix options usage within http adapter

omsmith:
@daleharvey noticed some oddities in the diff from 3c964adbfb9a187af79d1685d63497bdc6b9ecc6

1st commit: `getHost` no longer uses `opts`. So in particular, we definitely dont need to clone them. Still passes `opts` to `getHostFn` in case the passed in implementation uses it.

2nd commit: after the change, the input `opts.ajax` is being modified within the `if (opts.auth || host.auth)` so this just moves the `clone` earlier

Can fix-up commit messages / add regression test for the second case, but was able to just do this quick to get it in front of you

### DIFF
--- a/lib/adapters/http/index.js
+++ b/lib/adapters/http/index.js
@@ -63,7 +63,7 @@ function preprocessAttachments(doc) {
 
 // Get all the information you possibly can about the URI given by name and
 // return it as a suitable object.
-function getHost(name, opts) {
+function getHost(name) {
   // Prase the URI into all its little bits
   var uri = utils.parseUri(name);
 
@@ -83,8 +83,6 @@ function getHost(name, opts) {
   // Restore the path by joining all the remaining parts (all the parts
   // except for the database name) with '/'s
   uri.path = parts.join('/');
-  opts = opts || {};
-  opts = clone(opts);
 
   return uri;
 }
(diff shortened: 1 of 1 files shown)


daleharvey:
This looks sensible, thanks for catching, will see what CI says now

lib/adapters/http/index.js
```diff
@@ -63,7 +63,7 @@ function preprocessAttachments(doc) {
 
 // Get all the information you possibly can about the URI given by name and
 // return it as a suitable object.
-function getHost(name, opts) {
+function getHost(name/*, opts*/) {
```
daleharvey: Should get rid of this

daleharvey:
Awesome green run :) not sure whats up with the allowed failures but its probably nothing, restarted.

If you want to address the nit and squash these up with the right commit message, I dont think tests are absolutely required for this change, but I am mostly not committing already in case you do want to add them.

omsmith:
removed the commented param, squashed, and reworded

daleharvey:
Awesome, thanks - https://github.com/pouchdb/pouchdb/commit/d9e9726f6cc3a9b7a1b00408d9e333c586b5a64a

Issue #4315, opened 2015-09-11
Firefox: specify persistent storage


scottmtp:
Firefox has a non-standard API for specifying persistent storage when opening an IndexedDB. It would be helpful to expose this in PouchDB.

See:
https://developer.mozilla.org/en-US/docs/Web/API/IDBFactory/open

scottmtp:
I updated the PR to fix the line length in case that's helpful. Also, it is not clear to me how I'd be go about testing this change, but if anyone has thoughts about how to test, I'd be happy to update the PR with tests as well.

daleharvey:
This is pretty much impossible to test so I am happy enough without a test, this is great thanks. One thing that will likely need following up is a note in the documentation.

https://github.com/pouchdb/pouchdb/commit/eeb6a242ac63f52b44606e956b5d3aec5750c331

Thanks

PR #4316, opened 2015-09-11
Add support for a 'persistentIndexedDB' option when creating a PouchDB

scottmtp:
This option will enable persistent storage on Firefox only. See https://github.com/pouchdb/pouchdb/issues/4315.

### DIFF
--- a/lib/adapters/idb/index.js
+++ b/lib/adapters/idb/index.js
@@ -806,7 +806,12 @@ function init(api, opts, callback) {
     return;
   }
 
-  var req = indexedDB.open(dbName, ADAPTER_VERSION);
+  var req;
+  if (opts.persistentIndexedDB) {
+    req = indexedDB.open(dbName, {version: ADAPTER_VERSION, storage: 'persistent'});
+  } else {
+    req = indexedDB.open(dbName, ADAPTER_VERSION);
+  }
 
   if (!('openReqList' in IdbPouch)) {
     IdbPouch.openReqList = {};
--- a/lib/adapters/idb/index.js
+++ b/lib/adapters/idb/index.js
@@ -808,7 +808,8 @@ function init(api, opts, callback) {
 
   var req;
   if (opts.storage) {
-    req = indexedDB.open(dbName, {version: ADAPTER_VERSION, storage: opts.storage});
+    req = indexedDB.open(dbName, {version: ADAPTER_VERSION,
+      storage: opts.storage});
   } else {
     req = indexedDB.open(dbName, ADAPTER_VERSION);
   }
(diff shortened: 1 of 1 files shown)


daleharvey:
Nice thanks a lot, this looks great. however to bikeshed on this api, I am thinking 

```
new PouchDB('name', {storage: 'permanent'}); 
```
with the code looking a little like
```
opts.storage = opts.storage || 'temporary';
var req = indexedDB.open(dbName, {version: ADAPTER_VERSION, storage: opts.storage});
```
Its flexible in case new types get added and a little cleaner imo, what do you think? I think `storage` would be enough for the key name but its possible we may want to specify its for indexeddb, if so I cant remember if we are snake case or camel case these days, @nolanlawson ?

nolanlawson:
Snake case is more my cup of tea, because it matches the CouchDB options.

And yeah, I would definitely be down for a `storage` option on the IndexedDB adapter; it would also open us up to support things like `fakeIndexedDB` or `shimIndexedDB` if we really wanted to.

Out of curiosity, why did Moz create this separate IndexedDB API?

daleharvey:
Mozilla devs are worried about content filling up indexedDB to the point where the device is out of space. The basic semantics are that `permanent` will cause an initial prompt, but is never evicted, and `temporary` will not cause a prompt, but will be evicted when the device storage limit is reached based on LRU

scottmtp:
Good suggestion, we may as well allow the user to specify the 'storage' value we're giving Firefox.

Is this what you were thinking:
https://github.com/pouchdb/pouchdb/commit/3e4e42351b892331b994d898ab2e60529e00cff6

I think we still need the test for opts.storage so we know whether the second arg to open should be an integer or an object.

daleharvey:
Actually you are right, I didnt consider that other idb implementations may not support the properties object, probably safer to only use it if its specified

daleharvey:
I think this commit is good, just need to reformat it to not break the linter, we have an 80char limit on line length, Thanks :)

daleharvey:
Awesome cheers, merged in https://github.com/pouchdb/pouchdb/commit/eeb6a242ac63f52b44606e956b5d3aec5750c331

nickcolley:
Do we need to document this?

CatchSandeepVaid:
I think we are talking about providing persistent indexeddb storage for firefox only (as Firefox has a non-standard API for specifying persistent storage when opening an IndexedDB) BUT
This link [https://demo.agektmr.com/storage/](url) says that even we can use indexeddb for unlimited storage. Cant we use this in pouchdb and  provide persistent (or unlimited storage) for chrome on android as well?

PR #4336, opened 2015-09-15
(#4195) - remove vuvuzela

nolanlawson:
Been dong some more performance profiling, and this is
still a drag, mostly due to the try/catch deoptimization
in V8. I would also like to reduce the bundle size.

If this is passing in all our Saucelabs browsers, then I
am happy to remove it. If we're lucky, the recent fixes to
actually respect revs_limit at 1000 will mean this will
work now.

Keep in mind, though, that these tests are disabled for
Safari/iOS due to the popup, so we should probably double-check
them manually (specifically the #2543 tests).

### DIFF
--- a/lib/adapters/idb/utils.js
+++ b/lib/adapters/idb/utils.js
@@ -54,7 +54,7 @@ exports.idbError = function (callback) {
 // format for the revision trees other than JSON.
 exports.encodeMetadata = function (metadata, winningRev, deleted) {
   return {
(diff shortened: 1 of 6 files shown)


nolanlawson:
To wit:

screenshot 2015-09-15 00 41 50

nolanlawson:
Getting "too much recursion" errors in Firefox. It was a beautiful dream.

Issue #3535, opened 2015-02-13
Document skipSetup


nolanlawson:
Also while we're at it, can we alias it to `skip_setup` and document that instead? I am pretty annoyed by our mixed camelCase and snake_case.

dholth:
+1

daleharvey:
Done - https://github.com/pouchdb/pouchdb/commit/d1760732b8b95fdd92117d8781b6e5f0fa705e29

PR #4338, opened 2015-09-15
(#3535) Document and alias skipSetup (skip_setup)

janraasch:
This tries to solve https://github.com/pouchdb/pouchdb/issues/3535.

I left `skipSetup` in here for compatibility reasons, but added a `TODO` to remove this in a future (major) release.

Let me know, if this works for you.


nolanlawson:
This looks great to me; Travis seems to have run into an intermittent error, but I will restart it, and I'm definitely +1 when it's green.

`camel_case` isn't very JavaScripty, but it's very Couchy, so for consistency's sake I'm in favor of removing all camelCase stuff from the options. Next one that ought to go is `returnDocs`.

daleharvey:
Awesome, thanks a lot, merged in https://github.com/pouchdb/pouchdb/commit/d1760732b8b95fdd92117d8781b6e5f0fa705e29

Issue #4339, opened 2015-09-15
Unclear error when trying to use instance after .destroy


gr2m:
We have a failing test in https://github.com/hoodiehq/pouchdb-hoodie-api after upgrading to 4.0.2: https://github.com/hoodiehq/pouchdb-hoodie-api/issues/78

I could track down the issue to this

```js
// run with node.js
var PouchDB = require('pouchdb')
var db = new PouchDB('test', {
  db: require('memdown')
})

// just to make sure the init of the db is finished properly
// this can be left out, it has no effect in that context
db.info()

.then(function () {
  return db.put({
    _id: 'cleanTest'
  })
})

  .then(function () {
    console.log('cleanTest created')
    return db.destroy()
  })

  .then(function () {
    console.log('db removed')

    // PROBLEM:
    // this one does never settles settled.
    return db.get('cleanTest')
  })

  .then(function () {
    console.log('.then called after db.get("cleanTest")')
  })
  .catch(function () {
    console.log('.catch called after db.get("cleanTest")')
  })

```

It's very odd. The node script ends without either calling neither `.then` or `.catch` at the end. How is that even possible?

nickcolley:
I think the constructor now doesnt return a promise, but rather the api will is destroy stopping .get from setting up a new db?

nickcolley:
Confirmed with @gr2m that any api methods fail, for instance a .put

Also this is only an issue with leveldown and *down based adapters.

nickcolley:
https://github.com/pouchdb/pouchdb/blob/f8c06ac76d321e84bdcab25f9cf02f5d648e06af/lib/adapters/leveldb/index.js#L345

Doesn't return at all after db.destroy

nickcolley:
1. check this out https://github.com/pouchdb/pouchdb/pull/4340
2. run GREP=test.basics.js npm test
3. The get never returns

nickcolley:
Once the database is destroyed https://github.com/dominictarr/level-sublevel/blob/master/nut.js#L116 doesnt get called anymore

nolanlawson:
Awesome work tracking this down! So I definitely think this is more of a documentation issue than a bug in PouchDB. My opinion about the `destroy()` API is that it's the event horizon - once you call it, you never hear from that database object again. The solution is:

```js
var db = new PouchDB('foo');
return db.destroy().then(function () {
  db = new PouchDB('foo');
  return db.get();
});
```

[...]

PR #4340, opened 2015-09-15
(#4339) - Failing get after destroy test

nickcolley:


lib/adapters/leveldb/index.js
```diff
@@ -342,6 +342,7 @@ function LevelPouch(opts, callback) {
     opts = utils.clone(opts);
 
     stores.docStore.get(id, function (err, metadata) {
+      console.log(id);
```
gr2m: don't forget to remove this one again ;)

nickcolley: left it in so you could see that it never returns

daleharvey:
Yup this has come up before, the behaviour this is testing was never one that we have catered for, it would have sometimes worked (and sometimes not) n the http adapter, I think almost always break in any of the other adapters, when you call `db.destroy` we generally expect that to be the end of the life of the object, if you need another db handle you need to reopen it.

We can wipe out the handle when `destroy` is called, make it throw an obvious error if someone calls any further functions on it, we could also support this use case (being able to reuse the handle after `destroy` by making all the constructor stateless.

nickcolley:
It fails silently at the moment so no error message :(

nolanlawson:
Maybe we should put a `help wanted` on this and open up an issue? Should be a pretty straightforward fix to have the promise reject with a helpful error message.

nolanlawson:
Since nobody seems to be attacking this, I'm going to take it to mean that everybody is happy with the workaround where you just don't try to re-use a destroyed database.

Issue #4329, opened 2015-09-13
Add FruitDOWN as a supported adapter


nolanlawson:
I wrote this thing: https://github.com/nolanlawson/fruitdown

Presumably this is much more useful to our users than idb-alt was, so now that we've gotten rid of it, we can replace it with `pouchdb.fruitdown.js` and offer a very useful adapter that actually works in environments like WKWebView and allows unlimited storage in iOS/Safari.

I might not have time to do this today, but anybody who wants to add this thing only needs to look at https://github.com/pouchdb/pouchdb/commit/427efbcdf71e965a3aa34fc68641f948a66ae8b6 (which removed idb-alt) and basically revert it, but use fruitdown instead of level-js. Also for the adapter name, I'd suggest just calling it `'fruitdown'`.

nolanlawson:
fixed by 25da9e5 and a0fc6a9

PR #4343, opened 2015-09-16
(#4329) - Add fruitdown as a supported adapter

nolanlawson:
@nickcolley I added some more documentation about FruitDOWN
and upped it to 1.0.1, which should be passing the tests now,
even after a refresh. I also set Travis to test FruitDOWN
using Safari 8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,82 +54,83 @@ env:
 
   matrix:
   - CLIENT=node COMMAND=test
-  # - CLIENT=node LEVEL_PREFIX=foo_ COMMAND=test
+  - CLIENT=node LEVEL_PREFIX=foo_ COMMAND=test
 
   # Test against pouchdb-server
-  # - CLIENT=node SERVER=pouchdb-server COMMAND=test
-  # - CLIENT=selenium:firefox SERVER=pouchdb-server COMMAND=test
-  # - SERVER_ADAPTER=memdown LEVEL_ADAPTER=memdown SERVER=pouchdb-server COMMAND=test
+  - CLIENT=node SERVER=pouchdb-server COMMAND=test
+  - CLIENT=selenium:firefox SERVER=pouchdb-server COMMAND=test
+  - SERVER_ADAPTER=memdown LEVEL_ADAPTER=memdown SERVER=pouchdb-server COMMAND=test
 
   # Test against pouchdb-express-router
-  # - CLIENT=node SERVER=pouchdb-express-router COMMAND=test
+  - CLIENT=node SERVER=pouchdb-express-router COMMAND=test
 
   # Test in firefox/phantomjs running on travis
-  # - CLIENT=selenium:firefox COMMAND=test
-  # - CLIENT=selenium:phantomjs ES5_SHIM=true COMMAND=test
+  - CLIENT=selenium:firefox COMMAND=test
+  - CLIENT=selenium:phantomjs ES5_SHIM=true COMMAND=test
 
   # Test auto-compaction in Node, Phantom, and Firefox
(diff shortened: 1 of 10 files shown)


nolanlawson:
Also if anyone would like to try the fruitdown adapter with fruitdown 1.0.1:  [pouchdb.fruitdown.js](https://nolanlawson.s3.amazonaws.com/pouchdb/www/20150916/pouchdb.fruitdown.js)

daleharvey:
:+1: this is awesome

nickcolley:
Should we make the fruitdown adapter only valid in the broken implementations? This means people can use it almost like a polyfill..

ssured:
This is so awesome! Infinite thanks. I just tested on my device and it works like a charm. No more limits and popups on iOS.

:+1: You guys rock! :+1:

nolanlawson:
@nickcolley I am really hesitant to make it a supported part of core PouchDB. The WebSQL adapter is preferable in 99% of situations; it's just the odd ones like WKWebView or where you want to store >5MB where it's useful. FruitDOWN itself is slower, heavier, and more memory-intensive than either our WebSQL or our IndexedDB adapters.

I mostly see this as a temporary stopgap while we wait for the WebKit devs to fix IndexedDB. Notice I've been diligently filing issues on their Bugzilla, with test cases. :) E.g. [this one](https://bugs.webkit.org/show_bug.cgi?id=149205) and [this one](https://bugs.webkit.org/show_bug.cgi?id=149107).

nolanlawson:
If we want to make it easier for people, though, we could suggest something like

```js
 var isSafari = typeof openDatabase !== 'undefined' &&
    /(Safari|iPhone|iPad|iPod)/.test(navigator.userAgent) &&
    !/Chrome/.test(navigator.userAgent) &&
    !/BlackBerry/.test(navigator.platform);
var db;
if (isSafari) {
  db = new PouchDB('db', {adapter: 'fruitdown'});
} else {
  db = new PouchDB('db');
}
```

nickcolley:
I'm on your wave length I just wonder if we could do that automatically, I think I'm swaying towards no but just thought worth mentioning. :+1:

nolanlawson:
Ironically this is failing in Safari 8, and I think it might be genuine. Need to test to debug.

ssured:
I experienced a failure too; but when wiping the cache of my ios8 ipad mini 1 it did work for me. On chrome desktop I had the same behaviour. First I got a 'Cannot open indexeddb, are you in private browsing mode?' error. That page had a default adapter pouchdb running in a previous session, which for testing purposes I switched to fruitdown. Clearing the cached data in chrome made it work.

nickcolley:
@nolanlawson does iOS8 use the same IDB as Safari 8, wondering if it's worth testing iOS8 as well as Safari 8

nolanlawson:
@nickcolley I've never seen bugs in one that didn't also exist in the other. But in any case, we seem to have genuine failures here for some reason. :(

nolanlawson:
Something still wrong in Safari/iOS, need to investigate: https://github.com/pouchdb/pouchdb/pull/4350

nolanlawson:
if this build passes, then we're good: https://travis-ci.org/pouchdb/pouchdb/builds/83596028

daleharvey:
And all passing :) this is great, :+1: but needs some merge conflicts fixed

nolanlawson:
The merge conflicts were just in .travis.yml. Fixed them and merged to master as 25da9e52a34b20f92cb871d410affd94d717670f and a0fc6a972a3332b91877f5f5c23efad903b7dc84

Issue #4293, opened 2015-09-09
sync with live: true triggers "change" events twice


gr2m:
```js
var PouchDB = require('pouchdb').defaults({
  // memdown makes no difference, just added for sake of simplicity
  db: require('memdown')
})

var db1 = new PouchDB('db1')
var db2 = new PouchDB('db2')

db1.put({_id: 'a'})

.then(function () {
  return db2.put({_id: 'b'})
})

.then(function () {
  db1.sync(db2, {live: true}).on('change', function (change) {
    console.log(change.direction, change.change.docs)
  })
})
```

Output looks like this

```js
$ node debug.js 
push [ { _id: 'a', _rev: '1-4c355ea809e3720366f3ce63e030039f' } ]
pull [ { _id: 'b', _rev: '1-30da76a9658982ad291add2de458cb7c' } ]
push [ { _id: 'a', _rev: '1-4c355ea809e3720366f3ce63e030039f' } ]
pull [ { _id: 'b', _rev: '1-30da76a9658982ad291add2de458cb7c' } ]
```

WIthout the `live: true` option, it correctly triggers the "change" events only once per document. Is this by design? Or some kind of race condition maybe?

daleharvey:
Certainly not by design, just a straight up bug I think

Charlotteis:
I poked around a bit: I'm unfamiliar with how everything fits together but I can say with confidence that your journey to fix this may start with this line: https://github.com/pouchdb/pouchdb/blob/master/lib/sync.js#L80

```
  if (opts.live) {
    this.push.on('complete', self.pull.cancel.bind(self.pull));
    this.pull.on('complete', self.push.cancel.bind(self.push));
  }
```

Not a great lot of help, but a start :)

nolanlawson:
Mmm yeah if this only occurs with `sync()` but not with `replicate()`, then it's definitely a bug in `sync.js`.

daleharvey:
Added a test here, its a bug in `replicate` - https://github.com/pouchdb/pouchdb/pull/4348

daleharvey:
Fairly trivial fix - https://github.com/pouchdb/pouchdb/commit/764ba8a1f3b9be807e96c87d8c06bffd30e97c31, however looks like the retry tests are broken somehow (they depend on extra events being fired) and now failing

daleharvey:
[...]

PR #4348, opened 2015-09-17
(#4293) - Prevent replication firing extra events

daleharvey:
[skip ci]

### DIFF
--- a/lib/replicate/replicate.js
+++ b/lib/replicate/replicate.js
@@ -103,6 +103,12 @@ function replicate(src, target, opts, returnValue, result) {
   }
 
   function finishBatch() {
+    result.last_seq = last_seq = currentBatch.seq;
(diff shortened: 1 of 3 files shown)


daleharvey:
So the issue here was that we fired a `change` event when we had done duplicate `new_edits: false` writes (ie nothing had changed), We now only fire change events when something has changed. 

However some tests depending on this event because on the event of a a checkpoint failure we never sent a change event for the documents we have still actually written, so I moved the event trigger to before the checkpointing and it will now fire if the documents are written, regardless of whether the checkpoint was written

nolanlawson:
+1

daleharvey:
Awesome cheers - https://github.com/pouchdb/pouchdb/commit/fe4cc24dbe6a93ddc072c7dd3bacaec70120f74d

Issue #3852, opened 2015-05-16
"active" and "paused" fire too many times


nolanlawson:
I finally have tests that actually simulate offline state: https://github.com/pouchdb/pouchdb/pull/3851. Unfortunately it reveals (as noted in https://github.com/pouchdb/pouchdb/issues/3848), that we fire the `'active'` and `'paused'` events way too many times.

Even if we never go offline, those active and paused events fire 5 times in that sample test, which indicates to me that it's just firing for every `'change'`. Definitely something we want to fix.

nolanlawson:
I think in order to solve this problem, the entire replicator should probably be rewritten. I've gone in a couple of times and tried to get my hands dirty, but it's just too complicated. Really hard to tell what's going on in `replicate.js`.

nolanlawson:
^ I've started writing some unit tests that I think ought to be fixed when this issue is fixed.

* If no retries occur, then there should be 1 active event and 1 paused event.
* If 1 retry occurs, then there should be 2 active events and 2 paused events.
* If 2 retry occurs, then there should be 3 active events and 3 paused events.

And so on. Right now there are like 10 active events and 10 paused events even with no retries. I don't know why.

daleharvey:
So there have been a bunch of event changes and these tests seem to confirm the events are looking ok, if we see any more issues can open bugs for them directly, cheers - https://github.com/pouchdb/pouchdb/commit/a17be0693f325449ed7716d8dc49a7c3a0069b78

PR #4352, opened 2015-09-18
(#3852) - New tests for replication events

daleharvey:


daleharvey:
So these tests found one bug, it looks like the event listener fixes have fixed most of the problems with the multiple `paused` / `active`, I wouldnt be surprised if there were some more hidden cases but it looks like we are generally ok, will use these as the basis for testing https://github.com/pouchdb/pouchdb/issues/4251 too

nolanlawson:
Awesome, +1.

tests/integration/test.replication_events.js
```diff
@@ -0,0 +1,174 @@
...
+            db.bulkDocs([{_id: 'e'}, {_id: 'f'}]);
+          } else if (counter === 6) {
+            // All good, wait for pause
+          } else {
+            done('active fired incorrectly');
+          }
+        });
+
+        repl.on('paused', function(err) {
+          counter++;
```
nolanlawson: ^ apparently Sinon can get rid of this manual counter stuff.

daleharvey: Its possible, I try to reserve sinon for actually mocking implemetation in unit tests, I find a lot of times when people put sinon in integration tests they hide fairly large errors, but will see how sinon is working in the other PR for the replication options

daleharvey:
Sweet cheers - https://github.com/pouchdb/pouchdb/commit/a17be0693f325449ed7716d8dc49a7c3a0069b78

Issue #4018, opened 2015-07-01
PUTting to a full database hangs


alxndrsn:
## Environment

Android Emulator with platform 4.4.2, API Level 19; both of
* Default browser (Chrome?)
* WebView in java app
* pouch-db 3.6.0

## Observed behaviour

When a database has too much stuff in it, no callback is made from the final call to `db.put()`.

## Expected behaviour

Callback should be called with an error, or doc detailing successful put

## Steps to reproduce:

1. call `db.put()` a lot of times (around 3,500 times with a doc size of approx 2KB)
2. call `db.put()` one more time

Example code at: https://github.com/alxndrsn/pouch-db-capacity-testing/tree/full-pouch-demo

## Extension

If a further `db.put()` call is made while there is one "in progress", the following error is given to callback:

> `"InvalidStateError: Failed to execute 'transaction' on 'IDBDatabase': The database connection is closing."`

If the page is reloaded, and a `put()` or `drop()` call is made, the following error is received by the callback:

> `{"status":500,"name":"error","message":"unknown"}`

alxndrsn:
To reproduce on desktop (Chromium 44.0.2395.0 (64-bit)):

1. `db.put()` until out of device space
2. observe that `db.put()` call has hung
2. free up some space
3. click "add doc" button
4. observe error:

> `InvalidStateError: Failed to execute 'transaction' on 'IDBDatabase': The database connection is closing.`

alxndrsn:
Seems to be a Chrome issue - with Firefox 38.0.5:

The following error is returned when space runs out on device:

> `{"status":500,"name":"error","message":"unknown"}`

Subsequent `db.put()` calls return:

> `"InvalidStateError: A mutation operation was attempted on a database that did not allow mutations.`

alxndrsn:
# Safari 8.0.6:

 ## When space runs out on device:

1. This error is reported in the console from `pouchdb-3.6.0.min.js:8:10761`:

> `IndexSizeError: DOM Exception 1: Index or size was negative, or greater than the allowed value.`

2. This error is passed to the callback:

> `{"status":500,"name":"Object","message":"unknown"}`

After freeing space, subsequent calls to `db.put()` work fine.

daleharvey:
@alxndrsn Thanks so much for this report, I was worried about how hard this would be to reproduce but you already have a test repo up and running, thats amazing.

Ill take a shot at running those tests / possible getting them running on travis, would you be up for us merging them into pouchdb?

[...]

PR #4354, opened 2015-09-20
(#4018) - catch errors from websql openDatabase()

nolanlawson:
This is impossible to test in integration as far as I can tell,
but I can confirm manually that this will catch synchronous
errors thrown by `openDatabase()` in cases like e.g. DOM Exception
18 (in WKWebView or Android WebView if WebSQL is disallowed) or
if the wrong arguments are passed in to `openDatabase()`.

### DIFF
--- a/lib/adapters/websql/bulkDocs.js
+++ b/lib/adapters/websql/bulkDocs.js
@@ -18,7 +18,7 @@ var ATTACH_AND_SEQ_STORE = websqlConstants.ATTACH_AND_SEQ_STORE;
 var select = websqlUtils.select;
 var stringifyDoc = websqlUtils.stringifyDoc;
 var compactRevs = websqlUtils.compactRevs;
-var unknownError = websqlUtils.unknownError;
+var unknownError = websqlUtils.websqlError;
 
 function websqlBulkDocs(req, opts, api, db, Changes, callback) {
   var newEdits = opts.new_edits;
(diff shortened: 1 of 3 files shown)


lib/adapters/websql/bulkDocs.js
```diff
@@ -18,7 +18,7 @@ var ATTACH_AND_SEQ_STORE = websqlConstants.ATTACH_AND_SEQ_STORE;
 var select = websqlUtils.select;
 var stringifyDoc = websqlUtils.stringifyDoc;
 var compactRevs = websqlUtils.compactRevs;
-var unknownError = websqlUtils.unknownError;
```
nolanlawson: `unknownError` is the wrong name for the this thing; it is a generic handler for WebSQL errors.

lib/adapters/websql/utils.js
```diff
@@ -147,8 +147,9 @@ function compactRevs(revs, docId, tx) {
   });
 }
 
-function unknownError(callback) {
+function websqlError(callback) {
   return function (event) {
+    console.error('WebSQL threw an error', event);
```
nolanlawson: These kinds of things can be really tricky to debug, so I think it's best to just log an error so the user can see it immediately. We do a similar thing in idb as well.

daleharvey:
:+1: this makes sense, although even though this doesnt add tests would still prefer it be landed after being green, just to ensure no regression, cheers

nolanlawson:
Rebased to force Travis to run the tests.

daleharvey:
Sweet cheers - https://github.com/pouchdb/pouchdb/commit/a4282217e345068af5b07fe1004dbb8554321a87

PR #4357, opened 2015-09-22
(#4018) - Ensure proper handling of quote exceeded errors

daleharvey:


daleharvey:
So we arent going to be able to test this in automation without some serious work, but added some manual tests in there to help verify, and verifying this failed in both firefox and chrome previously and is passing now

nickcolley:
Does this swallow the error?

daleharvey:
Nope, this means we properly catch the error from indexeddb and report it in the error callback / rejected promise

nolanlawson:
Fantastic catch. So since either `txn.onabort` or `txn.onerror` means something has gone horribly wrong, should we catch both cases? I noticed you are replacing some `onerror`s with `onabort`s; make me nervous in case some browsers throw `txn.onerror` instead.

Either way, +1. Both `txn.onabort` and `txn.onerror` are callbacks that never occur in normal usage of PouchDB (hell, idbError never even gets called in our integration tests), so whatever we do in there is just icing on the cake to communicate errors to devs.

daleharvey:
`txn.onerror` does exist, but it would only ever be called as part of bubbling req.onerror and `txn.onabort` should also be called in this instance so I think we are good.

Hopefully https://github.com/pouchdb/pouchdb/pull/4357/files#diff-7e3f9be9833af207db13e42a5c6c2ec7R870 should also make a big difference, any transaction could have aborted and it would be silently hidden, cheers, will wait for green

daleharvey:
Sweet cheers, merged - https://github.com/pouchdb/pouchdb/commit/6c14a1a6df52ce5d06c89bd006c3d96c09183bfe

nolanlawson:
:+1: makes sense

PR #4360, opened 2015-09-22
fixes #4251 with test

### LINKED ISSUE #4251: 'paused' event not firing during sync

colinskow:
Fires "paused" and "active" changes on `db.sync`

### DIFF
--- a/lib/sync.js
+++ b/lib/sync.js
@@ -89,9 +89,15 @@ function Sync(src, target, opts, callback) {
     return function (event, func) {
       var isChange = event === 'change' &&
         (func === pullChange || func === pushChange);
+      var isDenied = event === 'denied' &&
+        (func === pullDenied || func === pushDenied);
+      var isPaused = event === 'paused' &&
+        (func === pullPaused || func === pushPaused);
+      var isActive = event === 'active' &&
+        (func === pullActive || func === pushActive);
       var isOtherEvent = event in listeners;
 
-      if (isChange || isOtherEvent) {
+      if (isChange || isDenied || isPaused || isActive || isOtherEvent) {
         if (!(event in removed)) {
           removed[event] = {};
         }
--- a/lib/sync.js
+++ b/lib/sync.js
@@ -30,6 +30,9 @@ function Sync(src, target, opts, callback) {
   this.push = replicate(src, target, opts);
   this.pull = replicate(target, src, opts);
 
+  this.pushPaused = true;
+  this.pullPaused = true;
+
   function pullChange(change) {
     self.emit('change', {
       direction: 'pull',
@@ -52,6 +55,30 @@ function Sync(src, target, opts, callback) {
     self.emit('denied', {
       direction: 'pull',
       doc: doc
+    });
+  }
+  function pushPaused() {
+    self.pushPaused = true;
+    if(self.pullPaused) {
+      self.emit('paused');
+    }
+  }
+  function pullPaused() {
+    self.pullPaused = true;
+    if(self.pushPaused) {
+      self.emit('paused');
+    }
+  }
+  function pushActive() {
+    self.pushPaused = false;
+    self.emit('active', {
+      direction: 'push'
+    });
+  }
+  function pullActive() {
+    self.pullPaused = false;
+    self.emit('active', {
+      direction: 'pull'
     });
   }
 
@@ -89,6 +116,12 @@ function Sync(src, target, opts, callback) {
     } else if (event === 'denied') {
       self.pull.on('denied', pullDenied);
       self.push.on('denied', pushDenied);
+    } else if (event === 'active') {
+      self.pull.on('active', pullActive);
+      self.push.on('active', pushActive);
+    } else if (event === 'paused') {
+      self.pull.on('paused', pullPaused);
+      self.push.on('paused', pushPaused);
     }
   });
 
@@ -96,6 +129,15 @@ function Sync(src, target, opts, callback) {
     if (event === 'change') {
       self.pull.removeListener('change', pullChange);
       self.push.removeListener('change', pushChange);
+    } else if (event === 'denied') {
+      self.pull.removeListener('denied', pullDenied);
+      self.push.removeListener('denied', pushDenied);
+    } else if (event === 'active') {
+      self.pull.removeListener('active', pullActive);
+      self.push.removeListener('active', pushActive);
+    } else if (event === 'paused') {
+      self.pull.removeListener('paused', pullPaused);
+      self.push.removeListener('paused', pushPaused);
     }
   });
 
--- a/tests/integration/test.sync_events.js
+++ b/tests/integration/test.sync_events.js
@@ -1,0 +1,60 @@
+'use strict';
+
+var adapters = [
+  ['local', 'http'],
+  ['http', 'http'],
+  ['http', 'local'],
+  ['local', 'local']
+];
+
+if ('saucelabs' in testUtils.params()) {
+  adapters = [['local', 'http'], ['http', 'local']];
+}
+
+
+adapters.forEach(function (adapters) {
+  var title = 'test.sync_events.js-' + adapters[0] + '-' + adapters[1];
+  describe('suite2 ' + title, function () {
+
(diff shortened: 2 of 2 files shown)


colinskow:
Some tests are failing on Travis, but 100% pass on my local machine.

daleharvey:
The main use case for the `active` / `paused` is to drawing ui to indicate what is happening with 2 main use cases. a UI that shows any activity and no activity, and a ui that should actively pushing, actively pulling and no activity. The current patch would require a user who wanted to show any activity vs no activity to manage the state of each replication to do so.

We could add a `status` to the events (and the sync object) that was `waiting`, `pushing`, `pulling` and `active` so to update the UI would look like

```
function updateUI(evt) { 
  if (evt.status === 'waiting') { 
    showInactive(); 
  } else { 
    showActive(); 
  }
}
sync.on('active', updateUI);
sync.on('paused', updateUI);
```
To be honest I think this is still a little unwieldy, but would like to hear what others think

foucdeg:
I personally am interested in both use cases, but maybe the priority is to fix those events not firing at all. Still not working with 4.0.3.

daleharvey:
@foucdeg If the events dont serve their purpose then its better they dont exist than to rush in broken ones, this isnt a large change so should be fairly safe for the next release, but would like to get them landed fulfilling their use case first time, if anyone wants to track events from the underlying replications then they can do so now with `sync.pull.on('active', fun)`

foucdeg:
@daleharvey You're right, and I didn't know about this workaround. I'll use it, thanks!

colinskow:
Paused only fires when both directions are paused at the same time. `active` events fire on both push and pull. So with the above PR you can do:
```js
sync.on('active', function(status){
  // push or pull
  console.log('Active', status.direction);
});
```

This is behavior is consistent with the replication events. So this PR will handle both use cases.

colinskow:
Here's a [tool](https://github.com/colinskow/ng-pouch-mirror/blob/master/src/ng-pouch-mirror.js) I made for Angular that tracks replication status and updates the UI. It worked with PouchDB 4.0.0, but has been broken since then.

This PR will restore the original behavior and enable what is already [documented](http://pouchdb.com/api.html#sync).

Status is very easy to construct from `active` and `paused` events, and clients may want to do it in different ways.

daleharvey:
@colinskow The case I was wondering about was going from active in both directions then one direction cancelling, a `paused` event firing in that case seems confusing, but it not firing makes it impossible to construct a UI that shows the direction of activity.

However I am thinking that is somewhat out of scope and if users what that they can track push / pull seperately, we can also look at adding status fields in a future change, so will take another look at this as is, cheers

colinskow:
@daleharvey what I care about is that:
1) There is a way to update the UI during sync based on an event emitter
2) It works as documented

If you want me to make other changes to the PR, let me know. It needs a test for memory leaks, but I am not entirely familiar with how to do that.

lib/sync.js
```diff
@@ -54,6 +57,30 @@ function Sync(src, target, opts, callback) {
...
+    }
+  }
+  function pullPaused() {
+    self.pullPaused = true;
+    if(self.pushPaused) {
+      self.emit('paused');
+    }
+  }
+  function pushActive() {
+    self.pushPaused = false;
```
nolanlawson: I had a question about what would happen if only one direction became paused, but this flag seems to address that. Nice.

lib/sync.js
```diff
@@ -54,6 +57,30 @@ function Sync(src, target, opts, callback) {
...
+    self.emit('active', {
+      direction: 'push'
+    });
```
nolanlawson: So this will emit two `active` events but one `paused` event every time network conditions change? Why not emit a single `active` event the same way you are doing for `paused`?

lib/sync.js
```diff
@@ -62,9 +89,15 @@ function Sync(src, target, opts, callback) {
     return function (event, func) {
       var isChange = event === 'change' &&
         (func === pullChange || func === pushChange);
+      var isDenied = event === 'denied' &&
+        (func === pullDenied || func === pushDenied);
```
nolanlawson: I don't believe this `denied` stuff is tested, although it's undoubtedly necessary. Would be nice to add a test to confirm this is the right behavior (you can run `COVERAGE=1 npm t` to verify that this code is being hit).

tests/integration/test.sync_events.js
```diff
@@ -0,0 +1,60 @@
+'use strict';
```
nolanlawson: This is a new test file, so it would need to be added to `tests/integration/index.html` in order to actually be tested in the browser. As is, it is only tested in node, because we do `mocha tests/integration/test*js`

tests/integration/test.sync_events.js
```diff
@@ -0,0 +1,60 @@
...
+      });
+
+      repl.on('active', function(evt) {
+        activeCount++;
+        console.log('active', activeCount);
+      });
+
+      repl.on('paused', function(evt) {
+        pausedCount++;
+        console.log('paused', pausedCount);
```
nolanlawson: please remove console logs before committing

tests/integration/test.sync_events.js
```diff
@@ -0,0 +1,60 @@
...
+        if(pausedCount > 0 && activeCount > 0) {
+          console.log('finished');
+          repl.cancel();
```
nolanlawson: I'm confused about why this would even call `paused` in the first place, since it's never going into an "offline" state. You can simulate an offline state by throwing errors in the db methods; see [test.retry.js](https://github.com/pouchdb/pouchdb/blob/f8c06ac76d321e84bdcab25f9cf02f5d648e06af/tests/integration/test.retry.js) for some examples of how we do this

daleharvey: You dont need to be offline to be paused, paused gets fired when we are up to date and waiting on future changes

nolanlawson:
@colinskow Most memory leaks should be caught more-or-less automatically thanks to https://travis-ci.org/nolanlawson/throw-max-listeners-error. You can also use `listeners('eventname').length` to count the number of listeners before and after the test and verify that the number didn't increase.

I agree with @daleharvey that this is potentially confusing. You are ensuring that `'paused'` will only fire once (i.e. after both directions go into a paused state), but you are not doing the same thing for `'active'`, so it emits twice for one but once for the other. We should probably just do one for both.

Also as noted in the comments above, the new test isn't actually being run in the browser. I also am not seeing how it is testing very much; I would much prefer to see something more like what we do in `test.retry.js`, testing the exact number of events.

Then again, this is definitely a step in the right direction and brings us closer to compliance with the stated behavior in the documentation. As you no doubt noticed, our replication events are kind of messy and we are in dire need of somebody going in there to add tests and clean things up, so I'm glad somebody decided to roll up their sleeves. :)

tests/integration/test.sync_events.js
nolanlawson: ah yeah, forgot about that

daleharvey:
So for the tests, I wrote https://github.com/pouchdb/pouchdb/blob/master/tests/integration/test.replication_events.js with this patch in mind, so will add some tests to this PR

And I agree on consistency, we should have the same semantics firing active events as paused ones, so I think the best thing is 1 event in each direction.

daleharvey:
Filed follow ups @ https://github.com/pouchdb/pouchdb/pull/4363

daleharvey:
Awesome thanks for this @colinskow, You have made a few contributions now so I would like to invite you to have commit access, this means you can push other peoples pull requests and review pull requests etc, thanks so much for your contributions :) 

https://github.com/pouchdb/pouchdb/commit/473c98755197463945459ea296d152235afa1087

scholtzm:
Is there a reason why `paused` event is missing direction while `active` event includes `direction` property?
When I'm using `PouchDB.sync`, it's impossible to tell which direction is actually paused.

PR #4363, opened 2015-09-25
Follow up 4360

daleharvey:


daleharvey:
@colinskow How about this? the tests are a little more comprehensive and I made the change so we only fire active events once we are paused

colinskow:
@daleharvey looks good. I think it is easier to test with only one active firing. I was having the problem that the event counters were inconsistent across adapters. Thanks for the tips @nolanlawson . I'm new to PRs ;-)

colinskow:
I agree with @nolanlawson that denied events should be tested, but don't know how to mock the 403 response. These will only fire with the HTTP adapter.

daleharvey:
There are tests in test.sync and test.replication for denied, new tests are always good but I think those ones dont need to block getting the events landed. When there are problems getting the events working across adapters that is a good thing to be specific about in the tests, we have one bug here where on http sync `paused` gets fired before `active` when you start a replication containing data, will definitely need looked at but one we can possibly live with when landing and following up

colinskow:
The sync `denied` events are just pass-throughs from replication, so if replication-denied is well-tested, then I don't think we need to go out of our way.

daleharvey:
Followed up and remove the extra objects

nolanlawson:
> The sync denied events are just pass-throughs from replication, so if replication-denied is well-tested, then I don't think we need to go out of our way.

Always nice to see the code coverage numbers go up, though. :wink:

nolanlawson:
LGTM :+1:

daleharvey:
Merged - https://github.com/pouchdb/pouchdb/commit/473c98755197463945459ea296d152235afa1087

Issue #4289, opened 2015-09-08
Minor enhancement to sync() to support to/from replication options (e.g. filters)


tohagan:
Would be nice if the sync() opts could include to/from specific options that are combined with the other "root" options.  Here's the code change to sync() 

      var optsPush = utils.extends({}, opts, opts.to);
      var optsPull = utils.extends({}, opts, opts.from);
  
      // deleted are not strictly needed 
      delete optsPull.to;  
      delete optsPull.from;
      delete optsPush.to;
      delete optsPush.from;

      this.push = replicate(src, target, optsPush);
      this.pull   = replicate(target, src, optsPull);

daleharvey:
Yup, right now you can do this by constructing seperate db object, but agree that we should add this, and probably should have done instead of pass all the options to both db's

nickcolley:
@tohagan if you have the time do you fancy doing a PR? Let us know :)

tohagan:
Yes sorry I should have done that :). 

FYI ... I've just watched [Nolan's videos on Pouch internals](https://www.youtube.com/watch?v=ven5f4wCE4k).

nickcolley:
I didn't mean any pressure just if you were interested we could help out! :+1:

tohagan:
I'm going to have a crack doing a PR for this one as I need it rather badly for an app.

Did you want test to be added to tests/integration/test.sync.js or somewhere else?
Not sure it will be an integration test.

daleharvey:
I think `tests/component` would be a good place for this, we just need to test that the correct options get sent to the constructor

tohagan:
Sure ... I wonder if that might be the case.  Thanks Dale.

tohagan:
Dale .. what are you using for spying on function calls?

I'm used to using Jasmine 2.x that has spy features but you're using Mocha for unit testing.

You could add [Sinon](https://github.com/mochajs/mocha/wiki/Spies).
But would need to add this lib to your package.json

nolanlawson:
We're not using spies, but I'm fine with adding Sinon as a devDependency if you really want it for the tests.

daleharvey:
Yup sorry I missed you in IRC but thats totally fine by me

tohagan:
[...]

PR #4365, opened 2015-09-26
(#4289) - Added push/pull replication options (e.g. filters) to sync().

tohagan:
Attempted to write a unit test using a SinonJS spy to watch the replication() calls in sync(), however this won't work as these lines in sync.js ... 

    var replication = require('./replicate');
    var replicate = replication.replicate;

.. bind the replicate() function before it can be intercepted by the unit test.

I have run the `sync()` code via unit tests and added console.log(pushOpts)` and `console.log(pullOpts)`  to confirm that the correct args are being sent to the `replicate()` method.

### DIFF
--- a/lib/sync.js
+++ b/lib/sync.js
@@ -27,8 +27,11 @@ function Sync(src, target, opts, callback) {
   var self = this;
   this.canceled = false;
 
-  this.push = replicate(src, target, opts);
-  this.pull = replicate(target, src, opts);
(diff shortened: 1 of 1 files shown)


daleharvey:
Awesome thanks, so yeh I may have sent you down the wrong whole with component / unit tests, it may just be easier to an integration test that exposes this, I can give it a shot if you fancy

daleharvey:
https://github.com/pouchdb/pouchdb/pull/4366 contains this PR + ones for tests + docs, so closing in favour

Issue #4346, opened 2015-09-17
Deprecate returnDocs in favor of return_docs


nolanlawson:
Ala https://github.com/pouchdb/pouchdb/pull/4338, I would like us to be consistent about our naming conventions for options. `camel_case` is what we use almost everywhere (because it fits the CouchDB conventions), but there are odd exceptions, the last of which (I believe) being `returnDocs` (see [changes()](http://pouchdb.com/api.html#changes)). 

I think we should support both `returnDocs` and `return_docs` but mark `returnDocs` for deprecation in the docs.

nolanlawson:
So basically this would involve making a similar change to https://github.com/pouchdb/pouchdb/pull/4338, except in this case we also need to add a deprecation notice to the docs, because `returnDocs` has been documented for awhile.

nolanlawson:
By the way the deprecation notice should go in [api.html](http://pouchdb.com/api.html), which can be modified [here](https://github.com/pouchdb/pouchdb/edit/master/docs/_includes/api/overview.html). Just find the part about `returnDocs` and add a note that it's now `return_docs`.

xelawafs:
Hey Nolan,

I guess you mean the returnDocs bit can be modified under the changes section on api.html which is [here] (https://github.com/pouchdb/pouchdb/blob/master/docs/_includes/api/changes.html). Given it a try and submitted a [commit](https://github.com/pouchdb/pouchdb/commit/b015b2c2c8956b5fa0be55e93fd51a7989d643ad)

christopherreay:
https://en.wikipedia.org/wiki/CamelCase ?????????????????????

nickcolley:
yo @christopherreay , I think Nolan just typoed when he said 'camel_case' he meant 'snake_case'.

CouchDB uses snake_case and since PouchDB is trying to be as compatible as possible here we're trying to make the options similar too.

Hope that makes sense :+1:

christopherreay:
it says "firsttimeusers" in one of the tags....

pity. I use a mix of camelAndSnake_forComplexNames, lol.

Great tool guys, thank you so much.

On 13 October 2015 at 15:54, Nick Colley <notifications@github.com> wrote:

> yo @christopherreay <https://github.com/christopherreay> , I think Nolan
> just typoed when he said 'camel_case' he meant 'snake_case'.
>
> CouchDB uses snake_case and since PouchDB is trying to be as compatible as
> possible here we're trying to make the options similar too.
>
[...]

PR #4367, opened 2015-09-26
Convert camelCase instances of returnDocs to snake_case per issue #4346

mwksl:
Changed all functions and variables referencing returnDocs to return_docs. This was my best interpretation of the issue. :koala:


daleharvey:
Hey thanks so much :) few small issues with this PR that need addressed.

We never modify those `*-postfixed.js` files, they are previous releases that have been built and we do them to test backwards compatibility / migrations so they need to be identical to what they were.

`lib/replicate/getDocs.js` doesnt need changed, we only want to change the api arguments, camelcase is still the convention for functions so `returnDocs` should still be used there.

And the last thing is that we still need to support `returnDocs` for backwards compatibility reasons, the easiest way to do that is to copy and paste the 'supports return_docs=false' test and change the new instance to `returnDocs`. Now to actually support both options we will need to do something like

    var return_docs = true;
    if ('return_docs' in opts) { 
      return_docs = opts.return_docs;
    } else if ('returnDocs' in opts) { 
      return_docs = opts.returnDocs;
    }

mwksl:
Thanks for your patience, @daleharvey I'm new to this, and maybe a bit too eager to contribute. I'll see if I can make the suggested changes!

daleharvey:
@mwksl Its no bother at all, our issues arent always the most explanatory and it can take a few back and forths to get on the same page, thats what the review process is for. Hopefully the above helped and please feel free to ask more questions if you have any

Issue #3955, opened 2015-06-11
Race condition in clone


nolanlawson:
From a Twitter comment: https://twitter.com/ermouth/status/608849583729467393

willholley:
It looks like there are a few places where this could happen but I think the simplest solution would be to have argsarray deep clone objects before passing them on.

daleharvey:
Added a test for this https://github.com/pouchdb/pouchdb/commit/ee770fefb7409b2955dc514121c2279f39d33632

daleharvey:
Closed by https://github.com/pouchdb/pouchdb/commit/44b3429b4fe10d1f9a76d758e96ce3f51dde835e

Issue #4273, opened 2015-09-04
We modify the users document when they .put


daleharvey:

daleharvey:
So this only fails when you try to save an object with a property `hasOwnProperty` because the clone will early return due to https://github.com/pouchdb/pouchdb/blob/master/lib/deps/clone.js#L9

daleharvey:
So this is probably super easy or super hard, I tried copying the current jquery code but that breaks down with a `constructor` property

daleharvey:
Ok will describe this in more detail. So users pass in objects to pouchdb to save, we do not want to modify those users objects with our metadata so we clone the object. Currently this works fine except when users give us weird objects that have 'constructor' or 'hasOwnProperty' user defined properties, you can see a failing test @ https://github.com/pouchdb/pouchdb/blob/4273/tests/integration/test.basics.js#L1020

Currently this fails because we copy and old version of jquery's clone in https://github.com/pouchdb/pouchdb/blob/master/lib/deps/clone.js, specifically we check for 'hasOwnProperty' property existing and if it is there, do not clone the object. If we try upgrading to the latest jquery clone, it also fails this time due to the `constructor` property, you can see the failure @ http://jsbin.com/bipedorimo/edit?js,console

Been trying a few different clone implementations right now, none seem to be able to handle this case

daleharvey:
So https://www.npmjs.com/package/clone will strip the object of all these properties (possibly the way we might want to go)

daleharvey:
Changing `clone.js` to 

```
  module.exports = function(obj) {
  if (typeof obj !== 'object') {
    return obj;
  }
  return Object.assign({}, obj);
};
```
Passes our test, but fails somewhere else in attachments, which will be fun

daleharvey:
ok, test that is failing is `issue 2803 should throw 412 part 3` because `issue 2803 should throw 412 part 1` modified the contents of `binAttrDoc`

daleharvey:
dumb mistake, Object.assign is shallow

daleharvey:
and JSON.stringify will not work because we use this function to clone various things that include functions. Also assuming we need clone for `_attachments` where we would like to save blobs directly in indexeddb

ermouth:
When I made custom PouchDB fork after known events, I had to embed two cloners to solve this problem. One works fine for Safari (I experirenced troubles with .constructor in it), and second – in other browsers.

This approach is far from being elegant, but it works, at least for my tasks.

[...]

PR #4368, opened 2015-09-26
(#3955) - Fix race condition in writes

daleharvey:


daleharvey:
Causing a permanent failure in `putAttachment and getAttachment with plaintext` in chrome

daleharvey:
Found it, should be good to go now

daleharvey:
Accidently pushed this to the same branch, but meh they are pretty related and can see the first one had a green run so can +1 that separately if there are issues.

https://github.com/pouchdb/pouchdb/pull/4368/files#diff-58235c2e06758d108ddf0403b20f58fcL41 was the main cause of the problems and happy to get rid of that

daleharvey:
Woops, reverted that last commit and left this only #3955 which is ready for review, #4273 still needs work

lib/deps/clone.js
```diff
@@ -38,6 +38,10 @@ module.exports = function clone(object) {
     return object.toISOString();
   }
 
+  if (object instanceof ArrayBuffer) {
+    return object.slice(0);
+  }
+
```
nolanlawson: If we're doing this fix for the browser, won't we also need it for Node? I.e. Buffers?

nolanlawson: Actually I have a better idea - we should really avoid the additional global `arraybuffer-slice` polyfill. Since we're always calling `slice(0)`, we can just do `return new Uint8Array(object)`, which is the same as `return object.slice(0)`

nolanlawson:
+1 after removing the slice shim as described. I suppose the reason it's working in Node is because of something internal to the leveldb adapter, but as long as there's a test, it LGTM.

lib/deps/clone.js
daleharvey: Yeh we have another around this code that I am working on in https://github.com/pouchdb/pouchdb/issues/4273

daleharvey:
Nice idea, will give that a shot

daleharvey:
`return new Uint8Array(object)` isnt working, I dont know all the weird details about array buffer etc seems like it must be returning just a vew on the same data

nolanlawson:
My bad, well in that case whatever the minimal shim is that doesn't require us to include a browser polyfill (pretty proud that we don't have any polyfills that modify globals outside PouchDB). I'll look at the polyfill code and try to provide a good one.

nolanlawson:
I went ahead and pushed directly to this branch. Based on my tests, this should do the trick.

daleharvey:
Hrm, I was wondering why you were wanting to avoid a polyfill, I like the approach because makes the intent clear and it is code that can just be straight deleted when we dont need to support older browser. However I do agree that its best to avoid changing globals in a permanent way. Cheers for rewriting, will watch for it going green

daleharvey:
Has to switch to `slice(0)` instead of `slice()`, safari doesnt like it despite the standard saying specifically it defaults to 0

nolanlawson:
yeah, the only thing that bothered me was 1) overwriting globals, and 2) including more shims than we needed. thanks for the `slice()` fix.

daleharvey:
Sounds great, cheers for the help - https://github.com/pouchdb/pouchdb/commit/44b3429b4fe10d1f9a76d758e96ce3f51dde835e

PR #4387, opened 2015-10-01
(#4273) - Ensure we dont modify objects in place

daleharvey:


lib/deps/clone.js
```diff
@@ -54,13 +39,19 @@ module.exports = function clone(object) {
...
-  if (!isPlainObject(object)) {
-    return object;
+  if (typeof Blob !== 'undefined' && object instanceof Blob) {
+    return object.slice(0, object.size, object.type);
+  }
+
+  if (typeof Buffer !== 'undefined' && object instanceof Buffer) {
+    var copy = new Buffer(object.length);
+    object.copy(copy);
+    return copy;
```
nolanlawson: ^ this will browserify all of `Buffer`. probably the best is to use a snippet of code with the `-browser` trick to exclude it. I'll submit a PR

nolanlawson:
It's amazing how difficult it is to clone stuff properly

nolanlawson:
I went ahead and pushed directly to this branch again. Checked the pouchdb.min.js size and it is still ~48KB after minification/gzip, so we're good.

daleharvey:
Yeh I should have marked this as WIP, I wanted to check if it was failing in any browsers, cancelled the previous run but it says everything except phantomjs is passing, wait to see what this one says about safari (safari 7 is passing locally)

nolanlawson:
`blob.slice` is undefined for PhantomJS >_<. We will probably need to call `createBlob()` and use `BlobBuilder` to build a new one.

daleharvey:
So yeh, safari and ie are fine but phantomjs is failing, I think we should cull phantomjs or if not then at the least, move it to allowed failures and whoever wants to support it can do so

nolanlawson:
That person will be me, and I'll do it before you need to add it to allowed failures

I do think that the day will come soon when we'll have to drop phantom support, but I don't think we're there yet. Sadly it seems pretty much unmaintained, but I know a lot of test tools are built on top of it.

nolanlawson:
k, looks like `blob.webkitSlice()` does the trick

daleharvey:
Heh thanks :)

daleharvey:
Also the previous version was showing up untested since node wasnt hitting that path, so thats nice

https://github.com/pouchdb/pouchdb/commit/338ccf97140ee8b8a8d25ca7dc0f52bdebf518b6

PR #4388, opened 2015-10-01
Install Erlang R18 for travis

garrensmith:
CouchDB master installs rebar as part of the install process. No longer
need to do this manually.

### DIFF
--- a/bin/run-couch-master-on-travis.sh
+++ b/bin/run-couch-master-on-travis.sh
@@ -4,14 +4,12 @@ set -x
 
 CWD=$(pwd)
 
-# Rebar isnt in apt
-git clone git://github.com/rebar/rebar.git ~/rebar
-cd ~/rebar
-./bootstrap
-export PATH=$(pwd):$PATH
+sudo apt-get remove erlang-base erlang-crypto erlang-base-hipe
+sudo apt-get -y install haproxy default-jdk libwxgtk3.0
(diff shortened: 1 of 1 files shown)


bin/run-couch-master-on-travis.sh
```diff
@@ -4,12 +4,6 @@ set -x
...
 CWD=$(pwd)
 
-# Rebar isnt in apt
-git clone git://github.com/rebar/rebar.git ~/rebar
-cd ~/rebar
-./bootstrap
-export PATH=$(pwd):$PATH
-
 # Sweet, build CouchDB
 cd ..
```
kxepal: This `cd  ..` should get also removed since it's about get out from rebar directory.

garrensmith: Yes good spot. I'll fix that.

daleharvey:
This is awesome, we are going to need to do some more work when we turn back on the docker infrastructure as they do not allow sudo, but lets get it working now and then fix it, couch master is still red, but is building and looks like a fairly trivial error

daleharvey:
Merged in https://github.com/pouchdb/pouchdb/commit/b6522c2272c6eba077dae49767a4e504eb420a23

garrensmith:
Cool thanks. @daleharvey once the Couchdb 2.0 docker files is done we can build off that for the Pouchdb work. I think the failing issue is unrelated to the Erlang install.

Issue #4395, opened 2015-10-03
Explain how to merge pull requests on this repository


Charlotteis:
It wasn't clear that I couldn't click the big green merge button, as the instruction not to is hidden in CONTRIBUTING.md. I would like to make this clearer, and give instructions on how to actually merge something.

nolanlawson:
eb1ff8631d03a91404209ddc4eaeed7d4ac285e2

PR #4396, opened 2015-10-03
(#4395) - Clarify the merge procedure

Charlotteis:
I still think we run the risk of new commiters not knowing they cant press the merge button, but I think we can at least clarify how to do things as I would have had no idea on how to merge a pull request.

I fleshed out instructions given to me by @daleharvey, so hopefully these are right.

👋

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -115,11 +115,22 @@ With great power comes great responsibility yada yada yada:
 
  * Code is peer reviewed, you should (almost) never push your own code.
  * Please don't accidentally force push to master.
- * Cherry Pick / Rebase commits, don't use the big green button.
+ * Cherry Pick / Rebase commits, **don't use the big green button**, see below for instructions on how to
+ merge a pull request.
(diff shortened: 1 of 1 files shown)


nolanlawson:
:+1:, you can add `[skip ci]` to the commit for docs changes, since there's no need to wait for a Travis build.

These new instructions are great; I've also been meaning to take a look at [apply-pr](https://github.com/petkaantonov/apply-pr) which is supposed to automate this process, but I wasn't sure how to do it with rewriting the commit (also it uses `git am` under the hood, which is a command I never use so I have no idea what to do with it :stuck_out_tongue: ).

nolanlawson:
OK, I just tested `apply-pr` for this very PR and it was as easy as:

```
apply-pr 4396
git push origin master
```

Think this is my new workflow; you can also change commit messages or squash using `--interactive`, so it seems to cover all the bases. I'll wait for everyone to try it out, though, and if everybody likes it we can update the docs again. :)

nolanlawson:
eb1ff8631d03a91404209ddc4eaeed7d4ac285e2

Issue #4276, opened 2015-09-05
401 unauthorized firing 'complete' instead of 'error' or 'denied'


colinskow:
I want my app to send the user back to the login screen when a live replication fails due to a 401 unauthorized response.

Based on your documentation: "In a live replication, only cancelling the replication should trigger this [complete] event." I would expect this to be handled by either 'error' or 'denied'.

With the 'complete' event, I have to iterate through the errors on both 'push' and 'pull' properties to tell if an unauthorized occurred... more complicated than it should be.

daleharvey:
Can you describe the situation a little more? I believe this may be due to the fact that replication can reject writes with a 401 and replication will carry on by design, people may have a `validate_doc_update` to do per document authorisation which will reject some writes but not others.

You mentioned the denied event, that is supposed to be fired in this situation are you not seeing it? a 'denied' event isnt a completion one and the complete event will still fire after denied does

colinskow:
As a security feature, when a user logs out I remove them from the _security doc of their database. This means that replication will fail with `401 unauthorized` on any other sessions that user has opened (intended behavior). This sets off the 'complete' event and nothing else.

In contrast, a validate_doc_update failure should result in `403 forbidden` and replication continues.

I am listening for the 'denied' event but it does not fire in the case of `401 unauthorized`. Although I am using PouchDB 4.0.0 because 4.0.1 hasn't been firing events at all.

colinskow:
So I did a bit of further testing. When PouchDB live replication receives an `unauthorized` or `forbidden` error from the process of attempting to find or write a checkpoint (`/_local`) then replication is terminated and the `complete` event is fired. This seems like rational behavior because CouchDB doesn't like the credentials. Although based on your documentation, it should be firing the `error` event instead.

When a normal `validate_doc_update` fails with `unauthorized` or `forbidden` then the `denied` event correctly fires and replication continues.

colinskow:
By the way, here is the source code that I'm working on that sets up 3-way sync and tracks events...
https://github.com/colinskow/ng-pouch-mirror

daleharvey:
@colinskow It seems like you are wanting to get your bugs fixed fairly quickly and fixing them yourself, so thanks :), and leaving these here in case you are looking to take a look at this (or anyone else)

[...]

PR #4398, opened 2015-10-03
(#4276) - Maintain cancelled state in replicator better

daleharvey:


daleharvey:
Ok, not particularly intuitive, but we were setting a local cancellation on `completeReplication` regardless of whether anything was cancelled, that cancellation wouldnt apply to retried replications but was the only thing the current replication was checking for and it basically confused everyone.

I had a ton more complicated patches for this but everything kept breaking somewhere, hoping this is the smallest working version

nolanlawson:
Nice, the new code is much easier to understand than the old code. That `state` object always scared me. :+1:

nolanlawson:
31cdcf8993924414d13f0d3df080d60220ba2b87

Issue #2982, opened 2014-11-12
Filtered replication on large database failure


mainakae:
Hello,

I've been testing serveral approaches to partially load a very big CouchDB database (> 100k documents) using PouchDB filtered replication.

The first problem came when I tryed to use the method "sync" and passing the filter and params as options. It failed. As the PouchDB database was empty, I assume that when it tries to replicate from the PouchDB to the CouchDB, it tries to use the filter, but it still doesn't exist... thus it fails.

I tried to overcome this situation making first a "replicate.from(couch...". Then I finally could do the filtered replication of small databases, but not the big one. After some fiddling and testing, I realized that the problem arises when the filter function takes a really long time to process all the documents in the DB: as it contains 100k documents, even though the filter only returns a bunch of documents, it takes ages to complete. 

The message after failing (catched "db.replicate.from(couch,...).on('error'...), would be:

ERROR: r {message: undefined, status: 405, statusText: "Method Not Allowed", result: Object, name: "unknown_error"…} 

I hope this is not some mistake by mi side making a missuse of the library.

Congratulations for the great work!
Pedro.

daleharvey:
Ok, this is fixed by adding a `heartbeat` option (we can default it to 10 seconds)

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/db6eadc3dee6109381319a829955fc8d3137c107

amkonieczka:
The heartbeat option is working great to keep the connection open indefinitely which is what I want. However, some people may not want the connection to remain open indefinitely. Either there should be no default and the developer must explicitly specify a value, or perhaps allow heartbeat: false to turn off the feature. Something to consider.

Yougz:
Hello,
As mentioned by amkonieczka, I really need a way to disable this option. Long polling requires a very specific server configuration, especially if you use a reverse-proxy in front of the couchDB server... I do not have access to the server's configuration, therefore the only option left to me is to disable the heartbeat.

daleharvey:
Ok, filed a new issue to disable it, will put a quick patch in for that now - https://github.com/pouchdb/pouchdb/issues/4490

PR #4404, opened 2015-10-04
(#2982) - Add heartbeat to keep alive long connections

daleharvey:


daleharvey:
This is extremely hard to test end to end, basically need to create a huge database (I copied npm), then make a changes request with a filter that takes a long time, so I think given the code is very simple a component test is ok here

nolanlawson:
I'm totally fine with the weak test; this is a neat feature and very useful. Would be nice to also add to `replicate()` and `sync()` as well.

+1 when green.

daleharvey:
Awesome cheers, pushed in https://github.com/pouchdb/pouchdb/commit/db6eadc3dee6109381319a829955fc8d3137c107

PR #4405, opened 2015-10-04
test fruitdown in a bunch of safaris

nolanlawson:
I cannot reproduce any of the failures locally, so let's
try a bunch of safaris/iphones and see if any pass.


nolanlawson:
ok, looks like if I break it up, it should pass in ios 8.4

Issue #4410, opened 2015-10-05
Upgrade Firefox in Travis


nolanlawson:
Opening a generic issue about this. I tried to solve it in https://github.com/pouchdb/pouchdb/pull/4403, but frustratingly, could not figure out at all how to get Travis to actually upgrade from Firefox 31.

nolanlawson:
Will fix in #4403

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/2457a2b2d91d080a18ec10720fd6531cd6db3b08

PR #4403, opened 2015-10-04
(#4402) - fix Firefox FileReader in a WW

nolanlawson:
This also re-enables the web worker tests for Firefox. Web worker
tests should only be running in Chrome and Firefox now.

PR #4411, opened 2015-10-05
(#4410) - attempt 1 to upgrade firefox

nolanlawson:
don't merge, this is just for testing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,8 @@ git:
 #     - /tmp/phantomjs
 
 addons:
-  firefox: latest
+  firefox:
+    latest
   apt:
     sources:
       - ubuntu-toolchain-r-test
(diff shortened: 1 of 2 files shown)


nolanlawson:
check the user agent to see if this is working

daleharvey:
If you expand the "Build system information" at the top you get some more info
```
Installed Firefox version

firefox 31.0esr
```

daleharvey:
So actually there is a bit below that, an `install_firefox` step which shows 41 being installed fine, the issue is that selenium isnt picking it up

nolanlawson:
Weird - it turned out to be the selenium version. 2.45 works; 2.46 doesn't. Specifying the firefox version in `firefox:selenium:41` also works.

nolanlawson:
OK, I figured out something that works. I'll just apply it to the #4403 PR.

Issue #4412, opened 2015-10-05
5.0.0


nolanlawson:
Let's do this.

nickcolley:
Looks great, would be great to follow this up later with a bulk get introduction.

Charlotteis:
@nolanlawson added two comments onto the blog post commit, but apart from that you have a +1 from me!

nolanlawson:
cbc2543ed28955f7c2bba8c1a4d5fbdf3febaff2

PR #4413, opened 2015-10-05
(#4412) - pouchdb 5.0.0 blog post

nolanlawson:
Not my favorite blog post ever. I wish I had had the time to run some numbers on just how much faster the new `_bulk_get` replication is, but I just didn't have time this weekend. So I wrote a pretty compulsory blog post that just lists out the deprecations/bugfixes/features.

If anybody would like to write a separate blog post about the new replication with some numbers about how much faster it is, that would be much appreciated. :)

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "pouchdb",
-  "version": "4.0.4-prerelease",
+  "version": "5.0.1-prerelease",
   "description": "PouchDB is a pocket-sized database.",
   "repo": "daleharvey/pouchdb",
   "keywords": [
--- a/component.json
+++ b/component.json
@@ -1,6 +1,6 @@
 {
   "name": "pouchdb",
-  "version": "4.0.4-prerelease",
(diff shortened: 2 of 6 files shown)


nolanlawson:
Also I would be pleased to delay 5.0.0 if we could get #4403 in.

daleharvey:
Agree with changing the wording of `should be easy`, `should look like this` seems good. I would also like to make it ultra clear that _bulk_get is an experimental api designed to improve replication performance. It may get changed and would really prefer people avoid using it directly.

I think its worth doing a follow up blog post purely about the _bulk_get performance by itself. This is awesome, :+1: from me.

Happy to wait for #4403, we can always land it with the tests disabled on travis for now and fix the fx / travis version issue after

nolanlawson:
OK, since the other stuff is merged, I'm going to modify the blog post as suggested above and do the release this morning. :)

nolanlawson:
cbc2543ed28955f7c2bba8c1a4d5fbdf3febaff2

Issue #4418, opened 2015-10-06
Remove Glob


daleharvey:
looks unused

daleharvey:
and done - https://github.com/pouchdb/pouchdb/commit/4e8f3cf576e5729805c9522d9d313d86710f4d67

PR #4419, opened 2015-10-06
(#4418) - Remove glob and do a few updates

daleharvey:


daleharvey:
Now that greenkeeper is going to keep us up to date with deps, My ocd wants me to make sure that our current set of deps are also up to date, picked a few that are hopefully harmless. I cant see where we use glob at all

daleharvey:
And that is them all, split the test runs up so if one fails it wouldnt kill the rest of them

nolanlawson:
:+1:

daleharvey:
Awesome, squashed and merged - https://github.com/pouchdb/pouchdb/commit/4e8f3cf576e5729805c9522d9d313d86710f4d67

PR #4423, opened 2015-10-06
(#4423) - fix timeouts in mapreduce tests

nolanlawson:
`worker-pouch`'s tests are failing, because these timeouts
are actually lower than the global default for mapreduce
(100000). (And `worker-pouch` takes a bit longer to run the tests.)

I suggest removing them and then upping the
one that is likely to take a long time (the "many persisted"
test).


daleharvey:
This looks great, cheers - https://github.com/pouchdb/pouchdb/commit/48717326d050be6c73f59e6fd7bc494334e0e462

Issue #4422, opened 2015-10-06
CouchDB Master is perma fail


daleharvey:
Looks like a trivial fix

nolanlawson:
#4430

PR #4430, opened 2015-10-07
(#4422) - Dont test for specific error messages

daleharvey:
Couch 2.0 changed its error message for these, we have had problems checking for specific error messages before so removed the check


nolanlawson:
:+1:, these always made me nervous anyway

nolanlawson:
b01c32c8cf0957a8319227d1e02370eb57057049 and 82188b477b43f2c50fdb19442bbf04db290a0107

PR #4434, opened 2015-10-07
(#4434) - add more plugins to the plugins page

nolanlawson:
I'm sure I'm still missing some. There are getting to be too many to keep on one page.


daleharvey:
:+1: , added skip ci since its docs only - https://github.com/pouchdb/pouchdb/commit/69787f059b5145924695cad2ab7cc6bc46436d48

Issue #4435, opened 2015-10-07
Explain normal 40x's in console.info when _bulk_get API is not supported


janraasch:
Since as of right now the _bulk_get API is not supported by most CouchDB services out there in the wild, it would be nice to have an explanation in the browser console for the http error a user receives when the _bulk_get API is not supported by the remote. Similar to #2835.

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/e164a33739be1f61b560266668950f71499eb1ef

PR #4436, opened 2015-10-07
(#4435) explain http error caused by check for _bulk_get

janraasch:
I kept `explain404` around in order to not break the public `utils` API.

### DIFF
--- a/lib/adapters/http/index.js
+++ b/lib/adapters/http/index.js
@@ -189,7 +189,9 @@ function HttpPouch(opts, callback) {
     setupPromise = ajaxPromise({}, checkExists).catch(function(err) {
       if (err && err.status && err.status === 404) {
         // Doesnt exist, create it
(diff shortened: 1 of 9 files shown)


daleharvey:
@janraasch Thanks so much

So first issue is this is breaking the linting, can see this issues @ https://travis-ci.org/pouchdb/pouchdb/jobs/84124307 and test its working with `npm run jshint`

The next is `utils` is only public facing by accident, it is not a supported part of our API so I would prefer to just switch everything over to explainError and drop explain404 completely.

Cheers, hope that is ok, if you have any questions then feel free to ask

janraasch:
@daleharvey thank you for your detailed response. I fixed the linting problems and removed `explain404`. Please have another look.

daleharvey:
This is perfect and all tests are passing thanks a lot :+1: - https://github.com/pouchdb/pouchdb/commit/e164a33739be1f61b560266668950f71499eb1ef

PR #4438, opened 2015-10-08
(#4071) - Ignore ie polyfill

daleharvey:


lib/deps/functionName.js
```diff
@@ -7,6 +7,8 @@ if ((function f() {}).name) {
     return fun.name;
   };
 } else {
+  // We dont run coverage in IE
+  /* istanbul ignore next */
```
nolanlawson: I believe you can do `istanbul ignore else` to ignore the whole else block

daleharvey:
Updated to use `ignore else`, also skipped ci since there is no point in running the full suite for a comment change, will verify it does skip in coverage though

nolanlawson:
:+1:

daleharvey:
Sweet cheers - https://github.com/pouchdb/pouchdb/commit/137376d5cb59ff29bafe522b7263fe4e659192e5

PR #4441, opened 2015-10-08
(#4071) - Omit some browser only ajax options from coverage

daleharvey:


daleharvey:
Hrm, this https://github.com/pouchdb/pouchdb/commit/3dec222e83a8dc4b71c6dda7aed5a633e6f2367c is managing to break test.events.js with ```(node) warning: possible EventEmitter memory leak detected. 11 destroyed listeners added. Use emitter.setMaxListeners() to increase limit.
Trace```

The test is fine so we have a bug somewhere else it looks like, @nolanlawson you did the max event listeners stuff, any ideas here? it looks tricky

daleharvey:
Ok figured it out, we were both inheriting the event emitter and manually assigning it, should be good now

daleharvey:
Nope not sorted, will skip this part for now

daleharvey:
Filed a new bug and worked around the issue

daleharvey:
So for https://github.com/pouchdb/pouchdb/commit/a25265713055d42e9271643d5062ce65ce07d6b9, its pretty much impossible to test that the cancellation has happened precisely between different stages (ie after bulkDocs) so ignoring those clauses for now

daleharvey:
So I did these as individual commits so if there is an issue with one the rest can still go in. The lines left oncovered are mostly quite tricky, for things in adapters/leveldb/index.js where we are handling errors we do not encouter in the test I think its likely best we skip them, for the ones in replication it would be good to get a test up that simulates the failure, but will probably have to be decided on case by case, so going to get this lot landed then look at the rest

daleharvey:
Closing since this is hard to manage seperate PR's - https://github.com/pouchdb/pouchdb/pull/4445

PR #4445, opened 2015-10-09
(#4071) - Improve code coverage 

daleharvey:


daleharvey:
I split this into its own PR since its a little less sure whether we want to go this way, so the vast majority of these are handling error conditions that we never get while running the tests and shouldnt be expecting to get at all, instrumenting the errors to exercise the code paths is going to be hard to somewhat impossible and if we remove the callbacks then it makes life harder for when something does go disastrously wrong

daleharvey:
Ok this is like a pick and mix, they are all seperate commits so if there are issues with some the rest can still land, all together this gets us to `Lines        : 99.37% ( 4420/4448 )`

nolanlawson:
Awesome work, really happy to see the coverage go up, whether it's through dead code removal or reasonable `istanbul ignore`s.

lib/adapters/leveldb/index.js
```diff
@@ -145,6 +147,7 @@ function LevelPouch(opts, callback) {
   }
 
   if (typeof leveldown.destroy !== 'function') {
+    /* istanbul ignore next */
     leveldown.destroy = function (name, cb) { cb(); };
   }
```
nolanlawson: ^ FYI this was an old fix for old MemDOWN and probably isn't required anymore.

lib/adapter.js
```diff
@@ -279,6 +281,7 @@ AbstractPouchDB.prototype.removeAttachment =
       callback(errors.error(errors.REV_CONFLICT));
       return;
     }
+    /* istanbul ignore if */
     if (!obj._attachments) {
       return callback();
     }
```
nolanlawson: ^ this is weird and should probably just be removed.

Issue #3110, opened 2014-12-04
Is there a way to force-install leveldown in Windows?


nolanlawson:
In Node, PouchDB has a dependency on leveldown, which depends on a non-JS build process and is therefore brittle. It seems to especially break a lot on Windows.

This is a shame, because there are many valid uses for PouchDB in Node that don't involve leveldown (http, other *down modules, pouchdb-replication-stream).

So question: is there a way to force-install pouchdb even if leveldown is not compiling? If so, let's add it to the FAQs. If not, maybe we should publish a `pouchdb-sans-leveldown` fork?

/cc @jacargentina

PR #4448, opened 2015-10-10
Add friendly error message for common leveldown error case

cdaringe:
- first off. holy tests!  wow.  so many tests!

# problem statement
`leveldown not available` could be more user friendly when user believes that it should be available

# context
let's get real.  sometimes we have to change node environments. it's terrible, it really is. it's also wonderful, & we all get it.  the detection of leveldown was intended to help alleviate #3110.  however, it's unintended consequence is that it swallows other errors that happen during `require('leveldown')`.

# solution
Detect a common case were leveldown is available, but incompatible.  The intent of the PR is not to cover all cases... just this little guy!  However, we could instead listen for the specific `not found` error:

```bash
cdieringer@Snapper-osx:~/node/pouchdb$ node
> require('abcd')
Error: Cannot find module 'abcd'
```

# test discussion
it's a tricky one to write tests for, as it requires different prebuilt leveldown versions.  not sure if folks think it's a worthwhile thing to test.  regardless, i installed pouch with node 0.12.7, ran the tests (pass), switched to iojs 2.5.0, ran the tests, and confirmed that the helpful error was present:

1_tests_pass
2_switch_to_iojs
3_observe_new_error


lib/adapters/leveldb/index.js
```diff
@@ -10,7 +10,16 @@ function requireLeveldown() {
...
-  } catch (e) {}
+  } catch (err) {
+      if (err && err.message && err.message.match('Module version mismatch')) {
+          throw new Error([
+            err.message,
+            'This generally implies that leveldown was built with a different',
+            'version of node than that which is running now.  You may try',
+            'fully removing and reinstalling PouchDB or leveldown to resolve.'
+          ].join(' '));
+      }
```
nolanlawson: I *love* this new error message and think it's much more helpful, but if we `throw` within a `catch`, then that removes the whole value of the `try/catch`. The point here was that, if you were on a system that didn't support leveldown, you could use some other *down (sqldown, memdown, etc.).

I'd suggest:

```js
if (typeof console !== 'undefined' && typeof console.warn === 'function') {
  console.warn(errorMessage);
}
```

nolanlawson:
Also great explanation of the problem and how to test it. :)

lib/adapters/leveldb/index.js
cdaringe: Thanks Nolan!

> but if we throw within a catch, then that removes the whole value of the try/catch

my take was that that this was a valid case to throw, as I assumed that the "Cannot find module" error was the _actual error that this block was intended to catch_.  if this were the case, in retrospect, it may have sniffed more specifically for that error.  however, i'm betting that it was an oversight [when it was implemented](https://github.com/pouchdb/pouchdb/commit/da09b3e6fcf3d4cdf04880ce9433d8e47600b36f).

I'm totally down with the `console.warn` suggestion, but i'm now thinking that the actual ideal solution would be to:

1. test if `opts.db` is passed, then on falsy,
1. try to import leveldown
1. sniff for the module not found error
1. if module is not found, we should throw the `"leveldown not available "` error per the status-quo, otherwise,
1. throw the actual nodejs produced error, unsquashed

right now, importing leveldown in the try catch doesn't offer much value as the next executing control block just throws the "not available" error (in probably ~99% of use cases).  let me know your thoughts!

nolanlawson: Yeah it's true; that `try/catch` is pretty generic and will throw on just about any error. That said, I think it's less dangerous to have a general `try/catch` in this case, because I don't believe npm specifies exactly what the name of the error you get when a module fails to load (correct me if I'm wrong).

Also keep in mind that we make a distinction here between synchronous throwing and asynchronous callbacks; in general we prefer asynchronous callbacks because synchronous throwing might get swallowed in a promise chain.

For a concrete suggestion, let's just reword [the current error](https://github.com/pouchdb/pouchdb/blob/79785ee234b77af36137969b73f5d80dd7f0b0e4/lib/adapters/leveldb/index.js#L140-L147) to be your more helpful version. The `console.warn` stuff is unnecessary if users are already seeing the error.

cdaringe: [Check this out](https://github.com/nodejs/node/blob/master/lib/module.js#L338).  Module not found is part of nodejs core, and can be probably be relied on.  It's [been there since 2011](https://github.com/nodejs/node/commit/3f987cd925dd342b4695afd9c221bc19e76bdf7b), and I wouldn't forsee it changing.  Also, ["Module version mismatch"](https://github.com/nodejs/node/blob/master/src/node.cc#L2128) is also part of node core.  It doesn't have a standard ERROR code, which is a bummer, but is also stable over the past 5 years.

Although grateful for the support :), I think putting my Error where the current error block is not a good idea.  This is because there are a variety of _other_ errors that it could happen while doing a `require('leveldown')`, and I'd hate to misdirect users by stating their problem is possibly related to a build problem if the root cause is something else entirely.  For instance, perhaps there's an `EACCES` error because someone has fubared their node_modules permissions, etc.  When I learned node, I did that all of the time :)

With this in mind, how would you feel about the follow up commit?  Test's re-ran, using async CB only now.

```js
// example.js
var P = require('./'); // ran in pouchdb root
var p = new P('dbname');
p.get('abcd').catch(function(err) {
    console.log(err.message);
   // => mismatch error if leveldown built on different node version OR,
   // 'Error: EACCES, permission denied ... ' if theres a perm error, for example, OR,
   // "the 'leveldown' package is not available. install it, or, specify another storage backend using the 'db' option" if leveldown missing
   // i tested each case
});
setInterval(function() {}, 1000);
```

finally, [optional](https://github.com/tony-o/node-optional/blob/master/optional.js) offers these features, but it doesn't have a great API and doesnt have test cases.  this whole thought process may inspire me to write a better version of that

cdaringe: again today, i had a [leveldown error swallowed](https://gist.github.com/cdaringe/e25beed841c756c4e7dd).

daleharvey:
So this seems pretty reasonable, having it rethrow means its backwards compatible with anything that is expecting it to throw and also deals with my main concern that when we try to provide helpful error message we have often made things more confusing. If this could be fixed downstream it would be ideal however.

I am :+1: although really confused at the travis failure, can you squash these into a single commit and do `npm run jshint` on them?, I dont think we need to test this case but we do need an istanbul ignore for that function clause since we wont hit that in the tests (you can run the coverage tests with `COVERAGE=1 npm test`

Thanks so much

cdaringe:
rad+.  thanks guys.  istanbul ignore added.  additionally, jshint was barking about `atob` being redefined, which I suppressed.  rebase was being a little bully as  f70fc9b was a merge commit with a conflict resolution embedded.  i wasn't comfortable rebasing that one :/

daleharvey:
Ok so lets give this a shot, if it causes problems then its easy to revert. I dont know git well enough to know what went wrong with your rebase, but the `atob` didnt need supressed, I took that out and landed in https://github.com/pouchdb/pouchdb/commit/c22bc9d4e4e8c92ac021f027adbc518c9e263aaa

Thanks so much :)

Issue #4449, opened 2015-10-10
PouchDB custom build


nolanlawson:
I think our build has gotten a bit too big (~47KB). There's nothing we can reasonably remove without removing serious functionality, though.

So I would like to create a custom build. You should be able to build a PouchDB that is missing:

* idb
* websql
* migrations
* http
* sync/replicate
* mapreduce

I think this is technically possible both with a browserifable script and a web UI builder. So the web UI builder would both allow you to download a file and display a browserifiable file that would probably look like this:

```js
var PouchDB = require('pouchdb/custom/pouchdb');
require('pouchdb/custom/idb');
// no websql
require('pouchdb/custom/mapreduce');
// no http
// no migrations
// no replicate/sync
module.exports = PouchDB;
```

I'll take a stab at it this weekend.

nickcolley:
We could think about https://github.com/pouchdb/pouchdb/issues/4199 alongside this.

daleharvey:
So I think a custom build is a nice thing to offer to users, however I do not think it is a solution for having to large a distribution. I would still like to make sure we have a "this is pouchdb" distribution that has sensible defaults for most users while being as small as possible.

nolanlawson:
I wish we could make it smaller, but aside from removing stuff for edge cases (vuvuzela, websql, blob shims), the only big win I see would be from using rollup, which means switching to ES6 modules: https://github.com/nolanlawson/rollup-comparison.

I consider this more of a stopgap while our build is too big; also a good transitional step. E.g. when we drop WebSQL support, that will just mean that the "this is pouchdb" distribution will lose the "require websql" line from its custom build.

nolanlawson:
Just investigated, and unfortunately Rollup and tree-shaking don't help us out much here: https://github.com/nolanlawson/test-tree-shaking

I think the best solution is to basically ask users to write their own `index.js` and exclude what they don't want. It's the simplest to support.

nickcolley:
:+1: for having our normal builds plus the instructions to build their one index.js

kidwm:
@nolanlawson is it possible to release an ES Modules build like Redux did ?
BTW, I tried to bundle my pouchdb app but always failed with these setting:
```
plugins: [
		babel({
			"presets": ["es2015-rollup"]
		}),
		nodeResolve({
			jsnext: true,
			main: true
		}),
		commonjs({
			include: 'node_modules/**'
		})
	]
```

nolanlawson:
[...]

PR #4452, opened 2015-10-10
(#4449) - add custom builds

nolanlawson:
Adds the ability to customize the PouchDB build, using either Browserify/Webpack or by just downloading from the website.

Adds some complexity to our build process (see changes to `CONTRIBUTING.md`) at the expense of an arguably very very useful feature for our users.

There is a new `npm run build-custom` that builds up all possible combinations of builds, calculates the min/gzip size, and moves them to the `static/js` directory. These builds are not published to npm/bower; I think they're too big for that (~12MB).

Not implemented yet: separating out migrations, separating out `lie`.

animated


lib/index.js
```diff
@@ -1,23 +1,10 @@
 "use strict";
```
nolanlawson: Note that PouchDB itself is now a custom build. It's important to eat your own dog food. :smiley:

nolanlawson:
Also `require('pouchdb/custom'leveldb')` is in kind of a weird position, because I can't imagine why any npm user would want to use it, but it's in there for our own custom build. My current solution is to just not document it, because I don't think it's necessary.

nolanlawson:
I should also do a blog post explaining what you might use this for:

* Using only in-memory, WorkerPouch, or some other adapter? Skip both the IndexedDB and the WebSQL adapters.
* Want to use PouchDB just as an interface to CouchDB? Uncheck everything except "HTTP."
* Want to skip map/reduce? Go ahead; you'll save 4KB.

nickcolley:
This is really cool, can you give me some time to review (hopefully later today)

Thinking it'd make more sense to have this in the downloads page?

nolanlawson:
@nickcolley Yeah, I wasn't really sure where to put it. Also would definitely like your help with the design/CSS/etc. :)

nolanlawson:
Rebased, also fixed it so that `ajax` is not included if you don't choose "HTTP". Makes that build a bit smaller.

nickcolley:
Started making some changes to make this a downloads page.

One thing I want to add is the ability to set the build via params in the URL to make it easier to share a custom build.

nolanlawson:
> One thing I want to add is the ability to set the build via params in the URL to make it easier to share a custom build.

I worry that's potentially dangerous, because then people might start using `pouchdb.com` as a CDN, which is a problem we ran into with `pouchb.com/pouchdb-nightly.js`. I kind of intended to make the url non-human-readable (using the base64 hash), so that people wouldn't feel encouraged to use them in that way. :/

nickcolley:
I meant the URL to the page, not the physical build itself. Agree with what you're saying.

e.g. `pouchdb.com/downloads?custom=http` or whatever. So we can link to the download page.

nolanlawson:
Ah okay, so it would pre-check the boxes depending on the query params? Sounds like a nifty feature.

nolanlawson:
I think I would like to put custom builds on the back burner for the time being. From hanging out with @rich-harris and @reconbot and talking to them about Rollup/ES6

PR #4453, opened 2015-10-11
(#4453) - simpler blobSupport check

nolanlawson:
I know @daleharvey has wanted something like this for awhile. I've been holding onto it mostly due to sentimentality about the amount of time I had to spend to write the fix :stuck_out_tongue:.

According to [caniuse stats](http://caniuse.com/usage-table), Chrome 37-42 is at 1.36% of all browser users, which is a lot, but not enough to justify the expensive ajax-based check here. I also don't like the dependency of `idb` on `ajax` given the new custom builds, where local-only builds should not have to include `ajax`.

The `userAgent` check I'm proposing is honestly only slightly less hacky than the present solution, since we are using one bug to detect the presence of another. Let's simplify the code and do a straight-up `userAgent` check.

Note this will also get rid of the confusing "PouchDB is only detecting blob support..." message, as well as the fetching of the Blob URL (which also confused people IMO).

### DIFF
--- a/lib/adapters/idb/blobSupport.js
+++ b/lib/adapters/idb/blobSupport.js
@@ -1,29 +1,25 @@
 'use strict';
 
-var utils = require('../../utils');
 var createBlob = require('../../deps/binary/blob');
-
+var Promise = require('../../deps/promise');
 var idbConstants = require('./constants');
 var DETECT_BLOB_SUPPORT_STORE = idbConstants.DETECT_BLOB_SUPPORT_STORE;
 
 //
-// Detect blob support. Chrome didn't support it until version 38.
-// In version 37 they had a broken version where PNGs (and possibly
-// other binary types) aren't stored correctly, because when you fetch
-// them, the content type is always null.
+// Blobs are not supported in all versions of IndexedDB, notably
+// Chrome <37 and Android <5. In those versions, storing a blob will throw.
 //
-// Furthermore, they have some outstanding bugs where blobs occasionally
-// are read by FileReader as null, or by ajax as 404s.
+// Various other blob bugs exist in Chrome v37-42 (inclusive).
+// Detecting them is expensive and confusing to users, and Chrome 37-42
+// is at very low usage worldwide, so we do a hacky userAgent check instead.
 //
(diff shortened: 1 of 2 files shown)


lib/adapters/idb/blobSupport.js
```diff
@@ -35,38 +31,14 @@ function checkBlobSupport(txn, idb) {
...
-            }
-          }
-          URL.revokeObjectURL(url);
-        });
-      };
+      if (window.chrome) {
+        var matched = navigator.userAgent.match(/Chrome\/(\d+)\./);
+        var matchedVersion = matched && parseInt(matched[1], 10);
+        if (matchedVersion < 43) {
+          return resolve(falase);
```
nickcolley: lil false typo

nolanlawson: yeah, just caught that :)

nickcolley:
+1 for less console spam!

nolanlawson:
LOL, the Travis build for Android 5.1 is stuck on Chrome 39 for some reason, so it is actually repro'ing this bug and revealing that I did the fix wrong. :stuck_out_tongue:

nolanlawson:
OK, apparently `window.chrome` is undefined in Android; that's why it was failing. Fixed it and verified in an Android emulator; should be good now.

nolanlawson:
I also manually tested to confirm that modern Chrome and Firefox are still using native Blobs.

nolanlawson:
let it be known that this got a full green run in travis: https://travis-ci.org/pouchdb/pouchdb/builds/84776163

now I'm going to fix the merge conflicts

daleharvey:
Yup was watching, +:100: getting rid of that warning is going to be super nice

nolanlawson:
OK, I'll just fix and merge then. the conflicts are simple

nolanlawson:
Merged as a1dcfcb73dad6350e64999e41e6fb9cb3e3777a0, also added an Edge check because apparently Edge spoofs Chrome 42. We should start testing Edge at some point, but for now I just tested it manually.

PR #4454, opened 2015-10-11
(#4454) - remove more stuff from PouchDB.utils

nolanlawson:
This removes nearly all the unnecessary stuff from
`utils.js`. The remaining stuff is either dangerous to
remove because of widespread (mis-)use (myself included here),
or because it's in the tests somewhere. Ultimately I hope
to start browserifying the tests themselves, so we don't
need to hook stuff onto this global `utils` object.

### DIFF
--- a/lib/adapter.js
+++ b/lib/adapter.js
@@ -1,6 +1,11 @@
 "use strict";
 
 var utils = require('./utils');
+var pick = require('./deps/pick');
+var collections = require('pouchdb-collections');
+var inherits = require('inherits');
+var getArguments = require('argsarray');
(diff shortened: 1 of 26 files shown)


nolanlawson:
Bah, min+gz size increased from 48404 to 48457. I was hoping it would go down.

daleharvey:
:+1: Even with the slight increase this is far cleaner, I have been meaning to do it myself at some point

nolanlawson:
Sweet, 4007ee58d6a21a1fbd14a12a8810f9133c31d1bc

PR #4455, opened 2015-10-11
100% Coverage

daleharvey:
Some of this really should have coverage, the bulkGet stuff for example, however bulk_get is going to be actively worked on. I think if we land this, we have 100% coverage and we can maintain it easier at 100% and remove istanbul clauses as we work on those areas of code 

(its hard to maintain < 100% because patches that remove tested code will make coverage worse despite making it better tested)

### DIFF
--- a/lib/adapters/http/index.js
+++ b/lib/adapters/http/index.js
@@ -322,7 +322,9 @@ function HttpPouch(opts, callback) {
     if (typeof supportsBulkGet !== 'boolean') {
       // check if this database supports _bulk_get
       doBulkGet(function (err, res) {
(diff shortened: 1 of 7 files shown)


nolanlawson:
Nice, let's do it. I agree that just requiring 100% code coverage is way easier to manage.

nickcolley:
This isn't really that transparent but I think the point about regressions and 100% is sound. :+1:

daleharvey:
Awesome, will start working on removing them now :)

https://github.com/pouchdb/pouchdb/commit/79785ee234b77af36137969b73f5d80dd7f0b0e4

Issue #4450, opened 2015-10-10
putting a document with attachments fails to pass authentication headers


tyler-johnson:
The following should fail, assuming that the database `mydb` has at least one member in the security document.

```js
var db = new PouchDB("http://localhost:5984/mydb", { 
  auth: { username: "admin", password: "12345" }
});

db.put({
  "_id": "doc",
  "title": "Legendary Hearts",
  "_attachments": {
    "att.txt": {
      "content_type": "text/plain",
      "data": new Blob(['Is there life on Mars?'], {type: 'text/plain'})
    }
  }
});
```

It seems the issue is the default request header fields are being replaced by the multipart request headers, instead of being merged.

https://github.com/pouchdb/pouchdb/blob/master/lib/adapters/http/index.js#L667
https://github.com/pouchdb/pouchdb/blob/master/lib/adapters/http/index.js#L145

daleharvey:
Yup, confirmed we a test 

  it('4450 Test headers are sent correctly on put', function() {
    var db = new PouchDB('http://127.0.0.1:' + PORT);
    var opts = {auth: {username: 'foo', password: 'bar'}};
    return db.put({
      _id: 'doc',
      _attachments: {
        'att.txt': {
          content_type: 'text/plain',
          data: new Buffer(['Is there life on Mars?'], {type: 'text/plain'})
        }
      }
    }).then(function() {
      should.equal(headers.authorization, 'Basic Zm9vOmJhcg==');
    });
  });

daleharvey:
hrm, as far as I can tell our close is shallow

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/6a10dff0734f29910eb296d4b3b777b9f1232357

PR #4458, opened 2015-10-11
(#4450) - Use js-extend for deep merging

daleharvey:


daleharvey:
So the `js-extend` module is fairly minimal, it increases the total build slightly but seems to compress to the same .min size

```385848 12 Oct 15:22 pouchdb.js
151075 12 Oct 15:22 pouchdb.min.js

386520 12 Oct 15:25 pouchdb.js
151075 12 Oct 15:22 pouchdb.min.js```

nolanlawson:
Do we still need a custom clone if we are using someone else's extend? Also I noticed you didn't delete `deps/extend.js`? Otherwise this seems reasonable to me.

nolanlawson:
Actually wait, I remember now, our clone does some fancy stuff with `Date`s to fix those inconsistencies. So this makes sense.

daleharvey:
Yup, although may be an idea to get our clone / extend based on the same code, will take a look, cheers, https://github.com/pouchdb/pouchdb/commit/6a10dff0734f29910eb296d4b3b777b9f1232357

PR #4462, opened 2015-10-13
Experiment with ES6 modules and Rollup

nolanlawson:
Not a real pull request, just a demo to start a conversation with @Rich-Harris.

I managed to convert PouchDB to ES6 modules; got the tests passing and everything. My strategy is to build one giant CJS file with all external npm modules as `--external` dependencies. Then I just browserify that huge file like normal.

This works great for the non-minified build:

```
380K	/tmp/pouchdb-browserify.js
364K	/tmp/pouchdb-rollup.js
```

But oddly is actually worse after `uglifyjs -mc`:

```
148K	/tmp/pouchdb-browserify.min.js
160K	/tmp/pouchdb-rollup.min.js
```

Gzip sizes are 48356 vs 49070, so again browserify is winning, by ~700 bytes.

Is there something I'm doing wrong in the build process? Or is it a problem with the minifier, that it somehow can't do more aggressive optimizations if everything is in the same scope? (I do notice that top-level variable names are not getting mangled for some reason.) You can check out [./bin/build-rollup.sh](https://github.com/pouchdb/pouchdb/blob/9d29b1b4f80b6ad752a8df74619d0e840a5a89fc/bin/build-rollup.sh) for how I'm doing the build.


nolanlawson:
build canceled to avoid Travis spam

nolanlawson:
Also [here are the built files](https://www.dropbox.com/sh/w6b1uroad9yt9a4/AABXi2TquNM2T2xdE-4iDa38a?dl=0) if anyone wants to inspect them.

lib/adapters/idb/constants.js
```diff
@@ -1,26 +1,35 @@
...
+var META_STORE = 'meta-store';
 // Where we store local documents
-exports.LOCAL_STORE = 'local-store';
+var LOCAL_STORE = 'local-store';
 // Where we detect blob support
-exports.DETECT_BLOB_SUPPORT_STORE = 'detect-blob-support';
\ No newline at end of file
+var DETECT_BLOB_SUPPORT_STORE = 'detect-blob-support';
+
+export default {
```
marten-de-vries: Isn't it possible to do this add variable definition?

```js
export var DETECT_BLOB_SUPPORT_STORE = 'detect-blob-support'; //etc?
```
I seem to recall there is something like that from some babel experiments (although I used ``const`` instead of ``var`` then). Saves some manual code.

marten-de-vries:
Very interesting!

@reconbot @nolanlawson mentioned ``uglifyjs -mc`` in his post, which is what PouchDB uses everywhere.

Rich-Harris:
@reconbot has eagle eyes – in utils.js, you're exporting a default object with several properties (such as `Promise` and `clone` and `extend`, and every time you to `var Promise = utils.Promise` you're creating a new variable.

The ES6-y way would be do have named exports from utils.js:

```js
// this...
export { ajax, parseUri, uuid, Promise ....}

// NOT this
export default {
  ajax: ajax,
  parseUri: parseUri,
  ...
};
```

And then you'd do `import { clone, Promise } from './utils'` on the importing side. That way, you're always referring to the same thing, plus you don't have any `utils.unmangleablePropertyName` references in your code. (You could also do `import * as utils from './utils' and then do `utils.Promise` – the result would be the same. But I prefer the aesthetics of named imports.)

There are similar situations in base64.js, constants.js, parseDoc.js, errors.js and a few others (found by searching `lib` for `export default {`). In some cases they're presumably supposed to be configurable objects, so you'd need to keep those ones as they are.

That should shrink things down a little bit further. The non-mangling is more of a mystery – your guess that uglify maybe gives up with large scopes is interesting, I wonder if that's right?

Rich-Harris:
By the way I couldn't figure out how to generate those `tmp` files – what are the steps?

lib/adapters/idb/constants.js
nolanlawson: Yeah I am very new to ES6 modules, and I was in a hurry, so the format I used was mostly determined by a search-and-replace.

Also I had to be careful to only use ES5 features *outside* the modules, because I found that using `babel` to transpile from ES6 to ES5 added a huuuge amount of overhead, even if I was only using one or two features in one or two places.

nolanlawson:
@reconbot @Rich-Harris Yeah that is definitely a bad pattern that I'm using, but what surprises me so much is that browserify+uglify consumes the same thing (more or less), and ends up with much smaller code in the end.

As for `import { clone, Promise }`, I'm actually explicitly avoiding *any* ES6 syntax (outside of import/export) because I discovered that running `babel` actually adds a significant number of bytes, even if I'm only using a few features, and uglify will throw a fit if it's not consuming ES5.

To build PouchDB normally, check out the master branch and do

```
npm i && npm run build-main-js
```

Which will create a `dist/pouchdb.js` file. To minify, run `npm install -g uglify-js` and run `uglifyjs -mc < dist/pouchdb.js`.

To build with this Rollup-based branch, check out the branch and then run 

```
npm i && ./bin/build-rollup.sh
```

Which will also create a `dist/pouchdb.js` file.

nolanlawson:
whoops, forgot to `npm install --save-dev rollup`

nolanlawson:
OK, I might have been mistaken about what I said about Babel adding bytes, because I just re-tested:

```
babel < dist/pouchdb.js | uglifyjs -mc | gzip -c | wc -c
49194
cat dist/pouchdb.js | uglifyjs -mc | gzip -c | wc -c
49192
```

So it adds 2 bytes. :stuck_out_tongue:

nolanlawson:
OK, this definitely seems to be an issue with Uglify. If I run the code through the Google Closure Compiler, it ends up smaller.

John David Dalton has recommended that running `uglify -> closure compiler -> uglify` results in the smallest code, so I tested that here. Results:

Before gzipping:

```
150854	/tmp/pouchdb-browserify-ug.min.js
146490	/tmp/pouchdb-browserify-ug-cc.min.js
145970	/tmp/pouchdb-browserify-ug-cc-ug.min.js
161820	/tmp/pouchdb-rollup-ug.min.js
138884	/tmp/pouchdb-rollup-ug-cc.min.js
138382	/tmp/pouchdb-rollup-ug-cc-ug.min.js
```

After gzipping:

```
48497 /tmp/pouchdb-browserify-ug.min.js
48773 /tmp/pouchdb-browserify-ug-cc.min.js
47367 /tmp/pouchdb-browserify-ug-cc-ug.min.js
49217 /tmp/pouchdb-rollup-ug.min.js
46153 /tmp/pouchdb-rollup-ug-cc.min.js
45089 /tmp/pouchdb-rollup-ug-cc-ug.min.js
```

Rich-Harris:
Babel doesn't need to get involved – `import { ... }` and `export { ... }` aren't examples of destructuring and object shorthand, they're just the import/export syntax. It's the difference between [this](http://bit.ly/1K6urI5) and [this](http://bit.ly/1Oz7weF).

Looks like maybe Closure Compiler is succeeding where Uglify failed – could that be related to the mangling issue?

nolanlawson:
It's gotta be the mangling issue. When I look through the uglified code, I see lots and lots of un-mangled variable names, but not if I use the browserify version. I tried uglify with verbose mode, but it's not giving any insights. Weird.

Rich-Harris:
Have opened https://github.com/mishoo/UglifyJS2/issues/830 to see if it sheds any light

nolanlawson:
Gonna close for now; I think 2-3KB is not enough to justify a big architectural change plus needing the closure compiler, but this is definitely interesting and I will be keeping my eye on this space.

As an aside, I found it surprisingly cumbersome to translate ES5 to ES6. `5to6` did barely anything; I mostly had to hack together some find-and-replace scripts in order to get everything converted. Maybe I was missing some tool that everyone already knows about.

Also I notice Rollup doesn't support `"browser"` switching in `package.json` ala webpack/browserify; may be a nice feature if you think it's in Rollup's scope @Rich-Harris.

Rich-Harris:
> Also I notice Rollup doesn't support `"browser"` switching in `package.json` ala webpack/browserify

Interesting, I hadn't considered that – largely because I use Rollup for bundling Node projects as well (results in faster startup time and smaller installations, plus no surprises when dependencies change in a breaking way despite semver, which happens). Will look into it though.

Updating projects to use ES6 *is* cumbersome, I've found myself writing [fairly substantial quantities of code](https://github.com/rollup/three-jsnext/tree/master/scripts/generate) in the past to do so. I don't think there's a tool that is going to help automate it – at least, if there is, no-one told me! Setting things up in such a way that you can do it piecemeal is probably the sanest route in many cases.

Rich-Harris:
> I wanted to just add them to externals but it didn't seem to work based upon path name

Yeah, it currently only attempts to resolve modules as external if they have non-relative sources. You'd have to add a `resolveId` option that returned `false` if it resolved to those modules, I think.

Using `new Function ()` instead of `eval`, or `var eval2 = eval`, might both solve the issue (since in either case you cause the eval'd code to run in the global scope), though I haven't tried it. Using named imports/exports ought to crunch things down a bit further still, but probably not by a huge amount.

nolanlawson:
Thanks @reconbot ; that makes sense to externalize `eval()`! Although I think ultimately 2KB is not enough for us to justify the switch over to ES6 just yet.

This was very interesting experiment, though, in that this is the first I've seen that ES6 transpilation actually *reduces* the bundle size rather than increasing it. Great motivator to get off of ES5. Bravo to Rollup!

PR #4466, opened 2015-10-13
(#4346) Document and alias return_docs (returnDocs)

bdchauvette:
Adds `options.return_docs` to the Changes API, and adds a deprecation
notice for `options.returnDocs` in the docs.

### DIFF
--- a/docs/_includes/api/changes.html
+++ b/docs/_includes/api/changes.html
@@ -33,7 +33,7 @@ All options default to `false` unless otherwise specified.
 
 **Advanced Options:**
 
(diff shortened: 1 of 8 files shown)


daleharvey:
This looks great thanks, :+1: the only thing I think we may want to change is the way the documentation is posted, maybe 

 * `options.return_docs` (previously `returnDocs`): Is available ...

daleharvey:
since it was a very minor change, I amended the commit and pushed it, hope you dont mind and thanks so much :) - https://github.com/pouchdb/pouchdb/commit/609cfc2ed5b4f1fb7887c68739343f4c021c47c3

Issue #4467, opened 2015-10-13
Dont create a database just to delete it


daleharvey:

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/188395194c213c26538b6db26fe9957056ab0b71

PR #4468, opened 2015-10-13
(#4467) - Dont create a database just to delete it

daleharvey:
Was something I noticed when doing the stateless constructor, may speed up our tests a fair bit, is cleaner and removes an untested code path :)


nolanlawson:
Nice catch, +1 when green

daleharvey:
Cheers, green - https://github.com/pouchdb/pouchdb/commit/188395194c213c26538b6db26fe9957056ab0b71

PR #4470, opened 2015-10-14
(#4470) - remove more utils usage

nolanlawson:
Also break up `base64` into a browser version and node version.
This PR reduces the min+gz size from 48583 to 48436.


nolanlawson:
small mistake in atob/btoa, should be fixed now

daleharvey:
:+1: this is great, cheers

Issue #4474, opened 2015-10-15
_bulk_get sometimes requests an empty docs array


nolanlawson:
This is a nice optimization we could sneak in; if the `docs` array is empty, there is no point in sending the `_bulk_get` request:

<img width="670" alt="screenshot 2015-10-15 15 35 46" src="https://cloud.githubusercontent.com/assets/283842/10525397/87f72c02-7352-11e5-8561-c48a2c0ce4d1.png">

Might even be a good first commit.

daleharvey:
fixed in https://github.com/pouchdb/pouchdb/commit/8132c76b0a79b75a1ca7f19d1b85943869f6e47d

PR #4475, opened 2015-10-16
(#4474) - avoid empty bulkGet()

nolanlawson:
Kinda embarrassed we missed this one, but it's fixed now!
Avoids an empty request when there's no point in doing
a `bulkGet()`.

### DIFF
--- a/lib/deps/upsert.js
+++ b/lib/deps/upsert.js
@@ -7,10 +7,6 @@ var Promise = require('./promise');
 // the doc, or false if it doesn't need to do an update after all
 var upsert = module.exports = function upsert(db, docId, diffFun) {
   return new Promise(function (fulfill, reject) {
-    if (typeof docId !== 'string') {
-      return reject(new Error('doc id is required'));
-    }
-
     db.get(docId, function (err, doc) {
       if (err) {
         /* istanbul ignore next */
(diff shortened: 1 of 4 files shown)


daleharvey:
:+1: however coverage regressed so that needs fixed, I dont think this case is urgent for a test, up to you

daleharvey:
I r=me'd a one like istanbul comment so we should be back at 100%

nolanlawson:
Fixed the coverage, should be back to 100 lines now. Turned out we were looking for `docs.length` in other places as well.

lib/deps/upsert.js
```diff
@@ -7,10 +7,6 @@ var Promise = require('./promise');
 // the doc, or false if it doesn't need to do an update after all
 var upsert = module.exports = function upsert(db, docId, diffFun) {
   return new Promise(function (fulfill, reject) {
-    if (typeof docId !== 'string') {
-      return reject(new Error('doc id is required'));
-    }
-
```
nolanlawson: I have no idea what this has to do with this PR, but it makes sense to me to remove it because it's only used internally, so I'm sure we're not passing in any non-string _ids anywhere.

daleharvey:
Heh we have an intermittent in our coverage, will take a look at that

daleharvey:
Awesome, https://github.com/pouchdb/pouchdb/commit/8132c76b0a79b75a1ca7f19d1b85943869f6e47d

PR #4476, opened 2015-10-16
fix jspm - main pointing to dist

stephenlautier:
Fixed for JSPM users to point directly to the dist/ without affecting NPM usage


daleharvey:
Awesome, this just need a sanity check so cancelled the full browser run, merged in https://github.com/pouchdb/pouchdb/commit/9263a3cb6f0296fd21dc35b10ec9e5fcdff3f0b7

Thanks a lot :)

stephenlautier:
Welcome :)

PR #4477, opened 2015-10-16
Remove vuvuzela and implement revs_limit

daleharvey:


daleharvey:
This lowers the revs limit to a 100 which should very comfortably be able to be traversed, the main side effect of this is our rev_tree will be smaller so real life use cases (lots of updates to the same doc) should be significantly faster, the other side effect is that revs_limit is the number of updates you can do while disconnected before introducing a conflict, so if the users updates a document 100+ times then syncs with an existing db, then it becomes a conflict. So its possible this change will introduce more conflicts but given our main use case I think its fairly negligible and I think the positives outweight the negatives. Also implemented revs_limit in case a user wants to keep long revision trees

daleharvey:
Another side effect is a chunk off the build

```
-rw-r--r--   1 daleharvey  staff  385849 16 Oct 15:10 pouchdb.js
-rw-r--r--   1 daleharvey  staff  150658 16 Oct 15:10 pouchdb.min.js

-rw-r--r--   1 daleharvey  staff   377K 16 Oct 15:10 pouchdb.js
-rw-r--r--   1 daleharvey  staff   147K 16 Oct 15:10 pouchdb.min.js
```
vs 

```
-rw-r--r--   1 daleharvey  staff  380498 16 Oct 15:08 pouchdb.js
-rw-r--r--   1 daleharvey  staff  148728 16 Oct 15:08 pouchdb.min.js

-rw-r--r--   1 daleharvey  staff   372K 16 Oct 15:12 pouchdb.js
-rw-r--r--   1 daleharvey  staff   145K 16 Oct 15:12 pouchdb.min.js
```

We can also remove the stringifying of data before it gos to indexeddb, as they have to marshall the data between processes etc it may not improve tons but is fairly certainly going to be faster than doing it in our own code

nolanlawson:
So a few things:

1. Removing vuvu only saves 615 bytes after gzip
2. Reducing revs_limit from 1000 to 100 is a breaking change, will require a 6.0
3. This doesn't really solve the problem, because if anybody sets revs_limit to ~500 then they will run into exactly the same error and will file exactly the same bug on us
4. Making lots of revisions to the same doc is an unfortunate (mis-)use of PouchDB, but it's one that I've seen a lot from users

So in general I am -1 on this PR; I don't think the benefits either to code reduction or to speed are worth the potential bugs and the major version bump. (The biggest speed hit we currently take from vuvu is from the try/catch deoptimization in Chrome/V8, which is going away with Turbofan anyway.) If we really want to remove vuvu, what I'd like to see is a migration to a different metadata model that doesn't require so much recursion, which would probably have benefits during stringification and traversal anyway.

I'd prefer if we broke this PR into 1) implementing revs_limit and 2) removing vuvu/reducing revs_limit to 100. I'd be +1 on the former but I would push back very much on the latter.

BTW as for stringification in IndexedDB, I have it from @inexorabletash at least that cloning in Chrome is [slower than stringification/parsing](https://code.google.com/p/chromium/issues/detail?id=536620#c11), so we may actually already be doing the smartest thing in terms of performance.

tests/integration/test.bulk_docs.js
```diff
@@ -958,5 +954,44 @@ adapters.forEach(function (adapter) {
...
 
+    it('2839 implement revs_limit', function (done) {
+
+      // We only implement revs_limit locally
+      if (adapter === 'http') {
+        return done();
+      }
+
+      var LIMIT = 50;
+      var db = new PouchDB(dbs.name, {revs_limit: LIMIT});
```
nolanlawson: Nice, I like having it as a constructor option. It needs to be passed to the server in `http/index.js` too, though, doesn't it? Also needs docs.

daleharvey:
> Removing vuvu only saves 615 bytes after gzip

There arent going to be a lot of big wins but removing 1.3% of the code we ship is pretty significant, we do that a bunch of times and its going to get a lot smaller.

> This doesn't really solve the problem, because if anybody sets revs_limit to ~500 then they will run into exactly the same error and will file exactly the same bug on us

`revs_limit` is a level of internal implementation that almost no users care about, to be honest I only implemented on the chance it would make it easier to get this merged. Not letting it be a user configurable option or giving it a max limit would be fine by me. probably preferable.

> Making lots of revisions to the same doc is an unfortunate (mis-)use of PouchDB, but it's one that I've seen a lot from users

Updating documents is not a misuse of PouchDB in any way and I would really like to avoid going down that path. CouchDB in the past has replied to a lot of "we dont handle this use case well so you are misusing it" and it leads to ignoring the important use cases.

Its not a big enough improvement that I will stand my ground on this PR, (at least until we bump up the performance suite) but I would really like to make the point that the best way we are going to reduce the code size, reduce the amount of bugs and increase the performance is by doing less things. Instead of looking at things to add (es6 transpilers, lru memory cache) we should be looking for things to remove and how to remove them (`http.put`, taskqueue, vuvuzela, debug etc)

nolanlawson:
The Rollup ES6 transpiler actually removed 2-3KB of code, but yeah, it complicated the source code, which is why I backed off.

I agree that we should be relentless about removing dependencies, but I don't want to remove stuff at the expense of introducing (fairly) serious bugs. Also it's good to keep perspective: Ember is 177KB, Angular is 51KB, React is 53KB; we are still smaller than all of them. Also I really like the direction with the custom build, because it gives us a path e.g. to removing websql: 1) websql becomes custom build option, 2) websql becomes plugin 3) websql becomes removed from core.

For reducing code size, we could also add `debug` to the custom build options or fork `spark-md5` to remove the stuff we're not using (e.g. only use the ArrayBuffer stuff, not the string stuff). I did measure this recently, so to recap, after min+gz our deps take:

- spark-md5: 2.2KB
- debug: 1.4KB
- vuvuzela: 0.6KB

> CouchDB in the past has replied to a lot of "we dont handle this use case well so you are misusing it" and it leads to ignoring the important use cases.

Strongly agree with you here; I misspoke a bit, was talking more about the case where a user writes e.g. a document editor where it adds a revision for every keystroke. Sloppy and unperformant use of PouchDB, but I've seen a lot of stuff like that.

How about this: I will pledge to migrate our metadata model to a more efficient, non-recursive format. I've been meaning to go in there anyway and use some kind of class-based model so that we can benefit from things like V8's hidden classes. (I might also just avoid it, though, because again it adds bytes, so will need testing.) Then we can remove vuvu but still not explode if revs_limit is >1000. Sound good?

daleharvey:
I dont expect anybody to configure `revs_limit` (or would be happy removing the option entirely) So I wouldnt expect any bugs to come out of this change, but yup a nicer way to represent metadata would be an ideal solution, I expect once we do performance testing and see the cost of reading/writing/traversing a 1000 deep tree we will probably want to be reducing our 1000 default, but can hold off this commit for now.

PR #4479, opened 2015-10-16
(#xxx) - try to repro empty results query bug

nolanlawson:
Bug reported on IRC. I tried to reproduce it but did not succeed; the test passes for me on Mac OS 10.10 using Node 4.0.

### DIFF
--- a/tests/mapreduce/test.mapreduce.js
+++ b/tests/mapreduce/test.mapreduce.js
@@ -94,6 +94,37 @@ function tests(suiteName, dbName, dbType, viewType) {
         });
       });
     });
+
+    it("Test collation empty results bug", function () {
+      return new PouchDB(dbName).then(function (db) {
+        return createView(db, {
+          map: function (doc) {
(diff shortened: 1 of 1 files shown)


sajinshrestha:
@nolanlawson I was using CouchDB 1.6.0 and upgrading to 1.6.1 fixed the issue.

nolanlawson:
Looks like this issue was resolved by the user by upgrading from CouchDB 1.6.0 to 1.6.1, but I still think this is a good test to add.

nolanlawson:
@sajinshrestha OK, thanks anyway for the report! :)

nolanlawson:
Interesting - this is a straight-up bug in CouchDB. Travis says it's using 1.6.1, and yet it's failing. Locally I can't reproduce - it passes on my Homebrew version of CouchDB (1.6.1_3).

Since there doesn't seem to be anything we can do to use an updated version of CouchDB, I'm just going to close this PR.

Issue #2513, opened 2014-07-22
Add option to set withCredentials = false for ajax requests


nolanlawson:
In `ajax-browser.js` we're setting `withCredentials` to be true every time. We need some way to programmatically set this, although TBH I have no idea how.

Without this feature, we won't be able to set up the npm-browser to replicate from skimdb.npmjs.com, since credentials will be set to false on that server.

xMartin:
This is not limited to CouchDB configuration. For instance I have a setup with some auth proxy. How about renaming the issue to something like 'make `withCredentials` configurable' or something like that.

anaran:
I arrived at this solution to make auth work for all my use cases (Firefox OS app, Desktop packaged app, served as hosted app via python -m SimpleHTTPServer

Using a customized xhr field I was able to debug this.

I leave the if statement in for historical background:

```
    var myXHR = function () {
      var request;
      if (false && /* false && */window.location.protocol == "app:") {
        request = new XMLHttpRequest({ mozSystem: true, mozAnon: true });
      }
      else {
        request = new XMLHttpRequest({ mozSystem: false, mozAnon: false });
      }
      return request;
    }
    var opts = {
      ajax: {
        xhr: myXHR,
        timeout: 30000
      }
    };
    var remoteOptionsDB = new PouchDB(destination + optionsDB._db_name, opts);
    var remoteDB = new PouchDB(destination + db._db_name, opts);
```

nolanlawson:
woot, efe5dd8aa645e349a46a8e293703310c8da06ea2

xMartin:
Nice, thanks!

PR #4483, opened 2015-10-17
(#2513) - Add option to configure withCredentials

daleharvey:


daleharvey:
So testing this would require a fairly large chunk of work as the option doesnt make any other effect than to control how the browser handles requests, we already have tests to ensure the options get passed through to the requests so I skipped it.

I know we lean towards `snake_case`, but that is for consistency with CouchDB, in this case I think the only sensible option is to be consistent with the property it is setting `xhr.withCredentials`

nolanlawson:
Hm, how about just `options.credentials`? I just get a bit nitpicky about this, would prefer to keep consistent case. :P

daleharvey:
I am usually fine with being nitpicky about consistency, but this is directly setting the `withCredentials` parameter, calling it anything else is asking for users to be confused, its what they will be searching for.

nolanlawson:
Fair enough, we do a similar thing for the SQLite plugin due to its superReminiscentOfJavaOptionNames, so it makes sense here too. +1

nolanlawson:
efe5dd8aa645e349a46a8e293703310c8da06ea2

PR #4484, opened 2015-10-18
(pouchdb/express-pouchdb#247) - fix http url with port 80

nolanlawson:
This is one of those issues that's so fundamental, it makes me
wonder if anybody's actually using our library.

If you create a PouchDB over http without a port specified,
it won't work. This PR fixes that, which I've confirmed manually.

Kinda hard to write a test for this one, because we'd have to
depend on a CouchDB exposed via port 80. I'd prefer not to add a test;
Istanbul 100% coverage won't break, though, because this switch is
a one-liner.

### DIFF
--- a/lib/adapters/http/index.js
+++ b/lib/adapters/http/index.js
@@ -113,8 +113,9 @@ function genUrl(opts, path) {
 
   // If the host already has a path, then we need to have a path delimiter
   // Otherwise, the path delimiter is the empty string
-  return opts.protocol + '://' + opts.host + ':' + opts.port + '/' +
-         opts.path + pathDel + path;
+  return opts.protocol + '://' + opts.host +
+         (opts.port ? (':' + opts.port) : '') +
+         '/' + opts.path + pathDel + path;
 }
 
 function paramsToStr(params) {


daleharvey:
Where are you seeing this fail? in node / chrome and firefox I see `http://url:/db` falling back and working fine?

```
~/src/pouchdb (master)$ node
> var PouchDB = require('./lib');
undefined
> var db = new PouchDB('https://skimdb.npmjs.com/registry');
undefined
> db.info().then(console.log.bind(console)).catch(console.error.bind(console));
Promise { <pending> }
> https://skimdb.npmjs.com:/
{ db_name: 'registry',
  doc_count: 193878,
  doc_del_count: 2613,
  update_seq: 383314,
  purge_seq: 0,
  compact_running: false,
  disk_size: 848629883,
  data_size: 695420757,
  instance_start_time: '1443816481491613',
  disk_format_version: 6,
  committed_update_seq: 383314,
  host: 'https://skimdb.npmjs.com:/registry/',
  auto_compaction: false,
  adapter: 'http' }
```

:+1: on the change though, the old parseUri library used to default to port 80 which is why there is no check there

nolanlawson:
I dunno, maybe it's a DNS thing? In any case, urls like `http://foobar.com:/path` just don't work at all for me on my machine (Mac Yosemite, Node 4.0).

nolanlawson:
0cbb9f1d4890f4850c6a831bd4fa82d38b71cb56

Issue #4451, opened 2015-10-10
Invalid view fails in pouch, returns in couch


daleharvey:

daleharvey:
Heh realised I put no context here, there is a failing test @ https://github.com/pouchdb/pouchdb/blob/master/tests/integration/test.changes.js for this

daleharvey:
Fixed by https://github.com/pouchdb/pouchdb/commit/6baa2b1cf6b392b8e185ca6e8f3e1619932f6aaf

PR #4492, opened 2015-10-20
(#4451) - Handle invalid view name consistently

daleharvey:


nolanlawson:
So invalid views don't emit in `on('error')`; it just gets ignored? If that's what Couch does, then sure, +1.

daleharvey:
Yup consistent with couch and couch master, this isnt an area of the api I am worried about getting super right right now, just consistent, cheers - https://github.com/pouchdb/pouchdb/commit/6baa2b1cf6b392b8e185ca6e8f3e1619932f6aaf

Issue #2344, opened 2014-06-04
put() has inconsistent arguments


nolanlawson:
As reported [by Lukas Rieder](https://groups.google.com/forum/#!topic/pouchdb/pQUq-nOBPhQ):

>If I may, I've come across another potential problem: It seems to me that `db.put()` does not handle its arguments in a safe way. Let's consider the following:
>
>var doc = { _id: 'some-id', foo: 'bar' };
>var rev = 'some-rev'  // this was obtained by a get() somewhere
>db.put(doc, doc._id, rev);
>
>Apparently, PouchDB sees the `_id` in the document and hence does not look for it in the argument. It then considers `some-id` the revision, which fails, of course. Leaving out the second parameter works, but seems very fragile.

daleharvey:
Closing as dupe https://github.com/pouchdb/pouchdb/issues/5154

PR #4499, opened 2015-10-21
(#1526) - Remove put implementation and fix local doc handling

daleharvey:
I know and agree we want to keep multipart, but its main benefit is during replication and right now it is not being used during replication. I dont want implementing it in bulkDocs to block improvements to the rest of the code.

Our special handling of these API's in the http adapter is causing bugs to be hidden, with this one we were previously sending back a `{ok: true}` for local documents written via bulkDocs and away to file another one in which we handle deletions in bulkDocs inconsistently (as well as consistency issues like https://github.com/pouchdb/pouchdb/issues/2344)

### DIFF
--- a/lib/adapters/http/index.js
+++ b/lib/adapters/http/index.js
@@ -32,9 +32,7 @@ var btoa = base64.btoa;
 var atob = base64.atob;
 var errors = require('../../deps/errors');
 var log = require('debug')('pouchdb:http');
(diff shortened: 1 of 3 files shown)


nolanlawson:
OK, I'll just open an issue to add multipart back to bulkDocs. Nice PR, agree the code simplification and bugfixes are worth it.

nolanlawson:
ef0944f72429915c7248689cc6950895de78982e

Issue #4504, opened 2015-10-22
Re-introduce multipart uploads to the http adapter


nolanlawson:
It was removed in #4499 (ef0944f). Ideally we should be doing this for all put/post/bulkDocs, if there are any attachments inside the request.

I also have hope that someday someone will get around to https://github.com/pouchdb/pouchdb/issues/3913

nolanlawson:
WTF, looks like `_bulk_docs` only accepts `application/json` anyway. `PUT /doc` is the only one that accepts multipart.

Ugh, looks like we might as well just do `application/json` everywhere. :(

nolanlawson:
I still agree with your commit in ef0944f @daleharvey, if only because _bulk_docs is the primary way that data is going to be sent to the server (via replication) for most of our users. Just disappointed that multipart isn't doing us any favors either for downloads (due to gzip) or uploads (due to _bulk_docs not supporting it).

daleharvey:
Yup that sucks, cheers for looking though, Closed - https://github.com/pouchdb/pouchdb/commit/b6a6aa3d2ecd665fba9eeeb46de8717cc76de923

willholley:
I've seen a few use cases recently where multi-part uploads would be beneficial - primarily where the CouchDB JSON document size limit is hit due to inline attachments. Is there any reason not to re-introduce this feature, perhaps only for documents where attachment size is greater than a (configurable) value?

nolanlawson:
Does CouchDB support multipart/mixed for `_bulk_docs` yet? And yeah, I agree that uploading base64 data is less-than-ideal.

kxepal:
@nolanlawson suddenly, no.

PR #4507, opened 2015-10-24
(#4504) - remove dead multipart.js file

nolanlawson:
No point in keeping this around if it's not doing anything.


daleharvey:
Yeh was hoping implementing in _bulk_docs wouldnt be too tricky, but looks like it is, :+1:  - https://github.com/pouchdb/pouchdb/commit/b6a6aa3d2ecd665fba9eeeb46de8717cc76de923

Issue #4519, opened 2015-10-28
Document bulkGet()


nolanlawson:
The bulkGet() API is currently undocumented, but since it's technically user-facing, it ought to be.

nolanlawson:
Some background on bulkGet():

* https://github.com/pouchdb/pouchdb/pull/3783
* https://github.com/pouchdb/pouchdb/pull/4169
* https://github.com/apache/couchdb-chttpd/pull/33
* https://issues.apache.org/jira/browse/COUCHDB-2310

willholley:
I'm not entirely convinced - we don't document all the APIs used only for replication (e.g. revs_diff) and I don't think it's very useful as a standalone endpoint.

kxepal:
PouchDB users cannot use bulkGet from their apps for good deeds? It useful not only for replication use case.

willholley:
Appreciate that it might be useful but it's not something I've heard people ask for outside of replication and we don't document the other replication-related endpoints (maybe we should?).

If we do add documentation, we should provide some guidance around when to use _bulk_get vs  other bulk fetch methods (e.g. POST to primary/secondary index with multiple keys).

daleharvey:
We document revsDiff - http://pouchdb.com/api.html#revisions_diff, I believe we document everything considered public facing (aside from this of course), I do think this needs documented

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/c9143d64dbdd381bf969bd5066d0cee6a4c2fcf7

PR #4532, opened 2015-11-02
(#4519) add documentation for bulk_get endpoint

willholley:
Add documentation for the bulk_get endpoint, introduced in PouchDB 5.0.

The PouchDB implementation currently differs from the CouchDB implementation slightly (no atts_since support and rev is a required parameter for each id/rev pair). The documentation reflects the current PouchDB behaviour.


daleharvey:
Apologies for taking so long, this looks great, thanks - https://github.com/pouchdb/pouchdb/commit/c9143d64dbdd381bf969bd5066d0cee6a4c2fcf7

PR #4557, opened 2015-11-09
Firefox in "private browsing" mode always fails to open idb.

rubensayshi:
'Common' way to detect "private browsing" (that I found) is by checking if idb can open a DB.

Not sure if there's any way to add a test for this, and I'm having a hard time running the tests to begin with ... :/

### DIFF
--- a/lib/adapters/idb/constants.js
+++ b/lib/adapters/idb/constants.js
@@ -24,3 +24,5 @@ exports.META_STORE = 'meta-store';
 exports.LOCAL_STORE = 'local-store';
 // Where we detect blob support
 exports.DETECT_BLOB_SUPPORT_STORE = 'detect-blob-support';
+// Where we detect general support
+exports.DETECT_SUPPORT_STORE = 'detect-support';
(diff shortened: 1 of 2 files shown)


nickcolley:
Hey @rubensayshi what issues are you running into when trying to run the tests?

rubensayshi:
hmm idk, `npm test` and `npm run dev` execute `bin/run-tests.sh`, when I look at `bin/run-tests.sh` it seems to me `$SERVER` is supposed to be set to something, but it's not ... ?!

daleharvey:
This doesnt work for me, at least in Firefox nightly, unfortunately we made this a sync api and catching a database failing to open is async
```
try { 
  indexedDB.open("dummy"); 
} catch(e) { console.log('caught'); }
```
Gives an `InvalidStateError` thrown as an uncaught async error (you can catch it with `openReq.onerror`

daleharvey:
I will see if I can try to get a little more activity on the Firefox bug, ideally we shouldnt need to work around anything here since this is is a browser bug, however in the meantime it is possible to detect us failing to open the database
```
    new PouchDB('testing', function(err, db) { 
      if (err) {
        console.log('Probably private browsing?');
        return;
      }
      db.post({a: 'doc'}, function(err, doc) {
        console.log('I POSTED!', doc);
      });
    });
```

daleharvey:
@rubensayshi What do you think? is the ability to detect the pouch failure to open good enough for your use case? in this situation I would expect authors would either say 'Doesnt work in private browsing mode in firefox', or to switch out the local database with the remote version.

Changing valid to be async would be a fairly large change and this will not work reliably without it, I am also generally wary of attempting to open database to test for compatibility, I believe we used to do that and it has caused problems.

Hopefully I have tried pinging that bug and hoping we can get it fixed in firefox sooner rather than later

daleharvey:
ok going to close this out @rubensayshi apologies but this way isnt going to work, if there is something you still need for your use case could you open an issue and we can talk about it, cheers

PR #4559, opened 2015-11-10
don't 'wait for host' when there's no $SERVER.

rubensayshi:
`npm run dev` and `npm test` both exec `bin/run-tests.sh` without a `$SERVER`, how is it even working for anyone? because to me it looks like it needs `$SERVER` to not exit?

### DIFF
--- a/bin/run-test.sh
+++ b/bin/run-test.sh
@@ -51,20 +51,20 @@ if [[ ! -z $SERVER ]]; then
     echo -e "Unknown SERVER $SERVER. Did you mean pouchdb-server?\n"
     exit 1
   fi
-fi
 
-printf 'Waiting for host to start .'
-WAITING=0
-until $(curl --output /dev/null --silent --head --fail --max-time 2 $COUCH_HOST); do
-    if [ $WAITING -eq 4 ]; then
-        printf '\nHost failed to start\n'
-        exit 1
-    fi
-    let WAITING=WAITING+1
-    printf '.'
-    sleep 5
-done
-printf '\nHost started :)'
+  printf 'Waiting for host to start .'
+  WAITING=0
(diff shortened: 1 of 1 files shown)


rubensayshi:
also the first test fails for me

```
  1) test.all_docs.js-http Testing allDocs opts.keys:

      Uncaught AssertionError: expected [ '0', '1', '2', '3' ] to deeply equal [ '3', '1' ]
      + expected - actual

       [
      -  "0"
      +  "3"
         "1"
      -  "2"
      -  "3"
       ]
```

rubensayshi:
also travis seems to err randomly? https://travis-ci.org/pouchdb/pouchdb/builds/90292098 or did I really mess something up with this change?

daleharvey:
@rubensayshi Yup our automated tests arent 100% reliable, these fails are very likely intermittents and we just restart them (working on getting them more reliable)

However I dont particularly think we want to make this change, the check for the host being up is checking against `$COUCH_HOST` which defaults to `http://127.0.0.1:5984` (https://github.com/pouchdb/pouchdb/blob/master/bin/run-test.sh#L4) We do this to ensure there is a host running to test against which is a requirement for the tests to run. If you are having problems running the tests it is likely because there is no host running. 

If I misunderstood some use case then please feel free to explain, but if it is an issue for you getting the tests running we can fix that in the other thread, cheers

daleharvey:
Oh wait there isnt another thread, you need a CouchDB compatible server running to be able to run the test suite, if you install the default install of CouchDB and `npm test` then everything should work, if you still run into problems feel free to reply here / open a new issue, this is something we should like to make simple

rubensayshi:
yea, I understand now ;-)

Issue #4314, opened 2015-09-11
PouchDB wrongly flags + as illegal character for database name


colinskow:
```js
new PouchDB('http://localhost:5984/my+database')
  .catch(function(err) {
    console.log(err);
  });
```

Responds...

```
{ [illegal_database_name: Database encountered an unknown error]
  status: 400,
  name: 'illegal_database_name',
  message: 'Database encountered an unknown error',
  error: true,
  reason: 'Name: \'my database\'. Only lowercase characters (a-z), digits (0-9), and any of the characters _, $, (, ), +, -, and / are allowed. Must begin with a letter.' }
```

willholley:
This error is just passed through from CouchDB in this instance. It will work if you percent-encode the +:

    new PouchDB('http://localhost:5984/my%2Bdatabase')
    .catch(function(err) {
       console.log(err);
    });

daleharvey:
I am fairly sure we purposefully dont url encode the database name, but I cant quite think of why right now

daleharvey:
Nope, cant think of any reason to not do this

daleharvey:
fixed in https://github.com/pouchdb/pouchdb/commit/94c7276adfe0ce36b6471874cb60d5bdbcebdacf

lupomontero:
This breaks things when a database name contains a `/`, which is valid as far as CouchDB is concerned (ie: user databases in Hoodie are named `user/randomid`).

Before [this change](https://github.com/pouchdb/pouchdb/commit/94c7276adfe0ce36b6471874cb60d5bdbcebdacf) we could work with database names like this, but not now. Consider a CouchDB server being served in a subdirectory (ie: `http://foo.com/_couch/`), and a database called `user/bar`, so the URL would be `http://foo.com/_couch/user%2Fbar`.

```js
// pre 5.1.0 this worked, now it gets double encoded
var db = new PouchDB('http://foo.com/_couch/user%2Fbar');

// now, without encoding PouchDB will think `bar` is the database name
var db = new PouchDB('http://foo.com/_couch/user/bar');
```

This is because `PouchDB` is splitting the url using `/` as the separator, and then encoding the last part. From [source](https://github.com/pouchdb/pouchdb/blob/master/lib/adapters/http/index.js#L88):
```js
  // Split the path part of the URI into parts using '/' as the delimiter
  // after removing any leading '/' and any trailing '/'
  var parts = uri.path.replace(/(^\/|\/$)/g, '').split('/');

[...]

PR #4563, opened 2015-11-11
fix double encoding database name in http adapter

lupomontero:
See https://github.com/pouchdb/pouchdb/issues/4314#issuecomment-155931616

### DIFF
--- a/lib/adapters/http/index.js
+++ b/lib/adapters/http/index.js
@@ -89,7 +89,11 @@ function getHost(name) {
 
   // Store the first part as the database name and remove it from the parts
   // array
-  uri.db = encodeURIComponent(parts.pop());
+  uri.db = parts.pop();
+  // Prevent double encoding of URI component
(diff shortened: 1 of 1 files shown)


daleharvey:
This is great thanks, the code looks good, it definitely needs a test though, if you want to add a test that would be great, otherwise I will try to get to it, the test would be pretty much identical to 

https://github.com/pouchdb/pouchdb/blob/master/tests/integration/test.basics.js#L38

but with a url encoded name given

daleharvey:
I added a test @ it, it failed before, passes with this fix, fairly certain we are good to go but will make sure its green then merge, thanks

daleharvey:
Ok thats great, test is all green so merged, thanks so much - https://github.com/pouchdb/pouchdb/commit/ba4df1166d1554a0b26b9714d991ba534b297377

lupomontero:
@daleharvey nice one! many thanks! (sorry for the slow reply)

PR #4566, opened 2015-11-12
fixed cors error

wilk:
There's no way to check if the backend has CORS enabled or not.
Checking the origin for a cross-domain request is a bad solution: in fact, if the server has CORS enabled, it will throw a CORS error/warning and that's wrong.

The problem happens with PouchDB 5.1 on the _bulk_get request because sometime's ignored and sometime's not: when it's not ignored, the replication stops (db.replicate.from).
Using status 400 with response.statusCode equal to 0, it works fine and the replication goes on.

### DIFF
--- a/lib/deps/ajax/ajaxCore.js
+++ b/lib/deps/ajax/ajaxCore.js
@@ -87,19 +87,9 @@ function ajax(options, callback) {
 
   return request(options, function (err, response, body) {
     if (err) {
-      // Node request only ever fires error or response, never both
-      /* istanbul ignore if */
-      if (response) {
-        var origin = (typeof document !== 'undefined') &&
-          document.location.origin;
-        var isCrossOrigin = origin && options.url.indexOf(origin) !== 0;
-        if (isCrossOrigin && response.statusCode === 0) {
-          explainCors();
-        }
-        err.status = response.statusCode;
-      } else {
-        err.status = 400;
-      }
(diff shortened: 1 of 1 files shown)


daleharvey:
Thanks for this @wilk 

Again it would be useful to explain the problem you are seeing in more details, so you do a replication and sometimes the `_bulk_get` request fails with `status === 0` and sometimes with 400 and depending on which the replication fails?

I would not like to change the default error to `400`, telling the user that it was `0` tells them that it was a network request error and not something returned by the server.

CORS is designed to be undetectable which is why the given message is only a warning not and an instruction

wilk:
Thank you too for answering quickly ;)

Well, in my case the server always responds with a 400 (it uses CouchDB 1.5) but the error is not handled as expected ("the above 400 error is normal..." and the replication goes on): the replication is blocked because of the status code (0).

Again, this happens on iOS 9 not in Chrome.

Now, I don't know why the status code is set to 0 when it has to be set to 400 but this is what I get debugging the app with Safari (remote debugging).

daleharvey:
ok, so one I think fairly safe way considering the issues we have already had with this bulk_get call is if there is any error at all from bulkGet, just consider it unusable - https://github.com/pouchdb/pouchdb/blob/master/lib/adapters/http/index.js#L323

daleharvey:
Closing this one out, I am going to take another look at the cors error since others are still seeing it misreported I definitely think just disabling bulk_get on any error is the safest option

PR #4567, opened 2015-11-12
fixed _bulk_get headers

wilk:
If host.headers is undefined, don't add it to the ajax call because it will perform the _bulk_get call without basic headers (like the authorization one).

When this problem happens, _bulk_get is done without authorization header and a CORS is thrown even if credentials are passed from the application and CORS are enabled on the server.

### DIFF
--- a/lib/adapters/http/index.js
+++ b/lib/adapters/http/index.js
@@ -272,19 +272,23 @@ function HttpPouch(opts, callback) {
     var self = this;
 
     function doBulkGet(cb) {
-      var params = {};
+      var params = {},
+          ajaxParams = {
+            method: 'POST',
+            body: { docs: opts.docs}
+          };
       if (opts.revs) {
         params.revs = true;
       }
       if (opts.attachments) {
         params.attachments = true;
       }
-      ajax({}, {
-        headers: host.headers,
-        url: genDBUrl(host, '_bulk_get' + paramsToStr(params)),
-        method: 'POST',
-        body: { docs: opts.docs}
-      }, cb);
+      if (host.headers) {
+        ajaxParams.headers = host.headers;
+      }
+
+      ajaxParams.url = genDBUrl(host, '_bulk_get' + paramsToStr(params));
+      ajax({}, ajaxParams, cb);
     }
 
     function doBulkGetShim() {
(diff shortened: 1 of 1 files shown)


daleharvey:
We use this pattern in all of our calls, so I would like to understand the issue a bit better. Are you saying you do a 

`new PouchDB('name', {ajax: {headers: { 'Auth': ',,,'}});`

then when making the bulk_docs request the headers are not sent?

wilk:
I tried with both:

```
new PouchDB('name', {ajax: {headers: {Authorization: ...}}})
db.replicate.from(url, {ajax: {headers: {Authorization: ...}}})
```

The problem occurs only on iOS (9) because in Chrome it works fine.
In fact, headers are well sent with allDocs call but not with bulk get.

daleharvey:
ok so this is actually a curious bug, `_bulk_get` landed around the same time I refactored the way some of the header merging works, previously every call to `ajax(` from inside http used to have to pass in global headers but now they are merged within the ajax function (https://github.com/pouchdb/pouchdb/blob/master/lib/adapters/http/index.js#L156).

So if you just remove `host.headers` here completely you should be good, @wilk do you fancy testing that and opening a PR for it if so?

wilk:
You mean, removing it from doBulkGet?
Is it correct?
I mean, I've already tested it without *host.headers* and it worked but I refactored just for backwards compatibility.
Are *host.headers* useless?

daleharvey:
Yup pretty much, we moved where we defined the `host.headers` that are used, every call to `ajax` used to have them, now none of them do and the `ajax` function handles that for them all

daleharvey:
@wilk Hey sorry to be hassle, however the refactoring here is different to how most of our code is structured, we dont use comma seperated `var`'s and we generally stick to calling `ajax` with its options inline, partially to make up for the confusing api to that that we should fix.

If you could edit it to just take out the `host.headers` line from the original, that would be perfect, thanks

daleharvey:
Awesome, thanks for that, apologies it was a little roundabout, merged in https://github.com/pouchdb/pouchdb/commit/3011e714633238eb4c4c74280346010bb190c8e5

PR #4569, opened 2015-11-13
Remove fauxton from CouchDB build

garrensmith:
Fauxton isn't needed for the Travis build and it should save about 5
minutes on the build process.


daleharvey:
Yup makes sense, thanks, https://github.com/pouchdb/pouchdb/commit/683893e51968534506c464f0e7edb17393b67d54

Issue #4545, opened 2015-11-05
Spurios CORS warnings


marcelgoya:
I just upgraded to the new version and I'm constantly getting the following warning:

PouchDB: the remote database may not have CORS enabled.If not please enable CORS: http://pouchdb.com/errors.html#no_access_control_allow_origin_header

I've enabled CORS and it worked fine with 5.0.0. I'm also experiencing a never ending loop during the sync now

nickcolley:
There was a patch to CORS https://github.com/pouchdb/pouchdb/issues/4501 that might be why you're getting a behaviour change, maybe there was a mistake there.

Can you provide a code snippet to demonstrate the never ending loop for sync?

marcelgoya:
Sure, below my code. It normally fires up the complete listener but he's getting into a loop between active and changes. I don't know if this is only a CB related problem or a general one.

```
       var localDB = pouchDB(window.userId, {
			auto_compaction: true
		});

		$rootScope.sync = localDB.replicate.from('https://' + window.userId + ':' + window.password + '@mydomain.com:4984/default', {
			live: false,
			retry: true,
			since: 0			
		}).on('change', function(changes) {
			console.log('PouchDB syncing inital changes');
		}).on('paused', function(err) {
			console.log('PouchDB inital sync paused');
		}).on('active', function() {
			console.log('PouchDB inital sync active');
		}).on('complete', function() {
			console.log('PouchDB inital syncing complete');
                 
                       // Do something

		}).on('denied', function() {
			console.log('PouchDB inital sync denied');
		}).on('error', function(err) {
			console.log('syncing error: ');
			console.log(err);			
		});
```

daleharvey:
So the never ending loop is very likely https://github.com/pouchdb/pouchdb/issues/4542, I will rename this and look into the CORS issue though, not the first time that has been reported

daleharvey:
Removed in https://github.com/pouchdb/pouchdb/commit/56da7075224da9b74981308c898c69e66a92ee2e

PR #4572, opened 2015-11-14
(#4545) - Remove CORS explanation

daleharvey:
So the fact is CORS errors are designed to be undetectable, the false positives that are pretty much impossible to avoid were confusing uses, and the browsers are doing a better job at making cors errors more visible now anyway

### DIFF
--- a/lib/deps/ajax/ajaxCore.js
+++ b/lib/deps/ajax/ajaxCore.js
@@ -6,7 +6,6 @@ var errors = require('./../errors');
 var clone = require('../../deps/clone');
 var applyTypeToBuffer = require('./applyTypeToBuffer');
 var defaultBody = require('./defaultBody');
-var explainCors = require('./explainCors');
 
 function ajax(options, callback) {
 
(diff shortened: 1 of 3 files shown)


lib/deps/ajax/ajaxCore.js
```diff
@@ -87,19 +86,7 @@ function ajax(options, callback) {
 
   return request(options, function (err, response, body) {
     if (err) {
-      // Node request only ever fires error or response, never both
```
marten-de-vries: Maybe keep this comment as it's still relevant? Not too important though.

marten-de-vries:
Code looks good (I added one minor nit comment inline).

I haven't kept up with reports of people finding this confusing, but if it is then I'm fine with removing it. So consider this my :+1:. :)

lib/deps/ajax/ajaxCore.js
daleharvey: Nah the comment was there only to mention that the following code path would not be followed in node

daleharvey:
Sweet cheers - https://github.com/pouchdb/pouchdb/commit/56da7075224da9b74981308c898c69e66a92ee2e

lib/deps/ajax/ajaxCore.js
marten-de-vries: Well, technically that code path is still there, just as part of the ternary expression below. But it's not important enough, so no need to follow up on it. :)

PR #4576, opened 2015-11-14
(#2035) - Remove cors proxy

daleharvey:


marten-de-vries:
Looks good; seems like this requires CouchDB installs of devs to have CORS enabled, though. Might be good to mention that in the docs. Apart from that, :+1: when green.

daleharvey:
Hrm this is extremely strange, passing in ie 10 normal, but not ie 10 with the memory adapter

nolanlawson:
I'm seeing failures in every Saucelabs browser, which is the same thing I saw the last time I tried to fix this CORS issue. :crying_cat_face: 

screenshot 2015-11-21 16 10 24

daleharvey:
Saucelabs chrome / safari and very strangely internet explorer are passing, its not blanket saucelabs but definitely something strange is up

nolanlawson:
Just re-ran all the tests, looks like iPhone is the one browser that's consistently failing.

I'm happy to keep the CORS proxy around for now, rather than "rabbit hole" on this issue. But yeah, it's annoying.

Issue #4540, opened 2015-11-04
Add a timeout option to replication


garethbowen:
I'm using server side filtered replication which can be very slow, particularly when replicating since `0` in a large remote database. Furthermore I'm using live replication with a heartbeat to keep the databases synced. In both of these cases the default 30 second timeout is too short. Add an option to replication to override the default changes feed timeout.

daleharvey:
Fixed by https://github.com/pouchdb/pouchdb/commit/9cc5d0c4eb1dc935eb7ddec6e27e819917e1f2f2

PR #4581, opened 2015-11-16
(#4540) - Add timeout option to replication

garethbowen:
Passes through a timeout option to the changes feed so longpoll requests can be kept alive for longer than the default 30 seconds.

### DIFF
--- a/docs/_includes/api/replication.html
+++ b/docs/_includes/api/replication.html
@@ -26,6 +26,7 @@ All options default to `false` unless otherwise specified.
 
 * `options.since`: Replicate changes after the given sequence number.
 * `options.heartbeat`: Configure the heartbeat supported by CouchDB which keeps the change connection alive.
+* `options.timeout`: Request timeout (in milliseconds).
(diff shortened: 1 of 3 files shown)


daleharvey:
This is great @garethbowen, dont worry about the failure, it was intermittent.

I think if we are adding this to more places we should have it be consistent with `heartbeat`, which means we should allow setting `timeout: false` which will disable us sending the timeout value at all, does that sound right @garethbowen?

Awesome job on the test, thanks

garethbowen:
@daleharvey Yes being able to disable the timeout altogether by passing `false` would be awesome. Do you think it should be done as part of this issue or in a new one?

daleharvey:
Yeh at the very least it needs to be changed in there to allow the option to pass through `timeout in opts` etc, the actual removing the param could happen in another bug

garethbowen:
I've updated the PR to pass through the `timeout` option even if it's set to `false` and raised #4583 to modify the Changes implementation to respect it.

daleharvey:
Perfect, thanks for this, merged in https://github.com/pouchdb/pouchdb/commit/9cc5d0c4eb1dc935eb7ddec6e27e819917e1f2f2

Issue #4549, opened 2015-11-07
Replicate.on('beforeChange')


filoozom:
Hello,

I'd like to get the difference between the old and the new documents that get replicated via ```PouchDB.replicate()```. I need to remove old files from the local storage, but I need to know the name of the file for that, which is contained in the old document.

I tried something like : 
```
localDB.replicate.from(remoteDB).on('change', function(info) {
    info.docs.forEach(function(change) {
        localDB.get(change._id, function(doc) {
            // Already the new data
            removeFile(doc.picture);
        });
    });
});
```

But ```doc``` already contains the new data.
How could I do this ? Would it be possible to implement a ```.on('beforeChange')``` event ?
I guess I could just use ```localDB.changes()```, but the replication process gets kind of complicated then because I'd need to always fetch the current revision of a document before updating it, and the sequence numbers of both databases wouldn't be in sync.
Or should I just replicate one document after acknowledging a change ? But that doesn't seem like a good solution.

tohagan:
An alternative approach that you might consider is to treat your docs as a sequence of delta changes and then use a map/reduce view to collapse the changes to the current state of each doc.  To apply a change to the view you now send add/remove delta docs. The localDB.changes() is now a sequence of add and remove deltas that will include the doc.picture file paths you want to add/remove.

nolanlawson:
The way the changes feed is currently implemented, I think it would be prohibitively complicated to add this feature. It also introduces some questions that can't be answered very easily - for instance, if a change came in because you are resolving conflicts, then which document was the "previous" document? The previous winner, or the parent revision of the current winner? (They're not the same thing necessarily!) Also what happens if you're doing auto_compaction? That previous document no longer exists, because we just removed it.

I would recommend @tohagan's approach, which I think is well-documented here: http://pouchdb.com/2015/02/28/efficiently-managing-ui-state-in-pouchdb.html. You can also use `stale:ok` to pull out stale data from map/reduce queries, meaning that if you absolutely need to see some old view of the data, you can use that. You can also just not remove the fields when you update your documents, i.e. add an `"oldFilename"` field

tohagan:
You might also want to have a look [Delta Pouch](https://github.com/redgeoff/delta-pouch)

PR #4588, opened 2015-11-18
(#4549) Include old versions of replicated documents

filoozom:
Option to include old versions of replicated documents

### DIFF
--- a/lib/replicate/replicate.js
+++ b/lib/replicate/replicate.js
@@ -30,6 +30,7 @@ function replicate(src, target, opts, returnValue, result) {
   var checkpointer;
   var allErrors = [];
   var changedDocs = [];
+  var oldDocs = [];
   // Like couchdb, every replication gets a unique session id
   var session = uuid();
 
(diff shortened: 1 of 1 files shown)


nolanlawson:
It was very kind of you to write up this pull request, but I think there are a few issues with the proposed change:

1. Discussion in #4549 didn't really reach a consensus on why this feature was even needed,
2. This adds no new tests, meaning coverage dropped from 100%
3. This new feature was not documented

Other than that, the proposal kinda makes sense, although I'm hesitant about adding additional code that we then need to maintain. In general I would vote no on this PR even if we fixed the 3 issues above, because I think there are other ways to solve this problem (e.g. listening to the changes feed, fetching the old revisions in your own `on('change')` event, etc.) without needing to bloat PouchDB's core codebase.

Waiting for @daleharvey's input, but I am -1.

filoozom:
After the answers I got in #4549, I get that this issue could be resolved in different ways.
But could this really be solved with a custom ```on('change')``` ? I mean, maybe it could with ```auto_compaction```disabled, but ```replicate.on('change')``` fires after a change was written if I'm not mistaken, so there are chances that the old document doesn't even exist anymore. With my pull request, I fetch the old documents before the new documents are written, so it doesn't break ```auto_compaction```.

But yes, this probably isn't needed, and I implemented it in another way to make it work in my app.

daleharvey:
Yeh I agree, thanks very much for the PR, but I think there are problems with the semantics of this and it is best solved in user code, I wouldnt necessarily change everything to a delta but possibly store a `toDelete: ['somefile.png']` field in the updated document that gets deleted when the file has gone.

PR #4600, opened 2015-11-23
(#4675) - skip invalid tests for CouchDB 2.X

willholley:
CouchDB 2.X adds validation of design documents at save time.
This prevents issues at query time due to malformed docs but also
causes some of our tests to fail where they explicitly test the
behaviour of the changes endpoint when malformed design documents
exist.

I think the correct thing for now is just to skip these tests when
running against CouchDB master as we still want to check the behaviour
against local adapters.

### DIFF
--- a/tests/integration/test.changes.js
+++ b/tests/integration/test.changes.js
@@ -645,6 +645,12 @@ adapters.forEach(function (adapter) {
       // worth our time. This test does increase code coverage for our
       // own local code, though.
       it('Changes with invalid ddoc with no map function', function () {
+        // CouchDB 2.X does not allow saving of invalid design docs,
+        // so this test is not valid
+        if (testUtils.isCouchMaster()) {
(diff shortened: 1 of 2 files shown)


daleharvey:
Tests are red, returning `Promise.resolve()` should work for that case. Ideally we should be matching Couch 2.0 functionality, but +1 on getting the tests less red first (when there is a green PR)

daleharvey:
If you want to call `done()` to early return, you also need to make sure it is called in the success case

willholley:
There are still some failing tests against CouchDB master but I'm unable to reproduce them.

willholley:
re-ran the tests in Travis and they passed :(

daleharvey:
Awesome thanks, possibly an intermittent, will wait to close the bug on checking, thanks these look good - https://github.com/pouchdb/pouchdb/commit/02f35041d00b46e98d78b52b9ddb745518548285

Issue #4578, opened 2015-11-15
Coerce options which should be numbers to number


wheresrhys:
e.g limit is ignored in the following 
```
db.query(`${indexName}/index`, {
    limit: "4"
})
```
Either this should error/warn or coerce to a number (which would be nice as these values will often come from url query strings), but just getting ignored isn't particularly helpful.

daleharvey:
For a simple case like this, I think coercing them would be fine, cheers for the report, if you wanted to make PR for it would be very happy to help

wheresrhys:
Great. I'll have a go some time in the next few days. If a value can't be
coerced to a number what are your thoughts on behaviour. I'd favour
erroring - particularly as its wrapped in the safety of a promise - but a
warning seems reasonable too

daleharvey:
Yup if we are given a value and expecting a certain type then I would throw an error. One thing I would like to mention is that the check should be done as deep as possible to where it is actually used and not checking on the way in. I think that way has a lot of benefits

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/112faf9945e8fc746cff20f62df47a5e4f4425a6

PR #4601, opened 2015-11-23
(#4578) - Coerce options which should be numbers to number

wheresrhys:
Addresses https://github.com/pouchdb/pouchdb/issues/4578 @daleharvey 

Let me know if I've missed any other points in the code where there should be additional changes to address this

### DIFF
--- a/lib/mapreduce/index.js
+++ b/lib/mapreduce/index.js
@@ -190,6 +190,38 @@ function addHttpParam(paramName, opts, params, asJson) {
   }
 }
 
+function coerceInteger (integerCandidate) {
+  if (typeof integerCandidate !== 'undefined') {
+    var asNumber = Number(integerCandidate);
+    if (!isNaN(asNumber) && asNumber === parseInt(integerCandidate, 10)) {
+      return asNumber;
+    } else {
+      return integerCandidate;
+    }
+  }
+}
+
+function coerceOptions(opts) {
+  return Object.assign({}, opts, {
+    group_level: coerceInteger(opts.group_level),
+    limit: coerceInteger(opts.limit),
+    skip: coerceInteger(opts.skip)
+  });
+}
+
+function checkPositiveInteger (number) {
+  if (number) {
+    if (typeof number !== 'number') {
+      return  new QueryParseError('Invalid value for integer: "' +
+      number + '"');
+    }
+    if (number < 0) {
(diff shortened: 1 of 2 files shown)


lib/mapreduce/index.js
```diff
@@ -190,6 +190,38 @@ function addHttpParam(paramName, opts, params, asJson) {
...
+    group_level: coerceInteger(opts.group_level),
+    limit: coerceInteger(opts.limit),
+    skip: coerceInteger(opts.skip)
+  });
+}
+
+function checkPositiveInteger (number) {
+  if (number) {
+    if (typeof number !== 'number') {
+      return  new QueryParseError('Invalid value for integer: "' +
```
daleharvey: tiny nit: extra space `return new QueryParseError`

daleharvey:
@wheresrhys Thanks for the PR, the new testing is great

The tests are failing on this however though, I am assuming from the error that `Object.assign` is going to have spotty support between browsers and node versions. Even fixing that doing whole object clone to work on 3 values seems wrong

```
opts.group_level = ('group_level' in opts) ? coerceInt(opts.group_level) : undefined;
// Same for skip etc
```

or similiar for each of the 3 values seems clearer and much less risky code here. 

Last question would be for coerce, why would a simple `parseInt(number, 10)` not suffice?

wheresrhys:
Object assign is used to avoid mutating the options object. Perhaps that is
a bit too purist as it's only primitive values being altered.

E.g. Parseint('1 acacia road') will return 1, so that's why the extra check
is there. Would seem wrong to allow the above but error on e.g. 1.2

daleharvey:
Ok thanks, `coerce` vs `parseInt` makes sense (it may be useful to add a comment to that function to explain it). 

The code + tests look good for me now, the tests however are failing so will need to take a look at that, if you have problems running the test suite then just ask, thanks a lot

wheresrhys:
I forgot to return the object from coerceOptions. Fixed now

wheresrhys:
I can't recreate that failed build locally ^

daleharvey:
Yup this is looking great now, thanks, merged in https://github.com/pouchdb/pouchdb/commit/112faf9945e8fc746cff20f62df47a5e4f4425a6

daleharvey:
I forgot to mention sorry, our tests arent 100% reliable and occasionally fail with an intermittent, trying to chase them down but restarting the test and it was green so you are all good

PR #4615, opened 2015-11-26
4530 - do not require explicit revs to be specified when calling bulkGet

willholley:
This PR adds some basic tests for bulkGet (which triggered some documentation updates!) and allows bulkGet to be called without explicitly specifying the rev for each document (similar to CouchDB).

### DIFF
--- a/docs/_includes/api/bulk_get.html
+++ b/docs/_includes/api/bulk_get.html
@@ -12,9 +12,9 @@ Given a set of document/revision IDs, returns the document bodies (and, optional
   - `id`: ID of the document to fetch.
   - `rev`: Revision of the document to fetch. If this is not specified, all available revisions are fetched.
   - `atts_since`: Optional and supported by the http adapter only. Includes attachments only since specified revisions. Doesn’t includes attachments for specified revisions.
-* `options.revs`: Each returned revision body will include its revision history as a `_revisions` property. Default is `true`.
-* `options.attachments`: Include attachments in the response. Default is `true`.
-* `options.binary`: Return attachment data as Blobs/Buffers, instead of as base64-encoded strings.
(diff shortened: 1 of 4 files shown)


daleharvey:
Sorry some linting errors here, looks like a good start though

willholley:
oops - linting errors should now be fixed (and I squashed the commits).

daleharvey:
This looks good to me but it does seem to have regressed the node couchdb-master tests? (I gave it a restart)

nolanlawson:
couchdb-master is indeed failing; pouchdb-server seems to be failing only in Firefox (not in Node).

kxepal:
Sorry, I have a filter on "couchdb-master fail" key phrase (: Looking in the build log I see the following reason:
> node: ../deps/leveldb/leveldb-1.18.0/db/dbformat.h:99: leveldb::Slice leveldb::ExtractUserKey(const leveldb::Slice&): Assertion `internal_key.size() >= 8' failed.
> Aborted

Looks as something internal, right?

daleharvey:
pouchdb-server has been red a while, couchdb-master was green in node and red in firefox apart from this change, it looks like an internal issue but caused by this commit

willholley:
I think couchdb-master tests are failing intermittently for other reasons. This code path should not get exercised against couchdb-master (since bulk_get is supported directly).

daleharvey:
ok cool, will give it another restart, you and @kxepal have been doing most of the work for couch 2.0 compat and its in expected failures, so if yall think this is good to merge then I am happy

daleharvey:
You were right, its intermittent, :+1: and merged in https://github.com/pouchdb/pouchdb/commit/aea279a26c3b214ab37c82175ea612036a8fd74b

PR #4617, opened 2015-11-26
Fixed options for http adapter.

### LINKED ISSUE #4616: Destroying a db with http adapter is failing.
With PouchDB 5.1.0, destroying a database using the http adaptor without passing any option fails with this message:

    Uncaught (in promise) TypeError: Cannot read property 'ajax' of undefined
        at ajax (http://localhost:8000/assets/vendor/pouchdb/dist/pouchdb.js:1066:27)
        at PouchDB.HttpPouch.api._destroy (http://localhost:8000/assets/vendor/pouchdb/dist/pouchdb.js:1855:5)
        at destroyDb (http://localhost:8000/assets/vendor/pouchdb/dist/pouchdb.js:864:10)
        at PouchDB.<anonymous> (http://localhost:8000/assets/vendor/pouchdb/dist/pouchdb.js:875:12)
        at PouchDB.<anonymous> (http://localhost:8000/assets/vendor/pouchdb/dist/pouchdb.js:5997:21)
        at PouchDB.<anonymous> (http://localhost:8000/assets/vendor/pouchdb/dist/pouchdb.js:11044:18)
        at http://localhost:8000/assets/vendor/pouchdb/dist/pouchdb.js:8253:21
        at PouchDB.<anonymous> (http://localhost:8000/assets/vendor/pouchdb/dist/pouchdb.js:8240:19)
        at PouchDB.destroy (http://localhost:8000/assets/vendor/pouchdb/dist/pouchdb.js:11044:18)
        at http://localhost:8000/services/Household.js:505:24

jerem:
Fixes #4616.

### DIFF
--- a/lib/adapter.js
+++ b/lib/adapter.js
@@ -847,6 +847,8 @@ AbstractPouchDB.prototype.destroy =
     opts = {};
   }
 
+  opts = opts || {};
+
   var self = this;
   var usePrefix = 'use_prefix' in self ? self.use_prefix : true;
 


daleharvey:
Hrm, we use `destroy()` without a promise extensively in the test suite (not least @ https://github.com/pouchdb/pouchdb/blob/master/tests/integration/test.basics.js#L59)

I would really like to know what situation this would fail since a bug fix like this needs tests to ensure we dont regress it. Is there anything specific to your setup that you think could cause this to fail, what browser / environment did you see it in?

I am assuming `var reqAjax = userOpts.ajax || {};` is the line that is failing when opts are undefined, little confused at why it ever works. I assumed we passed in a default function somewhere but doesnt seem so. Also just to check, the call stack seems to suggest you are calling `PouchDB.destroy()`, that has been removed from 5.X, are you calling `PouchDB.destroy` or `db.destroy()`?

daleharvey:
Oh, and dont worry about the failing test suite, looks like intermittents, and thanks for the PR

daleharvey:
So took an extra look at this, `adapterFun` will ensure that `AbstractPouchDB.prototype.destroy` will receive at the least a callback parameter, then https://github.com/pouchdb/pouchdb/blob/master/lib/adapter.js#L845 will ensure that there is a default options provided into the http adapter

@jerem This should work and is currently a heavily tested code path, do you know of anything about your setup that would cause something to be different?

nolanlawson:
Closing unless we get more clarification, because I agree that this code path should never get hit. Last I checked, #4616 was a bug in pre-5.1.0 that has long since been fixed.

Issue #4560, opened 2015-11-10
xhr.upload detection is failing


mudcube:
I was having trouble with uploading to my server. After 10 seconds the request would be canceled and start again. I read elsewhere about setting the ajax.timeout. This helped, but it did not deal with the core issue; during an active upload the xhr should abort unless things have stalled.

For me, the test for 'hasUpload' returns false in Firefox/Chrome on mac. Specifically it's this tidbit of code that's causing the problem. For me this returns false on all devices/browsers tested:

```
Object.keys(xhr).indexOf('upload') !== -1
```

I noticed a note in the code about how 'IE throws an error...'. Do you know the version of IE has an error with the code:

```
typeof xhr.upload !== 'undefined'
```

I tested in IE8+ via VMWare and did not have a problem.

daleharvey:
It is a problem with ie10 inside a service worker, https://github.com/pouchdb/pouchdb/issues/2132 has more details

daleharvey:
You are right though, I am seeing that test fail in my browsers. If you could test switching it to `typeof` and open a PR / see if the tests complain?

mudcube:
@daleharvey Sorry for the delayed response I've been in Hawaii ;) Just sent in that PR.

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/0be4b4d4c030c29f0645f490c1f866117ec19711

nolanlawson:
There was a reason I used that fancy-schmancy `Object.keys()` trick - it was to avoid breaking in IE9 when emulated in IE10+. Arguably that is not enough to justify blocking this fix, though, so +1.

PR #4619, opened 2015-11-30
Fix for xhr.upload detection failing (#4560)

mudcube:
As described in https://github.com/pouchdb/pouchdb/issues/4560


daleharvey:
Given that this has been tested in the original browser mentioned, I think we are good here, thanks so much for the PR, https://github.com/pouchdb/pouchdb/commit/0be4b4d4c030c29f0645f490c1f866117ec19711

Issue #4621, opened 2015-11-30
'storage' option not documentated


mudcube:
`storage` option does not show up in the documentation. Additionally, when `storage` option is used in any browser that doesn't support this non-standard feature PouchDB bails.

PR #4622, opened 2015-11-30
(#4621) - add documentation for 'storage' option, add try/catch

mudcube:
As referenced in #4621:

* Added documentation for the `storage` option which allows developers to choose `persistent` or `temporary` storage in Firefox v26+. 
* Added a try/catch so using this feature will not break indexedDB.open on browsers that do not support this non-standard feature. 

More information on Firefox's implementation can be found here; https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API/Browser_storage_limits_and_eviction_criteria

btw, if an issue has not been reported yet, do you prefer an issue and a pull request to be opened, or just to create a pull request without a reference id (and you can use the pull request id in the merge)?


lib/adapters/idb/index.js
```diff
@@ -816,10 +816,14 @@ function init(api, opts, callback) {
...
-      storage: opts.storage
-    });
+    try { // option only available in Firefox 26+
+      req = indexedDB.open(dbName, {
+        version: ADAPTER_VERSION,
+        storage: opts.storage
+      });
+    } catch(err) {
+        req = indexedDB.open(dbName, ADAPTER_VERSION);
+    }
```
nolanlawson: Nice! Can you please move this try/catch to a separate function to avoid deoptimizations in V8? Try/catches [deoptimize their entire parent function](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers).

nolanlawson:
I'm +1 after moving to a separate function. Travis failures seem spurious.

mudcube:
Ahh, interesting, good to know! Let me know if you were thinking the tryCatch wrapper function should appear inside the init() or whether outside is what you were imagining, as it is in my latest commit.

daleharvey:
There is a minor linting error here, but aside from that it looks good to go for me. In terms of issue vs PR, I find its usually easier to open the issue straight away then there can be discussion before deciding on the correct fix, but its not a hassle to open a PR straight away and we all occasionally do it

mudcube:
@daleharvey  Ahha! Thanks for the good advice.

daleharvey:
Awesome thanks, all passing now, :+1: - https://github.com/pouchdb/pouchdb/commit/2433610e283f37496325311535ee34c54214209f

Another thing that makes it easier to merge is to have your branch rebased on master, we cherry-pick / git am directly on master to avoid merge commits. Thanks :)

PR #4636, opened 2015-12-06
Let users know when they submit invalid base64-encoded attachments

### LINKED ISSUE #4208: Error message is too opaque: "Cannot base 64 encode full string"
If you look at the source code where this error is thrown, there is a much better explanation of the problem. It would be nice to surface this to users so they don't get confused.

tmct:
To fix #4208.

The error in that issue's title ("Cannot base 64 encode full string") is never shown to users - it's thrown then discarded, and only in node. The best users get is this error:
```
{ [badarg: Some query argument is invalid]
  status: 500,
  name: 'badarg',
  message: 'Some query argument is invalid',
  error: true,
  reason: 'Attachments need to be base64 encoded' }
```
The `reason` would benefit from more specificity though, so I propose that we change it to 'Attachment is not a valid base64 string'. This commit makes that change and tests for the new string.

### DIFF
--- a/lib/deps/binary/base64.js
+++ b/lib/deps/binary/base64.js
@@ -7,7 +7,7 @@ exports.atob = function (str) {
   // Node.js will just skip the characters it can't decode instead of
   // throwing an exception
   if (base64.toString('base64') !== str) {
-    throw 'InvalidBase64Error';
+    throw new Error("attachment is not a valid base64 string");
   }
   return base64.toString('binary');
 };
--- a/lib/deps/binary/base64.js
+++ b/lib/deps/binary/base64.js
@@ -4,10 +4,10 @@ var buffer = require('./buffer');
 
 exports.atob = function (str) {
   var base64 = new buffer(str, 'base64');
-  // Node.js will just skip the characters it can't encode instead of
-  // throwing and exception
+  // Node.js will just skip the characters it can't decode instead of
+  // throwing an exception
   if (base64.toString('base64') !== str) {
-    throw ("Cannot base64 encode full string");
+    throw 'InvalidBase64Error';
   }
   return base64.toString('binary');
 };
--- a/lib/deps/docs/preprocessAttachments.js
+++ b/lib/deps/docs/preprocessAttachments.js
@@ -20,7 +20,7 @@ function preprocessAttachments(docInfos, blobType, callback) {
       return base64.atob(data);
     } catch (e) {
       var err = errors.error(errors.BAD_ARG,
-        'Attachments need to be base64 encoded');
+        'Attachment is not a valid base64 string');
       return {error: err};
     }
   }
@@ -30,7 +30,7 @@ function preprocessAttachments(docInfos, blobType, callback) {
       return callback();
     }
     if (typeof att.data === 'string') {
-      // input is a base64 string
+      // input is assumed to be a base64 string
 
       var asBinary = parseBase64(att.data);
       if (asBinary.error) {
--- a/tests/integration/test.attachments.js
+++ b/tests/integration/test.attachments.js
@@ -2142,12 +2142,12 @@ adapters.forEach(function (adapter) {
       });
     });
 
-    it('Test putAttachment with invalid base64', function (done) {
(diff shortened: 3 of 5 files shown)


lib/deps/docs/preprocessAttachments.js
```diff
@@ -20,7 +20,7 @@ function preprocessAttachments(docInfos, blobType, callback) {
       return base64.atob(data);
     } catch (e) {
       var err = errors.error(errors.BAD_ARG,
-        'Attachments need to be base64 encoded');
+        'Attachment is not a valid base64 string');
```
tmct: This is the only change I haven't been able to add a test for. I can't break any of the `npm test` tests from within this file, and the whole file seems to be ignored for Istanbul... Are there any tests for this?

lib/deps/binary/base64.js
```diff
@@ -4,10 +4,10 @@ var buffer = require('./buffer');
 
 exports.atob = function (str) {
   var base64 = new buffer(str, 'base64');
-  // Node.js will just skip the characters it can't encode instead of
-  // throwing and exception
+  // Node.js will just skip the characters it can't decode instead of
+  // throwing an exception
   if (base64.toString('base64') !== str) {
-    throw ("Cannot base64 encode full string");
+    throw 'InvalidBase64Error';
```
tmct: This still gets discarded immediately, so I'm not sure what error to throw... Maybe just `throw new Error();`?

tests/integration/test.attachments.js
```diff
@@ -2142,13 +2142,12 @@ adapters.forEach(function (adapter) {
...
-    it('Test putAttachment with incorrect base64', function () {
+    it('Test putAttachment with invalid base64', function (done) {
       var db = new PouchDB(dbs.name);
-      return db.putAttachment('doc', 'att', null, '\u65e5\u672c\u8a9e', 'text/plain').then(function () {
-        throw new Error('shouldnt have gotten here');
-      }, function (err) {
-        should.exist(err);
-      });
+      db.putAttachment('doc', 'att', null, '\u65e5\u672c\u8a9e', 'text/plain')
+        .should.eventually.be.rejected
```
tmct: Looks like some of the tests have failed. Maybe that's because Chai As Promised is set up in the node tests but not the browser ones? I'll try to set it up for the latter.

lib/deps/binary/base64.js
nolanlawson: I'd say `throw new Error("attachment is not a valid base64 string")`. Throwing a string is ugly and weird IMO, but throwing an empty Error makes it impossible to debug.

nolanlawson:
The new test looks good, except for the fact that `reason` is non-standard (we introduced it erroneously long ago... long story), so it would be better to test for `error.name` or `error.message` or something else that we get for free with JavaScript `Error`s.

The Travis tests seem to be erroneous, so I restarted them.

lib/deps/binary/base64.js
nolanlawson: Oh I see, it gets discarded no matter what. Then yeah, it doesn't matter what we throw; it will never get used. I would still say `throw new Error("same string we use elsewhere")`, since gzip will compact that string down anyway, and it's helpful to have a message in case we ever need to debug.

daleharvey:
Apologies for the delay, so the failures here were due to an upgrade in travis that our tests didnt agree with, hopefully once we merge https://github.com/pouchdb/pouchdb/pull/4665 we can get them in a better state.

Agree with @nolanlawson's nits that the throw string should be the same and we should be checking for `err.message` does that sounds good @tmct? if so when you make the change could you rebase on top of latest master and hopefully the tests will all go well. Thanks :)

tmct:
Have made those changes, thanks! Fingers crossed this one'll build...

daleharvey:
It did, the only failure there is another bug, thanks so much for following up on this and sorry it took so long / there were build hassles, trying to iron them out

:+1: - https://github.com/pouchdb/pouchdb/commit/653b99125451975156a2ea61bdb23479274f9316

Issue #4321, opened 2015-09-12
Unable to upgrade Travis to container-based due to OOM in Istanbul


nolanlawson:
We would like to upgrade to the new container-based Travis system, but unfortunately Instanbul fails with an out-of-memory. See https://github.com/pouchdb/pouchdb/pull/4312#issuecomment-139645784 and https://github.com/pouchdb/pouchdb/pull/4250 for the backstory.

nolanlawson:
Jobs that reproduced this:

* https://travis-ci.org/pouchdb/pouchdb/jobs/79838632
* https://travis-ci.org/pouchdb/pouchdb/jobs/78384248
* https://travis-ci.org/pouchdb/pouchdb/jobs/78384249

nolanlawson:
@daleharvey What's the status of this? Did we file an issue on Travis? Can't remember.

daleharvey:
hrm I thought we did but cant find it, either way the legacy infrastructure is completely busted right now so we will likely need to upgrade even if it means losing coverage for now

nolanlawson:
:(

daleharvey:
And upgraded - https://github.com/pouchdb/pouchdb/commit/203044c9824d8eb444640df12a8ea1b69260103e

Issue #3934, opened 2015-06-06
Intermittent test failures


nolanlawson:
Trying to track them here.

MemDOWN test fails for `#2913 massively parallel compaction`:

```
events.js:72
        throw er; // Unhandled 'error' event
              ^
EncodingError: Unexpected token u
    at /home/travis/build/pouchdb/pouchdb/node_modules/levelup/lib/read-stream.js:91:28
    at /home/travis/build/pouchdb/pouchdb/node_modules/pouchdb-server/node_modules/memdown/node_modules/abstract-leveldown/abstract-iterator.js:24:16
    at Object._onImmediate (/home/travis/build/pouchdb/pouchdb/node_modules/pouchdb-server/node_modules/memdown/memdown.js:86:30)
    at processImmediate [as _immediateCallback] (timers.js:354:15)
```

IE10:

```
{ failures: 
   [ { message: 'expected { Object (_id, _rev, ...) } to deeply equal { Object (foo, _id, ...) }',
       stack: 'AssertionError: expected { Object (_id, _rev, ...) } to deeply equal { Object (foo, _id, ...) }\n   at assert (http://127.0.0.1:8000/node_modules/chai/chai.js:869:7)\n   at assertEql (http://127.0.0.1:8000/node_modules/chai/chai.js:1412:5)\n   at Anonymous function (http://127.0.0.1:8000/node_modules/chai/chai.js:3627:5)\n   at assertEqual (http://127.0.0.1:8000/node_modules/chai/chai.js:1378:7)\n   at Anonymous function (http://127.0.0.1:8000/node_modules/chai/chai.js:3627:5)\n   at Anonymous function (http://127.0.0.1:8000/tests/integration/test.issue3179.js:298:11)\n   at Anonymous function (http://127.0.0.1:8000/dist/pouchdb.js:22961:7)\n   at nextTick (http://127.0.0.1:8000/dist/pouchdb.js:22993:7)\n   at onreadystatechange (http://127.0.0.1:8000/dist/pouchdb.js:23077:7)',
       title: '#3179 conflicts synced, live repl' } ],
  lastPassed: '#3179 conflicts synced, non-live replication',
  failed: 1,
  passed: 715,
  browser: 'Mozilla/5.0 (compatible; MSIE 10.0; Windows NT 6.2; WOW64; Trident/6.0; .NET4.0E; .NET4.0C)' }
```

Android 5.1:

```
{ failures: 
[...]

PR #4665, opened 2015-12-19
(#4321) - Upgrade travis infra

daleharvey:
See https://github.com/pouchdb/pouchdb/issues/3934#issuecomment-165106843 for the reasoning

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,12 +9,12 @@ services:
 git:
   depth: 30
 
-# sudo: false
+sudo: false
 
-# cache:
-#   directories:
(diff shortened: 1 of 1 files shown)


daleharvey:
So we lose CouchDB master tests with this but coverage is actually working now :) taking a look at the couch master stuff now, would appreciate an early +1 on this though since there are quite a few test related issues building up now (phantomjs is also super faily)

daleharvey:
Sweet, got couch working as well, closing in favour of https://github.com/pouchdb/pouchdb/pull/4668

Issue #4673, opened 2015-12-20
phantomjs is permafail


daleharvey:

daleharvey:
Closing as dupe https://github.com/pouchdb/pouchdb/issues/4686

PR #4674, opened 2015-12-20
(#4673) - Add phantomjs to allowed failures

daleharvey:
Master - https://travis-ci.org/daleharvey/pouchdb/builds/97905051 - 10/10 failures
Reverted last phantom - https://travis-ci.org/daleharvey/pouchdb/builds/97906682 - 10/10 failures
And reverted last es5 shim - https://travis-ci.org/daleharvey/pouchdb/builds/97908086 - 10/10 fails

I dont know what causing these failures, they arent particularly consistent and I dont want to spend any more time on this. I would vote to remove phantom but at the very least it should go into allowed failures

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -120,6 +120,8 @@ matrix:
   - env: CLIENT=selenium:firefox:41.0.1 SERVER=pouchdb-server COMMAND=test
   - env: SERVER_ADAPTER=memdown LEVEL_ADAPTER=memdown SERVER=pouchdb-server COMMAND=test
   - env: COMMAND=report-coverage
+  - env: CLIENT=selenium:phantomjs ES5_SHIM=true COMMAND=test
+  - env: AUTO_COMPACTION=true CLIENT=selenium:phantomjs ES5_SHIM=true COMMAND=test
 
 
   fast_finish: true


nolanlawson:
Most of the failures are stuff like this:

```
 title: 'returnValue doesn\'t leak "change" event w/ onChange',

       message: 'timeout of 100000ms exceeded. Ensure the done() callback is being called in this test.',
```

My guess is that it's a regression in Browserify's EventEmitter shim, which we are now catching in our tests. Nothing else changed recently, so it's the only thing I can think of.

nolanlawson:
In general, I'm +1 with adding this to allowed failures, but we might as well just remove it entirely. Everything that we've ever added to allowed failures just ends up becoming perma-red. :(

I would still like to understand how this is happening, though. I can't repro on my own machine, so maybe it's due to flakey Travis infrastructure?

nolanlawson:
Tried three times, cannot repro locally. My new theory is that this is a race condition that only appears on slow hardware. Either way, +1.

daleharvey:
Will follow up to remove it in a while if nobody wants to maintain it.

I have been making sure pouchdb-express-router has stayed green, couch-master and coverage I both plan on taking a look at once the rest of the suite is stable

daleharvey:
https://github.com/pouchdb/pouchdb/commit/5e4143e3af02ebd4c1e9f3a46b7cd8f3d6ed0650

nolanlawson:
yeah, I feel a responsibility to get pouchdb-server back up to green, but for phantomjs I have very little sentimentality. I just worry it's indicative of a more general problem.

Issue #4678, opened 2015-12-20
 suite2 test.replication.js-local-http Test live push checkpoint


daleharvey:

daleharvey:
https://travis-ci.org/pouchdb/pouchdb/jobs/97945071

daleharvey:
This one I have been seeing forever, so going to take a look now

daleharvey:
Ok I think I have found it, the calls to `db.replicate` and `db.changes` can kick off concurrent constructions, this is something we try to protect against but have never fix it 100% so for now we can ensure the database is created before kicking off both of those functions, doing a run of 40 to make sure its fixed.

daleharvey:
So I can workaround it in the test, 40 runs showed it not coming up again but it did seem there might be more with the same symptoms and we really should be handling this so looking at fixing it proper

nolanlawson:
Nice work! I've been seeing this fail constantly as well.

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/1cbcaeef98bcd3b6c817417db409dae13206eea2

PR #4683, opened 2015-12-20
(#4678) - Use the same database instance and dont reconstruct the db

daleharvey:
So I was more right the first diagnosis, using `dbs.remote` triggers a full parallel reconstruction which we cant entirely guard against, I did want to make sure that it couldnt be triggered by just calling 2 methods on an uninitialized db though and that is working good.

### DIFF
--- a/tests/integration/test.replication.js
+++ b/tests/integration/test.replication.js
@@ -505,7 +505,7 @@ adapters.forEach(function (adapters) {
     it('Test live pull checkpoint', function (done) {
       var db = new PouchDB(dbs.name);
       var remote = new PouchDB(dbs.remote);
-      remote.bulkDocs({ docs: docs }, {}, function (err, results) {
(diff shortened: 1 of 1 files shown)


daleharvey:
Doing another 40 run @ https://travis-ci.org/daleharvey/pouchdb/builds/98001657, but this should be good to go, tested it extensively locally

daleharvey:
ok 40 runs and no more instances of this error - https://travis-ci.org/daleharvey/pouchdb/builds/98001657

nolanlawson:
:+1:, you're doing God's work getting rid of these intermittents

daleharvey:
Not totally finished the suite, but its gone through 40 runs and travis time is precious right now, cheers - https://github.com/pouchdb/pouchdb/commit/1cbcaeef98bcd3b6c817417db409dae13206eea2

Issue #4677, opened 2015-12-20
 test.changes.js-http Kill database while listening to live changes


daleharvey:

daleharvey:
https://travis-ci.org/pouchdb/pouchdb/jobs/97945070

daleharvey:
http adapter, complete isnt being called in some cases of the destroy path, looking now

daleharvey:
I have no idea how this can fail, the changes object on setup adds the event listener for `destroyed` (https://github.com/pouchdb/pouchdb/blob/master/lib/changes.js#L46), this is during the constructor so there is no way it cannot be listening by the time the database gets destroyed, its also never reproduced locally

daleharvey:
Added some debugging @ https://travis-ci.org/daleharvey/pouchdb/builds/98008401

daleharvey:
omg 

```
  1) test.changes.js-http Kill database while listening to live changes:
     Error: Parse Error
      at onError (lib/deps/ajax/ajaxCore.js:52:18)
      at Request._callback (lib/deps/ajax/ajaxCore.js:90:14)
      at self.callback (node_modules/request/request.js:198:22)
      at Request.onRequestError (node_modules/request/request.js:820:8)
      at Socket.socketOnData (_http_client.js:322:9)
      at readableAddChunk (_stream_readable.js:163:16)
      at Socket.Readable.push (_stream_readable.js:126:10)
      at TCP.onread (net.js:540:20)
```

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/7ede502720dea30c87b02608c5710ec62698cf58

PR #4684, opened 2015-12-20
(#4677) - Consider changes erroring a valid result

daleharvey:
What was happening was that when we destroyed the database we kill the changes connection, most of the time it exited cleanly however occasionally socket could not be closed cleanly and triggered an xhr error in turn an changes error.

We cant make assuptions about the underlying state of the socket connection, its pretty much always possible for an abort to trigger a parse error in the xhr request so throwing an error is valid here as long as something is notified to the changes user

### DIFF
--- a/tests/integration/test.changes.js
+++ b/tests/integration/test.changes.js
@@ -1044,17 +1044,12 @@ adapters.forEach(function (adapter) {
 
     it('Kill database while listening to live changes', function (done) {
       var db = new PouchDB(dbs.name);
-      var count = 0;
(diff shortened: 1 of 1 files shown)


nolanlawson:
Awesome work, makes sense to me. +1

daleharvey:
Same, Not totally finished the suite, but its gone through 40 runs and travis time is precious right now, cheers - https://github.com/pouchdb/pouchdb/commit/7ede502720dea30c87b02608c5710ec62698cf58

PR #4687, opened 2015-12-20
Fix segfault sanssync

daleharvey:


daleharvey:
Meant to be on my own repo

Issue #4679, opened 2015-12-20
Immediately cancelling replication causes segfault


daleharvey:
https://travis-ci.org/pouchdb/pouchdb/jobs/97955573

daleharvey:
```
*** Error in `/home/travis/.nvm/versions/node/v0.12.9/bin/node': free(): corrupted unsorted chunks: 0x00007fd044002cb0 ***
Aborted (core dumped)
```

Pretty interesting

daleharvey:
Started noticing a pattern, these all seem to happen in test.sync.js, or after ' Test immediate replication canceling'

daleharvey:
Changed the title to the unlikely sounding but true bug, merged https://github.com/pouchdb/pouchdb/commit/f471a78426599636e40770b58280d4b7c58ea818 to skip the offending tests to give us some time to fix this

daleharvey:
I did a large test run again, problem still exists across all versions of node.js - https://travis-ci.org/daleharvey/pouchdb/builds/132496526

daleharvey:
Ugh gonna look into this again, @mikeal does this sound familiar at all? basically it seems like when we do

```
var req = request(opts);
req.abort();
```
We see either request throw an error
```
 1) test.changes.js-http it handles a bunch of individual changes in live replication:
     Error: socket hang up
      at onError (packages/pouchdb/lib/index.js:5095:18)
      at Request._callback (packages/pouchdb/lib/index.js:5137:14)
      at self.callback (packages/pouchdb/node_modules/request/request.js:200:22)
      at Request.onRequestError (packages/pouchdb/node_modules/request/request.js:831:8)
      at Socket.socketOnEnd (http.js:1624:9)
      at _stream_readable.js:944:16
```
or a segfault, sometimes the segfault doesnt even seem to be inside request code, but it only happens directly after this sync `abort` happens
```
node: ../deps/leveldb/leveldb-1.18.0/db/dbformat.h:99: leveldb::Slice leveldb::ExtractUserKey(const leveldb::Slice&): Assertion `internal_key.size() >= 8' failed.
```

This didnt always happen, it seemed to get introduced sometime around the end of last year, but it is affecting all versions here. 

Will try to narrow it down and hopefully get a request only case that fails but that sounds unlikely, it seems like the `abort` is somehow affecting other io that node is currently working on.

daleharvey:
[...]

PR #4688, opened 2015-12-21
(#4679) - Skip broken sync cancel tests

daleharvey:


daleharvey:
In an interesting turn of events, it turns out by doing 

``` 
var sync = db.sync();
sync.cancel();
```

We can segfault node.js completely, this definitely needs fixing, however we should disable the tests in the meantime because we are over a 50% failure rate with them and its getting in the way of any other work, with these skipped we are pretty close to 100% passing

willholley:
curious, but +1 to the PR

daleharvey:
Yeh finding the root cause of this is going to be interesting, thanks - https://github.com/pouchdb/pouchdb/commit/f471a78426599636e40770b58280d4b7c58ea818

PR #4695, opened 2015-12-23
fixes bug when using with pouchdb-async-storage

joshblour:
addresses https://github.com/nolanlawson/pouchdb-async-storage/issues/6

### DIFF
--- a/package.json
+++ b/package.json
@@ -139,7 +139,7 @@
     "crypto": false,
     "fs": false,
     "leveldown": false,
-    "request": "./lib/deps/ajax/request-browser.js"
+    "request": "./pouchdb/lib/deps/ajax/request-browser.js"
   },
   "browserify": {
     "transform": [


nolanlawson:
Thanks! However, this seems to be a bug caused by subtle differences between webpack/browserify. So your PR is failing the tests because it breaks Browserify.

I suggest a better way to solve this: create our own `requestWrapper.js` that simply wraps request, then create a `requestWrapper-browser.js`. The reason I am guessing this will work is because it is the strategy we use everywhere else (see all the various `-browser.js` files), and it seems to be working for webpack users.

Maybe we should also start testing Webpack in Travis.

joshblour:
Ok. I'd love to submit a pr with that fix but I'm still really new to all these components. Maybe once I get a bit more comfortable.




Also, not to hijack this thread, but the other issue is that even once I'm able to run it on react native, the promises never resolve. Any ideas what could be causing this?

nolanlawson:
I'm not sure. Is that on iOS or Android? If you can try to debug, that would be helpful. Very few people are playing around with PouchDB+ReactNative in my experience. (Also please open a separate issue for that. :))

nolanlawson:
Opening a new issue as #4700, thanks for the clues about what's going wrong here.

PR #4698, opened 2015-12-24
Dont cache node_modules

daleharvey:


daleharvey:
The node_modules caching was breaking leveldown on different versions of node as it download prebuilt versions depending on the nodejs version

nolanlawson:
Ahhh, that explains a lot. +1

Issue #4700, opened 2015-12-25
Webpack ignores request-browser.js


nolanlawson:
See https://github.com/nolanlawson/pouchdb-async-storage/issues/6 and https://github.com/pouchdb/pouchdb/pull/4695.

Webpack is trying to browserify `request` itself instead of using our `request-browser.js`, which causes failures.

nolanlawson:
Well, I am working on writing a test for Webpack build, and discovering that https://github.com/pouchdb/pouchdb/commit/f95aca6147402d82a23633c2e42d0a9911564333 actually broke Webpack. So it's good we're testing this.

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/e586068e6486474ab210e50223d45c1bb82663c1

PR #4701, opened 2015-12-25
(#4700) - fix and test Webpack

nolanlawson:
This fixes Webpack support, and additionally it tests Webpack in CI.

There were (surprisingly) quite a few changes that needed to be made to support this:

* `require('request')` needed to be isolated to its own file and then swapped with the `-browser` version at that level
* `browserify-versionify` wasn't working at all with Webpack and was breaking the tests, since Webpack just left the `'__VERSION__'` string in there as-is. So I swapped it out for a new Browserify transform that I wrote - [package-json-versionify](https://github.com/nolanlawson/package-json-versionify), which simply removes everything from `package.json` except for `"version"`. Webpack will ignore this transform, meaning that Webpack users are punished with a larger bundle size, but it still works.
* Unfortunately, once you start doing `require('../package.json')`, Webpack loses its mind and you suddenly need a `json-loader`. AFAICT from [this discussion](https://github.com/webpack/webpack/issues/378), this is expected behavior and they intend for end-users to include custom loaders to deal with this stuff. So unfortunately our users are going to have to do that. Hopefully they've run into it enough times before that it won't surprise them. (WTF, why can't you just `require()` a JSON file without any additional configuration?)

I think moving forward it is super important that we test Webpack in CI, because I *cannot believe* how much bending over backwards you have to do to support it and Browserify simultaneously.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,6 @@
   ],
   "dependencies": {
     "argsarray": "0.0.1",
-    "browserify-versionify": "^1.0.6",
     "debug": "2.2.0",
     "double-ended-queue": "2.0.0-0",
     "es3ify": "0.1.4",
@@ -31,6 +30,7 @@
     "lie": "3.0.1",
     "localstorage-down": "0.6.3",
     "memdown": "1.1.0",
+    "package-json-versionify": "^1.0.0",
     "pouchdb-collate": "1.2.0",
     "pouchdb-collections": "1.0.1",
     "request": "2.67.0",
@@ -60,6 +60,7 @@
     "istanbul": "0.4.1",
     "istanbul-coveralls": "1.0.3",
     "jshint": "2.8.0",
+    "json-loader": "^0.5.4",
     "less": "2.5.3",
     "mkdirp": "0.5.1",
     "mocha": "2.3.4",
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -116,15 +119,14 @@
...
     "./lib/deps/ajax/createBlobOrBufferFromParts.js": "./lib/deps/ajax/createBlobOrBufferFromParts-browser.js",
     "./lib/deps/ajax/createMultipartPart.js": "./lib/deps/ajax/createMultipartPart-browser.js",
     "./lib/deps/ajax/defaultBody.js": "./lib/deps/ajax/defaultBody-browser.js",
-    "./lib/deps/ajax/explainError.js": "./lib/deps/ajax/explainError-browser.js",
     "./lib/deps/ajax/explainCors.js": "./lib/deps/ajax/explainCors-browser.js",
-    "./lib/deps/env/hasLocalStorage.js": "./lib/deps/env/hasLocalStorage-browser.js",
-    "./lib/deps/env/isChromeApp.js": "./lib/deps/env/isChromeApp-browser.js",
+    "./lib/deps/ajax/explainError.js": "./lib/deps/ajax/explainError-browser.js",
+    "./lib/deps/ajax/prequest.js": "./lib/deps/ajax/prequest-browser.js",
+    "./lib/deps/ajax/request.js": "./lib/deps/ajax/request-browser.js",
```
nolanlawson: I sorted these alphabetically; the only true change is the above line.

daleharvey:
:+1: this is awesome, thanks, merry christmas :)

daleharvey:
Merged in https://github.com/pouchdb/pouchdb/commit/e586068e6486474ab210e50223d45c1bb82663c1

nolanlawson:
Heh, Merry Christmas to you too. :)

Issue #4595, opened 2015-11-21
Firefox throws TypeError for invalid URLs


filoozom:
This issue only occurs on Firefox, and works as expected in all the other browsers I tried.
When you try to replicate to an invalid URL, you get an error on https://github.com/pouchdb/pouchdb/blob/d2ec5f9e3ecff4f373e88473260208e586ddd998/lib/replicate/replicate.js#L219 : 
```
TypeError: setting a property that has only a getter
```

```console.log(err);``` gives something like : 
```
Exception { message: "", result: 2152398858, name: "NS_ERROR_MALFORMED_URI", filename: "http://cdn.jsdelivr.net/pouchdb/lat…", lineNumber: 9, columnNumber: 0, inner: null, data: null, stack: "i@http://cdn.jsdelivr.net/pouchdb/l…" }
```

You can replicate the error on Firefox by providing an invalid URL, like :
```
PouchDB('test').replicate.to('http:///')
```

**Comment by nolanlawson** [comment]
This is a good first-timers bug, because it should be very very easy to reproduce since you kindly provided the one-liner test case:

```js
new PouchDB('test').replicate.to('http:///')
```

If anybody wants to take this up, they are more than welcome. Presumably you would just need to add a test to one of our integration tests (I recommend `test.basics.js`), then follow the instructions in TESTING.md to run the tests, and then fix the issue. It should throw an error, but the error should be the same in all browsers and should be something useful to users (e.g. `"Your URL is invalid"`).

daedlock:
@nolanlawson, I am working on closing this issue. I added url validation in `replicate.to` and `replicate.from` that throws an error on invalid urls. The problem is some tests fail because they are using invalid urls for replication and yet the error gets thrown. Please advise

nolanlawson:
fixed by d3aa2c92b0a5bf461384184211bde68ae4037eec

PR #4705, opened 2015-12-26
(#4595) -  validate urls in replicate.from and replicate.to

daedlock:


daleharvey:
Thanks for the PR :) the setup looks great however I am not certain what we want to do is validate urls here, validating urls is an extremely tricky business and the url here isnt exhaustive, I think what we should be doing is ensuring any error thrown by ajax is properly returned back to the user

Currently it looks like replicate.to will not either complete or throw when hitting this error, so a valid test can look like 

```
    new PouchDB(dbs.name).replicate.to('http:///').then(function() { 
      done();
    }).catch(function(e) {
      done();
    });
```

daleharvey:
Talked to @daedlock on irc and said they are working on changing the PR to avoid reporting error, but pinging @nolanlawson just to check you agree since we didnt say the same thing on the original issue. If we were to do url testing then it should likely go in the http adapter during construction, but I think we want to avoid that really, url parsing is complicated and we have specific features to override ajax so that users dont even need to provide valid urls. I think its best we make sure we propagate the error correctly.

daedlock:
As discussed with @daleharvey. The problem with propagating the error is `NS_MALFORMED_URL` gets thrown earlier in Firefox, so I think the only solution here is to do some sort of a obvious url validation like checking for illegal characters, number of slashes..etc. The validation checks make sense to be done in the `replicate()` constructor and if it fails it would reject the promise and emit an error. So basically I am talking about a wrapper that detects the problem before it happens.

daleharvey:
Sorry little miscommunication, I dont think we should be doing pre validation, I think whatever line throws the malformed url error (I am guessing xhr.send @ https://github.com/pouchdb/pouchdb/blob/master/lib/deps/ajax/request-browser.js#L199) we should catch that error and propogate it back to the user. 

It should hopefully be the matter of a try / catch around the send and calling `callback(err)` when caught

lib/deps/ajax/request-browser.js
```diff
@@ -120,7 +120,13 @@ function xhRequest(options, callback) {
     xhr = new XMLHttpRequest();
   }
 
-  xhr.open(options.method, options.url);
+  try {
+    xhr.open(options.method, options.url);
+  }
+  catch(exception) {
```
daleharvey: nit: We put the catch on the same line generally
``` 
} catch (exception) { 
```

tests/integration/test.basics.js
```diff
@@ -83,6 +83,22 @@ adapters.forEach(function (adapter) {
       });
     });
 
+    it('[4595] should reject xhr errors', function(done){
+      var invalidUrl = 'http:///';
+      new PouchDB(dbs.name).replicate.to(invalidUrl,{})
+      .catch(function(err,changes){
+        done();
+      });
```
daleharvey: nit: this doesnt quite match our code style, should look as follows
```
new PouchDB(dbs.name).replicate.to('http:///').catch(function() {
  done();
});
```

daleharvey:
There are tiny nits on the code style, but I think this looks great, can edit the code style issues on the way in, lets just wait to see if the tests are green

tests/integration/test.basics.js
daedlock: It wont pass  jshint "too long line". Should we ignore?

nolanlawson:
Sweet, tests are green! By the way, I have a huge PR coming in #4710, but I'll consider it my responsibility to fix merge conflicts.

tests/integration/test.basics.js
nolanlawson: hm, it does if you remove the args from the function (which is fine)

nolanlawson:
Fixed the nits and merged as d3aa2c92b0a5bf461384184211bde68ae4037eec. Nice work!

Issue #4652, opened 2015-12-13
Migrate to ES6, use Rollup to build for Node


nolanlawson:
See https://github.com/pouchdb/pouchdb/pull/4452#issuecomment-164276341 for why I want to do this.

Basically our structure would be as follows:

```
src/
lib/
dist/
```

`src` will be the ES6 source. `lib` will be the ES5, as output in one gigantic `index.js` by Rollup. `dist` will be the browser builds.

In our package.json, we'll have an `esnext:main` pointing to `src` and a `main` pointing to `lib`, so Node users will typically get `lib`, but ES6 module bundlers will know to use `esnext:main`.

Browserify will simply build as normal, using `lib` as its source and `dist` as its output.

For our officially supported `extras` API, we will have to build that separately and include it in `lib`, but it should be fairly easy. One additional benefit is that anything *not* in the extras **will immediately stop working**. So e.g. if you are being naughty and trying to use private methods, such as `require('pouchdb/lib/utils/private-thingie')`, it will stop working.

This has the potential to break some of our naughty consumers (I know both Hoodie and my plugins have abused `PouchDB.utils` in the past), so we may want to consider it a breaking change.

nolanlawson:
fixed by https://github.com/pouchdb/pouchdb/pull/4710

reconbot:
:clap: :clap: :star2: :+1:

jarrodek:
@nolanlawson PouchDB 7.0.0 has only `dist/` and `lib/` folders in installed package. The `package.json` file clearly describes to not include `src/` folder as it ids missing in `files` array. Therefore it is impossible to use the library in ES6 modules. Naturally you can import PouchDB in the web app and this will work but that's not the point. Will this be fixed?

PR #4710, opened 2015-12-30
(#4652) - Migrate to ES6/Rollup, build one index.js

nolanlawson:
`src` is the new `lib`. `src` is our source, `lib` is consumed by Node/Browserify/Webpack.

As described in #4652, I'm proposing this not so much because I'm enamored of ES6 and want to use all the fancy features, but because

1) It's getting harder and harder to support both Webpack and Browserify.
2) Rollup allows us to build one giant `index.js`.
3) Having one giant `index.js` allows us to avoid Webpack/Browserify headaches (e.g. with `require('./package.json')`.

I also see this as [kind of inevitable](https://twitter.com/nolanlawson/status/680942231126020096). Most JS projects I see nowadays are building from ES6/Typescript/CoffeeScript, for both Node and the browser. This puts us firmly on that bandwagon, although at least we're doing it in a way that will reduce code size and make us less susceptible to subtle bugs in Webpack/Browserify/Babel/Rollup/JSPM/whatever our users are using to consume us.

The big advantage is that we are insulated from weird bugs, optimizations, and tricks in our user's build scripts. They just see PouchDB as one big `index.js` file (and `index-browser.js` file), and we can build it however we want instead of being forced to also use it as the source. E.g. this means we can include the version number from `package.json` by just doing a search-and-replace as part of the build script.

The big downside I see is the added complexity of the build process and lack of source maps while debugging. The sourcemaps don't bug me so much, though, because when I'm debugging in the browser, I can just search through the one big `index-browser.js`, which is arguably more searchable than many tiny files anyway.

Another possible downside is larger bundle sizes for users using the `extras` API. However, gzip should help out with a lot of that, and many of those benefits were theoretical anyway (and would only occur if both the plugin's PouchDB dep and the user's PouchDB dep were exactly the same version). Also, with Rollup building our primary `index.js` and `index-browser.js`, we end up with overall smaller bundle sizes for all users, regardless of what build system they're using (it decreases the size of all three of Webpack, Browserify, and straight-up `<script>` tags).

Overall, this reduces the size of `dist/pouchdb.min.js` from 48010 to 44815 (3.2KB saved, or 6.7%).


nolanlawson:
Rebased.

daleharvey:
This PR is epic. So I havent been the one maintaining webpack support so not in the best position to say whether or not this is the best way, completely trusting that you know the best thing to do here. 

The smaller dist file is awesome, and I think basing the code on ES6 modules seems like a good bet, its the standard. Our sourcemap support has never been superb, I usually just try to recognise where code that is throwing is coming from anyway, it would be nice to be improving it but its a developer convenience which isnt as important as less end user bugs. My only big worry about this is the build complexity.

I am fairly ignorant about the rollup / webpack / babel infrastructure and a little confused at what is actually happening here, so I have a few questions but I think one change we should make is to add some documentation around whats actually happening during the build process. Possibly in the CONTRIBUTING file or even in comments in the build script.

So we have es6 code in `src/`, rollup then takes all those imports and bundles them into a single file, why do we then need to browserify that file? does rollup not understand external dependencies? 

For the extras, I think we should seriously consider dropping them, anything we expose under extras we create an npm module for, how does that sound?

Lastly what does the simplication of this look like in future, are we going to be able to get rid of browserify? with platforms that fully support ES6 are we going to be able to remove any build step entirely?

This PR is unavoidably going to be a huge hassle if it sits around for any period of time, between people writing PR's and having to constantly rebase it so would be looking to get this in asap, the test run is going pretty well so honestly, I dont understand everything about this right now, but trust its the right thing to do so :+1: when you think it is good to land, I will catch up :)

nolanlawson:
> So I havent been the one maintaining webpack support so not in the best position to say whether or not this is the best way...

I feel like I just see the writing on the wall. Browserify/Webpack are diverging. Babel, Rollup, JSPM (and any newcomers) add to the mess. I don't trust these tools to "do the right thing" anymore; bugs often happen at the last mile of the user's build process.

> add some documentation around whats actually happening... to CONTRIBUTING file...

Will do! :)

>  why do we then need to browserify that file? does rollup not understand external dependencies? 

Yes, because Rollup doesn't understand `"browser"` switches (https://github.com/rollup/rollup-plugin-npm/issues/8). So e.g. when we build an external module (like "buffer") that has a browser vs Node version, Webpack and Browserify know what to do, but Rollup doesn't.

Also we need Browserify support anyway, because so many of our users are using it to consume our `lib` directory, so I just find it convenient to use it for the "final build" (`lib` to `dist`). Even if we used Rollup, we'd probably want to test Browserify as we're testing Webpack.

> For the extras, I think we should seriously consider dropping them, anything we expose under extras we create an npm module for, how does that sound?

I'm kinda against this, because 

1. Lots of dependents in the ecosystem (including the new Hoodie and my many plugins)
2. Would be awkward as external modules (e.g. `checkpointer`).
3. The `plugins` are special because they currently work *only* as a part of the `pouchdb` module. Basically it's a CommonJS/npm hack to be able to do:

```js
// module/plugin.js
require('module').addModule(...);
```

Then consumers can do:

```js
var module = require('module');
require('module/plugin');
```

And it just works &trade;. We're exploiting this to make it dead-simple to add the in-memory/localstorage/fruitdown adapters for Browserify/Webpack users. It would not work at all as a separate npm module, because the plugin and the client would not share the same `module` object (unless we use `peerDependencies`, which come with their own set of problems...).

That said, I'd be open to deprecating them at some point in the future, if we can slowly move them out of the `extras` and into external modules where it makes sense. (`ajax` seems appropriate for that...) Another option is simply e.g. `require('pouchdb').extras`.

Anyway, I'll clean up this PR and try to get it in soon; agree that the potential for bitrot is massive.

nolanlawson:
Rebased, also made some simplifying changes where basically the full build is done in `build.sh` (both browserify and rollup).

CONTRIBUTING.md
```diff
@@ -59,7 +59,9 @@ All dependancies installed? great, now building PouchDB itself is a breeze:
 
 You will now have various distributions of PouchDB in your `dist` folder, congratulations.
 
- * If you are on windows, you will need `node-gyp` to install levelup, visit https://github.com/TooTallNate/node-gyp#installation for installation instructions.
+Note that the source code is in `src/`, which is built by [Rollup](http://rollupjs.org/) as a
+Node module to `lib/`, which is then built by [Browserify](http://browserify.com/) as a browser-ready
+UMD module to `dist/`. All of this logic is in `bin/build.sh`.
```
nolanlawson: ^ Explanation of build setup.

Also I understand this Github diff is impossible to read, but you can just ignore everything in `src`/`lib`, because all the action is in `bin` anyway. `src`/`lib` stuff is just a conversion from ES5 to ES6 (which I did with simple search-and-replace plus some manual tweaks).

bin/build.sh
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env bash
```
nolanlawson: this file builds everything, and is re-run when doing `npm run dev`

nolanlawson:
Ugh, I just realized this is going to mess up our code coverage tests, since the unit tests are tested using a separate build style.

Only solution I can think of is to expose unit-testable modules on the exported `PouchDB` object. (This can be reverted before publishing.) I'll take another stab at this later.

nolanlawson:
OK, I came up with a solution. This should pass all the tests now, including coverage.

nolanlawson:
^ 100% lines coverage for me with this one.

daleharvey:
> Yes, because Rollup doesn't understand "browser" switches (rollup/rollup-plugin-npm#8). So e.g. when
> we build an external module (like "buffer") that has a browser vs Node version, Webpack and 
> Browserify know what to do, but Rollup doesn't.

> Also we need Browserify support anyway, because so many of our users are using it to consume 
> our lib directory, so I just find it convenient to use it for the "final build" (lib to dist). Even if 
> we used Rollup, we'd probably want to test Browserify as we're testing Webpack

Agree that we should add tests to ensure browserify consumers work, I have been meaning to get on that for a long time. However if/when rollup supports browser switches and anything else we get from browserify, it seems like removing a build step should very much be a goal.

I think leaving `extras` as is but with a goal to reduce them, expose them as packages when it makes sense etc sounds like a good idea.

Looks like you missed checking in something for the coverage to work but otherwise looking great :)

nolanlawson:
All green except for coverage; I still messed something up. One more shot...

nolanlawson:
Also I agree about eliminating Browserify for our `dist/` builds if possible. Maybe I'll take it as an opportunity to help implement the necessary Rollup plugin. :)

daleharvey:
suweet, :+1:

nickcolley:
This is mammoth, thanks @nolanlawson ! :+1: :tada: :100:

Issue #4712, opened 2015-12-31
Inserting a new document with an invalid _rev succeeds (against local databases)


willholley:
Against local databases, the bulkDocs function allows insertion of documents with invalid _revs.

For example:

    var db = new PouchDB("test");
    db.bulkDocs([{_id:"foo2",_rev:"1-bar"}]).then(function(results) { console.log(results);});

results in:

    [ { id: "foo2", ok: true, rev: "2-f126b7c8c8c80fa0d7bd8d989d38cd56" } ]

I'd expect a conflict error in this case. Presumbably, the invalid rev is being used to generate the incremental revision (2-f...) and treated like an update.

willholley:
it seems that CouchDB 1.6.1 behaves in the same way:

     $ curl 'http://127.0.0.1:5984/revtest' -XPUT
     {"ok":true}

     $ curl 'http://127.0.0.1:5984/revtest/_bulk_docs' -H 'Content-Type: application/json' --data-binary '{"docs":[{"_id":"foobar","_rev":"1-123"}]}'
     [{"ok":true,"id":"foobar","rev":"2-9af8fdb20d72f07cc94e26da4857e3ae"}]

however, Cloudant / CouchDB 2.0 correctly return an error:

     $ curl 'http://127.0.0.1:15984/revtest' -XPUT
     {"ok":true}
     
     $ curl 'http://127.0.0.1:15984/revtest2/_bulk_docs' -H 'Content-Type: application/json' --data-binary     '{"docs":[{"_id":"foobar","_rev":"1-123"}]}'
     [{"id":"foobar","error":"conflict","reason":"Document update conflict."}]

@janl / @kxepal is the CouchDB 2.0 behaviour expected / correct?

willholley:
This also occurs with single doc insertions:

     var db = new PouchDB("test");
     db.put({_id:"foo",_rev:"1-bar"}).then(function(results) { console.log(results);})

returns:

     {ok: true, id: "foo", rev: "2-fb61ce36c134030c8ad3923c259cd60f"}

kxepal:
cc @rnewson here as well, but for me new 2.0 behaviour is correct and the right.

daleharvey:
Fixed by https://github.com/pouchdb/pouchdb/commit/70271ec769c6bc93909ee315f98ec1ac72ab1862

PR #4713, opened 2015-12-31
(#4712) - generate conflict error when using inserting doc with unrecognised _rev

willholley:
When inserting a new document with a specific _rev, we previously
allowed the insertion instead of generating a conflict error
(as CouchDB does).

It was a bit fiddly to figure out where to add this check but I settled
on processDocs.js. This commit adds a test that when inserting a document
new to the database (and new_edits is true), the root of the revision tree
must not be missing. Normally, PouchDB will generate a new, available
root node so the lack of this indicates that the user specified a _rev
that the database doesn't recognise.

Updates to a document which the database already knows about will not
flow down this code path - instead they go to updateDoc.js which performs
its own conflict test.

### DIFF
--- a/lib/deps/docs/processDocs.js
+++ b/lib/deps/docs/processDocs.js
@@ -15,12 +15,25 @@ function processDocs(revLimit, docInfos, api, fetchedDocs, tx, results,
   // Default to 1000 locally
   revLimit = revLimit || 1000;
 
+  function rootIsMissing(docInfo) {
+    return docInfo.metadata.rev_tree[0].ids[1].status === "missing";
+  }
+
   function insertDoc(docInfo, resultsIdx, callback) {
     // Cant insert new deleted documents
     var winningRev = calculateWinningRev(docInfo.metadata);
     var deleted = isDeleted(docInfo.metadata, winningRev);
     if ('was_delete' in opts && deleted) {
       results[resultsIdx] = errors.error(errors.MISSING_DOC, 'deleted');
+      return callback();
+    }
+
+    // 4712 - detect whether a new document was inserted with a _rev
+    var inConflict = newEdits && rootIsMissing(docInfo);
+
+    if (inConflict) {
+      var err = errors.error(errors.REV_CONFLICT);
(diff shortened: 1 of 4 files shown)


tests/integration/test.basics.js
```diff
@@ -1064,6 +1064,26 @@ adapters.forEach(function (adapter) {
...
+      var newdoc = {
+        '_id': 'foobar',
+        '_rev': '1-123'
+      };
+
+      db.put(newdoc, function (err, result) {
+        should.not.exist(result);
+        err.should.have.property('name', 'conflict');
+        done();
+      });
```
nolanlawson: let's please stop introducing new tests that use callbacks instead of promises

lib/deps/docs/processDocs.js
```diff
@@ -15,6 +15,10 @@ function processDocs(revLimit, docInfos, api, fetchedDocs, tx, results,
   // Default to 1000 locally
   revLimit = revLimit || 1000;
 
+  function rootIsMissing(docInfo) {
+    return docInfo.metadata.rev_tree[0].ids[1].status === "missing";
+  }
+
```
nolanlawson: this would be better for performance to hoist to the outer function, since it doesn't need any local variables

tests/integration/test.bulk_docs.js
```diff
@@ -996,6 +996,23 @@ adapters.forEach(function (adapter) {
...
+      }
+
+      var db = new PouchDB(dbs.name);
+      var newdoc = {
+        '_id': 'foobar',
+        '_rev': '1-123'
+      };
 
+      db.bulkDocs({ docs: [newdoc] }, function (err, results) {
+        results[0].should.have.property('name', 'conflict');
```
nolanlawson: typically we test `status` instead of `name`, but this is fine

nolanlawson:
I have some nits, but LGTM. I may wait for #4710 because it will be easier to rebase this on top of that than vice-versa.

nolanlawson:
Closing in favor of #4714

PR #4715, opened 2016-01-02
(#4715) - simpler/faster `npm run build`

nolanlawson:
`npm run build` is now responsible for building everything
(including the performance bundle and adding the license comments).
The script is also called from both `npm run build` and `npm run dev`.

Additionally, it runs faster due to efficient parallelization
with subprocesses. On my MacBook Air, it cut the build time from
0m37.793s to 0m19.963s.


nickcolley:
This looks great, will let someone else :+1: too as it's a big diff.

Looks like there's a bug somewhere as it's not passing in CI.

nolanlawson:
Yeah I messed up the order of build operations. Should be fixed now ^

nolanlawson:
Ugh, something weird with the Travis setup. I can't reproduce the failures.

nolanlawson:
OK, looks like I need to limit the number of concurrent tasks per # of CPUs, because the Travis build environment doesn't have very many.

nolanlawson:
Also, we were missing tests for the `require('pouchdb/[fruitdown|memory|localstorage]')`, so I added them.

nolanlawson:
This is a Travis-only failure, and it's driving me nuts.

nolanlawson:
Decided to remove all my overambitious parallelization optimizations and just do everything sequentially. Incidentally it's still completing in 21s.

nolanlawson:
Looks this is failing only because the `leveldown` prebuilt download is failing on Node stable...

```
Error: Cannot find module 'inherits'
    at Function.Module._resolveFilename (module.js:326:15)
    at Function.Module._load (module.js:277:25)
    at Module.require (module.js:354:17)
    at require (internal/module.js:12:17)
    at Object.<anonymous> (/home/travis/build/pouchdb/pouchdb/node_modules/are-we-there-yet/node_modules/readable-stream/lib/_stream_readable.js:45:17)
    at Module._compile (module.js:398:26)
    at Object.Module._extensions..js (module.js:405:10)
    at Module.load (module.js:344:32)
    at Function.Module._load (module.js:301:12)
    at Module.require (module.js:354:17)
npm WARN install:leveldown@1.4.2 leveldown@1.4.2 install: `prebuild --download`
npm WARN install:leveldown@1.4.2 Exit status 1
```

nolanlawson:
Sigh, looks like I'm going to have to break this up to figure out why *only Node stable* is failing with an inscrutable error message.

Issue #2594, opened 2014-08-13
API sidebar can overlap the kangaroo face


nolanlawson:
Steps to reproduce: scroll all the way to the bottom of the page, then quickly scroll all the way to the top. Seen in Chrome and Firefox.

Ping @nickcolley 

![screenshot 2014-08-12 23 36 21](https://cloud.githubusercontent.com/assets/283842/3900988/24af6fbe-229b-11e4-9be1-f17aa21eeeb4.png)

daleharvey:
@nickcolley Where abouts is the code that gives this the fixed position etc? seeing this be buggy quite a lot, fairly certain it can use `position: sticky` to work better, see the sidebar in github with labels on the right of this for an example

nickcolley:
https://github.com/pouchdb/pouchdb/blob/master/docs/_layouts/2ColLeft.html#L9

http://getbootstrap.com/javascript/#affix

Agree we should ditch the bootstrap plugin, I like the position sticky idea.

daedlock:
```
$(window).scroll(function() {
    if ($(this).scrollTop() == 0) {
      $("#sidebar > div").addClass("affix-top");
    }
});
```

How about that as a quick hack?

daleharvey:
It also hides itself when you scroll to the bottom and scroll half way up, its very broken and would rather fix it than work around it, also doing dom access / manipulation inside a scroll handler is a big cause of performance issues and best avoided 

Githubs sticky nav doesnt work in chrome, I think its fine we just implement it as position:sticky and if there is a polyfill that works nicely we can use it

daedlock:
I like this one https://github.com/wilddeer/stickyfill

nickcolley:
:+1: for using `position: sticky`.

As for a polyfill, as long as it's not too big I think it's a good idea.

@daedlock do you want to have a go doing a PR? Can help out if you do! :)

daleharvey:
Fixed in https://github.com/nickcolley/pouchdb/commit/709acfaf6c795c50fb8a32e70dc14dc868a439d0

PR #4717, opened 2016-01-02
(#2594) - Use position: sticky instead of bootstrap affix

nickcolley:
[skip ci]


nickcolley:
I was going to leave the polyfill out but without the sticky behaviour the nav is difficult to use so I think it's worth it in this case.

daleharvey:
:+1: this feels a ton better, merged in https://github.com/nickcolley/pouchdb/commit/709acfaf6c795c50fb8a32e70dc14dc868a439d0, thanks

PR #4720, opened 2016-01-02
(#4721) - http tests work offline

nolanlawson:
These tests fail when you're offline. It's not safe
to assume the name of the error that you get back.


daleharvey:
:+1: https://github.com/pouchdb/pouchdb/commit/5dc9350535587939f0cf87de38b7f98f9317e5fd

PR #4721, opened 2016-01-02
(#4721) - improve code coverage, functions at 100%

nolanlawson:
Here is the coverage after this fix:

```
Statements   : 100% ( 4954/4954 ), 162 ignored
Branches     : 95.53% ( 2328/2437 ), 124 ignored
Functions    : 100% ( 836/836 ), 20 ignored
Lines        : 100% ( 4899/4899 )
```

We are also no longer using the Babel `es2012` preset, because
we don't need it to simulate Rollup - we just need two small
plugins that convert ES6 modules to CJS.

### DIFF
--- a/bin/build-as-modular-es5.sh
+++ b/bin/build-as-modular-es5.sh
@@ -14,8 +14,7 @@ npm run build
 # that doesn't oblige use to do require('foo').default for every module.
 
 ./node_modules/.bin/babel \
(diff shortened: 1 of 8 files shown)


daleharvey:
Very nice - https://github.com/pouchdb/pouchdb/commit/ea33f785485db22a787578da975ef628c12ab162

PR #4724, opened 2016-01-02
"views" in design documents take a map function

fivetanley:
Before, the above code caused couchDB to return an error http status code.

This is because the example map function was not in the "map" key in the design
document. Therefore Couch saved a view without a map function, aka the
`views` property when querying couch for the document showed just an empty
object.

### DIFF
--- a/docs/_includes/api/replication.html
+++ b/docs/_includes/api/replication.html
@@ -261,11 +261,13 @@ First `put()` a design document in the remote database:
 {
   _id: '_design/mydesign',
   views: {
-    myview: function (doc) {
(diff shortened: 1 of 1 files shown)


nolanlawson:
Thanks for the PR! This is a classic *ouchDB newb mistake, but it was probably me who put that error in the docs! :stuck_out_tongue: 

I canceled the build, since this is just a doc change and we have lots of open PRs. I also amended the PR to match our code style. Thanks for submitting!

nolanlawson:
b670a992eacae2559ef5a48749d74c241b184d4e

Issue #4494, opened 2015-10-21
Add more users to the "who's using us" page


nolanlawson:
Known projects/products:

* Plunkr 2 (unreleased) https://twitter.com/plnkrco/status/641436637512245249
* Drupal 8 Relaxed Web Services: https://www.drupal.org/project/relaxed
* Standard Analytics science.ai platform: http://www.standardanalytics.io/

Any others we're missing? I feel like the list hasn't been updated in awhile.

marten-de-vries:
I stumbled upon http://delta.octavore.com/ today, which lists PouchDB in its credits section.

daleharvey:
hospital.io

nolanlawson:
https://pokedex.org ;)

nolanlawson:
fixed by bd64d75d8e397e5199ee26bd0be424747929d0c6

PR #4725, opened 2016-01-02
(#4494) - Add some more users to PouchDB who's using page

nickcolley:
Re-orders so they're alphabetical order to remove bias.

Adds the following:
- Pokedex.org
- HospitalRun
- Delta

From @nolanlawson 's suggestions:

- Plunkr 2 - Not sure this is shipped so didn't include it yet.
- Drupal 8 Relaxed Web Services - Looked like they're boasting compatibility but it's not actually part of the tech stack?
- Standard Analytics science.ai platform - No evidence of usage so not sure how interesting or credible this is?


nolanlawson:
@nickcolley I learned about Standard Analytics' use of PouchDB from some folks who work there. Also they mentioned it in job postings (may not be available anymore though).

nolanlawson:
Halting the build because I don't think it needs a new build. Anyway, +1, will merge. :)

nolanlawson:
Fixed by bd64d75d8e397e5199ee26bd0be424747929d0c6. I also appreciate the alphabetical ordering of users (more fair). Thanks!

PR #4726, opened 2016-01-03
Website typography

nickcolley:
This has been bugging me for a while.

This PR addresses accessibility and performance for our users, refactoring for us as maintainers:
- Body font size is now 16px instead of 14px, headings are relative to this, they're better balanced now and should be easier to read, I've also adjusted the vertical rhythm according to this change.
- I've removed the two Google web fonts we were using 'Lato' and 'Open Sans' this will make initial render faster as the page will not be blocked waiting for the sizeable downloads to finish.
- Check the commits for refactoring etc

In terms of performance, it's actually a good improvement, (if you're interested) i've ran an audit on the current live site to demonstrate this:
http://www.webpagetest.org/result/160103_W1_3WP/

[image](http://www.webpagetest.org/video/compare.php?tests=160103_W1_3WP-r%3A1-c%3A0&thumbSize=200&ival=100&end=doc)

Notice how we reach first paint at 5.2s but only actually see text at 5.6s, (this is on a slow cellular connection in India) this would be better or worse depending on connection.

Using system fonts we'd always see content at the 5.2s mark, not having to wait for the fonts to download.

This is WIP, I want to review spacing and heading sizes at this point but wanted to get this up incase anyone wants to review at this stage :+1:

Will rebase when we get close to merging.

(Open these up on their own to see the correct resolution)

type-updates

type-updates-guide


nolanlawson:
:+1: this is a LOT more legible and I appreciate the speed boosts as well. BTW feel free to commit with `[skip ci]`, since website changes don't need to trigger a full Travis build.

Just let me know when you feel it's done, and I'll be happy to review/merge. :)

daleharvey:
I think the improvements to the spacing and sizing are great, however the default fonts really do not look great (in particular looking at `Guides` in the navigation). Wondering if we could improve this by loading system fonts immediately and switching to webfonts when they are loaded? (https://www.filamentgroup.com/lab/font-events.html)

nickcolley:
Can you show me a screenshot of what you're seeing? The navigation looks clear to me.

In terms of lazy-loading the fonts, it comes with many pitfalls so I'd rather avoid it if possible.

daleharvey:
I was picking 'Guides' out specifically just from the screenshots, it looks less bad on the live site for me but it seems fairly clear how much better Open Sans looks here than Helvetica to me. Given this is only an issue for first time page load I would like to have a chance to see what the tradeoff is with lazy loading, Ill try a patch for it

nickcolley:
I think the performance improvements are more important anything subjective, if you can say why it will be worse for the user that'd be a good start, otherwise I can't accept a patch based on 'less bad' if that makes sense?

nickcolley:
I'm not going to do any more changes on this branch, I think I'm happy for now.

daleharvey:
So I do not agree that objective improvements automatically trump subjective ones, typography is a huge part of design and at least to me the new type is a fairly is a pretty large regression. In particular the "Introduction to PouchDB" and the already mentioned navigation menu as well as the general kerning and aliasing. 

If this were a large and constant performance improvement for all users / page loads it would be pretty big, but its a .4 seconds improvement (absolute maximum is 3 seconds) on a slow connection that will only ever apply to the first page load and that makes it quite a minor improvement with significant drawbacks.

My concern is only with the font face change, how about we merge the spacing improvements + refactoring which are great changes and we can discuss the font-face change on its own? (ie see if https://www.filamentgroup.com/lab/font-events.html makes reasonable performance improvements or look at alternative solutions)

nolanlawson:
I'm happy to defer to @daleharvey if he has strong feelings on this; I am definitely not an expert on typography, although it does seem vaguely like the new version is more legible to me.

nickcolley:
I agree with @nolanlawson, wonder if this is specific to how it's rendering on specific OS/Browser.

I still believe that our users do not care about kerning but rather how fast the page loads, this does not mean I think typography is not important I just place performance above minor type differences.

That said, willing to split this up however the spacing adjustments I made included the change in font so I'll have to do this again with the previous font.

zaccolley:
I think you can argue both ways.

As this is a site you'll be spending a lot of time reading initially (such as the guides) you want that experience to be good so a better typeface is ideal.

For users who are quickly checking the docs as reference the performance might be helpful, but then wouldn't they have the web fonts cached? Perhaps they're on a train? 🚂 

Without some structured user testing it's mostly assumptions though.

zaccolley:
To be clear, in the first point I don't think it's cut and dry to assume that Lato(?) is more legible than Helvetica.

If it comes down to aesthetics and not readability does it matter?

tinyhippw:
Given the changes are so minor to the overall design, layout and legibility, and imo actually improve things the increase in performance for such little adverse affect elsewhere is pretty incredible, so +1

jrichmond5:
+1 Big improvement IMO

kephail:
I have to say I agree with @nickcolley on this one, I think the performance increase trumps the negligible difference the font makes to the overall design of the page.

ermouth:
@nickcolley Arial in Windows looks definitely awful, imo Segoe UI is way better. 

Also I couldn‘t reproduce those visually disproportionate (may I say ugly) `ncuaBD` glyphs applying Helvetica under both Mac OS and iOS. Whats OS did you use making screenshots? Mb there exist more visually nice and better kerned/hinted system font in that kind of systems?

nickcolley:
@ermouth I'm running Linux which has awful font rendering compared with other operating systems, you'll see some changes depending on what OS you're using for sure.

Issue #4728, opened 2016-01-03
Test node stable by default


daleharvey:
It seems sensible for stable to be the default, of course still test the older versions

nolanlawson:
fixed by https://github.com/pouchdb/pouchdb/pull/4729

PR #4729, opened 2016-01-03
(#4728) - Test node stable by default

daleharvey:


daleharvey:
I did a big run @ https://travis-ci.org/daleharvey/pouchdb/builds/99940530, no more intermittent than master

nolanlawson:
:+1: when green, makes sense to me

PR #4731, opened 2016-01-03
(#4731) - remove lib/dist before publish

nolanlawson:
This is the safest thing to do, because other operations
(such as `run build-as-modular-es5`) might leave messy
stuff in the lib/ or dist/ directory.


nolanlawson:
Also the `npm run build` itself doesn't delete those directories, because if it did, it would mess up the `npm run dev` script, which is trying to watch files.

nolanlawson:
Actually wait, I think that's not true anymore. One sec, will try another PR.

Issue #4733, opened 2016-01-03
5.2.0


nolanlawson:
#4715 and #4732 are the last issues I'd like to get in before this release (fixes a bug in the plugins and requiring from npm via git). But I think we can do a release this week or next week.

nolanlawson:
OK, the only thing blocking this release now IMO is #4735. #4741 may not be ready in time; seems to be a subtle bug in node/npm.

daleharvey:
Awesome, I think its a good idea to roll in the rollup stuff with the release posts, its a little more incentive for people to read about the releases (ie come for the information about the rollup, stay for the release).

I think after such a big change it would be a good idea to aim to let the codebase settle a little bit and maybe aim to release on monday? everything is looking settled with the change but it gives time for people to come back from the holidays and a little more time for any subtle issues brought up with the switch to rollup to be noticed.

nolanlawson:
@daleharvey sounds good to me. I want time this weekend to speed up the build script anyway. I think I can just remove some deps added to `package.json` and work around the npm bug.

bbenezech:
@nolanlawson can you consider merging also #4753 ?

nolanlawson:
fixed by ef22e2eaca6795173151a29ff2f72c50f6ff130b

nolanlawson:
@bbenezech Sorry, it didn't make the cut, but we can always quickly follow up with a 5.2.1! :)

PR #4734, opened 2016-01-03
(#4733) - 5.2.0 blog post and version updates

nolanlawson:
Making Rollup the star of the show for this blog post,
if only because I think the reason we're using it is interesting.

I'm seeing lots of library authors complaining about Webpack/Browserify
differences. I think it's only going to get worse, so it's worth
blogging about.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "pouchdb",
-  "version": "5.1.1-prerelease",
+  "version": "5.2.1-prerelease",
   "description": "PouchDB is a pocket-sized database.",
   "repo": "daleharvey/pouchdb",
   "keywords": [
--- a/component.json
+++ b/component.json
@@ -1,6 +1,6 @@
 {
   "name": "pouchdb",
-  "version": "5.1.1-prerelease",
+  "version": "5.2.1-prerelease",
   "description": "PouchDB is a pocket-sized database.",
   "repo": "daleharvey/pouchdb",
   "keywords": [
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -4,7 +4,7 @@ url: http://pouchdb.com
 highlighter: pygments
 markdown: redcarpet
 baseurl:
-version: 5.1.0
+version: 5.2.0
 paginate: 5
 paginate_path: "blog/page:num"
 github:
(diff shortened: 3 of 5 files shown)


docs/_posts/2016-01-04-pouchdb-5.2.0-magical-mystery-pouch.md
```diff
@@ -0,0 +1,102 @@
...
+
+title: PouchDB 5.2.0&#58; Magical mystery pouch
+author: Nolan Lawson
+
+---
+
+> "Rollup! Rollup for the mystery tour!"
+> - Lennon and McCartney
+
+Besides a slew of bugfixes and the new `revs_limit` feature, the big news about 5.2.0 is that PouchDB is now authored in ES6 modules rather than CommonJS. Using [Rollup](http://rollupjs.org/), it's transpiled into one big `index.js` for Node and an `index-browser.js` for Webpack/Browserify.
```
nickcolley: could this be "with ES6 modules" using "in" makes it read as if PouchDB is written in ES6 now almost?

nickcolley:
Wow, really great stuff about Rollup, do you think it'd make more sense to have it as a separate blog post that is linked to?

docs/_posts/2016-01-04-pouchdb-5.2.0-magical-mystery-pouch.md
nolanlawson: Yeah you're right, sounds better that way.

nolanlawson:
@nickcolley I considered that, but I thought it would be more economical to make one post. I couldn't think of what to say for "just" the release blog post.

Maybe we could put the changelog at the end?

nolanlawson:
Or yeah, if somebody wants to write "just" the changelog post, then we could make this a separate post entirely. (Maybe with the title "PouchDB and JavaScript build systems")

nickcolley:
@daleharvey mentioned some good points on why to keep them together here https://github.com/pouchdb/pouchdb/issues/4733#issuecomment-168679749

I feel it's a tad clunky having them together but agree with Dale too so I'll leave it to you! :+1:

nolanlawson:
:+1: I'm fine with keeping it as one big blog post

nickcolley:
Let's do it! :)

nolanlawson:
Made the suggested change. Also changed the title.

nolanlawson:
ef22e2eaca6795173151a29ff2f72c50f6ff130b

Issue #4732, opened 2016-01-03
postinstall task to fix people who depend on pouchdb master


nolanlawson:
Since we're now transpiling our code, we'll need to do the build as a `postinstall` in order not to break people who add:

```
"pouchdb": "pouchdb/pouchdb"
```

to their `package.json`.

nolanlawson:
Or `prepublish` i guess: https://docs.npmjs.com/misc/scripts

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/2be37213b03c5d5da1d79bf8b714dd8fba8c09d1

PR #4735, opened 2016-01-03
(#4732) - add prepublish script

nolanlawson:
Not sure if this will work, so opening up a PR
to test.


.travis.yml
```diff
@@ -33,8 +33,6 @@ before_script:
     # Fail early so we dont run hours of saucelabs if we know there
     # is a lint failure
   - npm run jshint
-    # build before each test
-  - npm run build
```
nolanlawson: Travis should do this automatically after `npm install`.

nolanlawson:
Not working, so gonna fold this into #4715

PR #4736, opened 2016-01-04
DON'T MERGE - test node on travis

nolanlawson:


nolanlawson:
Crap, this means #4715 breaks Node stable for reasons I don't understand. Works in other versions of Node. :(

Issue #4738, opened 2016-01-04
plugins are broken


nolanlawson:
In current master, it's broken if you try to do `require('pouchdb/extras/fruitdown')` because we failed to test it in CI.

daleharvey:
Verified working in https://github.com/pouchdb/pouchdb/commit/3b1723de5bfda66df502b628df62c33b64579f2e

mciparelli:
In 5.2.0 it looks like it's trying to find `../lib/plugins/localstorage` but the only folder with the plugins inside `lib` is called `extras`

nolanlawson:
I can confirm we screwed this up in 5.2.0. We should probably release a 5.2.1 right away. No idea how this slipped through.

nolanlawson:
Yeah, okay, the `require.resolve()` test doesn't test anything, other than that Node can find the file. Unfortunately it seems we can't test it without just loading all the code

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/011a6f2ecb002085e9491a89af5bcf0e17226285

PR #4739, opened 2016-01-04
(#4738) - test plugin APIs exported correctly

nolanlawson:
Turns out I was wrong that this regressed, but I'm
adding a test so we can be sure about it.


daleharvey:
:+1: looks great thanks https://github.com/pouchdb/pouchdb/commit/3b1723de5bfda66df502b628df62c33b64579f2e

PR #4740, opened 2016-01-04
(#4732) - add `npm run build` as prepublish

nolanlawson:
This allows people to use the `pouchdb/pouchdb` version
of PouchDB. It will build it automatically for them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,8 +33,6 @@ before_script:
     # Fail early so we dont run hours of saucelabs if we know there
     # is a lint failure
   - npm run jshint
-    # build before each test
-  - npm run build
 
 script: npm run $COMMAND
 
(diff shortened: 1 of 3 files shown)


daleharvey:
Just so I understand, this change means `lib` will be checked in as we publish and users requiring pouchdb wont need to build anything?

nolanlawson:
Yeah, so the way `prepublish` works is this:

* If you check out the PouchDB code itself and `npm install`, it will run
* If you `npm link path/to/pouchdb` from another project, it will run
* If you `npm install pouchdb@pouchdb/pouchdb` (i.e. from Github), it will run

However, anybody using PouchDB normally (i.e. from an npm version) will not have to run the build script.

daleharvey:
ah ok sweet, :+1: then, thanks

daleharvey:
Merged in https://github.com/pouchdb/pouchdb/commit/2be37213b03c5d5da1d79bf8b714dd8fba8c09d1

Issue #4627, opened 2015-12-02
change event with 2 way replication issue + bugfix


mickael-kerjean:
Hello there,

First of all, thank you for having created such a cool and usefull project!

As I was having some fun with some code, I went into some issues with the live replication option.
The change event is quite powerfull to plug some application code that can refresh part of screen
depending on the type of document that is received.

With two way sync enable, Everytime some data is push, I receive a push then a pull.
The issue being the docs received while pulling has already been received before.

As a piece of code worth more than just plain text, here is an [example](http://jsfiddle.net/mikeulkeul/0dhft2pa/). The more we insert data, the worst it becomes: see this [other example](http://jsfiddle.net/mikeulkeul/zpc6qjx8/) when trying to insert way more documents.

As it was the perfect opportunity to browse some of pouchdb code, I
drilled down and found where my problem was coming: [/lib/replicate/replicate.js](https://github.com/pouchdb/pouchdb/blob/master/lib/replicate/replicate.js):
1) [Here](https://github.com/pouchdb/pouchdb/blob/8132c76b0a79b75a1ca7f19d1b85943869f6e47d/lib/replicate/replicate.js#L170), we can see how the batch is processed. The interesting functions are [writeDocs](https://github.com/pouchdb/pouchdb/blob/master/lib/replicate/replicate.js#L58) and [finishBatch](https://github.com/pouchdb/pouchdb/blob/master/lib/replicate/replicate.js#L105)
2) the [finishBatch](https://github.com/pouchdb/pouchdb/blob/master/lib/replicate/replicate.js#L105) function trigger the [actuel change event](https://github.com/pouchdb/pouchdb/blob/master/lib/replicate/replicate.js#L110) given a change object that [has the docs field](https://github.com/pouchdb/pouchdb/blob/master/lib/replicate/replicate.js#L109) I was interested in, this field being populated with the changedDocs variable.
3) [writeDocs](https://github.com/pouchdb/pouchdb/blob/master/lib/replicate/replicate.js#L58) is currently resetting the changedDocs array but not [in
every case](https://github.com/pouchdb/pouchdb/blob/master/lib/replicate/replicate.js#L59) which is causing the issue I was experiencing

In other words, for my use case I had to update the original writeDocs
function to become: https://gist.github.com/mickael-kerjean/a6fdda69372ace248b55

That fairly basic but it is now working as expected by my application.

Long life to pouchdb!

nolanlawson:
[...]

PR #4746, opened 2016-01-06
(#4627) - fix duplicate change events when using bi-directional live replication

willholley:
Added a test / fix as proposed by @mickael-kerjean. This test sets up a 2-way live replication which initially transfers documents from a remote to a local database. Once the initial sync is complete, it inserts a document locally - the change should propagate to the remote database and then back to the
local database via the live replications. Previously, this caused "change" events for existing documents to fire. This patch prevents the event duplication.

### DIFF
--- a/tests/integration/test.replication_events.js
+++ b/tests/integration/test.replication_events.js
@@ -172,69 +172,87 @@ adapters.forEach(function (adapters) {
 
 
     // this test sets up a 2 way replication which initially transfers
-    // documents from a remote to a local database. Once the initial
-    // sync is complete, it inserts documents locally - the changes
+    // documents from a remote to a local database.
+    // At the same time, we insert documents locally - the changes
     // should propagate to the remote database and then back to the
-    // local database via the live replications. However, we do not
-    // expect any "change" events to fire for these documents since
-    // they already exist locally.
-    it('#4627 Test no duplicate changes in live replication', function () {
-      return new PouchDB.utils.Promise(function (resolve, reject) {
-        var db = new PouchDB(dbs.name);
-        var remote = new PouchDB(dbs.remote);
-        var docId = -1;
-        var docsToGenerate = 10;
-        var lastChange = -1;
+    // local database via the live replications.
+    // Previously, this test resulted in 'change' events being
+    // generated for already-replicated documents. When PouchDB is working
+    // as expected, each remote document should be passed to a
+    // change event exactly once (though a change might contain multiple docs)
+    it('#4627 Test no duplicate changes in live replication', function (done) {
+      var db = new PouchDB(dbs.name);
+      var remote = new PouchDB(dbs.remote);
+      var docId = -1;
+      var docsToGenerate = 10;
+      var lastChange = -1;
+      var firstReplication;
+      var secondReplication;
+      var completeCalls = 0;
 
-        function generateDocs(n) {
-          return Array.apply(null, new Array(n)).map(function (e, i) {
-            docId += 1;
-            return {
-              _id: docId.toString(),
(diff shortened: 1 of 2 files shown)


tests/integration/test.replication_events.js
```diff
@@ -170,5 +170,71 @@ adapters.forEach(function (adapters) {
...
+            retry: true,
+            since: 0
+          }).on("change", function (feed) {
+            // attempt to detect changes loop
+            var ids = feed.docs.map(function (d) {
+              return parseInt(d._id, 10);
+            }).sort();
+
+            var firstChange = ids[0];
+            if (firstChange <= lastChange)
```
daleharvey: nit: our style is same line bracket

    if (firstChange <= lastChange) {

tests/integration/test.replication_events.js
```diff
@@ -170,5 +170,71 @@ adapters.forEach(function (adapters) {
...
+            live: true,
+            retry: true,
+            since: 0
+          });
+
+          var replicationHandler = remote.replicate.to(db, {
+            live: true,
+            retry: true,
+            since: 0
+          }).on("change", function (feed) {
```
daleharvey: nit: single quotes, `change`

tests/integration/test.replication_events.js
```diff
@@ -170,5 +170,71 @@ adapters.forEach(function (adapters) {
...
+
+    // this test sets up a 2 way replication which initially transfers
+    // documents from a remote to a local database. Once the initial
+    // sync is complete, it inserts documents locally - the changes
+    // should propagate to the remote database and then back to the
+    // local database via the live replications. However, we do not
+    // expect any "change" events to fire for these documents since
+    // they already exist locally.
+    it('#4627 Test no duplicate changes in live replication', function () {
+      return new PouchDB.utils.Promise(function (resolve, reject) {
```
daleharvey: Replication tests can very rarely use promises, in the other patch @nolanlawson was mainly referring to `db.post(obj, function`, for this test it looks cleaner to use `done()` instead of returning a promise

tests/integration/test.replication_events.js
```diff
@@ -170,5 +170,71 @@ adapters.forEach(function (adapters) {
...
+            };
+          });
+        }
+
+        remote.bulkDocs(generateDocs(docsToGenerate)).then(function () {
+          db.replicate.to(remote, {
+            live: true,
+            retry: true,
+            since: 0
+          });
```
daleharvey: So this is very tricky, I wrote a post about it @ http://pouchdb.com/2014/12/11/the-pains-of-being-async-at-heart.html (the last section), we can only finish this test once we are absolutely certain that both this and the previous replications are complete, its pretty much impossible to cleanup if not and it will lead to hard to detect intermittent failures. Once you are happy the test is complete then you need to `.cancel()` both the replications then call done when they have both fired their `complete` event handler, https://github.com/pouchdb/pouchdb/blob/master/tests/integration/test.replication.js#L534 may give you some help with that

daleharvey:
Awesome thanks for this, had a couple of notes on the test but they are mostly style issues, the test itself looks good, congrats on about the only valid reason to use a settimeout in a test :)

daleharvey:
Awesome almost there thanks for that, the problem now is the test will call `done()` 3 times, once inside the change event and twice on the replications completing. The easiest way to fix that is to have `complete` check and only call `done()` on the 3rd invocation.

daleharvey:
Awesome thanks so much for all the edits, :+1:  - https://github.com/pouchdb/pouchdb/commit/c905239093ed6e7c44b515e851b8cf7987110360

PR #4747, opened 2016-01-06
(#2035) - Remove cors proxy

daleharvey:


daleharvey:
Mostly just hoping something has changed and this just works now, but we were down to one platform failing so should be able to get it finally landed

daleharvey:
Oddly enough this is fixing pouchdb-server?

daleharvey:
and phantom, just broke ios, so weird

daleharvey:
`SAUCE_USERNAME=pouchdb SAUCE_ACCESS_KEY=xxx SKIP_MIGRATION=true CLIENT="saucelabs:iphone:8.1:OS X 10.10" npm test`

ie the exact same thing travis is doing works fine on my machine, I also used `add-cors-to-couchdb` to setup cors, the only error indication I see in any logs is `ThrowerFunction@[native code]`

daleharvey:
ok getting somewhere, starting to see this locally and as far as I can tell its not hitting the server at all (even the OPTIONS)

daleharvey:
So to explain this issue, currently we will open a browser to http://127.0.0.8000/ on saucelabs, that browser will make a request to http://127.0.0.1:2020 via ajax, that request is tunneled from saucelabs to the host (usually travis), on the host we have a reverse proxy from http://127.0.0.2020 -> http://127.0.0.1:5984, that reverse proxy adds cors requests.

We dont need a proxy to give us cors headers any more, CouchDB does that for us, so we change the requests from http://127.0.0.1:2020 to http://127.0.0.1:5984, doing this every one of our tests pass except ios. That would suggest the network and the tunnel are fine so likely it is an issue with the cors headers sent back by CouchDB. however CouchDB never receives the request at all and given the most basic xhr code

```
  var xhr = new XMLHttpRequest();
  xhr.open('GET', 'http://127.0.0.1:2020/');
  xhr.onreadystatechange = function () {
    if (xhr.readyState === 4) {
      window.results.log = 'GOT DATA: ' + xhr.status + ' - ' +
        xhr.response + ' - ' + xhr.responseText;
      window.results.completed = true;
    }
  };
  xhr.send();
```

ios will not make a connection with the server and fail with statusCode === 0

daleharvey:
ok possibly found the issue, when I change my couchdb to run on 2020 it makes the request fine, 5984 / 5985 / 2021 do not, I believe sauce connect has a detault set of ports it will proxy (https://github.com/saucelabs/sauce_ruby#with-sauce-connect) and if we arent using those ports ...

The confusing part is why this is not consistent across our builds (why android, chrome, safari etc all proxy fine), but maybe we can just switch to 2020 in travis and not care

daleharvey:
Only regressing test was performance tests in the browser, fixed now, woot

daleharvey:
Forgot to actually add the performance test fix, should be all good now

nolanlawson:
Nice!!!! Big :+1: from me, can't believe you finally figured this out. Weird Saucelabs bug, but I am beyond happy to finally get rid of the CORS proxy.

Small nit: this needs a change to `TESTING.md` to remove references to the CORS proxy and mention that you need to run `add-cors-to-couchdb`.

daleharvey:
Awesome, I was trying to keep track of ammendments and missed this one, will do a nit follow up - https://github.com/pouchdb/pouchdb/commit/2f7b57baca50e893a61c33f7de15835796f083a6

PR #4753, opened 2016-01-07
Fix for triple request with replication

bbenezech:
_edited after some sleep_

see below.
ref #3824

### DIFF
--- a/src/adapters/http/index.js
+++ b/src/adapters/http/index.js
@@ -875,12 +875,9 @@ function HttpPouch(opts, callback) {
             if (returnDocs) {
               results.results.push(c);
             }
+            opts.onChange(c);
           }
           return ret;
-        });
-
-        res.results.forEach(function (c, index) {
-          opts.onChange(c, index === (res.results.length - 1));
         });
       } else if (err) {
         // In case of an error, stop listening for changes and call
--- a/src/adapters/http/index.js
+++ b/src/adapters/http/index.js
@@ -875,9 +875,12 @@ function HttpPouch(opts, callback) {
             if (returnDocs) {
               results.results.push(c);
             }
-            opts.onChange(c);
           }
           return ret;
+        });
+
+        res.results.forEach(function (c, index) {
+          opts.onChange(c, index === (res.results.length - 1));
         });
       } else if (err) {
         // In case of an error, stop listening for changes and call
--- a/src/changes.js
+++ b/src/changes.js
@@ -46,12 +46,12 @@ function Changes(db, opts, callback) {
   }
   db.once('destroyed', onDestroy);
 
-  opts.onChange = function (change, flush) {
+  opts.onChange = function (change) {
     /* istanbul ignore if */
     if (opts.isCancelled) {
       return;
     }
-    self.emit('change', change, flush);
+    self.emit('change', change);
     if (self.startSeq && self.startSeq <= change.seq) {
       self.startSeq = false;
     }
--- a/src/changes.js
+++ b/src/changes.js
@@ -46,12 +46,12 @@ function Changes(db, opts, callback) {
   }
   db.once('destroyed', onDestroy);
 
-  opts.onChange = function (change) {
+  opts.onChange = function (change, flush) {
     /* istanbul ignore if */
     if (opts.isCancelled) {
       return;
     }
-    self.emit('change', change);
+    self.emit('change', change, flush);
     if (self.startSeq && self.startSeq <= change.seq) {
       self.startSeq = false;
     }
--- a/src/replicate/replicate.js
+++ b/src/replicate/replicate.js
@@ -264,7 +264,7 @@ function replicate(src, target, opts, returnValue, result) {
   }
 
 
-  function onChange(change) {
+  function onChange(change, flush) {
     if (returnValue.cancelled) {
       return completeReplication();
     }
@@ -274,7 +274,7 @@ function replicate(src, target, opts, returnValue, result) {
     }
     pendingBatch.seq = change.seq;
     pendingBatch.changes.push(change);
-    processPendingBatch(batches.length === 0);
+    processPendingBatch(flush !== undefined ? flush : batches.length === 0);
   }
 
 
--- a/src/replicate/replicate.js
+++ b/src/replicate/replicate.js
@@ -274,9 +274,7 @@ function replicate(src, target, opts, returnValue, result) {
     }
     pendingBatch.seq = change.seq;
     pendingBatch.changes.push(change);
-    if (changesOpts.live) {
-      processPendingBatch(true);
-    }
+    processPendingBatch(changesOpts.live);
   }
 
 
--- a/src/replicate/replicate.js
+++ b/src/replicate/replicate.js
@@ -264,7 +264,7 @@ function replicate(src, target, opts, returnValue, result) {
   }
 
 
-  function onChange(change, flush) {
+  function onChange(change) {
     if (returnValue.cancelled) {
       return completeReplication();
     }
@@ -274,7 +274,9 @@ function replicate(src, target, opts, returnValue, result) {
     }
     pendingBatch.seq = change.seq;
     pendingBatch.changes.push(change);
-    processPendingBatch(flush !== undefined ? flush : batches.length === 0);
+    if (changesOpts.live) {
+      processPendingBatch(true);
+    }
   }
 
 
--- a/tests/integration/test.replication.js
+++ b/tests/integration/test.replication.js
@@ -1636,20 +1636,16 @@ adapters.forEach(function (adapters) {
     it('Replication notifications', function (done) {
       var db = new PouchDB(dbs.name);
       var remote = new PouchDB(dbs.remote);
-      var changes = 0;
       var onChange = function (c) {
-        changes++;
-        if (changes === (adapters[1] === 'http' ? 1 : 3)) {
-          db.info(function (err, info) {
-            verifyInfo(info, {
-              update_seq: 3,
-              doc_count: 3
-            });
-            done();
+        db.info(function (err, info) {
+          verifyInfo(info, {
+            update_seq: 1,
+            doc_count: 1
           });
-        }
+          done();
+        });
       };
-      remote.bulkDocs({ docs: docs }, {}, function (err, results) {
+      remote.bulkDocs({ docs: [docs[0]] }, {}, function (err, results) {
         db.replicate.from(dbs.remote).on('change', onChange);
       });
     });
--- a/tests/integration/test.replication.js
+++ b/tests/integration/test.replication.js
@@ -1639,7 +1639,7 @@ adapters.forEach(function (adapters) {
       var changes = 0;
       var onChange = function (c) {
         changes++;
-        if (changes === 3) {
+        if (changes === (adapters[1] === 'http' ? 1 : 3)) {
           db.info(function (err, info) {
             verifyInfo(info, {
               update_seq: 3,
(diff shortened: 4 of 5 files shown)


bbenezech:
Hello guys, @nolanlawson @daleharvey 

I spent some midnight oil yesterday night on the `replicate.js` file. Sorry for the commit message in the PR, it was 4am.

I found the problem I was looking for.
When a replication starts, the adapter fetches up to 100 changes remotely, then yield those **one at a time** through the onChange callback in `replicate.js` https://github.com/pouchdb/pouchdb/blob/99c24a5861e19ac5ae78b35d71c9845a449ba9a0/src/replicate/replicate.js#L267-L278

Now the nasty stuff.

For some reason (happened in a big @daleharvey commit, refactoring the whole replication), a new batch is created each time there is no further batch : https://github.com/pouchdb/pouchdb/blob/99c24a5861e19ac5ae78b35d71c9845a449ba9a0/src/replicate/replicate.js#L277

Even if it makes no sense, it is needed for continuous replications as a side-effect to start up the pumping of changes.

It goes like this:

###### 1 change coming, do we have any batch next?
---> no, everything is pristine.

The change is thus stored in `pendingBatch`, then:

1. It is **immediatly** stored in the `batches` array https://github.com/pouchdb/pouchdb/blob/99c24a5861e19ac5ae78b35d71c9845a449ba9a0/src/replicate/replicate.js#L197 (`pendingBatch` is reset)
2. It is **immediatly** processed and removed from the `batches` array https://github.com/pouchdb/pouchdb/blob/99c24a5861e19ac5ae78b35d71c9845a449ba9a0/src/replicate/replicate.js#L167

###### 1 change coming, do we have any batch next?
---> no, `batches` is still empty!

Then the change is stored as a new batch in the `batches` array (since there already is a batch currently processed) 
But now `currentBatch` is true (the **first change** is being processed async.)
https://github.com/pouchdb/pouchdb/blob/99c24a5861e19ac5ae78b35d71c9845a449ba9a0/src/replicate/replicate.js#L160

###### 1 change coming, do we have any batch next?
---> yes, batch number 2 was delayed for processing. `immediate` is finally false!

Now at this stage, the changes finally begin to stack up in `pendingBatch`.
And we have the reason to the weird 1->1->n-2 batch size!

A simple solution is to trigger a flush when all changes have been fed back to the onChange callback.
It only matters for the http adapter, the local ones don't have any overhead and can be processed atomically. 

I added a simple flush flag to the http adapter, and everything went as expected.
The tests failures made perfect sense and were fixed with little hassle.
The code changes are minimal.

daleharvey:
Thanks so much for the PR + investigation

I am a little worried about this PR because it is exposing this `flush` parameter in the public api in a non standard way, I think the replicator needs to be smarter here.

it sounds like we have somehow managed to introduce an extra step of batching

```
1 change coming, do we have any batch next?
---> no, batches is still empty!
```
But there is data in `pendingBatch` at this point right? it sounds like that check is wrong (and possibly the existing of both `pendingBatch` and `batches` together, it seems like we should only need one queue to deal with incoming changes.

Also I would really like to think about how we would test this

bbenezech:
@daleharvey No, at this stage, the first change is in currentBatch, not in pendingBatch ;)

It is very easy to test: https://github.com/bbenezech/pouchdb/blob/fix-for-non-continuous-replication/tests/integration/test.replication.js#L1642

Here we actually count the number of time the replicator is firing the onChange event. It should only fire once for a batch of changes. (which it does now for the http adapter)

bbenezech:
I can't help you much on your replicator, I don't have a full view on it yet.

I can tell you is that the http adapter need a sort of onFlush CB (similar to the onComplete fired in non-continuous repl.) that can tell the replicator to go for a batch. This is needed for performance with the http adapter in continuous, since we can't rely on completion.
The `batches.length === 0` is meaningless here and needed for its side-effect on bootstrapping the batches dance :dancers:

Another solution would be to fire the onChange event with a batch of changes instead of just one.

daleharvey:
I dont think it should be firing per batch, the code and the test specifically test that it fires once per document, the tests should as much as possible never be testing that it is running against a specific adapter because we try to maintain a consistent api across all adapters.

bbenezech:
@daleharvey Damn you are right, I just notice onChange (the global one) is fired with only one change at a time. It breaks the API :cry:

bbenezech:
Ok.

There is an issue here. The number `3` is completely misleading here, see PR's title. Let's try with 7 docs ;)

```javascript
    var docs = [
      {_id: '0', integer: 0, string: '0'},
      {_id: '1', integer: 1, string: '1'},
      {_id: '2', integer: 2, string: '2'},
      {_id: '3', integer: 3, string: '3'},
      {_id: '4', integer: 4, string: '4'},
      {_id: '5', integer: 5, string: '5'},
      {_id: '6', integer: 6, string: '6'}
    ];

    it('Replication notifications', function (done) {
      var changes = 0;
      var onChange = function (c) {
        changes++;
        if (changes === (adapters[1] === 'http' ? 1 : 3)) {
          db.info(function (err, info) {
            verifyInfo(info, {
              update_seq: 7,
              doc_count: 7
            });
            done();
          });
        }
      };
      remote.bulkDocs({ docs: docs }, {}, function (err, results) {
        db.replicate.from(dbs.remote).on('change', onChange);
      });
    });
```

The test passes for both local and http

`http://127.0.0.1:8000/tests/integration/?grep=suite2%20test%5C.replication%5C.js-local-local%20Replication%20notifications`
`http://127.0.0.1:8000/tests/integration/?grep=suite2%20test%5C.replication%5C.js-local-http%20Replication%20notifications`


It means we are not talking about the API's global onChange event.
The number 3 was misleading, it was a consequence of the old weird behavior, not from the number of changes.

bbenezech:
With 7 docs:

For local adapters, the sequence of doc count in the test is: 1 -> 2 -> 7 (the old weird behavior, not an issue performance-wise for local)
For http adapters, the sequence of doc count in the test is: 7 (the new behavior)

3 docs was not enough to show the weird behavior.

daleharvey:
Yup that test is confusing, I was under the impression `changes().on('change'` fired once per doc and `replicate().on('change'`batched changes, its possible that test needs fixing but it definitely shouldnt have an `if adapter === http` in there, the behaviour should be consistent between adapters plus I am fairly certain we want to be doing this without modifying the public api

bbenezech:
Batching changes with the webSQL looks fairly easy, but for the others it looks really different. There are cursors involved.

Why do you consider adapter's Change.rb to be public API? Where is it documented?

daleharvey:
as far as I can tell https://github.com/pouchdb/pouchdb/pull/4753/files#diff-751bb0f46f782e6e1325ba289283ed26R54 changes the api to add the `flush` parameter

```
db.changes().on('change', function(change, flush) { 
});
```

which is a public api (http://pouchdb.com/api.html#changes), one of the features of replication is that it only uses the public api and does not have special access to the internals, thats also part of the reason we want to ensure the adapters behave consistently across api's

daleharvey:
ok so yup, https://github.com/pouchdb/pouchdb/blob/master/tests/integration/test.replication.js#L1636 is a broken test, it shouldnt be counting that there are 3 events, just any number of events that contain 3 documents withing `change.docs`

bbenezech:
Sorry I am so stupid!

I changed the public API here https://github.com/pouchdb/pouchdb/blob/master/src/changes.js#L54

At least `one of the features of replication is that it only uses the public api and does not have special access to the internals` makes it very clear:

We need to change the public API.

The code in the replicator is not at fault, it does need more info from the Change api in the case of continuous replication (for non-continuous one, the `complete` event triggers a flush with the `immediate` flag)

I can see these options:

1. add a flag in the `change` event (the one I chose, because it looked like the easy way to me)
2. emit a new event `batch_complete` that would replace the `complete` event, for continuous and non-continuous replications

Do you see any other? Which one should I go for? Is there anything unclear?
@nolanlawson do you have any opinion on this, since you tried to solve this issue a while ago?

daleharvey:
So I think there is actually a much simpler solution here, we dont use live replication until we have consumer the changes feed completely, `live` is only used when we are up to date and waiting on new changes. I think if https://github.com/pouchdb/pouchdb/blob/master/src/replicate/replicate.js#L278 only called `processPendingBatch()` when `changeOpts.live` was true, then we would wait until `onChangeComplete` was called before processing a batch and that would be a full one.

bbenezech:
`processPendingBatch(continuous)`?

daleharvey:
Its possible 

```
if (changesOpts.live) { 
  processPendingBatch(true);
}
```

Could be the right thing, but it needs experimenting with. I am fairly certain we dont need or want to add extra information to the public api for this. I think 1 -> 1 -> N shows there is a mistake when it comes to `currentBatch` vs `pendingBatch` vs `batches`. I would expect a naive implementation to at least only do 1 -> N.

Anything we do to fix that should be making this code simpler not more complicated.

bbenezech:
@daleharvey The fact that the first replication is not live makes your simple fix possible, awesome idea: it works great.
The `1 -> 1 -> N` versus  `1 -> N` issue is a different one. It does not collide with the one we are currently dealing with. I suggest we open a different issue for this one.

daleharvey:
Sorry what issue are we dealing with here then? I thought it was the fact we had an extra batch with only one item in it

bbenezech:
Ok, hang one! ;)

Same issue has 2 different causes:

- the one you just told, weird batching algorithm with 3 stages when 2 stages would probably suffice.
- the one I try to deal with in this PR, triggering a batch flush too early for no good reason, simply because we need to start the batch dance each time a new change arrives when there are no way to buffer them (continuous replication).

Each one of these issues adds an useless request, but they are unrelated, and solving the second one makes the first one invisible.

The second one cannot be fixed completely for continuous requests without an API change like the one I proposed earlier (making the replicator aware of the adapter's changes buffer's state). However dealing with the case of continuous request is good enough since we start with a non-continuous one and it is the one that is the most likely to have many changes at once.

I think the right course of actions is to merge the last fix you proposed (since it has the same behavior in fine as my first proposed fix in 99% of cases but does not mess with public API, which sounds nice) and open a separated issue for the batch dance over-engineering issue.

In a separate issue we could also discuss the API's problem and see if it worth the trouble!

Does it sound good?

bbenezech:
Hello @daleharvey 
I just committed one last adjustment. (read the commit message for insights).
Are you comfortable with the DIFF or is there something that is still unclear? I'd be happy to help.

daleharvey:
@bbenezech So the change looks great, thanks. I am worried about the tests however, both the tests you changed are working on the broken assumption that replication will trigger one change event per document so it isnt your changes fault, but I would like to fix these tests rather than just make them pass. so for the test.replication.js it will look something like

```
      var onChange = function (c) {
        // If I remember the change event contains a property listing the documents that have changes
        changes += c.docs.length; 

        if (changes === 3) {
          db.info(function (err, info) {
            verifyInfo(info, {
              update_seq: 3,
              doc_count: 3
            });
            done();
          });
        }
      };
```

and the `test.retry.js` should probably change to something like

```
      rep.on('change', function (c) {
        if (c.docs_written === 3) { 
          rep.cancel();
        }
      });
```

If you want to try those out and follow up it would be great, if not I should hopefully get to it soon

bbenezech:
I just tried, all good. Let's call it a day and wait for CI (GMT+1 here)

daleharvey:
So this had a little problem, my suggestion with the `retry` wasnt quite right and there was jshint errors, but ammended them and testing @ https://github.com/pouchdb/pouchdb/pull/4793. Thanks for the follow up, hopefully green now

daleharvey:
ok awesome, got a green run @ https://github.com/pouchdb/pouchdb/pull/4793, so :+1: and merged in https://github.com/pouchdb/pouchdb/commit/f3a72ea84c951a815f8dd7a002fb11aa17b82543. thanks for your patience with all the changes

nolanlawson:
Nice work!!

Issue #4756, opened 2016-01-07
Pouch is broken when running in iOS WebView when built with Cordova/PhoneGap


trygve-lie:
This issue has been discussed and solved on [slack](https://pouchdb.slack.com/archives/general/p1452128234003441) but this is the essential issue:

The following does not create a database in iOS (WebView) when built with Cordova/PhoneGap:

```js
var db = new PouchDB('myDB');
```

The above code does work when running it in Safari on the same iOS device. The following does also work when build for Android (WebView) with Cordova/PhoneGap. 

Iow: this is Cordova / PhoneGap iOS WebView specific issue. The rest of this issue refer to this environment only.

The above code will not create a database, and the `created` event will not fire. Appending an adaptor for specific telling Pouch to use ex websql does not work either. None of the API methods will perform correctly either.

Ex, doing a simple insert will fail silently:

```js
db.put({_id: 'test'}).then(function(result) { 
    console.log('I wrote', result); 
});
```

Though, trying to create a database this way:

```js
var db = new PouchDB('myDB', function(err, db) { 
    console.log(err, db);
});
```

yelds the following error message:

```sh
"TypeError: cordova.fireWindowEvent is not a function. (In 'cordova.fireWindowEvent(opts.name + "_pouch", {})', cordova.fireWindowEvent' is undefined);"
```

Removing [this block](https://github.com/pouchdb/pouchdb/blob/master/src/adapters/websql/index.js#L521) and [this block](https://github.com/pouchdb/pouchdb/blob/master/src/constructor.js#L170) does resolve the issue and Pouch works as expected.

A big thank you to @daleharvey for helping out debugging and resolving this issue on slack!! Mucho tusen takk!

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/b94f4b8827af3cb6da30272ec0061d8f06a29ade, Mange takk til deg :)

PR #4757, opened 2016-01-07
(#4756) - Fix for being borked when in Cordova/PhoneGap iOS

trygve-lie:
See #4756

### DIFF
--- a/src/adapters/websql/index.js
+++ b/src/adapters/websql/index.js
@@ -518,15 +518,7 @@ function WebSqlPouch(opts, callback) {
     });
   }
 
-  if (isCordova()) {
-    //to wait until custom api is made in pouch.adapters before doing setup
-    window.addEventListener(api._name + '_pouch', function cordova_init() {
-      window.removeEventListener(api._name + '_pouch', cordova_init, false);
-      setup();
-    }, false);
-  } else {
-    setup();
-  }
+  setup();
 
   api.type = function () {
     return 'websql';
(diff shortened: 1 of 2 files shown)


daleharvey:
@trygve-lie As mentioned on irc, will verify this locally but in the meantime there are jshint errors, you can check them with `npm run jshint`, looks like its just removing the jshint and import of cordova

trygve-lie:
Sorry about that. I'll fix

trygve-lie:
@daleharvey Any idea what is failing here?

daleharvey:
@trygve-lie apologies, our tests arent 100% stable, this was just an intermittent, restarted and it passed, all looks good, doing some local testing right now

daleharvey:
So it would be really great to have cordova in ci just to verify these changes, but without that will have to live with manual testing, this fixes a fatal bug for @trygve-lie and my local testing looks like everything is running fine.

From the code side of things, this was added to fix a race condition when starting up, I believe this was the same time we were implementing and seeing a lot of bugs with taskqueue (this code is pretty old). It really shouldnt be needed so happy to merge, thanks so much @trygve-lie

trygve-lie:
@daleharvey Thanks for merging.

I have been thinking a bit on how to run automated tests on iOS and Android devices in general. This is very early in my thought process so just throwing out an idea here.

What if one make a simple phonegap app that just executes the test. Then we use the node client for phonegap build to build the iOS and Android app and when they are built execute them is an environment (docker?) which then finally runs the tests. Not super optimal but an idea.

daleharvey:
We actually have most of that functionality in the tree - https://github.com/pouchdb/pouchdb/blob/master/TESTING.md#cordova-tests @nolanlawson wrote it while testing on various devices. I have been using it to test locally and although the command seems to have regressed (it exits early) the rest looks to still work.

It was a fairly huge amount of hassle to get it running on CI last time Nolan tried if I remember, but definitely something we would love to get working

daleharvey:
However if we can get a docker environment that we can just send an app and it will run the tests for us, that seems like it could be easier, we setup docker for getting couchdb master testing and its working out pretty handy for pouch

Issue #2824, opened 2014-10-04
Get console from running saucelabs / selenium instances


daleharvey:
@admc / @sourishkrout any ideas about this? a large percentage of our failures are onliy via saucelabs and its fairly hard to debug from one side

daleharvey:
Giving up on this

daleharvey:
Heh this is going to be the only way to debug https://github.com/pouchdb/pouchdb/pull/4747

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/b933b9847c1855e65d26e8bca9fb9c250e853b5b

nickcolley:
Thanks Dale! 🤓

PR #4760, opened 2016-01-07
(#2824) - Get console from running saucelabs failures

daleharvey:


nickcolley:
What's the best way to check this is working on Travis?

daleharvey:
If you put up a build that does a console.log somewhere inside the test run you can see if the test run fails, I was using it to debug https://github.com/pouchdb/pouchdb/pull/4747 so its definitely working.

I think I will amend this to also add the logs to passing commits, seems like it would also be useful

nolanlawson:
This is a great idea. :+1: from me, wonder why we didn't do it earlier.

daleharvey:
Awesome cheers - https://github.com/pouchdb/pouchdb/commit/b933b9847c1855e65d26e8bca9fb9c250e853b5b

PR #4764, opened 2016-01-09
Improve replication for read-only source databases

KlausTrainer:
This improves detection of read-only source databases. Before this
change, there were two different checks for read-only source databases.
One version was checking for `4xx` response status codes when trying to
update the replication log in the source database. The other version only
checked for a `401` status code when trying to create a new replication
log in the source database. We now always check for any client error
status code (`4xx`). This avoids problems when the source database
responds with client error status codes other than `401` to replication
log write requests.

This commit also adds an early return if the source database is
read-only. The goal is to avoid that the checkpointer unnecessarily
reattempts reading and writing checkpoints.

### DIFF
--- a/src/replicate/checkpointer.js
+++ b/src/replicate/checkpointer.js
@@ -95,9 +95,7 @@ Checkpointer.prototype.updateSource = function (checkpoint, session) {
   return updateCheckpoint(this.src, this.id, checkpoint,
       session, this.returnValue)
     .catch(function (err) {
-      var isForbidden = typeof err.status === 'number' &&
-        Math.floor(err.status / 100) === 4;
-      if (isForbidden) {
+      if (isForbiddenError(err)) {
         self.readOnlySource = true;
         return true;
       }
@@ -123,6 +121,10 @@ var comparisons = {
 Checkpointer.prototype.getCheckpoint = function () {
   var self = this;
   return self.target.get(self.id).then(function (targetDoc) {
+    if (self.readOnlySource) {
+      return Promise.resolve(targetDoc.last_seq);
+    }
+
     return self.src.get(self.id).then(function (sourceDoc) {
       // Since we can't migrate an old version doc to a new one
       // (no session id), we just go with the lowest seq in this case
@@ -151,7 +153,7 @@ Checkpointer.prototype.getCheckpoint = function () {
         }).then(function () {
           return LOWEST_SEQ;
         }, function (err) {
-          if (err.status === 401) {
+          if (isForbiddenError(err)) {
             self.readOnlySource = true;
             return targetDoc.last_seq;
           }
@@ -233,6 +235,10 @@ function hasSessionId (sessionId, history) {
   }
 
   return hasSessionId(sessionId, rest);
+}
+
+function isForbiddenError (err) {
+  return typeof err.status === 'number' && Math.floor(err.status / 100) === 4;
 }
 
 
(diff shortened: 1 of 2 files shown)


daleharvey:
This looks like a sensible change, however it has regressed coverage, the `sync.js` is intermittent and I will look at that now but there is now missing coverage in the checkpointer. 

I would love to have better ability to cover these types of patches, we dont do a great job testing error scenarios right now, we may have to just ignore it but would like to look at what we can improve as well

KlausTrainer:
I agree that it would be cool if these type of patches would come along with corresponding tests. The reason I haven't provided a test here is that it's tricky.

What makes it tricky? To have a database that's truly read-only, one would want to prevent clients from writing local documents (and replication logs with it). But PouchDB doesn't provide any feature for that (and CouchDB neither). So the way to go is to configure the reverse proxy respectively. For example with nginx, one could use the following rule to achieving that a database is read-only:

```
    location /db {
        limit_except GET {
            deny all;
        }

        proxy_pass http://couchdb;
    }
```

Now the problem is that as a user, I still expect replication to work with that type of setups. However, what makes those setups hard to test is that with the gateway they involve an extra component beside client and server.

daleharvey:
Yup, I have a fair idea of how to test this after thinking about it for a day though, will get a patch up today

daleharvey:
@KlausTrainer So I have started working on a test for this, I havent reproduced the exact error but I think it has most of the situation covered. Currently this only works in node but I have an idea how to make it work in the browser - If you could work on top of it or explain what would reproduce the error it would be great - https://github.com/pouchdb/pouchdb/commit/39d570d22ef0c9fa0008a8cee6bedc04fc636c27 is the test

KlausTrainer:
Thanks, @daleharvey! That's helpful. I don't worry about getting it to work in the browser for now, and I'll just try to get at least one failing test case with it.

KlausTrainer:
I've amended my previous commit adding a test proving that after successful replication `checkpointer.getCheckpoint` wrongly resolves with `0` instead of the actual last sequence number, if the source database responds with status code `403`.

KlausTrainer:
Just added another test case proving that `checkpointer.getCheckpoint()` doesn't make another request for reading or writing a replication log from or rather to the source database.

daleharvey:
Awesome, thanks so much, really glad to be landing these with tests - https://github.com/pouchdb/pouchdb/commit/bbc8dbcdf74a4c277d552c7b89a5693aadfb67c3

KlausTrainer:
Thanks so much for your help!

nickcolley:
Niiice @KlausTrainer :tada:

daleharvey:
@KlausTrainer also wanted to say, since you have made continued PR's to PouchDB I have invited you to become a PouchDB committer. This means you have write access to the repo and can merge other peoples pull requests (your own still require review, as do everyones)

Thanks so much for your contributions :)

nolanlawson:
Welcome @KlausTrainer ! :boom:

KlausTrainer:
I'm pleased about your invitation, which I gladly accept. I appreciate the confidence you're showing in me. Thank you, I will try my best!

ngbinh:
this is kick-ass! One of our pain point post `5.0.0`. For extra security, we only open up read-only couchdb source to the Internet. It broke after `5.0.0` dues to the introduction of `_bulk_get()`.

PR #4769, opened 2016-01-09
DONT MERGE try to repro #4715 (glob)

nolanlawson:


nolanlawson:
HAHA, reproduced! Merely adding "glob" to the devDependencies causes the error.

PR #4770, opened 2016-01-09
(#4770) - remove unnecessary deps from package.json

nolanlawson:
Our `package.json` is getting full of dependencies
that aren't strictly needed, and this is increasing
our `npm install` times. I'm removing some of them here.

### DIFF
--- a/bin/test-webpack.sh
+++ b/bin/test-webpack.sh
@@ -8,7 +8,6 @@
 
 npm run build
 ./node_modules/.bin/webpack \
-  --module-bind json \
   --output-library PouchDB --output-library-target umd \
   . pouchdb-webpack.js
 POUCHDB_SRC='../../pouchdb-webpack.js' npm test
--- a/package.json
+++ b/package.json
@@ -18,7 +18,6 @@
     "pouchdb"
   ],
   "dependencies": {
-    "add-cors-to-couchdb": "0.0.4",
     "argsarray": "0.0.1",
     "debug": "2.2.0",
     "double-ended-queue": "2.0.0-0",
@@ -67,7 +66,6 @@
     "istanbul": "0.4.1",
     "istanbul-coveralls": "1.0.3",
     "jshint": "2.8.0",
-    "json-loader": "0.5.4",
     "less": "2.5.3",
     "mkdirp": "0.5.1",
     "mocha": "2.3.4",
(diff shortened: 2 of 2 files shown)


bin/test-webpack.sh
```diff
@@ -7,8 +7,8 @@
 #
 
 npm run build
+npm install webpack
```
nolanlawson: just install webpack on the fly, not for every `npm install`

CONTRIBUTING.md
```diff
@@ -145,10 +145,10 @@ Release Procedure
 
  * Copy the last release post from ./docs/_posts/date-pouchdb-version.md, ammend date and version and fill in release notes
  * Push release post
- * `./node_modules/.bin/tin -v $VERSION`
+ * `npm install -g tin && tin -v $VERSION`
  * `npm run release`
  * Copy the `dist/pouchdb*` files from the $VERSION tag on github, paste the release notes and add the distribution files to Github Releases, rename `pouchdb.min.js` to `pouchdb-$VERSION.min.js` after you upload it.
- * `./node_modules/.bin/tin -v $VERSION+1-prerelease`
+ * `tin -v $VERSION+1-prerelease`
```
nolanlawson: `tin` even says as a warning that it should be installed globally

package.json
```diff
@@ -18,7 +18,6 @@
     "pouchdb"
   ],
   "dependencies": {
-    "add-cors-to-couchdb": "0.0.4",
```
nolanlawson: this is already in the `devDependencies`

bin/test-webpack.sh
```diff
@@ -7,8 +7,8 @@
...
 ./node_modules/.bin/webpack \
-  --module-bind json \
```
nolanlawson: `json-loader` no longer required thanks to one big ol' `index-browser.js`

nickcolley:
How long does tin take to install, I'd rather avoid global deps if possible as right now if you npm install you know everything will just work. Granted most won't be publishing ..

daleharvey:
Yup tin is wrong, if we use it then it should be installed and nothing should be installed globally aside from node/npm so :+1: to everything but the tin change here

nolanlawson:
We only use `tin` in the docs (i.e. as instructions for contributors). But it's not a big dep, so yeah, we can just keep it in, no biggie.

daleharvey:
:+1: sweet - https://github.com/pouchdb/pouchdb/commit/199f30e02d1f5e65ef2ffdd0430debc0b5f12a61

Issue #4762, opened 2016-01-08
all tests need a rebuild before testing


daleharvey:
I think with the extra compile step there are going to be a few of these. We can just put `npm run build` before test component however its a long build process

@nolanlawson in the work you are doing in https://github.com/pouchdb/pouchdb/pull/4741, would it be possible to split out the build steps so if we only need the babel conversion we can do only that, then only build `pouchdb.js` and a seperate one to do all the fruitdown and the plugins etc

daleharvey:
Just making note that `npm test` needs this too

nolanlawson:
Not sure what you mean by this? In Travis, we always do an `npm run build` after `npm install` (it's a `prepublish`, so it happens automatically), and for `npm run dev`, the whole codebase is rebuilt anytime a single file changes.

My idea for a workflow here was that, if you were just testing Node-related stuff (i.e. you didn't have `npm run dev` running), you'd just run `npm run build` before you do `npm t` or whatever your test step is. Doing it automatically would be nice for development, but it'd be hard to stop Travis from needlessly running the build twice.

daleharvey:
It wasnt for travis was meaning local workflows, its confusing to have to know which tests you need to do a `npm run build` before or not, I think `npm install && npm test` should basically always work. I have the PR for fixing it up now though

nolanlawson:
`npm install` will automatically do `npm run build`, yeah. The only really tricky one is the unit tests, since those need `npm run build-as-modular-es5`, which was a dirty hack, but I couldn't figure out a better way.

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/443cd66992337093ec963a61d1d3d333d63c6a4b

PR #4772, opened 2016-01-10
(#4762) - Add switch to only build for node

daleharvey:


daleharvey:
Hrm, I remember being surprised I could pass arguments to npm run, possibly what broke 0.10 here, its a stable failure

nolanlawson:
I think that was a feature added recently to `npm`. I'd be fine with dropping support for 0.10 BTW...

bin/build.js
```diff
@@ -179,41 +183,45 @@ function cleanup() {
...
+}
+
+if (process.argv[2] === 'node') {
+  buildForNode();
+  return;
 }
 
-Promise.resolve().then(function () {
-  return Promise.all(['dist', 'lib', 'src_browser'].map(function (file) {
-    return rimraf(file);
```
nolanlawson: Removing `rimraf()` here means that it's possible to accidentally commit unwanted files from the `dist ` directory

package.json
```diff
@@ -96,10 +96,11 @@
   },
   "scripts": {
     "build": "node ./bin/build.js",
+    "build-node": "node ./bin/build.js node",
     "build-as-modular-es5": "sh bin/build-as-modular-es5.sh",
     "test-unit": "mocha tests/unit/test.extras.js && npm run build-as-modular-es5 && mocha tests/unit",
-    "test-node": "sh bin/test-node.sh",
-    "test-component": "mocha tests/component",
+    "test-node": "npm run build-node && sh bin/test-node.sh",
+    "test-component": "npm run build-node && mocha tests/component",
```
nolanlawson: So this solves the problem of contributors not having to type `npm run build` before `npm test`, but it does an unnecessary double build on Travis

nolanlawson:
The double build is not such a big deal; I'm mostly just concerned about the `rimraf()` thing.  So I'd say let's fix the `rimraf()` and then :+1:.

bin/build.js
```diff
@@ -129,7 +129,9 @@ function doRollup(entry, fileOut) {
 
 // build for Node (index.js)
 function buildForNode() {
-  return doRollup('src/index.js', 'lib/index.js');
+  return mkdirp('lib').then(function() {
+    doRollup('src/index.js', 'lib/index.js');
```
nolanlawson: `return doRollup(...)`

daleharvey:
Yeh really not worried about travis doing a double build, but I want to make sure we have an easy to figure out workflow for local dev, remembering to do builds in between test runs is pretty confusing. Added the rm -rf dist and good catch on the `return doRullup`

nolanlawson:
Sorry, wasn't being clear - I'm also worried about `rm -rf lib`, because unnecessary files can end up in there as well (e.g. during `npm run build-as-modular-es5`).

Otherwise, looks good!

daleharvey:
Ok cool, I added the `rimraf('lib')`, cheers - https://github.com/pouchdb/pouchdb/commit/443cd66992337093ec963a61d1d3d333d63c6a4b

Issue #4748, opened 2016-01-06
Browser modal authentication window raised on normal detection 401 during sync with Cloudant


tdupard:
A modal authentication window is raised by the browser on "normal detection" 401 during sync with Cloudant.

I get the reason for the 401 from the log (see below message), but is there a way to suppress the modal window? If the window is cancelled away the sync happens normally.

"The above 401 is totally normal. PouchDB is just detecting if the remote supports the _bulk_get API."

Tested on: FF 43, IE Edge, Chrome 47

daleharvey:
Is your cloudant instance password protected? if you check the log do we not sent the correct credentials when checking for `bulk_get`?

tdupard:
That is correct, I can see the
   "Authorization: Basic ZDY2NDllMTItNmXXXXXX...." header in the other requests, like '_changes' but not in the '_bulk_get' request.

daleharvey:
How do you set the credentials, in the constructor or on the actual request?

tdupard:
I am passing the url with credentials embedded to the constructor:

var remoteCouch = 'https://XXXX:XXXXX.cloudant.com/' + dbName;

var sync = PouchDB.sync(dbName, remoteCouch, {
		  live: true,
		  retry: true
		}

daleharvey:
That is very strange, they should definitely be passed into that call then, taking a look, cheers for the follow up info

twitwi:
I observe the same problem in the same context (syncing to cloudant with credentials in the URL).

Also, I'm not sure it is totally related but when the same application is deployed on android using ionic (with the same cloudant sync), it also fails at subsequent synchronization. I didn't dig more on this one.

willholley:
I thought it was expected that we don't sent authorization to _bulk_get with Cloudant because the preflight response does not allow it (see https://github.com/pouchdb/pouchdb/issues/4613)? I still wouldn't expect a model dialog. @tdupard are you able to share your Cloudant username / raise a support ticket so I can check the logs on our side?

tdupard:
@willholley, sure, what is the best to do that? Do you want me to send my credentials to your Gmail?  But the authentication window is consistent across FF, Chrome and IE so far. My Cloudant creds were created in Bluemix if it maters...

daleharvey:
If we arent sending credentials then it seems like it has to be a bug on the pouch side, I was certain I fixed all of them but I will take another look

willholley:
[...]

Issue #4643, opened 2015-12-08
Error 500 with CORS warning


simonaberry:
I am getting repetitive 500 errors (v5.1.0) and would appreciate some guidance on how to debug further. 

```
Uncaught (in promise) e
error: true
message: "Database encountered an unknown error"
name: "unknown_error"
result: Object
doc_write_failures: 0
docs_read: 0
docs_written: 0
end_time: Tue Dec 08 2015 21:00:53 GMT+0200 (SAST)
errors: Array[1]...
ok: false
start_time: Tue Dec 08 2015 20:14:09 GMT+0200 (SAST)
status: "aborting"
status: 50
```
followed by a warning 
```
PouchDB: the remote database may not have CORS enabled.If not please enable CORS: http://pouchdb.com/errors.html#no_access_control_allow_origin_header
```

but I definitely have cors enabled 

```
curl --user XXX:XXX xxxx.cloudant.com/_api/v2/user/config/cors 
{
  "enable_cors": true, 
  "allow_credentials": true, 
  "origins": [
    "*"
  ]
}
```

this happens periodically, but especially when offline. I am canceling the sync as follows when offline is detected:

```
// on startup / online detected:
sync.company = db.sync(companyUrl, {live: true, retry: true})...

// if offline detected
console.log('offline detected - stopping sync');
if (sync.company) sync.company.cancel();
```

but am also seeing this error
```
offline detected - stopping sync
TypeError: o.company.cancel is not a function
    at e (vendor.js:62)
    at Object.fn (vendor.js:62)
    at n.$digest (vendor.js:5)
    at n.$apply (vendor.js:5)
````
(code has been minified)

inspecting the `sync.company` object i see it is a promise.... but the docs show that you should be able to call cancel() ??

http://pouchdb.com/api.html#sync

not sure if the sync cancel is in any way related to the 500 error, but thought it would be worth mentioning

**Comment by simonaberry** [comment]
so figured out the cancel()/promise issue.... the docs say 
```
var sync = PouchDB.sync('mydb', 'http://localhost:5984/mydb', {
  live: true,
  retry: true
}).on('change', function (info) {
  // handle change
}).on('paused', function () {
  // replication paused (e.g. user went offline)
}).on('active', function () {
  // replicate resumed (e.g. user went back online)
}).on('denied', function (info) {
  // a document failed to replicate, e.g. due to permissions
}).on('complete', function (info) {
  // handle complete
}).on('error', function (err) {
  // handle error
});

[...]

PR #4773, opened 2016-01-10
(#4748) - skip bulk_get detection for Cloudant

willholley:
A bit of a hacky fix. Cloudant always sets instance_start_time to "0"
so use this to shim bulk_get support. A request to /_bulk_get on a
Cloudant database today results in a 401 response with a
WWW-Authenticate: Basic header, generating a browser authentication
dialogue.

Remove this when Cloudant supports _bulk_get (or can avoid the
WWW-Authenticate response header for it).

### DIFF
--- a/src/adapters/http/index.js
+++ b/src/adapters/http/index.js
@@ -316,27 +316,40 @@ function HttpPouch(opts, callback) {
 
     if (typeof supportsBulkGet !== 'boolean') {
       // check if this database supports _bulk_get
-      doBulkGet(function (err, res) {
-        /* istanbul ignore else */
-        if (err) {
-          var status = Math.floor(err.status / 100);
-          /* istanbul ignore else */
-          if (status === 4 || status === 5) { // 40x or 50x
-            supportsBulkGetMap[dbUrl] = false;
-            explainError(
-              err.status,
-              'PouchDB is just detecting if the remote ' +
-              'supports the _bulk_get API.'
-            );
-            doBulkGetShim();
-          } else {
-            callback(err);
-          }
-        } else {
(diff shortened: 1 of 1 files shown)


src/adapters/http/index.js
```diff
@@ -316,26 +316,39 @@ function HttpPouch(opts, callback) {
 
     if (typeof supportsBulkGet !== 'boolean') {
       // check if this database supports _bulk_get
-      doBulkGet(function (err, res) {
-        /* istanbul ignore else */
-        if (err) {
-          var status = Math.floor(err.status / 100);
+      self.info().then(function (info) {
```
nolanlawson: no `catch()` on this promise means if `info()` fails it will get swallowed. :) I suggestion just adding `.catch(callback)`

nolanlawson:
Looks like a great fix for now; I was really wondering what was causing https://github.com/pouchdb/pouchdb/issues/4643.

daleharvey:
Yup this looks great, although I would like to check whether / when this will get fixed on the cloudant side first, would really like to avoid adding this as possible. Its great work @willholley but still the type of workarounds I would really like to avoid having around

willholley:
fully agree - hoping to know today/tomorrow if it's a quick fix on the server.

daleharvey:
@willholley any news from cloudant about whether this has been fixed on the server?

willholley:
I think _bulk_get is close to landing - maybe this week or next. Also, I can no longer reproduce the original problem and other users report that 5.2.0 fixes the issue so I suggest we park this.

daleharvey:
ok cool, closing out, cheers

PR #4792, opened 2016-01-15
Fix misspelling in API docs

pixelistik:


daleharvey:
Thanks so much, I amended the commit message to match our style and dont need a ci run for this change. I also learnt that apparently this is the second most popular typo on the internet :) :+1: - https://github.com/pouchdb/pouchdb/commit/86bed6c0e6837cf29631ab9cb166e16234a4b7f7

Issue #4791, opened 2016-01-15
Cant sync using filter. Syncs all records any way


Alekserok:
var filter =  function (doc) {
                return doc.name === 'name'};
            database.sync(remoteDatabase, {live: true, retry: true, filter:filter});

nolanlawson:
@Alekserok I'm sorry, but I can't reproduce (see tests in #4794). Can you find a test to reproduce, e.g. a sample HTML page that shows the error?

nolanlawson:
confirmed working as intended in https://github.com/pouchdb/pouchdb/pull/4794

nolanlawson:
Please reopen if you can find a reproducible test case! Thanks for reporting.

PR #4794, opened 2016-01-16
(#4791) - add tests for filtered sync

nolanlawson:
I realized from #4791 that this was untested. The tests pass.

### DIFF
--- a/tests/integration/test.sync.js
+++ b/tests/integration/test.sync.js
@@ -744,6 +744,7 @@ adapters.forEach(function (adapters) {
       });
     });
 
+
     it('4791 Single filter, live/retry', function () {
 
       var db = new PouchDB(dbs.name);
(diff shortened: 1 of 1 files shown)


daleharvey:
Looks like genuine failures

nolanlawson:
There were failures at first due to not using `PouchDB.utils.Promise`, but I think it ought to be fixed now. If not, then it's a bug in our sync.

nolanlawson:
Not sure if these are true failures or just strong intermittents.

daleharvey:
Theres definitely an issue with the test, a bug with `paused` seems most likely, will take a look at this

daleharvey:
Rerunning the tests just to get an idea if there are specific configurations failing

daleharvey:
So I made a silly mistake when I looked at this last time, switching the test to check for change events instead of 'paused' ones should be fine, its going to be hard to ensure the guarantees for the paused event that this test depends on but the change events assertions are very much ones we need to guarantee

nolanlawson:
:+1:

nolanlawson:
48774ae428533a6dd1d6411c333bcebe296360b4 and 0a66a762c76c0b56c39c4911f6444034a6a3e28c

PR #4798, opened 2016-01-17
(#4798) - use "files" instead of .npmignore, reduce tgz size

nolanlawson:
This reduces our npm tarball size from 7.7M to 800K, by
explicitly whitelisting the files to include instead of
ignoring them with `.npmignore`. (Most of the size comes
from docs, tests, and coverage output.)

The only downside is that projects like pouchdb-server
and express-pouchdb will no longer be able to pull
the tarball from npm and run pouchdb's tests. However,
they can just grab the source code from Github instead.


nickcolley:
This is great! :+1: from me, will let someone else review too.

daleharvey:
This is awesome, thanks :+1: https://github.com/pouchdb/pouchdb/commit/90445f946599ab70588576b9af2d79af5fa3e102

Issue #4801, opened 2016-01-18
PouchDB objected exported as "PouchDB$1"


nolanlawson:
Aesthetically this just bugs me. It also is Rollup's way of telling us that we are implicitly using a global `PouchDB` reference somewhere. So it's something we ought to fix.

![screenshot 2016-01-17 23 29 52](https://cloud.githubusercontent.com/assets/283842/12383297/55e4a1ee-bd72-11e5-87fd-5adf2714ef14.png)

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/9465e43c479d7976474e6bfa476e6f96d21e101e

PR #4802, opened 2016-01-18
(#4801) - fix global PouchDB object usage

nolanlawson:
I can confirm that this fixes the `PouchDB$1` issue.
Plus this is just clearly safer - we shouldn't rely
on `window.PouchDB` being globally available.

### DIFF
--- a/src/adapters/idb/index.js
+++ b/src/adapters/idb/index.js
@@ -57,7 +57,7 @@ function IdbPouch(opts, callback) {
     },
     callback: callback
   });
-  applyNext();
+  applyNext(api.constructor);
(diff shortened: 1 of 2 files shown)


nolanlawson:
And yeah, I know that the existing code is already pretty weird, but it seems to aid in debugging at least. Hopefully at some point we can diagnose the IndexedDB exceptions and just remove it.

daleharvey:
Looks reasonable, cheers - :+1: https://github.com/pouchdb/pouchdb/commit/9465e43c479d7976474e6bfa476e6f96d21e101e

Issue #4788, opened 2016-01-14
db.type() impredictible behavior regression (5.2.0) 


bbenezech:
Tested in latest chrome:

With PouchDB 5.2.0
```javascript
console.log(new PouchDB('1').type());
console.log(new PouchDB('2').type());
>> idb
>> Uncaught TypeError: (intermediate value).type is not a function(…)
```

A setTimeout to display `db2.type()` (sometime 0, sometime more) fixes it.
It worked as expected with 5.1.0

Tell me how I can help.

nolanlawson:
Interesting bug! Seems to be due to removing `type` from the prototype (it was marked as dead code by Istanbul).

Presumably we need to add it back to `adapter.js`. Here's [where it used to live](https://github.com/pouchdb/pouchdb/blob/57923d6c088f52adcc785f1dad693227fb209396/lib/adapter.js#L750-L753). It's probably easy enough to just add that code back in, but a slam-dunk would also be to add a test to `test.basics.js` that catches this bug.

nickcolley:
Should .type() be deprecated in favour of .info()?

bbenezech:
@nolanlawson Thanks for reviewing.
You may want to warn users in pouchdb-find's readme, since it tries to read `.type()` with `.getIndexes()`
It fails reliably when I instantiate/set up indices on my second collection.
I'll stay with 5.1.0 until `.type()` is moved back to `adapter.js`.

I can't express how much I love you guys for maintaining PouchDB, thanks a lot, really!

nolanlawson:
@nickcolley It's basically undocumented; it's one of those weird edge cases where it's widely used by plugins (naughty me), but probably should not exist or should be made official.

bbenezech:
I can't reproduce the issue in `test.basics.js`. (it is using the `leveldb` adapter, maybe not impacted?)

`.type()` is already covered there:

https://github.com/pouchdb/pouchdb/blob/b933b9847c1855e65d26e8bca9fb9c250e853b5b/tests/integration/test.basics.js#L1041-L1044

I don't get the async chicken and egg story with `adapter/type/_type` through `adpater.info()`
I guess moving `.type()` back would fix the issue, but it doesn't make much sense to me :D

Your call.

nolanlawson:
[...]

PR #4804, opened 2016-01-18
#4788 implement proposed fix

bbenezech:
Put back `.type()` in `adapter.js` to fix impredictible behavior with `idb` and `websql` adapters.

### DIFF
--- a/tests/integration/test.basics.js
+++ b/tests/integration/test.basics.js
@@ -1043,6 +1043,10 @@ adapters.forEach(function (adapter) {
       db.type().should.be.a('string');
     });
 
+    it('#4788 db.type() is synchronous', function () {
+      new PouchDB(dbs.name).type.should.be.a('function');
+    });
+
     it('replace PouchDB.destroy() (express-pouchdb#203)', function (done) {
       var old = PouchDB.destroy;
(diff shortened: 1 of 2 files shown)


tests/integration/test.basics.js
```diff
@@ -1043,6 +1043,10 @@ adapters.forEach(function (adapter) {
       db.type().should.be.a('string');
     });
 
+    it('#4788 db.type() is synchronous', function () {
+      new PouchDB(dbs.name).type.should.be.a('function');
```
nolanlawson: Unless I'm mistaken, this test will succeed both before and after the fix, so it doesn't catch it. Could you duplicate line 1047, so that the same thing is called twice (as in your example)? IME that reproduced the bug.

daleharvey:
I merged this but missed the need for 2 .type calls, will test and follow up if needed, apologies. Thanks for the PR @bbenezech  - https://github.com/pouchdb/pouchdb/commit/541b843fb16f34c263681851e3227a6fa05f2283

tests/integration/test.basics.js
bbenezech: No. You don't need to repeat it twice. I am not sure why.
I checked for failure in browser test like discussed.

PR #4805, opened 2016-01-18
(#4788) - Follow up to ensure test will fail before fix

daleharvey:


daleharvey:
Follow up to https://github.com/pouchdb/pouchdb/pull/4804, verified that the test fails before the fix and passes after

bbenezech:
I didn't need to call it twice to have a repro on my machine. Not sure why.

nolanlawson:
:+1:

@bbenezech Weird, well let's take the safe route and call it twice. It only failed for me on the second call (Chrome on OS X).

daleharvey:
green and landed, cheers - https://github.com/pouchdb/pouchdb/commit/7f2321b47ad69ca4214b0f6f03866a03dc21c587

Issue #4767, opened 2016-01-09
Our 100% coverage is intermittent


daleharvey:

nolanlawson:
I'm guessing this is because some code paths in `sync.js` or `replicate.js` are only occasionally hit, depending on the offline/retry tests.

daleharvey:
Yeh sorry meant to follow this up, its exactly that, one of the sync paused rarely gets hit, will put up a patch for it shortly

daleharvey:
Hrm, so https://github.com/pouchdb/pouchdb/commit/443cd66992337093ec963a61d1d3d333d63c6a4b has apparently regressed our coverage, confused how but looking into it, @nolanlawson if you have an idea would appreciate

nolanlawson:
Ah, I know what's happening.

So the issue is that the unit tests can _only_ work with `npm run build-as-modular-es5`, because they need to reach "inside" the rollup bundle to test things that normally aren't exported.

The problem comes when we then want to test for coverage, because "normally" we'd want to test on the bundle, but some functions are only covered by the unit tests, so we need to test on the modular ES5 instead.

When you added your change in 443cd66, it caused Travis to build once normally (on `npm install`), then once for modular ES5 (in `npm run test-unit`), then _once again_ normally, meaning that the `lib` directory is a mix of modular and non-modular code. The non-modular code is tested by the non-unit tests, whereas the modular code is tested by the unit tests.

TLDR: The ~97% we're now at represents what's covered by the non-unit tests.

daleharvey:
Ok the only one left is the sync the gets not hit very very occasionally, gonna try and get a test for that part

PR #4806, opened 2016-01-18
(#4767) - fix coverage tests in Travis

nolanlawson:
A somewhat hamfisted solution, but this should get our
coverage back to 100%.


daleharvey:
Would like to take a look at a cleaner solution, we have quite a lot of debt with the new build system but most important to get it working, great job - https://github.com/pouchdb/pouchdb/commit/76ea142f632ee4fc6fa846f488600291352a42d7

PR #4807, opened 2016-01-19
(#4767) - fix path to adapter plugins

nolanlawson:
I can confirm that this test fails before the fix
but succeeds after. We just messed up the path.


daleharvey:
This looks good but has killed our coverage again :(

nolanlawson:
Uggg, it's because suddenly we are "counting" the adapter plugins and they are largely untested because all I do is check that they exist. Now I remember why I only wanted to do `require.resolve()`. One sec.

nolanlawson:
I moved `test.extras.js` out of the unit tests entirely, and created a new `npm run test-extras`. They're not really unit tests anyhow; they're just smoke tests to ensure that we don't break our CommonJS-based `extras` API. Running them in the code coverage tests doesn't make any sense and penalizes us unnecessarily given the new Rollup build system.

nolanlawson:
yay, 100%!

daleharvey:
Yeh having them seperate makes sense, cheers :+1: https://github.com/pouchdb/pouchdb/commit/011a6f2ecb002085e9491a89af5bcf0e17226285

Issue #4821, opened 2016-01-25
Master is permafail


daleharvey:

nolanlawson:
Fixed by https://github.com/pouchdb/pouchdb/pull/4822

PR #4822, opened 2016-01-25
(#4821) - Try older node version

daleharvey:


daleharvey:
Very weird, same install error but removing phantom got the tests passing

nolanlawson:
Weird, looks like Travis has Phantom globally installed, because the Phantom tests are still passing. I don't understand what Phantom has to do with anything, though.

nolanlawson:
BTW if that's the fix, then I'm fine with removing phantomjs from `package.json` and just adding a note to TESTING.md saying "do `npm install -g phantomjs` to test phantomjs." After all, you need a browser to be globally installed to test that...

nolanlawson:
:+1: after documentation change; it's better to just unblock master than to dwell on this

daleharvey:
I dont love the idea of our tests depending on random travis globals because installing it the right way breaks stuff in a way we dont understand. I think if we want to keep phantom then we figure out whats wrong with it. But I would vote towards getting rid of it, amended the commit to do so.

nolanlawson:
I don't see any reason to remove it from the tests, especially if we're not replacing it with some other WebSQL-based browser. As is, that would leave us not testing various WebSQL code paths, such as mapreduce or autocompaction.

That said, I am really tired of the phantomjs dependency itself; it adds a huge amount of overhead to the install process and is weirdly causing `npm install` to fail in this case (for whatever reason).

I'm :+1: on this commit if it'll help us move forward and unblock the build, but at the least we need to add Safari tests for autocompaction/mapreduce (which we can do in another PR).

Issue #4323, opened 2015-09-12
destroy() doesn't (always?) accept an options argument, as documented


flimzy:
The `destroy()` method is [documented](http://pouchdb.com/api.html#delete_database) to take two optional arguments:

    db.destroy([options], [callback])

However, when I pass an options argument, my code crashes.  The documentation does mention that the `ajax` key in `options` is only used for remote databases, so I tried doing the same with a remote database call, and I get the same result.  Is the `options` argument now obsolete, and should it be removed from the documentation?  Or am I somehow using it incorrectly?

Note that I'm passing an empty options argument, and I would expect this to simply be ignored, the same as it is when passed to `new PouchDB()`.  But maybe that's an unwarranted expectation?

My code:

    var PouchDB = require('pouchdb');
    var db = new PouchDB("testdb")
    db.destroy({}, function() {
        console.log("in the callback")
    })

And the output:

    $ nodejs ./x.js 

    /home/jonhall/go/src/github.com/flimzy/go-pouchdb/node_modules/pouchdb/lib/adapter.js:806
        callback(null, resp || { 'ok': true });
        ^
    TypeError: object is not a function
        at /home/jonhall/go/src/github.com/flimzy/go-pouchdb/node_modules/pouchdb/lib/adapter.js:806:7

And my code with a remote db call instead of local:

    var PouchDB = require('pouchdb');
    var db = new PouchDB("https://xxxx:xxxx@xxxx.iriscouch.com/asdf")
    db.destroy({}, function() {
        console.log("in the callback")
    })

And the very similar (but more verbose) output:

    $ nodejs ./x.js 

    TypeError: object is not a function
        at /home/jonhall/go/src/github.com/flimzy/go-pouchdb/node_modules/pouchdb/lib/adapter.js:806:7
        at /home/jonhall/go/src/github.com/flimzy/go-pouchdb/node_modules/pouchdb/lib/adapters/http/index.js:1033:7
        at onSuccess (/home/jonhall/go/src/github.com/flimzy/go-pouchdb/node_modules/pouchdb/lib/deps/ajax/ajax-core.js:51:5)
        at Request._callback (/home/jonhall/go/src/github.com/flimzy/go-pouchdb/node_modules/pouchdb/lib/deps/ajax/ajax-core.js:118:7)
        at Request.self.callback (/home/jonhall/go/src/github.com/flimzy/go-pouchdb/node_modules/pouchdb/node_modules/request/request.js:123:22)
[...]

PR #4823, opened 2016-01-25
(#4323) - update localstorage-down/memdown to latest

nolanlawson:
I'm not sure why Greenkeeper didn't notify us, but this should
give us a faster memdown (1.1.2) and a (finally!) unbroken
version of localstorage-down (0.6.6).


nolanlawson:
greenkeeper caught up, let's just go with that

PR #4835, opened 2016-01-26
(#4835) - test in phantomjs again

nolanlawson:
I have high hopes that PhantomJS 2 is gonna work. As a bonus,
we no longer need the ES5 shims, because PhantomJS supports
ES5 fully. However, it does have the broken IndexedDB, so it
will test in WebSQL instead (I confirmed that it's doing this).

Basically I am proposing this because I want to test WebSQL, but
I don't want to test it in Chrome or Safari over Saucelabs, because
that's slow and flaky.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,16 +59,19 @@ env:
   # Test against pouchdb-express-router
   - CLIENT=node SERVER=pouchdb-express-router COMMAND=test
 
-  # Test in firefox running on travis
+  # Test in firefox/phantomjs running on travis
   - CLIENT=selenium:firefox:41.0.1 COMMAND=test
+  - CLIENT=selenium:phantomjs COMMAND=test
 
-  # Test auto-compaction in Node, and Firefox
+  # Test auto-compaction in Node, Phantom, and Firefox
   - AUTO_COMPACTION=true CLIENT=node COMMAND=test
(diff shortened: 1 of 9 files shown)


daleharvey:
Large :+1: for removing es5 shim stuff, however this doesnt specify which phantom we are installing? I think it would be best to be explicit about that as we seem to get bitten with phantom updates. It doesnt need to be in the `package.json` specifically, it could be a `before_install` / `before_script` if that works out better

nolanlawson:
>  this doesnt specify which phantom we are installing / testing against?

It's in the `package.json`; we're using phantomjs 2.1.2. Although yeah, I'd be fine with doing it explicitly in Travis (`npm install phantomjs@2.1.2`) to avoid paying the cost on every `npm install`.

nolanlawson:
PhantomJS tests are passing, but not Node, because of the same error we saw earlier: `Uncaught expected [Error: Error: Could not locate the bindings file. Tried:`.

Since it's a case of "either leveldown works, or either phantomjs works," I think installing phantomjs only for travis should work fine

nolanlawson:
Also this is clearly a bug in npm, so maybe we should file it on them.

nolanlawson:
It's green, and it follows the suggestion you made, and furthermore it unblocks express-pouchdb. So I'm gonna go ahead and merge this, cheers!

nolanlawson:
6567bae3d482e37a08111734b1c402d4f46bbf17

Issue #4372, opened 2015-09-28
revs_limit hides, but does not actually delete, old document data


elafargue:
I am using PouchDB for a data logging application, and one of my databases holds a single 'status' document. This document is updated up to twice per minute, so new revisions are generated fairly rapidly.

   I have noticed that even when using compaction - either automatic or explicit -, the storage requirements for this document get out of hand very quick: for a 100 byte JSON structure, I end up with several hundred megabytes of "*.ldb" files in the LevelDB storage directory. The server running for a bit more than a week now uses more than 1GB for the status document...
 
   My expectation was that once I reached the hardcoded 1000 revisions limit, the space required for this status document would be large but stop increasing, but that is not the case.

   When fetching all revisions of the 'status' document, I correctly only get the last 1000 revisions, all of them "missing" but the latest, since I use compaction. But exploring the ldb files, I can find references to every single revision ever created, which I suppose explains why my storage increases continuously.

   Is this the expected behavior, or is this a bug?

PouchDB version is the latest as of 2015.09.28
OS is Linux

nolanlawson:
Definitely a bug; sounds like when we trim 1000+ revisions, we don't actually remove the revisions; we just update the metadata.

elafargue:
@nolanlawson thanks for the feedback - let me know how I can help... is this independent from the underlying database system, or is this linked to LevelDB specifically? I can take a look if you point me in the right direction in the source tree...

nolanlawson:
Yep, so ideally this would need fixes in all three adapters and a new test (to guarantee that the fix worked). For LevelDB you would be looking at [the bulkDocs implementation](https://github.com/pouchdb/pouchdb/blob/07ccf50dc2220bea6e414f8572816b6bcd2aab1e/lib/adapters/leveldb/index.js#L406) and what happens when metadata gets trimmed, which currently happens in shared utility functions [here](https://github.com/pouchdb/pouchdb/blob/4ed6704f78fb75d80475390e40e2d86bf3e6d7f5/lib/deps/docs/processDocs.js#L88) and [here](https://github.com/pouchdb/pouchdb/blob/4ed6704f78fb75d80475390e40e2d86bf3e6d7f5/lib/deps/docs/updateDoc.js#L32). Note that in one case the code path applies for `new_edits=false`, whereas for the other one it applies to `new_edits=true`.

[...]

PR #4836, opened 2016-01-26
(#4372) Pruning rev_map at compaction for leveldb

sgenoud:
We only keep the revision - sequence map for revisions that have a
document in the database.

### DIFF
--- a/src/adapters/leveldb/index.js
+++ b/src/adapters/leveldb/index.js
@@ -582,7 +582,7 @@ function LevelPouch(opts, callback) {
       if (api.auto_compaction) {
         return autoCompact(complete);
       } else {
-          compact(stemmedRevs, complete);
+        compact(stemmedRevs, complete);
       }
     }
 
@@ -604,7 +604,7 @@ function LevelPouch(opts, callback) {
       }
 
       if (docInfo.stemmedRevs.length) {
-          stemmedRevs.set(docInfo.metadata.id, docInfo.stemmedRevs);
+        stemmedRevs.set(docInfo.metadata.id, docInfo.stemmedRevs);
       }
 
       var attachments = docInfo.data._attachments ?
@@ -953,7 +953,7 @@ function LevelPouch(opts, callback) {
         }
       }, function (next) {
         Promise.resolve().then(function () {
-          if (opts.include_docs && opts.attachments){
+          if (opts.include_docs && opts.attachments) {
             return fetchAttachments(results, stores, opts);
           }
         }).then(function () {
@@ -1200,7 +1200,7 @@ function LevelPouch(opts, callback) {
       if (err) {
         return callback(err);
       }
-      var seqs = revs.map(function(rev) {
+      var seqs = revs.map(function (rev) {
         var seq = metadata.rev_map[rev];
         delete metadata.rev_map[rev];
         return seq;
--- a/src/adapters/leveldb/index.js
+++ b/src/adapters/leveldb/index.js
@@ -1188,7 +1188,11 @@ function LevelPouch(opts, callback) {
       if (err) {
         return callback(err);
       }
-      var seqs = metadata.rev_map; // map from rev to seq
+      var seqs = revs.map(function(rev) {
+        var seq = metadata.rev_map[rev];
+        delete metadata.rev_map[rev];
+        return seq;
+      });
       traverseRevTree(metadata.rev_tree, function (isLeaf, pos,
                                                          revHash, ctx, opts) {
         var rev = pos + '-' + revHash;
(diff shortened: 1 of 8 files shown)


sgenoud:
My first PR - I did not know what to add as a test, feedback welcome!

nolanlawson:
I'm afraid this doesn't really fix #4372 - this only changes the behavior when `get`ting documents from the database, while leaving unresolved the problem that too much data stays written to the database.

The real fix would be to remove the old revs when they are inserted into the database, although unfortunately it's tricky to write a test for that, because the information isn't really exposed by the API in any way (the only way you'd know is by looking at the disk space used by the database). So ultimately if/when we do get a fix, I think a fix without a test would be fine.

The other tricky part of this fix would be fixing it for both the case of compaction (which is a manual operation done by the `compactDocument` function) and auto-compaction (an operation done as part of `bulkDocs`).

nolanlawson:
Also it's worth noting that the `rev_map` is merely a map of rev strings to integer seqs, meaning it doesn't actually contain the document data (although ideally it would be pruned as well).

sgenoud:
> I'm afraid this doesn't really fix #4372 - this only changes the behavior when getting documents from the database

I am not sure that I understands what you mean. Here is how I understand #4372 : it is two fold:

1. When a document reaches 1000 revisions it starts forgetting about the old ones (stemming) - but only in the rev_tree - but the actual documents should also be removed from the database. This is actually a problem not limited to the leveldb adapter. 

2. If autocompact is on, the first issue should not appear, but as reported the leveldb file size keeps getting bigger anyway. The reason for that is that on `doCompact` the `rev_map` is not updated - i.e. the document entry is removed, but the map keeps getting bigger (in a normal case with not that many revisions it is not an issue, but for the particular case of many many revisions it grows in size).

Anyway, I have implemented a solution to problem `1` (in the three different adapters) as well. I have seen that it fixes the leveldb issue but did not try to reproduce on the other adapters.

nolanlawson:
OK, nice writeup, and I see that you have some updates now that address the issue of other adapters. I'll take a look, and restart any failed tests (Travis typically flakes out on a couple tests; we can just restart them).

src/adapters/leveldb/index.js
```diff
@@ -571,11 +570,20 @@ function LevelPouch(opts, callback) {
...
+      fetchedDocs.forEach(function (metadata, docId) {
+        revsMap.set(docId, compactTree(metadata));
+      });
+      compact(revsMap, callback);
+    }
+
     function finish() {
       if (api.auto_compaction) {
         return autoCompact(complete);
+      } else {
```
daleharvey: Small nit, everywhere else in the code has 2 space indents and we would like to be consistent

daleharvey:
This looks great, its a pretty important issues so thanks a lot for taking a look at it. Sorry for taking so long to get round to review (holidays + moving home). 

Theres a few minor nits around the indentation but the code in general looks good. What we really need now is a way to test this which is going to be pretty tricky to do consistently, theres going to be 2 main ways, is there any way these 'leaks' are going to be exposed at the api level (open_revs etc) if so then we can do that. Otherwise we will likely be stuck with node only tests that check the database size (or even contents).

I will take another look at how to test this in the next few days, thats the only thing thats blocking this from landing as far as I can see, Cheers

sgenoud:
>  Sorry for taking so long to get round to review (holidays + moving home).

No problem!

> is there any way these 'leaks' are going to be exposed at the api level (open_revs etc) if so then we can do that in integration tests.

I added a test, based on @nolanlawson ideas in the original issue: revisions that are stemmed should not be available to directly get - I could reproduce the error with the test on master, while it passes on the fix branch.

It allowed me to find a bug for the websql implementation (I guess it is what was caught by travis), yeah!

> Theres a few minor nits around the indentation but the code in general looks good. 

Additionally, the version of jshint that we use do not catch many of the coding style errors - I brought that up on the Slack, should I open a separate issue for this?

daleharvey:
> Additionally, the version of jshint that we use do not catch many of the coding style 
> errors - I brought that up on the Slack, should I open a separate issue for this?

Definitely, the better job our linter can do the easier it is for us / constributors.

Will take a proper look at this tomorrow, but initially its looking great

daleharvey:
ok took another look, sorry I missed the test first time that is great, I still think we could do with another test that confirms no storage leaks but I think this is a great start. We cant copy couch's way of stemming those revs so I think this is the best way. :+1: thanks for the PR :)

daleharvey:
Merged in https://github.com/pouchdb/pouchdb/commit/4cccf617bce67351123a842f12ff0e694587f111

PR #4837, opened 2016-01-26
Update batch_create.html with explicit order

jacwright:
This information is found in the CouchDB docs but it would be nice to specify that guarantee here so users of PouchDB know there is a 1-1 in the output results array from the input docs array.

### DIFF
--- a/docs/_includes/api/batch_create.html
+++ b/docs/_includes/api/batch_create.html
@@ -92,6 +92,8 @@ will be provided individually like so:
 ]
 {% endhighlight %}
 
+The results are returned in the same order as the supplied "docs" array.
+
 Note that `bulkDocs()` is not transactional, and that you may get
 back a mixed array of errors/non-errors. In CouchDB/PouchDB, the smallest
 atomic unit is the document.


marten-de-vries:
Thanks for your PR!

Interesting. I thought the order was undefined. As a matter of fact, I'm relying on that in pouchdb-bulkdocs-wrapper meaning this is **not** the case for pouchdb-security, pouchdb-validation, and probably a couple more plugins.

Still, if it's in the CouchDB docs, the plug-in should adapt, not the other way around. Do you have a link for me so I can research some more on that?

jacwright:
See https://wiki.apache.org/couchdb/HTTP_Bulk_Document_API and search for "returned in the same order". This is important in using `bulkDocs` because without this guarantee it is impossible to know which errors when with which documents. The error returned has no `id` like the success response has.

daleharvey:
Yup we do work in the adapter to make sure they are returned in the correct order and there are tests that depend on it

jacwright:
@daleharvey happy to hear it!

marten-de-vries:
I see, my bad. I'll file an issue on the plugin. :+1: and merged (I added and [skip ci] and the commit number to the commit message while merging): 5972043ff6e1c2e705dfd4f554e9146bedf9f6ab

Thanks!

Issue #4480, opened 2015-10-16
Move website contents to Service Worker


nickcolley:
Right now we have 100+ assets for Appcache to download before the site will update, so it's not great on slow connections.

Thinking we move to SW for page contents or maybe everything.

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/be40f03b1427aeb76f29f6eca8ee878a93eae10d

PR #4846, opened 2016-01-28
WIP: (#4480) - Service workers instead of Appcache for the website

nickcolley:
Just started this WIP branch...

Approach with this:
- Critical assets that must be cached to install the service worker.
- Non critical assets such as the guide etc are optional
- If a page isn't cached it can be added if you visit it while online
- If you try to visit a page that is not cached while offline you'll see an offline message.

Some initial thoughts:
- Appcache seems to still do it's thing even though I thought Service Worker stopped it.
- [] Need to check the old caches are being removed properly
- [x] We would need to set pouchdb up with SSL
- [] Need to handle refresh of content like we do in appcache

### DIFF
--- a/docs/serviceWorker.js
+++ b/docs/serviceWorker.js
@@ -11,11 +11,6 @@ var criticalAssets = [
   '/offline.html'
   '/static/css/pouchdb.css',
   '/static/favicon.ico',
-  '/static/js/code.min.js',
-
-  'http://code.jquery.com/jquery.min.js',
-  'http://netdna.bootstrapcdn.com/bootstrap/3.1.1/js/bootstrap.min.js',
-  'http://cdn.jsdelivr.net/pouchdb/latest/pouchdb.min.js',
 ];
 
 var pages = [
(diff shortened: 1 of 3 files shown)


daleharvey:
Yeh the server will be fine, I have mostly been waiting for letsencrypt before deploying any serviceworker stuff, this is awesome

daleharvey:
Ill try and get https up by the weekend

nickcolley:
@daleharvey cool! Let me know and I'll work on getting this ready.

daleharvey:
`https` is all done :) (sorry it took a while)

nickcolley:
No worries, I think the only thing blocking this is adding the 'Content has updated' feature to the Service Worker implementation as I've said above.

docs/serviceWorker.js
```diff
@@ -0,0 +1,90 @@
...
+var criticalAssets = [
+  '/offline.html'
+  '/static/css/pouchdb.css',
+  '/static/favicon.ico',
+  '/static/js/code.min.js',
+
+  'http://code.jquery.com/jquery.min.js',
+  'http://netdna.bootstrapcdn.com/bootstrap/3.1.1/js/bootstrap.min.js',
+  'http://cdn.jsdelivr.net/pouchdb/latest/pouchdb.min.js',
+];
```
daleharvey: We need to be loading from `https` now, we use `//` in `default.html` so that we can develop locally, hopefully can do the same here

nickcolley: :+1:  This array is just critical assets, so fonts are not required to render the page in this case.

daleharvey: neither is the javascript really, the site fell back really well while these assets didnt load, however the main thing is that after first load are the fonts served from the cache? we definitely want that. It somewhat seems like it should from the `fetch` code below, but I wanted to make sure

nickcolley: Good point, I'll look into removing more that isn't 100% needed.

You're correct about the caching of requests automatically.

daleharvey:
Hey @nickcolley, getting this landed soon would be great, do you mind if I pick it up or do you want to finish it off? I dont just wanna pick up your work without asking

nickcolley:
@daleharvey will update within the week :+1:

nickcolley:
Sorry for the slow uptake, still working on this one..

nickcolley:
Closing this infavour of a branch on the pouchdb repo for my sanity...

PR #4849, opened 2016-01-28
(#4849) - test latest firefox instead of 41

nolanlawson:
If this fails, it's because Travis or Selenium is somehow
not choosing the latest firefox, and falling back to
the default Firefox 31, which doesn't pass our tests
due to bugs in the web worker implementation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ sudo:
   false
 
 addons:
-  firefox: "41.0.1"
+  firefox: "latest"
 
 before_install:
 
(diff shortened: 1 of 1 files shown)


nolanlawson:
Whoa, looks like certain versions of Firefox are just not working in Travis. No output received from the Selenium test. Maybe that explains https://github.com/pouchdb/express-pouchdb/pull/280

nolanlawson:
this isn't worth the headache of babysitting Travis

Issue #4858, opened 2016-01-30
extras/checkpointer and extras/generateReplicationId also need browser switches


nolanlawson:
This is a quite annoying problem that I just realized still exists for plugins like `pouchdb-load`.

nolanlawson:
The fact that nobody has noticed this until now (and the release has been out for a month) tells me that I may be the only person on the planet actually using the `extras` API.

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/b14ac3b491a728488d88788ca54cf6f597a15531

PR #4859, opened 2016-01-30
(#4858) - add browser alts for checkpointer/genReplId

nolanlawson:
Not adding a test because it would involve some kind of wacko test to browserify a package with a dependency on the extras and then grep the source code to make sure there's no `'crypto'` or anything like that.


nolanlawson:
FWIW though I tested this manually and it works. Will finally unblock `pouchdb-load`. Then later we can think about how we're going to deprecate these `extras`.

daleharvey:
:+1: looks good to me, also +1 on starting to deprecate these - https://github.com/pouchdb/pouchdb/commit/b14ac3b491a728488d88788ca54cf6f597a15531

Issue #4337, opened 2015-09-15
Remove sublevel


nolanlawson:
It is taking up a lot of space when we browserify with the memory/localstorage/fruitdown adapter, and I don't think it's buying us a whole lot. It basically just prefixes all our keys, which are just strings anyway.

![screenshot 2015-09-15 01 15 29](https://cloud.githubusercontent.com/assets/283842/9868495/64ed8634-5b47-11e5-802a-a7e7e25fa213.png)

Alternatively if we could get sublevel to upgrade to use `level-codec` and `level-iterator-stream` instead of pulling in all of `levelup`, then it would remove about 25% of the code that you see in the above image (basically everything that `level-sublevel` is pulling in).

nolanlawson:
This is also pulling in 3 versions of `readable-stream` from each of levelup, level-sublevel, and browserify (49kb each), which is pretty silly.

nolanlawson:
BTW I am using disc to generate these images. I'm testing the following script:

```js
var PouchDB = require('pouchdb');
require('pouchdb/extras/memory');
var db = new PouchDB('foo', {adapter: 'memory'});
```

nolanlawson:
49a1363a2246970259ff870a415754bd1b32d2ec

PR #4864, opened 2016-01-31
(#4337) - replace level-sublevel with sublevel-pouchdb

nolanlawson:
I forked [level-sublevel](http://github.com/dominictarr/level-sublevel) into [sublevel-pouchdb](http://github.com/nolanlawson/sublevel-pouchdb) and removed what we weren't using, including a lot of extra dependencies from `levelup` itself that were adding to the Browserify size. As a result, our level-based plugins have gone from 192K minified to 184K (gzipped: 56.2K -> 53.7K).

This also increases our performance a bit, because I was able to optimize two of the hot functions (`decode`/`encode` in `sublevel/codec/legacy.js`) which had some unnecessary checks for non-string keys and complex keys. Since all of our keys are just a single String prefix, I could replace all the array operations with basic string concatenation.

As a result, a `PERF=1` run using `LEVEL_ADAPTER=memdown` on Node 5.3.0 went from 24.67s to 22.65s. The biggest gain comes from reading the database as a stream (i.e. `allDocs()`); I saw that the `all-docs-skip-limit` test went from 4.726s to 3.763s. Upping the iterations on that test from 50 to 500 for better granularity, I got 41.080s -> 29.571s (28% faster), which is pretty awesome.

Keep in mind that all the above perf numbers are for the MemDOWN adapter, so the difference will probably be less stark for LevelDB/localstorage-down/fruitdown.

### DIFF
--- a/bin/build.js
+++ b/bin/build.js
@@ -25,7 +25,7 @@ var spawn = require('child_process').spawn;
 var pkg = require('../package.json');
 var version = pkg.version;
 var external = Object.keys(pkg.dependencies).concat([
- 'fs', 'crypto', 'events', 'path', 'pouchdb', 'level-sublevel/legacy'
+ 'fs', 'crypto', 'events', 'path', 'pouchdb'
 ]);
 
 var plugins = ['fruitdown', 'localstorage', 'memory'];
--- a/package.json
+++ b/package.json
@@ -25,7 +25,6 @@
     "fruitdown": "1.0.1",
     "inherits": "2.0.1",
     "js-extend": "1.0.1",
-    "level-sublevel": "6.5.4",
     "level-write-stream": "1.0.0",
     "levelup": "1.3.1",
     "lie": "3.0.2",
@@ -36,6 +35,7 @@
     "request": "2.69.0",
     "scope-eval": "0.0.3",
     "spark-md5": "2.0.0",
+    "sublevel-pouchdb": "1.0.0",
     "through2": "2.0.0",
     "vuvuzela": "1.0.2"
   },
--- a/src/adapters/leveldb/index.js
+++ b/src/adapters/leveldb/index.js
@@ -1,5 +1,5 @@
 import levelup from 'levelup';
-import sublevel from 'level-sublevel/legacy';
+import sublevel from 'sublevel-pouchdb';
 import { obj as through } from 'through2';
 
 import clone from '../../deps/clone';


nolanlawson:
Here's where you can see that `legacy.js` was taking up ~2.5% of self-time for our perf tests (using MemDOWN):

screenshot 2016-01-31 14 28 55

[cpuprofile file](https://nolanlawson.s3.amazonaws.com/www/cpuprofile/memdown-perf-basics-31012016.cpuprofile)

And here's the change I made:

screenshot 2016-01-31 14 31 03

nickcolley:
Numbers look great! What's the plan for getting upstream fixes to the original level-sublevel, wondering if it's possible to monkey patch the original or something?

nolanlawson:
Well, the reason I didn't contribute back to level-sublevel was a few things:

1. The format we currently use is known as the "legacy" format. It will be unsupported at some point, and we have no burning need to upgrade because it's fine for our use case (we don't need anything as complex as [bytewise](https://www.npmjs.com/package/bytewise) just to prefix a string.
2. My "fixes" were mostly about removing code PouchDB wasn't using, so not much of value to `level-sublevel` there.
3. My fix to remove levelup is not really consistent with what level-sublevel themselves will probably want to do, which is upgrade to the `level-codec` and `level-iterator-stream` modules (which were recently pulled out levelup). In this case, I just pulled the code out of levelup itself and put it in sublevel-pouchdb, because it was easier that way.

Again, `sublevel-pouchdb` is mostly about solving the problem of having a layer over levelup with prefxied keys (in our legacy format). That's it. It's very particular to PouchDB's use case and probably not of much interest to anyone else.

nickcolley:
That's totally reasonable, sorry my thoughts were forgetting the legacy nature of the package.

Hope that's okay :stuck_out_tongue: 

I'm :+1: to this otherwise!

daleharvey:
So this is a fairly complex dependency that we have given ourselves the maintainance burden of, by splitting it out into its own package at least we wont get bitten by upgrades but I do think long term we should be looking to migrate to using the plain leveldown api as I think sublevel is bringing quite a bit of complexity that we dont need.

:+1: to this, I think we should move `sublevel-pouchdb` to the pouchdb org before merging though.

nolanlawson:
I agree, it's a bit of a risk to take the maintenance burden on ourselves, but as-is this is just a very thin layer over vanilla levelup. And the possibility to remove the dependency entirely is still there, although TBH I'm not sure about the value of that, because this actually provides some nice abstractions that make it easier e.g. to write the `LevelTransaction` object.

I'll move `sublevel-pouchdb` to the PouchDB org and then merge this.

nolanlawson:
49a1363a2246970259ff870a415754bd1b32d2ec

PR #4872, opened 2016-02-07
(#4871) Added a default jscs file and npm scripts

### LINKED ISSUE #4871: Code style guide is not enforced by jshint anymore
The maintainer of JShint has decided to take it in a direction where it focuses on and do not implement coding style, see this [SO answer](http://stackoverflow.com/questions/23097316/how-can-i-make-jshint-indent-options-work)  (and the post it links to)

Why does that matter? Because pouchdb relies on jshint to enforce its coding style. Typically with the current version used the indent of 2 spaces does not show up in the posttest phase.

jshint proposes to use jscs for the task: http://jscs.info/overview, seems good enough to me (I spent a bit of time running it on the project: https://github.com/pouchdb/pouchdb/compare/master...sgenoud:adding-jscs I guess we should probably ignore the style for the comments (that adds lots of not that necessary changes).

sgenoud:
Here is the configuration that should fix #4871. Comparing to https://github.com/pouchdb/pouchdb/compare/master...sgenoud:adding-jscs I have not run the auto fix (and done the couple of manual fixes needed as well). Before adding it to the PR I would like to have the buy in of the configuration.

To run the fixes by the tool, you can just run `jscs -c .jscsrc bin/ src/ tests/`. Generally (for further use) `npm run jscs` will do the trick (without fixing the issues automatically).

Also it would be nice to have everything merged before running the auto fix (to avoid too many painful manual merge).


nolanlawson:
Thanks for the PR! Unfortunately there are still lots of style errors ([here's a log](https://s3.amazonaws.com/archive.travis-ci.org/jobs/107571676/log.txt)).

Question: is there any point in using both `jshint` and `jscs`, or can we just use `jscs`?

sgenoud:
I did not fix the errors - my aim was to come up with a proposition of defaults, wait for an agreement and then run the error fixing stuff.

> Also it would be nice to have everything merged before running the auto fix (to avoid too many painful manual merge).

by this I meant everything else, as well as an agreement on the configuration (to avoid doing the manual job too many times)

> Question: is there any point in using both jshint and jscs, or can we just use jscs?

Personnally I will continue to use jshint - as it finds syntax errors while I am developing. But obviously the tests would catch these. The reason I could see to keep using jshint is that it allows to make sure that pouch is written in es5 mostly.

nolanlawson:
OK, thanks for the clarification! Based on discussion in #4871 and this thread, I propose we move forward with an `.eslintrc`. One tool is better than two tools, and ESLint seems better maintained these days than JSHint.

I also propose, though, that if there's any disagreement, we just stick with the current system, because this is not worth bikeshedding over. I have zero problem with sticking with `.jshintrc` and JSHint nailed to 2.8.0.

nickcolley:
:+1: for eslint (for what it's worth semistandard is just an opinionated eslint setup)

daleharvey:
ok thanks for this @sgenoud however looks like the concencus is having eslint would let us do these checks and stick with a single tool, so I have https://github.com/pouchdb/pouchdb/pull/4897 to get it working with `eslint`, cheers for getting this started.

PR #4878, opened 2016-02-10
Missing key in ungrouped (custom) reduce functions

mroeder:
(Custom) CouchDB reduce functions get an array of (key, id) tuples and an array of values handed if `rereduce` is set to `false` (https://wiki.apache.org/couchdb/Introduction_to_CouchDB_views#Reduce_Functions).
The current implementation of PouchDB does not provide the correct key in those tuples if there is no grouping.
This fix corrects this and also restructures the code so the difference between the group's key and the key in the (key, id) tuple can be seen.

### DIFF
--- a/src/mapreduce/index.js
+++ b/src/mapreduce/index.js
@@ -600,38 +600,45 @@ function reduceView(view, results, options) {
   }
 
   var groups = [];
-  var lvl = options.group_level;
+  var lvl = isNaN(options.group_level) ? Number.POSITIVE_INFINITY :
+    options.group_level;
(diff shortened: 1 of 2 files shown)


daleharvey:
I am not super familiar with the reduce api, I generally avoid it, however the tests look grat are passing, code looks good, I am :+1: but will give others a few hours to comment if needed before merging

nolanlawson:
LGTM. Thank you!

nolanlawson:
8b1a50bad9993dda7fcf0f4729f2841c0c36b542

PR #4894, opened 2016-02-15
(#4894) - avoid unnecessarily exporting Changes object

nolanlawson:
There's no reason to export these `Changes` objects, and anyway it will minify smaller if we just use a regular variable instead of attaching it to the exported object.


daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/1f550da385a4410a0ec39c2be09864fefd901930

PR #4897, opened 2016-02-15
(#4871) - Switch to eslint

daleharvey:


.eslintrc
```diff
@@ -0,0 +1,49 @@
+{
+
+  "extends": "eslint:recommended",
+
+  "parserOptions": {
+    "ecmaVersion": 6,
+    "sourceType": "module",
+    "ecmaFeatures": {
+      "jsx": true
```
marten-de-vries: Why do we need jsx?

.eslintrc
```diff
@@ -0,0 +1,49 @@
...
+
+  "globals": {
+    "chrome": true,
+    "Promise": true,
+    "Uint8Array": true,
+    "ArrayBuffer": true,
+    "FileReaderSync": true,
+    "sqlitePlugin": true,
+
+    "it": true,
```
marten-de-vries: The mocha functions could be left out here by enabling the 'mocha' env (just like 'node' and 'browser')

.eslintrc
```diff
@@ -0,0 +1,49 @@
+{
```
marten-de-vries: I think the convention is now to name the file .eslintrc.json instead of .eslintrc

marten-de-vries:
Nice work looking into this. :)

daleharvey:
Thanks for the comments / help @marten-de-vries, never knew about some of those, fixed them up and pushed a new PR

daleharvey:
So one thing is this produces 453 warnings, almost entirely down to `no-unused-var` and `comma-dangle`. I would prefer to to have eslint fail on these which is why I set them to warn instead of ignore (the errors are almost entirely in the tests). I can follow up to remove all the warnings but wanted to get some input before doing that

nolanlawson:
I'm +1 on removing dangling commas and unused vars, although I imagine most of those warnings are going to come from `tests` rather than the `src` or `bin`. Still, if someone is willing to make the fixes, +1 to fixing them everywhere.

marten-de-vries:
If it's too much work to fix it everywhere at once it's also possible to selectively disable the warning in some files with a comment like:

```js
/* eslint comma-dangle: 0 */
```

nickcolley:
:+1: for eslint, I can make the changes if you want @daleharvey

daleharvey:
Didnt take a huge amount of time, I fixed all the warnings and turned them into errors :)

daleharvey:
This will bitrot to hell and got a bunch of :+1: 's pluss all green so landing now, cheers 

https://github.com/pouchdb/pouchdb/commit/d2d339f722e365cb797dcd6d25f355db0e8f911b

PR #4898, opened 2016-02-16
(#4898) - allow creating databases with reserved words

nolanlawson:
Just cleaning house, noticed that we were using regular objects here when we should be using `Map`s. Some other small fixes inside, although nothing earth-shattering.

### DIFF
--- a/src/adapter.js
+++ b/src/adapter.js
@@ -1,7 +1,7 @@
 import { extend as extend } from 'js-extend';
 import Promise from './deps/promise';
 import pick from './deps/pick';
-import collections from 'pouchdb-collections';
+import { Map } from 'pouchdb-collections';
 import inherits from 'inherits';
 import getArguments from 'argsarray';
 import adapterFun from './deps/adapterFun';
@@ -358,7 +358,7 @@ AbstractPouchDB.prototype.revsDiff =
   }
 
   var count = 0;
-  var missing = new collections.Map();
+  var missing = new Map();
 
   function addToMissing(id, revId) {
     if (!missing.has(id)) {
--- a/src/adapters/idb/index.js
+++ b/src/adapters/idb/index.js
@@ -1,6 +1,6 @@
 import clone from '../../deps/clone';
 import uuid from '../../deps/uuid';
-import collections from 'pouchdb-collections';
+import { Map, Set } from 'pouchdb-collections';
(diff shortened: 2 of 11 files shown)


src/adapter.js
```diff
@@ -1,7 +1,7 @@
 import { extend as extend } from 'js-extend';
 import Promise from './deps/promise';
 import pick from './deps/pick';
-import collections from 'pouchdb-collections';
+import { Map } from 'pouchdb-collections';
```
nolanlawson: Using this style is clearer (IMO) and leads to slightly more minified code.

src/adapters/idb/index.js
```diff
@@ -790,17 +791,16 @@ function init(api, opts, callback) {
     idbChanges.removeAllListeners(dbName);
 
     //Close open request for "dbName" database to fix ie delay.
-    if (IdbPouch.openReqList[dbName] && IdbPouch.openReqList[dbName].result) {
```
nolanlawson: There was no reason to export the `openReqList` object.

tests/integration/test.basics.js
```diff
@@ -697,6 +697,13 @@ adapters.forEach(function (adapter) {
       });
     });
 
+    it('Create a db with a reserved name', function () {
+      var db = new PouchDB('__proto__');
+      return db.info().then(function () {
+        return db.destroy();
+      });
+    });
+
```
nolanlawson: The one test I added. Hopefully it won't push us over the Safari limit.

src/constructor.js
```diff
@@ -29,8 +29,6 @@ function prepareForDestruction(self, opts) {
 
   function onDestroyed() {
     ctor.emit('destroyed', name);
-    //so we don't have to sift through all dbnames
-    ctor.emit(name, 'destroyed');
```
nolanlawson: This is an untested and undocumented feature that breaks if you try to emit a string like `'__proto__'`. I'm not aware of any code that is using it either in core or in plugin-land, so I propose we just remove it.

src/adapter.js
nickcolley: :+1:

daleharvey:
Looks good to me and all passing - https://github.com/pouchdb/pouchdb/commit/14449b7f80d27c6c08ec6c180dd8b5c1eb1439c0 :+1:

Issue #4867, opened 2016-02-03
Cache ddoc views and filters for performance


garethbowen:
Querying persistent views [first gets and parses the ddoc](https://github.com/pouchdb/pouchdb/blob/master/src/mapreduce/index.js#L887) to get the view implementation. With a large ddoc such as  a complex couchapp this can be slow, for example with my 9MB ddoc this was taking around 1 second. Because the ddoc is parsed for each query request it quickly becomes a performance bottleneck.

I wrote a dirty dirty patch to implement a `views` cache which greatly improves the performance of my app.

Other areas of the code also query the ddoc for `views` and `filters` properties and so would also benefit from this change.

My proposal is to cache the `views` and `filters` on the db object and then listen for changes to the ddoc to invalidate the cache.

garethbowen:
If this approach is acceptable I'm happy to help out with a PR for it.

nolanlawson:
I think this sounds like a reasonable fix, especially if it can be done in such a way that guards against odd edge cases like e.g. updating the ddoc. We already heavily cache inside of map/reduce, but it would make sense to go one step further and cache the ddoc itself as well.

If you open up a PR, it will run the tests against the full map/reduce test suite, and we can confirm that the fix works. I'd be happy to see your PR! :)

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/0ddeae6b6c44fa12b83c5348df6947e3ce28c28c

nolanlawson:
In retrospect I think this PR might have been merged too eagerly. Looking over the code now, I see several issues:

- adds undocumented `getView()` and `getFilter()` to the API, which could confuse users. These should at least be `_`-prefixed.
- no evidence that this speeds anything up, or by how much
- adds another dependency between mapreduce and pouchdb core, making it harder to untangle them (if we ever will)

In #5118 this commit has actually been accused of creating a performance _regression_. So I really think we need to re-evaluate this PR and determine whether it's worth the maintenance cost and the potential performance cost.

nolanlawson:
There's another fundamental problem with this PR, which is that it's not actually safe to cache DDocs in-memory and use that to determine whether or not an update is required. Hence why you have to do the `changes()` request for every ddoc `get()`.

It seems like if anything, this PR addresses the problem of having an enormous ddoc (e.g. one that is 9MB - why such a large DDoc?), whereas it introduces new performance problems due to:

1. having to do `changes()` for every map/reduce request, and
[...]

PR #4900, opened 2016-02-18
(#4867) - Cache ddoc views and filters

garethbowen:


src/adapter.js
```diff
@@ -185,6 +185,71 @@ function attachmentNameError(name) {
...
 
+function cacheUpdateRequired(api, cache, designDocName, callback) {
+  cache.seq = cache.seq || 0;
+  var changesOpts = {
+    doc_ids: [ '_design/' + designDocName ],
+    limit: 1,
+    since: cache.seq
+  };
+  api.changes(changesOpts)
+    .then(function(res) {
```
daleharvey: Small nit, general style is that as long as they can fit we do

```api.changes(changesOpts).then(function (res) { 
   // rest of code here
});
```

Theres a few other examples of this down below

garethbowen:
@daleharvey I've updated the code style as per your comment.

daleharvey:
Awesome thanks, I was supposed to do more of a review earlier sorry

I think this functionality is mildly worrying, creating additional changes handlers could have unforeseen consequences however it sounds like it creates a significant speedup that cant be done externally.

We should be trusting our tests however, its all passing, I would like performance tests to go along with this ideally, but our performance tests are not in a great state so I will focus on adding some for these soon.

I am :+1: but leave it a little for others to review

garethbowen:
I backed away from continuous changes handlers for that reason, and now it's querying for the latest ddoc when the view/filter is requested. This reduces performance slightly when querying but removes the overhead of listening for changes.

With our (admittedly massive) ddoc on our target phones rendering time for a specific page dropped from ~13 seconds to ~3 seconds with this change.

daleharvey:
Ah I should have read better, for some reason I always see `live: true` automatically when I see `changes()`, that looks like a much safer approach.

daleharvey:
Ok I think this is good to go, thanks so much for the PR @garethbowen - https://github.com/pouchdb/pouchdb/commit/0ddeae6b6c44fa12b83c5348df6947e3ce28c28c

nolanlawson:
This all looks great to me, except the I believe the changes to `test.persisted.js` indicate that we have introduced a bug. If it's not possible to run those promises using `Promise.all()`, then that means map/reduce is not robust against situations where users are querying the same ddoc multiple times in parallel. It indicates a race condition if this code doesn't pass without changing the test.

Issue #4575, opened 2015-11-14
CouchDB Master browser tests failing


daleharvey:
Again, it has been for a while, filing to take a look at

kxepal:
Reason?

daleharvey:
looks like a change in error code for the the first issue

kxepal:
Yes, there were recent changes that doest that. See [COUCHDB-2534](https://issues.apache.org/jira/browse/COUCHDB-2534).

willholley:
the breaking change is that CouchDB master rejects malformed design docs. The failing test ("Changes with invalid ddoc with no filter function") is explicitly using a malformed ddoc so fails to PUT the design document. 

Given that we already have a test to cover the same scenario with a valid ddoc ("Changes with `filters` key not present in ddoc"), I wonder what value the failing test provides? Ditto for "Changes with invalid ddoc with no map function".

@kxepal should this be noted as a breaking change in CouchDB 2.0 as ddocs which don't validate will not replicate from 1.X to 2.X?

daleharvey:
If there is already a test for how we handle invalid ddocs then I am good removing the broken tests

willholley:
I suggest we split out the test for "attempt to save invalid design doc" as we already have coverage for the invalid changes calls (missing view function, missing ddoc).

kxepal:
@willholley Technically, that's not a breaking change, but a bug fix. But yes.

daleharvey:
hrm, this is weird, seeing different errors locally than I do in travis even though I am running couchdb identically (via `docker run -d -p 3001:5984 klaemo/couchdb:2.0-dev --with-haproxy --with-admin-party-please -n 1`)

daleharvey:
will fix my local bugs first then see whats up

daleharvey:
@willholley Also @nolanlawson has added tests running against cloudant to the test suite, these and the couch 2.0 tests are the only failures, if there was anyone at cloudant that could take a look at why those tests are failing it would be awesome, ill be taking a look at the same time and happy to help out and hopefully we can get back to all green :)

daleharvey:
https://travis-ci.org/pouchdb/pouchdb/builds/109605078 - https://travis-ci.org/pouchdb/pouchdb/jobs/109605109 is a link to the tests running against cloudant

daleharvey:
Ok, so now running at 100% locally and not reproducing the perma fail that we are seeing on travis despite me running the exact same couchdb instance

[...]

PR #4905, opened 2016-02-19
(#4575) - Dont test for specific error message

daleharvey:


daleharvey:
First fail I ran up against running tests against local 2.0, in previous cases where couch is returning different error details we have just made the test less specific

daleharvey:
2nd failing test I found, on successful bulkDocs we ensure the order of the results, however couchdb master no longer does for others, I think its safer to pass through couch results than try sorting them ourselves so I think its best to remove the assertion

daleharvey:
So last change fixed `test.retry` when it was ran standalone but not when I run the full suite which is quite worrying as it usually indicates that there is something leaking across tests

daleharvey:
Ok still investigating why the paused events work differently however looking at the test I believe the fixed assertion is correct in that active and paused are supposed to fire at least twice. Locally the tests arent super stable but these changes get it to pass intermittently locally at least.

nolanlawson:
These changes all look good to me. In general I'm suspicious of testing for error names; it rarely works across node/browser/couch/etc. :+1: from me.

daleharvey:
Awesome thanks, merged in https://github.com/pouchdb/pouchdb/commit/90801c6ca892f9a3225d30304721f1e20c426a1c

PR #4909, opened 2016-02-19
avoid sh compatibility by explicitly invoke using bash

iivvoo:
Some testing related scripts invoked by 'npm test'  in bin/ use a bash-specific construct using double brackets (`if [[ .. ]]`).

This fails when invoking the scripts using 'sh' since that doesn't always point to bash (on ubuntu 14.04 it points to 'dash'. This results in

`bin/build-node.sh: 5: bin/build-node.sh: [[: not found
`
Fix is to explicitly invoke using bash.

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,15 +91,15 @@
   },
   "scripts": {
     "build": "node ./bin/build.js",
-    "build-node": "sh bin/build-node.sh",
-    "build-as-modular-es5": "sh bin/build-as-modular-es5.sh",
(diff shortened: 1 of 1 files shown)


daleharvey:
Nice, :+1: ill merge this along with others shortly

nolanlawson:
:+1:

daleharvey:
Merged in https://github.com/pouchdb/pouchdb/commit/31498609b8d465a28e95b196e7314a9eb8a4320a, thanks

PR #4914, opened 2016-02-21
Removing await* edit

tonylukasavage:
As it turns out, `await*` is no longer part of the async/await proposal and `Promise.all()` is the preferred way to handle this scenario. Not sure if you can still this with regenerator, but that would likely be the only place you could ever use it.


nolanlawson:
:+1: for this change. `await*` has been removed, so best just not to confuse people with it.

nolanlawson:
983343eb244970f76d096d90cd632d1407fce3db, cheers!

Issue #4915, opened 2016-02-21
Support the WebSQL adapter in Node


nolanlawson:
Lots of folks have trouble building LevelDB for various environments (Windows, IoT, etc.). Occasionally this had led them to the extreme measure of using something like [SQLdown](https://github.com/calvinmetcalf/SQLdown), which is unfortunate because it tends to be pretty inefficient due to using LevelDB-esque queries over SQLite.

To solve this problem, I recently reimplemented WebSQL in Node.js over [sqlite3](https://github.com/mapbox/node-sqlite3), and called it [node-websql](https://github.com/nolanlawson/node-websql). Using this, we can now offer our users an easier way to install PouchDB in Node, in case LevelDB is giving them troubles.

I anticipate this opens up PouchDB to more usage on resource-constrained devices, especially since SQLite is a well-established standard there, and typically easier to install than LevelDOWN (efforts like [leveldown-mobile](https://github.com/Level/leveldown-mobile) notwithstanding).

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/edcfd13bec4539fdf2f3d34f2fcc6b97499422d0

PR #4916, opened 2016-02-21
(#4915) - support WebSQL in Node via node-websql

nolanlawson:
Contains three major changes:

1. Modifications to the `websql` adapter to allow it to run in either Node or the browser. In particular, checks for `window.navigator` had to be made optional, and any manipulation of Blobs had to be made to support Buffers as well (see `preprocessAttachments.js`)
2. Changes to the Travis tests, to run the full PouchDB test suite against `node-websql`. I realize I could have made a separate repo for this and run my tests over there, but I think it's reasonable to add this to the PouchDB project, because it's just much simpler than maintaining a fork. The test is also pretty speedy.
3. Addition of a new `extras` API: `require('pouchdb/extras/websql')`, which allows users to use the WebSQL adapter in Node. I know we're planning on removing the `extras` API, but my current thinking about that is that the best approach will be a monorepo with separate modules (e.g. using [Lerna](https://www.npmjs.com/package/lerna), ala Babel), meaning that we would still want this adapter to be hosted in this repo.

I realize this is a big change to propose out of the blue, so I'm happy to discuss further, but the good news is that my weekend hacking has resulted in a fully working PouchDB adapter that directly uses SQLite in Node. :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,6 +51,9 @@ env:
   - CLIENT=node COMMAND=test
   - CLIENT=node LEVEL_PREFIX=foo_ COMMAND=test
 
+  # Test WebSQL in Node (using node-websql)
+  - CLIENT=node ADAPTER=websql COMMAND=test
+
   # Test against pouchdb-server
   - CLIENT=node SERVER=pouchdb-server COMMAND=test
   - CLIENT=selenium:firefox:41.0.1 SERVER=pouchdb-server COMMAND=test
--- a/TESTING.md
+++ b/TESTING.md
@@ -158,6 +158,10 @@ Some Level adapters also require a standard database name prefix (e.g. `riak://`
 
     LEVEL_PREFIX=riak://localhost:8087/
 
+To run the node-websql test in Node, run the tests with:
+
+    ADAPTER=websql
+
 ### Performance tests
 
 To run the performance test suite in node.js:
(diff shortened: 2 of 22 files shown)


nolanlawson:
Woohoo, it takes 6 min 38 seconds to run the node-websql test in Travis, whereas LevelDB is 6 min 9 seconds. :) Bet I could get that down lower if I do https://github.com/nolanlawson/node-websql/issues/3

nolanlawson:
Ah, coverage decreased. I need to figure out how to get it to ignore WebSQL for the purposes of the main test suite.

nolanlawson:
Coverage should be fixed now ^

daleharvey:
This is awesome, the only downside here is the number of new files, would like us to start looking at how to reduce them and the number of switches we do, but either way the new support gets preference, thanks https://github.com/pouchdb/pouchdb/commit/edcfd13bec4539fdf2f3d34f2fcc6b97499422d0

nolanlawson:
Yeah, I only did that because it made it easier to do the browser/node switching. Couldn't think of another way without duplicating code.

As an aside, about 95% of our browser-node switches right now are due to Blob/Buffer differences. It may behoove us to write a separate module called `bluffer` or something that normalizes the differences between the two.

nickcolley:
This is really cool! Thanks Nolan

PR #4924, opened 2016-02-22
(#4339) - Throw useful error when instance is destroyed rather than h…

nickcolley:
…anging

Something to note here is previously if you were using the HTTP adapter, it would return a 404 whereas the browser adapters will hang.

So this could technically be breaking functionality if someone was relying on that behaviour and they were *only* using PouchDB as a HTTP adapter.

This changes it so they all return the same result, a error message.

### DIFF
--- a/tests/integration/test.basics.js
+++ b/tests/integration/test.basics.js
@@ -73,6 +73,22 @@ adapters.forEach(function (adapter) {
       return new PouchDB(dbs.name).destroy({});
     });
 
+    it('4339 throw useful error if method called on stale instance', function () {
+      var db = new PouchDB(dbs.name);
+
+      return db.put({
+        _id: 'cleanTest'
+      }).then(function () {
+        return db.destroy();
+      }).then(function () {
+        return db.get('cleanTest');
+      }).then(function () {
+        throw new Error('.get should return an error');
+      }, function (err) {
+        should.equal(err instanceof Error, true, 'should be an error');
+      });
+    });
+
     it('destroy a pouch, with a promise', function (done) {
       new PouchDB(dbs.name, function (err, db) {
         should.exist(db);
--- a/src/adapter.js
+++ b/src/adapter.js
@@ -936,6 +936,7 @@ AbstractPouchDB.prototype.destroy =
       if (err) {
         return callback(err);
       }
+      self._destroyed = true;
       self.emit('destroyed');
       callback(null, resp || { 'ok': true });
     });
--- a/src/deps/adapterFun.js
+++ b/src/deps/adapterFun.js
@@ -31,6 +31,9 @@ function adapterFun(name, callback) {
     if (this._closed) {
       return Promise.reject(new Error('database is closed'));
     }
+    if (this._destroyed) {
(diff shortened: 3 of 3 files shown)


tests/integration/test.basics.js
```diff
@@ -73,6 +73,21 @@ adapters.forEach(function (adapter) {
...
+      }).then(function () {
+        return db.destroy();
+      }).then(function () {
+        return db.get('cleanTest');
+      }).catch(function (err) {
+        should.equal(err instanceof Error, true, 'should be an error');
+        done();
+      });
+    });
+
```
daleharvey: Since this is a promise based test I think we want to `return db.put` and not use `done` unless I am mistaken about something

daleharvey:
There is still an argument that people expect these instances to be reusable and I reworked the http adapter to make that possible (sounds like it was still a little broken) however its still a big amount of work to get that working across the other adapters and in the mean time being consistent with a clear error message is far better

I am :+1: on this, looks great, just a tiny note about the test format, thanks

tests/integration/test.basics.js
nolanlawson: Yes, if you're trying to verify that an error _was_ thrown, I like:

```js
somepromise.then(function () {
  throw new Error('expected an error!');
}, function (err) {
  should.equal(...);
});
```

nolanlawson:
Big +1 on this; I'm glad the fix ended up being only a few lines of code, but agree with Dale about the test.

nickcolley:
Thanks for the feedback, I've updated the test so hopefully this is good to go :)

tests/integration/test.basics.js
```diff
@@ -73,6 +73,20 @@ adapters.forEach(function (adapter) {
...
+      return db.put({
+        _id: 'cleanTest'
+      }).then(function () {
+        return db.destroy();
+      }).then(function () {
+        return db.get('cleanTest');
+      }).catch(function (err) {
+        should.equal(err instanceof Error, true, 'should be an error');
+      });
```
nolanlawson: It's a subtle difference, but this really ought to be

```js
.then(function () {
  throw new Error('expected an error');
}, function (err) {
  should.equal(err instanceof Error, true, 'should be an error');
});
```

Reason being: if the `db.get()` succeeds somehow, then the `catch()` function will never even run. It's not a big deal, though, and I can change it on the way in if you don't mind.

nickcolley: I'm not following why you'd throw an error in the test, surely if the get succeeds then the tests should fail since it should always error in this circumstance?

nickcolley: Sorry I missed this point because I was thinking of the catch behaviour when it's chained so I was confused why you'd be 'forcing' the catch the be called.

Infact here the error thrown if the get passes will not get caught as it's not chained, thanks for the tip.

daleharvey:
Awesome looks great, merged in https://github.com/pouchdb/pouchdb/commit/9549fda65e62e3806cc59b360a25ea4645d5d572

Issue #4903, opened 2016-02-18
fix btoa$1 error for non-ascii characters (such as Cyrillic)


unpete:
function `btoa$1(str)`  rise DOMException "The string to be encoded contains characters outside of the Latin1 range" if str contains non-ascii characters.
Replace `btoa(str)` to `btoa(unescape(encodeURIComponent(str)))` fixes this error

daleharvey:
Great catch, could you post any example code which highlights the issue?

unpete:
https://jsfiddle.net/oknosoft/dvjfzjev/

```javascript
var db = new PouchDB("https://fakepath.net/", {
	auth: {
  	username: "guest",
    password: "password"
    }						
});
```
all ok, but:

```javascript
var db = new PouchDB("https://fakepath.net/", {
	auth: {
  	username: "Иванов И.И.",
    password: "Секрет"
    }						
});
```
rises error

nolanlawson:
Could you open a PR with this change? Looks like the place to put the change would be [this line of code](https://github.com/pouchdb/pouchdb/blob/21e1346e11e611f1d312f468a711389038b04287/src/adapters/http/index.js#L150).

unpete:
Probably, the change in [that line of code](//github.com/unpete/pouchdb/blob/4903-btoa-outside-latin1/src/deps/binary/base64-browser.js#L6). PR will be added after the testing.

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/e45e7a04567025085a92a6ad0d95d59d68b84321

PR #4928, opened 2016-02-24
4903 btoa outside latin1

unpete:
#4903

### DIFF
--- a/src/deps/binary/base64-browser.js
+++ b/src/deps/binary/base64-browser.js
@@ -3,7 +3,7 @@ var thisAtob = function (str) {
 };
 
 var thisBtoa = function (str) {
-  return btoa(str);
+  return btoa(unescape(encodeURIComponent(str)));
 };
 
 export {
--- a/src/deps/binary/base64-browser.js
+++ b/src/deps/binary/base64-browser.js
@@ -3,7 +3,10 @@ var thisAtob = function (str) {
 };
 
 var thisBtoa = function (str) {
-  return btoa(unescape(encodeURIComponent(str)));
+  if (/[\u0080-\uFFFF]/.test(str)) {
+    str = unescape(encodeURIComponent(str));
+  }
+  return btoa(str);
 };
 
 export {


src/deps/binary/base64-browser.js
```diff
@@ -3,6 +3,9 @@ var thisAtob = function (str) {
 };
 
 var thisBtoa = function (str) {
+  if (/[\u0080-\uFFFF]/.test(str)) {
+    str = unescape(encodeURIComponent(str));
+  }
```
nolanlawson: is the `if` statement necessary? I think we can just call `unescape(encodeURIComponent())` every time, right?

Actually now that I think about it, is it more reasonable to ask users to do this themselves?

nolanlawson: I'd also be interested to know: if the user provides a string which is already `unescape(encodeURIComponent())`d, would this break for them or would it still work?

unpete: > is the if statement necessary?

imho, it will speed up code execution for a simple string

> Actually now that I think about it, is it more reasonable to ask users to do this themselves?

non-ascii characters can get on btoa() not only on authorization. [The "Unicode Problem"](https://developer.mozilla.org/en-US/docs/Web/API/WindowBase64/Base64_encoding_and_decoding#Solution_.232_.E2.80.93_rewriting_atob()_and_btoa()_using_TypedArrays_and_UTF-8)

>  if the user provides a string which is already unescape(encodeURIComponent())

double escaping will cause problems, but the rejection of the escaping throws DOMError

daleharvey:
So I think it would be our responsibility to encode the data properly, the encoding is a function of the transport (http) and we sit above that. However this looks like it is breaking a bunch of our tests so that definitely needs looked into

unpete:
If, before my issue was no problem, may be safer to leave the code as is. Way to circumvent non-ascii symbols authorization bugs can be described in the documentation

daleharvey:
We already test and handle unicode inside json and attachments so I dont think a this is best applied to the global `btoa`, I think it just needs done for the construction of the http header @ https://github.com/pouchdb/pouchdb/blob/master/src/adapters/http/index.js#L150

daleharvey:
Closed in favour of https://github.com/pouchdb/pouchdb/pull/4934

PR #4930, opened 2016-02-25
Fix Windows 10 Edge Cache Busting

nickcolley:
Adds Edge to the saucelabs setup, I've confirmed it failed without the cache busting.

Tried it in a VM, same failures, added the cache busting and no failures so fingers crossed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -77,6 +77,7 @@ env:
   - CLIENT=saucelabs:chrome COMMAND=test
   - CLIENT=saucelabs:safari:6 COMMAND=test
   - CLIENT="saucelabs:internet explorer:10:Windows 8" COMMAND=test
+  - CLIENT="saucelabs:MicrosoftEdge" COMMAND=test
   - CLIENT="saucelabs:internet explorer:10:Windows 8" ADAPTERS=memory COMMAND=test
   - CLIENT="saucelabs:iphone:8.4:OS X 10.11" ADAPTERS=fruitdown GREP=suite2 INVERT=true COMMAND=test
   - CLIENT="saucelabs:iphone:8.4:OS X 10.11" ADAPTERS=fruitdown GREP=suite2 COMMAND=test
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -77,6 +77,7 @@ env:
   - CLIENT=saucelabs:chrome COMMAND=test
   - CLIENT=saucelabs:safari:6 COMMAND=test
   - CLIENT="saucelabs:internet explorer:10:Windows 8" COMMAND=test
+  - CLIENT="saucelabs:MicrosoftEdge:20.10240:Windows 10" COMMAND=test
```
daleharvey: Did you have problems with the version or did you pick that from the local build you tested? if we could leave off the version here and pick up latest stable it would be ideal

nickcolley: No that's cool to change I think, if we put just `20` will it automatically update or do we need to use `latest`?

daleharvey:
This looks great, it looks like edge is failing in travis, but happy to take a locally verified fix and we can add edge to allowed failures for now and see if someone wants to get it green

nickcolley:
Okay let's do that, think it's more important to fix the caching issues :+1:

daleharvey:
Awesome, just one issue here, entry for Edge needs to be in both the matrix and the allowed failures list, the allowed failures is a list of things in the matrix that are allowed to fail, on their own they dont run

nickcolley:
Looks like the failures are not to do with Travis, I'm seeing them in a VM also

daleharvey:
Thats a bonus, easier to debug when they are reproducable, cheers for the fix :+1: I restarted the intermittents so we can merge when green, thanks for this :)

daleharvey:
Sweet all green, https://github.com/pouchdb/pouchdb/commit/94d1fe0583ba3f7a2929067849c33387c5b6d654

nolanlawson:
:+1:, appreciate the new Travis test; it's been sorely needed for awhile.

PR #4934, opened 2016-02-29
(#4903) - Escape unicode in auth header

daleharvey:


daleharvey:
@unpete Could you test this and confirm that it is the correct fix? Thanks

unpete:
Ok, this fix corrects a problem with the authorization

daleharvey:
Awesome thanks, will take as a :+1: but leave it for a while before merging to lets others to take a look

daleharvey:
And merged in https://github.com/pouchdb/pouchdb/commit/e45e7a04567025085a92a6ad0d95d59d68b84321, thanks

PR #4948, opened 2016-03-03
(#XXXX) - Dont need to split up test runs anymore

daleharvey:
We speed up the test suite quite a lot recently, hoping its a little simpler and faster to run these together so there is less startup time


nolanlawson:
Nice! +1

daleharvey:
Sweet all green, https://github.com/pouchdb/pouchdb/commit/3151a06ca11d4886292dcd86eafd08d87abe449d

PR #4961, opened 2016-03-07
Correct parameter names in replication example

adamlofts:
Use the same parameter names in the example as in the docs below. Also make clear that a paused event is emitted when the sync is up to date.


daleharvey:
This is great, thanks a lot for these improvements, ammended commit message to match guidelines and merged - https://github.com/pouchdb/pouchdb/commit/1d0a37061d3aae350cc7470a99e516d9ec073cbf

Thanks :)

PR #4962, opened 2016-03-07
(#4480) - Load external resources from https

daleharvey:
So https://pouchdb.com/ is up and running (having issues in safari but looking into them) but mixed content will block these resource from loading, all the https versions are available

### DIFF
--- a/docs/_layouts/default.html
+++ b/docs/_layouts/default.html
@@ -20,7 +20,7 @@
       var s = document.getElementsByTagName('script')[0]; s.parentNode.insertBefore(ga, s);
       })();
     </script>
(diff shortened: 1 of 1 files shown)


janl:
+1 unless you want to use `//` to allow `http` and `https`

janl:
in fact, +1 to `//`, so people working on the site locally don’t have to set up ssl

daleharvey:
Yup thats a great point, thanks, I switched to `//` and merged, cheers - https://github.com/pouchdb/pouchdb/commit/1a099485f47060ee28c22d042e73b2225bb5327a

PR #4992, opened 2016-03-23
Move websql as optional dependency

### LINKED ISSUE #4991: The mandatory websql dependency causes issues in some environments
Currently, the websql package is listed as a dependency and means that if that package (and all the packages it depends on) doesn't build, the installation of pouchdb package fails.

We are using pouchdb together with https://github.com/jxcore/jxcore and when using the SpiderMonkey version, the websql package doesn't install (because sqlite3 doesn't build against the SpiderMonkey engine).

A proposal would be to make websql an optional dependency similar to leveldown so that this issue could be avoided.

vjrantal:
This allows installing pouchdb in an environment where websql build would
not work and makes it possible to optionally omit websql installation by
passing `--no-optional` to `npm install`.

Fixes #4991.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,14 +37,14 @@
     "spark-md5": "2.0.2",
     "sublevel-pouchdb": "1.0.0",
     "through2": "2.0.1",
-    "vuvuzela": "1.0.2",
-    "websql": "0.1.0"
+    "vuvuzela": "1.0.2"
   },
   "jspm": {
     "main": "dist/pouchdb.js"
   },
   "optionalDependencies": {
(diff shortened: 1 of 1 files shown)


nickcolley:
Thanks @vjrantal this looks good to me, thanks for the write up also :)

Will let someone else have a look too :+1:

daleharvey:
Yeh I am :+1: on this, will wait to merge until everything else is green and do one big merge, thanks

nolanlawson:
Sweet, thanks. bf9acf51353be93fcf0bd0d7205b8ef134480176

Issue #4592, opened 2015-11-19
db.sync() pause event does not forward errors sent by db.replicate() pause events


tohagan:
The replicate `backOff()` function can fire a `'pause'` event with an error arg:
  https://github.com/pouchdb/pouchdb/blob/master/lib/replicate/backoff.js#L38

However this error is not sent to the db.sync() 'pause' events.

```
  function pushPaused() {
    self.pushPaused = true;
    if (self.pullPaused) {
      self.emit('paused');
    }
  }
  function pullPaused() {
    self.pullPaused = true;
    if (self.pushPaused) {
      self.emit('paused');
    }
  }
```

Code should be ...

```
  function pushPaused(err) {
    if (err) self.emit('error', err);
    self.pushPaused = true;
    if (self.pullPaused) {
      self.emit('paused');
    }
  }
  function pullPaused(err) {
    if (err) self.emit('error', err);
    self.pullPaused = true;
    if (self.pushPaused) {
      self.emit('paused');
    }
  }
```

Arguably it perhaps should also report if 'push' or 'pull' sent the error.

tohagan:
I had a look at the option of sending the error via the existing pause event like replicate() does but it get messy since you only want to fire a single 'pause' when both push and pull have paused so you might miss getting an important error. For example: push/pull Replication objects could fire any number of pause/active events with errors and yet never fire a pause event because they were not paused at the same time.  Safer to just emit errors as they happen. This perhaps alters the current API in that some devs may have assumed that only a single error can be emitted by sync() which is the current behaviour.

daleharvey:
So thinking about this, not sure this is an error. `paused` and `active` are sugar to give the state of the sync as a whole, not the individual replications and as an api it doesnt really make sense for it to do so, if you need to determine the error state of the individual replications you can do so with `sync.push.on('paused', ....` + `sync.pull.on('paused', ....`

daleharvey:
So we can discuss whether we want to improve the api here or not, but its definitely not a bug, reclassifying

dharders:
@daleharvey I think the confusion stems from the [documentation](https://pouchdb.com/api.html#sync) example code stating that db.sync.on('paused') handler is passed an err object, which is not true. It never gets passed. 

Is the db.sync.push, db.sync.pull a stable/public API ? 

[...]

Issue #3714, opened 2015-04-09
live replication with retry should emit 'paused' event when already paused && error


dholth:
If replication is paused and the remote changes feed is disconnected, the live replication with retry does not appear to fire any event. Instead, it should fire some event, maybe just an extra "paused" event with the connection error. This would be used to tell the user that they are offline.

If replication stops while it is active the paused event does include the error, and appropriate feedback can be given.

dholth:
Oh, it emits requestError. I'll just use that.

daleharvey:
@dholth hrm the replication should fire a paused event with an error, if it doesnt its a bug, for fixing the requestlisteners it should be based on this and not requestError, which is just spuriously fired for failing requests

dholth:
During a live replication with retry, if 'paused' and the remote changes feed is lost, the replication emits only the undocumented 'requestError' event. Instead, it should just emit ('paused', err). This matches what happens when replication is active and there is a similar connection error.

nolanlawson:
As commented in https://github.com/pouchdb/pouchdb/issues/3716, I think we should just follow the documentation and actually emit the error in `paused`. From the docs:

> This event fires when the replication is paused, either because a live replication is waiting for changes, or replication has temporarily failed, with err, and is attempting to resume.

`requestError` is undocumented, so let's let it die a silent death.

nolanlawson:
Er wait, this happens when it's already `paused`? Okay, so maybe we do need a separate event then. It would be odd if `paused` fired twice in a row with two different errors.

nolanlawson:
> This matches what happens when replication is active and there is a similar connection error.

The `paused` event indicates a change from `active` to `paused`, so no, in my mind it's not the same thing. :) Emitting `paused` multiple times is confusing; let's stick with `requestError`.

dholth:
In my mind it does make sense, paused (err), paused (even more paused,
err). requestError is a silly event.

It would make sense if 'paused' never included an error, otherwise you
need two kinds of error handling, one when active and one when not
active. What about using the regular error event, or is that reserved
for errors that make us give up? Or some new event on the off chance
that requestError is not the only retry-able error that we care about in
this general case? on('hiccup', function(err) { ... })

On Sun, Apr 12, 2015, at 12:21 PM, Nolan Lawson wrote:
>> This matches what happens when replication is active and there is a
>> similar connection error.


> The paused event indicates a change from active to paused, so no, in
[...]

PR #4998, opened 2016-03-24
Fix documentation, now we are same comments to sync or replicate

kosssi:


docs/_includes/api/replication.html
```diff
@@ -39,12 +39,12 @@
   retry: true
 }).on('change', function (info) {
   // handle change
-}).on('paused', function (err) {
+}).on('paused', function () {
   // replication paused (e.g. replication up to date, user went offline)
```
daleharvey: We should add `(err)` to the documentation of both the paused events, it is passed in the event that the replication was paused due to an error

daleharvey:
Having these be consistent would be ideal, good catch. However we shouldnt remove the documentation for the `err` parameter as commented, if you want to fix that then :+1: thanks :)

docs/_includes/api/replication.html
kosssi: Why it's not error event? You know where is the code?

daleharvey:
Thanks so much @kosssi, these changes look great, :+1: and merged in https://github.com/pouchdb/pouchdb/commit/10674e843e9401b13056a58541f9a5107238faf8

adamlofts:
I don't think the Sync object will ever emit an error object in the paused event. Arguably this is a bug in the code rather than the docs.

The relevant lines are:

https://github.com/pouchdb/pouchdb/blob/master/src/sync.js#L67

and

https://github.com/pouchdb/pouchdb/blob/master/src/sync.js#L73

kosssi:
@adamlofts it's not here? https://github.com/pouchdb/pouchdb/blob/99c24a5861e19ac5ae78b35d71c9845a449ba9a0/src/replicate/backoff.js#L36

adamlofts:
@kossi From my memory of looking at this before that fires paused on the replication object with the error but it does not get passed through to the paused event on the Sync object. Could be wrong tho. It was a couple of weeks ago that I looked at it.

daleharvey:
Yeh I think you are right, thats a bug, good catch

kosssi:
I think it's this bug https://github.com/pouchdb/pouchdb/issues/4592 and https://github.com/pouchdb/pouchdb/issues/3714

It's really problematic.

kosssi:
I will test to fix it ;)

PR #5008, opened 2016-03-30
Added Hoodie to the /users.html

turcapat:
Added Hoodie as a user of Pouch DB.


nickcolley:
Hey @turcapat thanks for the PR!

Do you think you could resize the hoodie logo so it's around 200-300 px wide? Since we're only showing it at 100px wide that'd account for high density screens (2-3x) and reduce the file size overall.

turcapat:
Of course :) Done.

nickcolley:
@turcapat sorry I wasn't very clear, the image can have a width of 100px in the site but the actual file should be 200px wide.

turcapat:
Ok, now I understand :) What about now ?

nickcolley:
Thanks @turcapat that's great I've merged it in! ada58b1864662b99c8

I squashed your commits together as per our convention https://github.com/pouchdb/pouchdb/blob/master/CONTRIBUTING.md#guide-to-contributions don't worry the first one is free :P

You'll see it live next time we deploy the site.

Issue #5010, opened 2016-03-30
WebKit Nightly broken due to CouchDB-related ETag bug


nolanlawson:
...but not broken in Safari latest. So we have a chance to catch this bug before it goes live.

~~I haven't been able to write a small reproducible test case yet. It's very tricky to nail down.~~ _see test in WebKit bug below_

To reproduce, run the tests with `http://127.0.0.1:8000/tests/integration/?grep=suite2%20test\.replication\.js-http-http` and notice that once you get into the `#3999` you start to get test failures. This only occurs in WebKit Nightly and Safari Technical Preview.

Note that this is not IndexedDB- or WebSQL- related. It seems to be HTTP-related. Also note that the bug is reproducible in CouchDB 1.6 but not PouchDB Server, so it may be related to the ETag (PouchDB Server doesn't use it).

nolanlawson:
Managed to reproduce and filed an issue on WebKit: https://bugs.webkit.org/show_bug.cgi?id=156048

nickcolley:
@nolanlawson thanks for putting this together! Hope it's not too hard for them to fix :+1:

nolanlawson:
Had a chat about this with @janl and @rnewson in IRC today. Seems it's more of a CouchDB bug than a WebKit bug. In the meantime we can workaround it by enabling the `_nonce` hack for Safari GETs.

I think we should definitely implement the workaround before Safari ships, because CouchDB 1.6.x is going to be out there in the wild for some time.

nolanlawson:
So for anybody interested in fixing this, the fix would be to change [the cache-busting logic](https://github.com/pouchdb/pouchdb/blob/94d1fe0583ba3f7a2929067849c33387c5b6d654/src/deps/ajax/prequest-browser.js#L11-L16) so that it applies to both IE/Edge and Safari, regardless of whether the request was a GET or a POST.

schamp:
I could take a stab at it, if it's as straightforward as it looks, it would be good practice contributing.

nolanlawson:
@schamp That's awesome! :) I updated the instructions above to make it a little clearer, but to confirm, the old behavior would be:

| Method | Safari | IE/Edge |
| --- | ---- | --- |
| GET | no cache-bust | cache-bust |
| POST | cache-bust | no cache-bust |

Whereas the new behavior would be:

| Method | Safari | IE/Edge |
| --- | ---- | --- |
| GET | cache-bust | cache-bust |
[...]

PR #5011, opened 2016-03-31
(#5010) update cache-busting for new version of Safari

schamp:
I don't have my full testing environment set up with Safari Technology Preview, but if someone else wants to run that test while I am getting it set up, perhaps this can be closed more quickly.

I'll post an update myself once I can run the tests successfully.

### DIFF
--- a/src/deps/ajax/prequest-browser.js
+++ b/src/deps/ajax/prequest-browser.js
@@ -1,6 +1,6 @@
 import ajaxCore from './ajaxCore';
 
-function ajax(opts, callback) {
+function ajax (opts, callback) {
 
   // cache-buster, specifically designed to work around IE's aggressive caching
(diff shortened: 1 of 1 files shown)


nolanlawson:
OK, I just tested this against the PouchDB test suite. It turns out it passes all the tests except for a single blob/attachments test, which I can nail down to the fact that Safari (and Chrome in fact) return 404s if you try to fetch `blob:` URLs with a query param added. So the fix will be to chck for `blob:` in the URLs and avoid the nonce in those cases.

nolanlawson:
Issues filed on Chromium and WebKit FWIW: https://bugs.chromium.org/p/chromium/issues/detail?id=600074 https://bugs.webkit.org/show_bug.cgi?id=156125

daleharvey:
Merged in https://github.com/pouchdb/pouchdb/commit/df0841ab38794fc2418f6a74a4e545cc245c6e37

PR #5016, opened 2016-04-01
4979 sqlite key fix

### LINKED ISSUE #4979: Add `key` support for SQLite Plugin
Hi,

I am using https://github.com/litehelpers/Cordova-sqlcipher-adapter plugin for sqlite support with encryption and I am trying the following code and PouchDB is not opening db as encrypted and displayed the following on console. 

> OPEN database: _pouch_userDB.txt
> open full db path: /var/mobile/Containers/Data/Application/99BA28C6-03BB-4636-9224-9EF708B79C94/Documents/_pouch_userDB.db
> Open DB with NO encryption
> 

Here is the code !
```
      var db2 = new PouchDB('userDB.db', {adapter: 'websql', key: 'password'});
      db2.info().then(console.log.bind(console));
      db2.put({_id: '1', name: 'username'})
        .then(usr => {
          console.log(`${usr} saved`);
        }).
        catch(ex=> {
          console.log(`${ex} occurred`)
        });

```

Where as the following works fine in the same app

```
      var db = window.sqlitePlugin.openDatabase({name: "my.db", key: "your-password-here"});
      db.transaction(function (tx) {
        tx.executeSql('DROP TABLE IF EXISTS test_table');
        tx.executeSql('CREATE TABLE IF NOT EXISTS test_table (id integer primary key, data text, data_num integer)');

        tx.executeSql("INSERT INTO test_table (data, data_num) VALUES (?,?)", ["test", 100], function (tx, res) {
          console.log("insertId: " + res.insertId + " -- probably 1");
          console.log("rowsAffected: " + res.rowsAffected + " -- should be 1");

          tx.executeSql("select count(id) as cnt from test_table;", [], function (tx, res) {
            console.log("res.rows.length: " + res.rows.length + " -- should be 1");
            console.log("res.rows.item(0).cnt: " + res.rows.item(0).cnt + " -- should be 1");
          });

        }, function (e) {
          console.log("ERROR: " + e.message);
        });
      });
```

And here is the console output!

> OPEN database: my.db
> open full db path: /var/mobile/Containers/Data/Application/99BA28C6-03BB-4636-9224-9EF708B79C94/Documents/my.db
> Open DB with encryption

**It will be really helpful if someone please point out the mistake. What am I missing to pass to PouchDB constructor ?**

mhartington:
Closes #4979

Cordova's sqlcipher-adapter requires a key to be passed in order
to encrypt the DB. This PR allows for not only the key to be passed,
but for any future options that could be added.

### DIFF
--- a/src/adapters/websql/index.js
+++ b/src/adapters/websql/index.js
@@ -120,6 +120,10 @@ function WebSqlPouch(opts, callback) {
 
   api._docCount = -1; // cache sqlite count(*) for performance
   api._name = opts.name;
+
+  opts['version'] = POUCH_VERSION;
+  opts['size'] = size;
+  console.log(opts);
   var openDBResult = openDB(opts);
   if (openDBResult.error) {
     return websqlError(callback)(openDBResult.error);
--- a/src/adapters/websql/index.js
+++ b/src/adapters/websql/index.js
@@ -120,12 +120,13 @@ function WebSqlPouch(opts, callback) {
 
   api._docCount = -1; // cache sqlite count(*) for performance
   api._name = opts.name;
-
+  console.log(opts)
   var openDBResult = openDB({
     name: api._name,
     version: POUCH_VERSION,
     description: api._name,
     size: size,
+    key: opts.key,
     location: opts.location,
     createFromLocation: opts.createFromLocation,
     androidDatabaseImplementation: opts.androidDatabaseImplementation
--- a/src/adapters/websql/index.js
+++ b/src/adapters/websql/index.js
@@ -123,7 +123,7 @@ function WebSqlPouch(opts, callback) {
 
   opts['version'] = POUCH_VERSION;
   opts['size'] = size;
-  console.log(opts);
+  
   var openDBResult = openDB(opts);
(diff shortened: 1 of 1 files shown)


nolanlawson:
Oops, looks like a lint error?

```
/home/travis/build/pouchdb/pouchdb/src/adapters/websql/index.js
   86:5  error  'POUCH_VERSION' is defined but never used  no-unused-vars
  117:7  error  'size' is defined but never used           no-unused-vars
```

nolanlawson:
I also noticed that the latest version of the SQLite Plugin is throwing errors about requiring `location` or `iosDatabaseLocation`. So it looks like this PR will also fix that. Thanks a bunch!

I think I'd feel a bit more secure if we cloned the object before passing it in to `openDB()`, but that's a change I can make on the way in. Thanks a bunch; I'll make that change and then merge, now that it's clear Travis is green.

nolanlawson:
d6876ab050ef5391bd64620d2406fb389c957aea and 45a2ee55cb85fb94a4b7874b435e0db3f3554b33

Issue #4963, opened 2016-03-08
Replication of documents with many attachments is unstable


ssured:
I have a database which contains map tiles at different zoom levels. All these tiles are stored in a single doc, resulting in a document with around 200 attachments of average 10kb each. My DB stores 2000 documents, of which 50 contain 200 attachments. All others documents hold 0 or 1 attachment of 100kb size. Total DB size is 80mb.

Replicating (parts of) these databases to other databases seems instable, sometimes documents with many attachments are not written to the destination, but no error is raised.

I have tested quite some scenarios, these are my observations: (all against CouchDB 1.6.1 instances)

 - documents with little or no attachments always sync perfectly
 - syncing using the futon interface (couchdb-couchdb) works fine
 - replication from localhost is more stable than a remote host, maybe a slow connection makes the problem more apparent
 - setting `require('http').globalAgent.maxSockets = 10;` increase reliability, which makes me think pouchdb maybe does too many concurrent requests
 - `PouchDB.replicate(fromDB, toDB, {batch_size: 2, batches_limit: 2})` works much more stable than without the extra options.
 - I get the same problems when using `pouchdb-dump-cli`; the dump file will not contain all documents
 - running `pouchdb-dump-cli > dump.data` multiple times for the same db results in different `dump.data` files, which is very weird

My gut feeling is that the replicator fetches all attachments at once, which leads to a massive data transfer peak, which might be unstable. Probably some attachments fail to load, which silently breaks something more downstream.

I don't know where to start debugging. For now I am working with `{batch_size: 2, batches_limit: 2}` which seems most stable.

nolanlawson:
Thank you; this is very valuable research! Can you please confirm which browsers/environments you've tested this on? Sounds like Node and LevelDB, but I'm not sure e.g. which Node version, which OS, etc.

Some of the issues you raise are unfortunately not solveable due to CouchDB limitations. For instance, we cannot download attachments using the more efficient `multipart` format because they ship `gzip` and we can't reasonably decode that in JS. However, for uploads it is possible for us to directly upload blobs.

I believe most of these issues could be addressed by #3913, which is on [my personal roadmap](https://gist.github.com/nolanlawson/7953d9aaebbbe8cb382b) but I can't promise when it will get implemented. After that feature is shipped, all attachments should be uploaded/downloaded directly as binary.

That said, we obviously should try to fix the base64 case as well if possible.

ermouth:
[...]

Issue #3156, opened 2014-12-12
Using pre-calculating md5 hash values for CouchDB attachments isn't a good idea


yaronyg:
__Executive Summary__: Please remove use of pre-calculated md5 hashes for attachments, especially in test.attachments.js.

#### Long Winded Version
CouchDB, as far as I understand it, doesn't guarantee anything about the digest returned for an attachment on a document other than if an attachment with the same name on the same document should change its content then the digest value will change. My understanding is that the guarantee is modeled after etags in HTTP/1.1 which are similar in their very restricted and scoped uniqueness guarantee.

Yet, in our tests, specifically in test.attachments.js, we depend on attachments on completely different CouchDB servers having the same digest. This blew up when we tested against the Windows release of CouchDB #3152 whose digest values were different than those on Linux.

The reason, I suspect, for the digest differences in this case is that the digest is calculated after compressing the content. So if the compression algorithm or its settings differ at all between two CouchDB servers then their digests will be different. This means that if the default compression settings in CouchDB ever change on Linux then all these tests will break.

So, put another way, if you are writing a test of attachments that depends on a pre-calculated md5 hash then you can be sure that you aren't testing what you think you are testing. :)

daleharvey:
Yeh I dont see this as being an issue we want to clean up, we can add the windows hashes if needed

PR #5018, opened 2016-04-01
(#2674) - avoid unneeded attachment GET requests

### LINKED ISSUE #2674: Use `atts_since` in the replication algorithm
Currently we don't do this, which I guess means we read in every attachment for every revision? Bad for perf.

From the [TouchDB wiki](https://github.com/couchbaselabs/TouchDB-iOS/wiki/Replication-Algorithm):

> Also use the “?atts_since” query parameter to pass a list of revisions that the target already has, so the source can optimize by not including the bodies of attachments already known to the target.

gr2m:
This is Work In Progress, it still needs test, just want to see if it is going in the right direction

ping @nolanlawson @janl

--- 

closes #2674

### DIFF
--- a/src/replicate/getDocs.js
+++ b/src/replicate/getDocs.js
@@ -90,12 +90,14 @@ function getDocs(src, target, diffs, state) {
       return Promise.all(bulkGetResponse.results.map(function (bulkGetInfo) {
         return Promise.all(bulkGetInfo.docs.map(function (doc) {
           var remoteDoc = doc.ok;
+
+          if (doc.error) {
+            // when AUTO_COMPACTION is set, docs can be returned which look
+            // like this: {"missing":"1-7c3ac256b693c462af8442f992b83696"}
+            ok = false;
+          }
+
           if (!remoteDoc || !remoteDoc._attachments) {
-            if (remoteDoc.error !== undefined) {
-              // when AUTO_COMPACTION is set, docs can be returned which look
-              // like this: {"missing":"1-7c3ac256b693c462af8442f992b83696"}
-              ok = false
-            }
             return remoteDoc;
           }
 
--- a/src/replicate/getDocs.js
+++ b/src/replicate/getDocs.js
@@ -1,8 +1,47 @@
 import clone from '../deps/clone';
+import flatten from '../deps/flatten';
 import Promise from '../deps/promise';
 
 function isGenOne(rev) {
   return /^1-/.test(rev);
+}
+
+function fileHasChanged(localDoc, remoteDoc, filename) {
+  return !localDoc._attachments ||
+         !localDoc._attachments[filename] ||
+         localDoc._attachments[filename].digest !== remoteDoc._attachments[filename].digest;
+}
+
+function getDocAttachments(db, doc) {
+  var filenames = Object.keys(doc._attachments);
+  return Promise.all(filenames.map(function (filename) {
+    return db.getAttachment(doc._id, filename);
+  }));
+}
+
+function getDocAttachmentsFromTargetOrSource(target, src, doc) {
+  var doCheckForLocalAttachments = src.type() === 'http' && target.type() !== 'http';
+  var filenames = Object.keys(doc._attachments);
+
+  if (!doCheckForLocalAttachments) {
+    return getDocAttachments(src, doc);
+  }
+
+  return target.get(doc._id).then(function (localDoc) {
+    return Promise.all(filenames.map(function (filename) {
+      if (fileHasChanged(localDoc, doc, filename)) {
+        return src.getAttachment(doc._id, filename);
+      }
+
+      return target.getAttachment(localDoc._id, filename);
+    }));
+  }).catch(function (error) {
+    if (error.status !== 404) {
+      throw error;
+    }
+
+    return getDocAttachments(src, doc);
+  });
 }
 
 function createBulkGetOpts(diffs) {
@@ -19,9 +58,7 @@ function createBulkGetOpts(diffs) {
 
   return {
     docs: requests,
-    revs: true,
-    attachments: true,
-    binary: true
+    revs: true
   };
 }
 
@@ -31,7 +68,7 @@ function createBulkGetOpts(diffs) {
 // changes to "cancelled", then the returned promise will be rejected.
 // Else it will be resolved with a list of fetched documents.
 //
-function getDocs(src, diffs, state) {
+function getDocs(src, target, diffs, state) {
   diffs = clone(diffs); // we do not need to modify this
 
   var resultDocs = [],
@@ -50,16 +87,34 @@ function getDocs(src, diffs, state) {
       if (state.cancelled) {
         throw new Error('cancelled');
       }
-      bulkGetResponse.results.forEach(function (bulkGetInfo) {
-        bulkGetInfo.docs.forEach(function (doc) {
-          if (doc.ok) {
-            resultDocs.push(doc.ok);
-          } else if (doc.error !== undefined) {
-            ok = false;
+      return Promise.all(bulkGetResponse.results.map(function (bulkGetInfo) {
+        return Promise.all(bulkGetInfo.docs.map(function (doc) {
+          var remoteDoc = doc.ok;
+          if (!remoteDoc || !remoteDoc._attachments) {
+            if (remoteDoc.error !== undefined) {
+              // when AUTO_COMPACTION is set, docs can be returned which look
+              // like this: {"missing":"1-7c3ac256b693c462af8442f992b83696"}
+              ok = false
+            }
+            return remoteDoc;
           }
-          // else: when AUTO_COMPACTION is set, docs can be returned which look
-          // like this: {"missing":"1-7c3ac256b693c462af8442f992b83696"}
-        });
+
+          return getDocAttachmentsFromTargetOrSource(target, src, remoteDoc).then(function (attachments) {
+            var filenames = Object.keys(remoteDoc._attachments);
+            attachments.forEach(function (attachment, i) {
+              var att = remoteDoc._attachments[filenames[i]];
+              delete att.stub;
+              delete att.length;
+              att.data = attachment;
+            });
+
+            return remoteDoc;
+          });
+        }));
+      }))
+
+      .then(function (results) {
+        resultDocs = resultDocs.concat(flatten(results).filter(Boolean));
       });
     });
   }
--- a/src/replicate/getDocs.js
+++ b/src/replicate/getDocs.js
@@ -15,7 +15,7 @@ function fileHasChanged(localDoc, remoteDoc, filename) {
 function getDocAttachments(db, doc) {
   var filenames = Object.keys(doc._attachments);
   return Promise.all(filenames.map(function (filename) {
-    return db.getAttachment(doc._id, filename);
+    return db.getAttachment(doc._id, filename, {rev: doc._rev});
   }));
 }
 
--- a/src/replicate/replicate.js
+++ b/src/replicate/replicate.js
@@ -151,7 +151,7 @@ function replicate(src, target, opts, returnValue, result) {
   }
 
   function getBatchDocs() {
-    return getDocs(src, currentBatch.diffs, returnValue).then(function (got) {
+    return getDocs(src, target, currentBatch.diffs, returnValue).then(function (got) {
       currentBatch.error = !got.ok;
       got.docs.forEach(function (doc) {
         delete currentBatch.diffs[doc._id];


gr2m:
I can reproduce the issue locally with `AUTO_COMPACTION=true CLIENT=node npm test` – but can someone explain to me what `AUTO_COMPACTION=true` does exactly, how it changes tests?

daleharvey:
So the tests with `AUTO_COMPACTION` set a global option on PouchDB that means documents will automatically compact on every write, so basically there are never any old revisions, only the current documents + conflicts stored, its essentially the same as calling `compact()` after every write

nolanlawson:
Ah yeah if your test relies on auto-compaction being false, then you need to wrap your test in an if-check to make sure that it's not run for the auto-compaction tests. there are examples of this in `test.compact.js`.

gr2m:
I didn’t write any tests yet. My code changes broke existing tests. I’m a bit confused. I’ll look into it of course but don’t know where to start. If you have any idea / pointers, please let me know

gr2m:
okay that _should_ fix the issue https://github.com/pouchdb/pouchdb/pull/5018/commits/35e10b7d35a9109cb89117fbc979710dfe7c8e89, I’ve missed the skipping of undefined docs here: https://github.com/pouchdb/pouchdb/pull/5018/files#diff-3b1908d46435f961119df0ed4762ec7fL54

yaronyg:
Did anyone figure out how revpos is supposed to work with this? Is revpos obsolete now that we have digest? Or is revpos only meaningful in the context of a particular server instance? In other words if my copy of the PouchDB DB has revpos X for a particular attachment and someone else's copy of the same PouchDB DB has revpos Y for the same version of the same attachment, is that o.k.? I've always been a bit confused by revpos.

nolanlawson:
This looks perfect to me, although I'm kinda surprised it's passing CI, since this contains a semicolon-less style and our eslint config is supposed to require semicolons.

@yaronyg This has nothing to do with revpos. It abandons the concept of revpos entirely. Instead, it just checks to see if there is an attachment available locally with the same digest as the thing we were going to fetch from the remote database, and in that case it skips fetching it.

I will note that this PR will not work for `text/plain` attachments, since those have md5sums generated based on Erlang strings, but that's a minor edge case and the end result is that the replication algorithm will be slightly less efficient for plaintext attachments – not that it will fail entirely. So this PR has a :+1: from me, although I'd like to see the semicolon issue fixed and ideally some more comments to explain the optimization. (Also no worries for a second Travis green checkmark; this has already passed once!)

src/replicate/getDocs.js
```diff
@@ -1,10 +1,49 @@
...
+function fileHasChanged(localDoc, remoteDoc, filename) {
+  return !localDoc._attachments ||
+         !localDoc._attachments[filename] ||
+         localDoc._attachments[filename].digest !== remoteDoc._attachments[filename].digest
+}
+
+function getDocAttachments(db, doc) {
+  var filenames = Object.keys(doc._attachments)
+  return Promise.all(filenames.map(function (filename) {
+    return db.getAttachment(doc._id, filename);
```
nolanlawson: We could also check all non-leaf non-deleted revisions here, but for now I think this is fine. That might be overkill anyway.

janl:
@yaronyg haven’t had time to dig into revpos yet.

Could you give this a spin maybe to see if it helps your use-cases? Thank you! :)

yaronyg:
@janl Sad to say but I'm a few weeks out before I have a solid test case I can use to test this. But given that most of what we do is move around binary attachments anything that stops them from being pulled down when they haven't changed is a huge win!

gr2m:
I added all missing semicolons. Wow, I missed quite a few, that was not intended. 
I’ve also started this PR to make sure that `npm run eslint` will catch missing semicolons in future:
https://github.com/pouchdb/pouchdb/pull/5036

gr2m:
@nolanlawson I suggest we create a follow up issue for this? 

> I will note that this PR will not work for text/plain attachments, since those have md5sums generated based on Erlang strings, but that's a minor edge case and the end result is that the replication algorithm will be slightly less efficient for plaintext attachments – not that it will fail entirely. So this PR has a :+1: from me, although I'd like to see the semicolon issue fixed and ideally some more comments to explain the optimization. (Also no worries for a second Travis green checkmark; this has already passed once!)

jo:
Please note that the attachment digest in Apache CouchDB is [generated from compressed file](https://issues.apache.org/jira/browse/COUCHDB-2251) and therefore depends on the compression configured (Snappy, deflate_N). If we adopt the digest on pull replication (what I think we do) there should be no problem. Also the `text/plain` attachment digest generation should not play a role here in that case. If thats not the case (digests are generated locally on pull replication) this would indeed render this optimisation useless.

gr2m:
I guess I should add tests to it, right? Any pointers on how I could measure the same binary attachment does not requested twice?

daleharvey:
So this is a little hacky, but we have a lot of tests depending on it and if we change how we do it we will just have to change all the tests at the same time, but I think an idea to test this is to overwrite the exposed ajax handler, there is an example @ https://github.com/pouchdb/pouchdb/blob/master/tests/integration/test.replication.js#L4059

This can then count the number of requests made to the attachment and you can assert that there are less requests (the test should fail before, pass after), how does that sound?

yaronyg:
I keep thinking about this PR and about digest. We have visited this issue before, for example, [here](https://github.com/pouchdb/pouchdb/issues/3156). In that bug we ran into a situation that because of how Windows works the digests it generates for its attachments are not the same as the digests generated for the same content on other platforms. The work around in that case was that we added the Windows digest values to our test infrastructure.

But what it means is that this optimization won't work if used against a Windows box or against any CouchDB box where the admin has used the publicly supported admin APIs to change their digest configuration. See [here](http://docs.couchdb.org/en/stable/config/misc.html#attachments) for the options.

This is so fragile it makes me sad.

If we had used revpos instead we wouldn't have these problems. But I recognize that revpos has its own issue that you can end up in a situation where two machines both have the same version of the same attachment but at different rev positions and so there will be a sync even though it isn't needed. The code here would avoid that problem.

But I suppose this code doesn't do any active harm if one accepts the reasonable assumption that even if different servers use different digest approaches with different configurations that the probability of an accidental digest collision is so low that we can ignore it.

So I suppose it's o.k. but it just makes me nervous.

Just thinking out loud in case anyone sees something I missed.

jo:
The question is whether the digest is recalculated on the way in (on http pull replication), which I doubt ot is. Please correct me if I am wrong.

The optimisation introduced in this pull does not depend on how the digest is generated, instead it just compares the ones stored in the docs meta.

Sorry if my previous comment was confusing.

yaronyg:
Sorry, I may be making a bad assumption. So let's expose my assumptions.

Assumptions:
1. The digest recorded locally for a particular attachment was generated locally.
2. The digest retrieved from the remote system via the attachment stub in _changes was generated on the remote system

So when the local logic runs and compares the local digest against the remote digest we are calculating two digests that as shown above can be generated using very different algorithms and settings. As such they may not be reasonably comparable.

That is the crux of my concern.

gr2m:
But aren’t the digests stored in the documents? Or are they generated on the fly? My understanding is that they get generated once and then stored in the documents, and we compare these values, so it doesn’t matter where it was generated in the first place

yaronyg:
No, I believe the hashes are regenerated locally. That is why the tests failed when I started testing on Windows. The remote (Linux) server was generating one ID and the local server was generating its own ID and so things didn't match. If I'm following the code correctly I believe one of the places we do this is [here](https://github.com/pouchdb/pouchdb/blob/9c25a2351c51507e75b5a785a1fcb2c6dad66c33/src/adapters/leveldb/index.js#L667) so we don't just take the remotely generated hash and save it. I believe we regenerate it locally.

BTW, if we were just saving the remote hash that would have some fairly scary security implications especially when replicating filtered DBs where we use filters to prevent send/receiving data from less than fully trusted remote dbs. Since PouchDB de-duplicates attachments via the hash it's critical that we generate the hash ourselves to prevent a remote from gaining access to information they should not have.

Note, btw, that because we use MD5 we are still susceptible to this attack which is why PouchDB really needs to change to a secure hash algorithm like SHA256.

mohlsen:
our team at [@project_stanton](https://twitter.com/project_stanton) had seen https://github.com/pouchdb/pouchdb/issues/2674 that this PR is trying to solve.  We tested this out today, and it solved the issue, but we did run into another bug that may have been caused by this.

We're working on the repro case, and will report here.  It seems to be inconsistent, so we're try to track it down.  It's related to syncing with a remote DB that has attachments that were deleted from docs in later revisions.

dakr0:
As @mohlsen mentioned we tried to test these changes and found some issue with syncing. In our test I used this version of pouchdb with express-pouchdb (that was used as remote server) and also with client app run in browser (that used local db). 
1. So I run client app on two browsers (let’s call them _A_ and _B_). 
2. Then on  client _A_ I created plain document (without any attachments). 
3. After that I put two attachments one by one by using `putAttachment` method. Let’s call them `image1.jpg` and `image2.jpg`. 
4. In next step I synchronized local db on _A_ with remote db (by using `sync` method without any additional options like live mode, retry, etc.). 
5. On client _B_ I synchronized its local db with remote db (in the same way like above).
6. On client _A_ I removed one attachment, for example `image1.jpg` 
7. On _B_ I modified plain content of this document (without touching attachments) and put to local db.
8. Then I synchronized dbs on _A_ as first
9. After that I synchronized dbs on _B_
10. I run client app on another browser (_C_) where local db was empty.
11. Then I started the synchronization process on _C_ and got an error (that `image1.jpg` was not found)
12. As the result the synchronization process failed and the data were not replicated at all (the local db on _C_ was still empty)

jo:
@yaronyg convinced me that attachment digest values are recalculated locally in PouchDB. And I understand the possible open attack vector if they would be taken over.

It would be interesting to know whether Apache CouchDB also recalculates attachment hashes on their way in. This could be easily tested by setting up two CouchDB servers with different compression settings and comparing the digests of a replicated doc.

In general it would be good if CouchDB could provide [reproducible digests, calculated from raw data](https://issues.apache.org/jira/browse/COUCHDB-2252).

yaronyg:
There is nothing to stop PouchDB from using reproducible digests everywhere. CouchDB makes no promises about how the digests are calculated anyway.

nolanlawson:
@yaronyg @jo This discussion has gotten a little off-topic; please follow up on md5sum issues in a separate issue. Yes, the plaintest issue is a problem; you can check the tests in `test.compaction.js` if you don't believe me; there are tests that demonstrate that CouchDB generates different md5sums than PouchDB for plaintext as opposed to non-text attachments ([here's an example](https://github.com/pouchdb/pouchdb/blob/29a0bcb7477cd0e6b88c00feccbca6fa1ee453ba/tests/integration/test.attachments.js#L1240-L1266)).

@gr2m I would be happy to merge this PR even without additional tests, but I'm troubled by folks saying that it caused new issues. @mohlsen @dakr0 if you could find tests to reproduce (as Mocha tests in the test suite preferably) then that would be very helpful. 😃

gr2m:
@dakr0 here is a test case for https://github.com/pouchdb/pouchdb/pull/5018#issuecomment-207550052 by @janl: https://gist.github.com/janl/0c461e810cb3e67e17f2bae61b57c415

it’s fixed via 6715907

I’ll see if I can add tests for the initial case (avoiding unneeded GETs) and Jan’s tests now

jo:
@nolanlawson sorry if I was unclear.

The problem with this approach is that CouchDB calculates digest *after* compression, and compression is configurable. Therefore CouchDB generates different md5sums than PouchDB *most of the time*, not only for plaintext attachments.

gr2m:
@jo so what we do with this PR is an optimisation that works in few cases, but falls back to standard behavior in most, right?

gr2m:
I’ve rebased my branch on latest master and there seems to be an issue with https://github.com/pouchdb/pouchdb/issues/4963 now. I’m looking int it

jo:
@gr2m This works only in the (rare imo) case that [`file_compression`](http://docs.couchdb.org/en/stable/config/couchdb.html#couchdb/file_compression) configuration has been changed to `none` (default is `snappy`). But yes, it falls back to previous behavior otherwise.

I'd argue that this only solves a small part of #2674 and does not justify the added complexity. I'd suggest to implement this optimization based on `revpos` instead, as [outlined in the issue](https://github.com/pouchdb/pouchdb/issues/2674#issuecomment-207604358).

janl:
@jo we've tested this on stock CouchDB and it does work with attachments coming from couch or pouch.

dakr0:
@gr2m I've checked this fix for #5018 and it works. Thanks.

nolanlawson:
Thanks for everyone's input and hard work on this. I will add @janl's test case to the tests and then I think this PR is good to go! @gr2m your `rev` fix seems exactly right; I suspected that would be a good addition as a performance improvement but didn't realize it would also be necessary for correctness. Nice catch.

nolanlawson:
See https://github.com/pouchdb/pouchdb/pull/5067; I think this PR needs a bit more work before we can merge.

jo:
As it turns out [COUCHDB-2252](https://issues.apache.org/jira/browse/COUCHDB-2252) is fixed in current CouchDB version 1.6.1. If anybody could approve it would be great.

This means that my objection is invalid and this digest approach is ok from my side.

:+1:

nolanlawson:
I'm sorry to hold off on this PR; I know everybody is very excited about faster replication! I just want to make sure we've nailed down these errors first. I will look into this the next chance I get unless someone else gets to it sooner.

gr2m:
@nolanlawson I think we nailed it now at https://github.com/pouchdb/pouchdb/pull/5067

nolanlawson:
Rebased on top of the new tests; waiting for a green, and then I'm :+1: https://github.com/pouchdb/pouchdb/pull/5085

janl:
Looks all good now :)

nolanlawson:
Everything looks great! Thanks for the PR, looking forward to faster attachment replication. :)

PR #5027, opened 2016-04-03
(#5027) - node build fix

gr2m:
`npm run build -- node` only created the `lib/` folder but nothing else because `process.exit(0)` was called before rollup could do its thing

### DIFF
--- a/bin/build.js
+++ b/bin/build.js
@@ -221,23 +221,27 @@ function buildPluginsForBrowser() {
 }
 
 if (process.argv[2] === 'node') {
-  buildForNode();
-  process.exit(0);
-}
-
-Promise.resolve()
-  .then(function () { return rimraf('lib'); })
-  .then(function () { return rimraf('dist'); })
-  .then(buildForNode)
-  .then(buildForBrowserify)
-  .then(buildForBrowser)
-  .then(buildPluginsForBrowserify)
-  .then(buildPluginsForBrowser)
-  .then(buildExtras)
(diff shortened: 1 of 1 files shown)


gr2m:
note that when I run `npm build -- node` twice in a row I get this error:

```
Error: ENOENT: no such file or directory, rename 'lib/index.js.tmp' -> 'lib/index.js'
    at Error (native)
```

`rm -rf lib` resolves the issue, but I don’t quite understand why, any idea?

gr2m:
this solves the problem, not sure if it is desired?

```diff
-  return buildForNode().then(function () {
+  return rimraf('lib').then(buildForNode).then(function () {
```

daleharvey:
I cant remember exactly why, but the full build needs the same thing @ https://github.com/pouchdb/pouchdb/blob/master/bin/build.js#L229

If you could add that in then that would be great and happy to merge this, good catch, thanks

daleharvey:
(also sorry for the delay, busy days)

gr2m:
@daleharvey updated the PR, thanks :)

daleharvey:
heh sorry, caught by eslint

gr2m:
oh my that’s embarrassing. I know added an `else` statement, please let me know if you would prefer a different style. As the build is now async, we have to make sure that the `Promise.resolve().then ...` part does not get executed

daleharvey:
Nope that looks perfect, thanks :)

PR #5029, opened 2016-04-04
(#4963) Fail entire replication batch if any docs within the batch fail

alxndrsn:
Fix for #4963

PR #5031, opened 2016-04-05
(#4963) Don't skip failed replication docs

garethbowen:
When replicating a batch and one doc fails save the docs that successfully downloaded up to that point, then proceed to the next batch starting with the one that failed.

This is an alternative approach to #5029 which attempts to solve the same problem by discarding the entire batch. The problem with that solution is if a large batch fails every time the replication will never make any progress. This patch attempts to resolve that issue by writing as many of the successful docs as possible so they don't need to be downloaded again next time.


garethbowen:
Made a mess of this one. I'll try again.

PR #5033, opened 2016-04-05
(#4963) Don't skip docs that fail to replicate

garethbowen:
When replicating a batch and one doc fails save the docs that successfully downloaded up to that point, then proceed to the next batch starting with the one that failed.

This is an alternative approach to #5029 which attempts to solve the same problem by discarding the entire batch. The problem with that solution is if a large batch fails every time the replication will never make any progress. This patch attempts to resolve that issue by writing as many of the successful docs as possible so they don't need to be downloaded again next time.

### DIFF
--- a/src/replicate/replicate.js
+++ b/src/replicate/replicate.js
@@ -101,7 +101,26 @@ function replicate(src, target, opts, returnValue, result) {
     });
   }
 
+  function findLatestSeq() {
+    var lastId = currentBatch.docs[currentBatch.docs.length - 1]._id;
+    for (var i = 0; i < currentBatch.changes.length; i++) {
+      var change = currentBatch.changes[i];
+      if (change.id === lastId) {
+        return change.seq;
+      }
+    }
+  }
+
   function finishBatch() {
+    if (Object.keys(currentBatch.diffs).length) {
+      if (!currentBatch.docs.length) {
+        // changes expected by none written - don't update checkpoint
+        return;
+      } else {
+        // check for partial success
+        currentBatch.seq = findLatestSeq();
+      }
+    }
     result.last_seq = last_seq = currentBatch.seq;
     var outResult = clone(result);
     if (changedDocs.length) {
--- a/src/replicate/replicate.js
+++ b/src/replicate/replicate.js
@@ -112,24 +112,22 @@ function replicate(src, target, opts, returnValue, result) {
   }
 
   function finishBatch() {
-    if (Object.keys(currentBatch.diffs).length) {
-      if (!currentBatch.docs.length) {
-        // changes expected by none written - don't update checkpoint
-        return;
-      } else {
-        // check for partial success
-        currentBatch.seq = findLatestSeq();
-      }
+    if (!Object.keys(currentBatch.diffs).length) {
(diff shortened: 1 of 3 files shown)


jo:
I really like this idea. Do you know how CouchDB handles such read failures? Do we need to differentiate between different errors?

src/replicate/replicate.js
```diff
@@ -101,7 +101,26 @@ function replicate(src, target, opts, returnValue, result) {
...
+        return change.seq;
+      }
+    }
+  }
+
   function finishBatch() {
+    if (Object.keys(currentBatch.diffs).length) {
+      if (!currentBatch.docs.length) {
+        // changes expected by none written - don't update checkpoint
+        return;
```
jo: You can't just return here, replication needs to be either [completed](https://github.com/pouchdb/pouchdb/pull/5033/files#diff-1b1f318b599e364220cdb8ac006c238bR135), [restarted](https://github.com/pouchdb/pouchdb/pull/5033/files#diff-1b1f318b599e364220cdb8ac006c238bR139) or [aborted](https://github.com/pouchdb/pouchdb/pull/5033/files#diff-1b1f318b599e364220cdb8ac006c238bR142). This might be the cause for [this reliably failing test](https://travis-ci.org/pouchdb/pouchdb/jobs/121032787#L3376).

I'd suggest to leave it up to the checkpointer to skip writing the checkpoint if nothing changed; this is out of the scope of this issue anyway imo.

garethbowen:
Thanks @jo . I've updated the code to not return prematurely.

willholley:
In response to "Do you know how CouchDB handles such read failures? Do we need to differentiate between different errors?", CouchDB would call _changes for the batch and then only fetch the content of those documents for those that didn't already exist in the target database (step 5 at http://www.replication.io/spec). Fetching _changes without document bodies is relatively cheap so it's not a huge issue to fetch _changes from a little further back than needed (i.e. start from the last checkpoint rather than the last change that was imported).

The issue for the PouchDB replicator is the use of _bulk_get which adds batch semantics to the document fetching and it looks like we haven't accounted for individual document failure/checkpointing appropriately. The CouchDB replicator doesn't (yet) use _bulk_get so doesn't suffer from the problem.

tests/integration/test.replication.js
```diff
@@ -2057,163 +2057,135 @@ adapters.forEach(function (adapters) {
...
+          })
+          .catch(function(err) {
+            console.error('Error saving doc', err);
+          });
+      }
+      save_doc(0);
+
+      var bulkGet = remote.bulkGet;
+      function first_replicate() {
+        remote.bulkGet = function () {
```
willholley: an issue here is that the order of results in _changes is not guaranteed. The test will fail against CouchDB 2.X and Cloudant because the result from bulkGet doesn't match the documents that were requested.

garethbowen: Right. I think the original bug only affects CouchDB 1.x as it's fetching individual documents that causes the inconsistent state.

nolanlawson:
Thanks for the PR! I think ultimately #5055 supersedes this one, although please let me know if there's anything covered in this PR that isn't covered in #5055.

PR #5036, opened 2016-04-06
eslint: enfore semicolons

gr2m:
semicolons are currently not enforced, `npm run eslint` does not throw errors when we miss them.
Docs for eslint option: http://eslint.org/docs/rules/semi#always

follow up for https://github.com/pouchdb/pouchdb/pull/5018#issuecomment-205357297

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -30,6 +30,7 @@
 
   "rules": {
     "no-empty": 0,
-    "no-console": 0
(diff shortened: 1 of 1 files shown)


gr2m:
lol go home coveralls you so drunk

daleharvey:
Awesome thanks, :+1: its best to be as strict with eslint as we can be

Issue #5039, opened 2016-04-07
WebSQL throws Security Error 18 on iOS UIWebView


nolanlawson:
This is one of those issues that's so fundamental, I'm astonished nobody every reported it up to now.

If you set `size` to `1` when calling `openDatabase()`, then apparently UIWebView on iOS really expects you to stick to that size, and throws a Security Error 18 if you exceed it. Apparently this behavior is different from iOS and desktop Safari, because we've been getting away with using `1` for years. Setting it to `5000000` (as we do for Android) works fine.

The only reason I can think that nobody ever reported this is that everybody writing Cordova apps just assumes they need the SQLite Plugin. It still kind of boggles my mind though.

nolanlawson:
OK, this is weird: I can no longer reproduce this. Instead I eventually get an error:

    "there was not enough remaining storage space, or the storage quota was reached and the user declined to allow more space"

...which is more of an expected error to me.

I guess users can always fix this by specifying their own `{size: 5000000}` or whatever when creating a `PouchDB`, so maybe this is a non-issue. Closing for now.

nolanlawson:
This actually ended up being a programmer error on my part. I was requesting too much data upfront.

PR #5040, opened 2016-04-07
(#5039) - fix UIWebView WebSQL

nolanlawson:
I would like someone else with an iDevice to please
test this and verify that my analysis is correct, since
we don't test UIWebView in CI.

To test it, you'll just need to create a "hello world"
Cordova project using UIWebView. You can detect UIWebView
vs WKWebView by checking `typeof indexedDB` since it's not
available in UIWebView.

And yeah, I am not proud of this UA-sniffing approach,
but I can't see any way around it since if we just use
5000000 everywhere, then it will break our tests due to
the popup.


nolanlawson:
Closing, see https://github.com/pouchdb/pouchdb/issues/5039#issuecomment-206946765

PR #5041, opened 2016-04-07
(#4920) Note reverse order of startkey and endkey

### LINKED ISSUE #4920: Documentation: note that startkey/endkey reverse when desending=true
Hi.

I've just realized, after few days, that... 

1. when using id prefix search
```
      opts.startkey = 'test';
      opts.endkey = 'test\\uffff';
```

2. and trying to use 
```
opts.descending = true;
```

You will get empty array. I just found an [SO](http://stackoverflow.com/questions/30952341/pouchdb-startkey-endkey-not-returning-docs) question with answer, if using startkey & endkey with combination of descending parameter, you should switch startkey and endkey, so that startkey becomes endkey and vice versa.

Please add this info into documentation for future reference or users, who'd like documents returned  or add automatically startkey and endkey reversing mehanism inside PouchDB.

:+1:

jkt001:
Documented the reverse ordering of startkey and endkey
when descending is set to true

Fixes #4920


daleharvey:
This is perfect, thanks :)

nickcolley:
Thanks @jkt001 :)

Issue #5042, opened 2016-04-07
Cannot increase the timeout on replication GETs made through bulkGetShim


alxndrsn:
I'm trying to change the timeout of individual document GETs in a replicate call against a remote CouchDB server.

At https://github.com/pouchdb/pouchdb/blob/master/src/deps/bulkGetShim.js#L85, the `ajax.timeout` option is not passed from the supplied `opts` object.

It seems that the default option (set when initially calling `new PouchDB('name', { ajax: { timeout: N } })` is not passed either.

The only value that gets used is the default of 10 seconds.

Am I missing something in my configuration, or is this a bug?

ig3:
I am seeing the same thing with bulkGetShim and http adapter. I added an ajax option to bulkGetShim call to db.get (I haven't looked at your commit) but this isn't sufficient.

db.get from bulkGetShim creates a lot of GET requests for documents and attachments. In my case, with default replication settings, several hundred (10 to 30 attachments per document). While I can eliminate or increase the timeout on GETs from bulkGetShim, these GETs also block other requests, including the next request for _changes, which has a 25 second timeout. So, unless everything downloaded by bulkGetShim completes in under 25 seconds, the GET on _changes times out.

ig3:
Sorry - clicked the wrong button...

alxndrsn:
@ig3 In the meantime, can you increase the 25s timeout using the `options.replicate` on your `PouchDB.replicate()` call?  You might also benefit from decreasing `options.batch_size` as well.  
ref. https://pouchdb.com/api.html#replication

nolanlawson:
fixed by https://github.com/pouchdb/pouchdb/pull/5043

PR #5043, opened 2016-04-07
WIP - (#5042) include timeout and cache options in bulkGetShim get() calls

alxndrsn:
Allow override of the default timeout in bulk-get calls.

Fix for #5042.

### DIFF
--- a/src/deps/bulkGetShim.js
+++ b/src/deps/bulkGetShim.js
@@ -77,7 +77,7 @@ function bulkGet(db, opts, callback) {
     }
 
     // globally-supplied options
-    ['revs', 'attachments', 'binary'].forEach(function (param) {
+    ['revs', 'attachments', 'binary', 'ajax'].forEach(function (param) {
       if (param in opts) {
         docOpts[param] = opts[param];
       }
--- a/src/adapters/http/index.js
+++ b/src/adapters/http/index.js
@@ -305,6 +305,7 @@ function HttpPouch(opts, callback) {
 
       for (var i = 0; i < numBatches; i++) {
         var subOpts = pick(opts, ['revs', 'attachments']);
+        subOpts.ajax = ajaxOpts;
         subOpts.docs = opts.docs.slice(i * batchSize,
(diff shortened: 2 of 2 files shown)


src/deps/bulkGetShim.js
```diff
@@ -77,7 +77,7 @@ function bulkGet(db, opts, callback) {
     }
 
     // globally-supplied options
-    ['revs', 'attachments', 'binary'].forEach(function (param) {
+    ['revs', 'attachments', 'binary', 'timeout', 'cache'].forEach(function (param) {
```
alxndrsn: I've discussed this with @nolanlawson, and we're not actually sure if we should be copying `timeout` here, or `ajax`.  I will test further locally.

alxndrsn: This fix doesn't work, but neither does including `ajax` in this list.  What _does_ work is adding a line which sets `docOpts = { ajax: { timeout: 12345 } };`.

ig3:
Function getDocs, called from replicate, doesn't accept any options so can't pass them to bulkGet even if there is an ajax option on the call to replicate.

alxndrsn:
@ig3 `getDocs()` calls `doBulkGetShim()`, which has access to `ajaxOpts`, so it can pick them up there.  This is what my latest PR does, and seems to work OK.

An alternative approach might be to add the `ajax` option to replicate's `createBulkGetOpts()`.  I will investigate this now.

alxndrsn:
`createBulkGetOpts()` doesn't seem suitable as there's no access to the `ajaxOpts` from there.  And it seems more related to getting options for the bulk get in total, as opposed to individual ajax requests within that (which is bulkGetShim-specific anyway).

nolanlawson:
This seems like exactly the right solution to me. Unfortunately `ajax`/`timeout` is notoriously tricky to test, so I'm fine to just manually inspect in this case without asking for a new test. Thanks for the PR!

nolanlawson:
Also I should mention that our commit message format actually goes `(#num) - message`, but I'm happy to fix on the way in, so no biggie!

Issue #4977, opened 2016-03-16
Consistently handle "onabort" events in IndexedDB


Heshyo:
If the user clears the cache of the browser then obviously he'll lose everything stored on Pouch. The issue is that I'd like to be able to detect such scenario. I've seen a few places where we get an error which is not accessible.
- First error:

```
    idb.onabort = function (e) {
      console.error('Database has a global failure', e.target.error);
      idb.close();
      cachedDBs["delete"](dbName);
    };
```

This displays the error `Database has a global failure DOMException: Connection is closing.`.
Couldn't we throw an error at the end, eg `throw new Error("Database has a global failure " +  e.target.error.message);`, so that we'd be able to access it in `window.onerror`?

I've tested throwing an error and I can see it in `window.onerror`, but I don't know if there are other things to consider.
- Second error:

```
function openTransactionSafely(idb, stores, mode) {
  try {
    return {
      txn: idb.transaction(stores, mode)
    };
  } catch (err) {
    return {
      error: err
    };
  }
}
```

The browser logs the error `Uncaught (in promise) DOMException: Failed to execute 'transaction' on 'IDBDatabase': The database connection is closing.`. Could we have something like

```
var txn = idb.transaction(stores, mode);
txn.onerror = function() {
    throw new Error("Database has a global failure.");
};
return {
  txn: txn
};

```

I've tested that code and I can see it in `window.onerror`, but again I don't know if there are other things to consider. Note that if in some parts of the code we don't want that behavior, we can always override the onerror (I've seen the `onerror` used once: `tx.onerror = idbError(callback);`, and as it's set afterwards, it would override the `throw`).

daleharvey:
So these sound like reasonable changes but in fairly tricky parts of the code, I cant see any objection to the first change it seems pretty sensible, maybe split them into seperate PRs submit them and see what the tests think?

nolanlawson:
:+1: to what Dale said, but I think it would be better to add `.onabort`s to separate usages of IndexedDB transactions rather than in one central place (unless it can be done in a clean and consistent way).

Right now it looks like we inconsistently apply `onabort` (see [this](https://github.com/pouchdb/pouchdb/search?utf8=%E2%9C%93&q=onabort)), but I would be okay with removing all those separate usages and just adding a global error handler in that one part of the code.

nolanlawson:
[...]

PR #5044, opened 2016-04-08
Throws an Error in txn.onerror

Heshyo:
Throws an Error in txn.onerror to let the rest of the code know about
it. Use "window.onerror" and look for "Database has a global failure"
to handle that error. Note that you can override the onerror if you
want.

This follows https://github.com/pouchdb/pouchdb/issues/4977

### DIFF
--- a/src/adapters/idb/utils.js
+++ b/src/adapters/idb/utils.js
@@ -243,7 +243,7 @@ function compactRevs(revs, docId, txn) {
 function openTransactionSafely(idb, stores, mode) {
   try {
     var txn = idb.transaction(stores, mode);
-    txn.onerror = function(e) {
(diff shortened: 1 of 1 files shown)


nolanlawson:
So the `onerror` is actually typically set inside of the consumer of this function. If there's anywhere in the code where we're _not_ setting `onerror` and bubbling it back to the user, then I'd prefer to add the code there rather than here.

Is there somewhere in the code that `onerror` is not being set, that you ran into?

nolanlawson:
See more comment in https://github.com/pouchdb/pouchdb/pull/5045#issuecomment-210861558 explaining why I am closing this. Thanks for reporting!

PR #5045, opened 2016-04-08
Throws an Error in idb.onabort

Heshyo:
Throws an Error in idb.onabort to let the rest of the code know about
it. Use "window.onerror" and look for "Database has a global failure"
to handle that error.

This follows https://github.com/pouchdb/pouchdb/issues/4977

### DIFF
--- a/src/adapters/idb/index.js
+++ b/src/adapters/idb/index.js
@@ -881,6 +881,7 @@ function init(api, opts, callback) {
       console.error('Database has a global failure', e.target.error);
       idb.close();
       cachedDBs.delete(dbName);
+      throw new Error("Database has a global failure " +  e.target.error.message);
     };
 
     var txn = idb.transaction([


nolanlawson:
Please see comment in https://github.com/pouchdb/pouchdb/pull/5044#issuecomment-210559621 about why this is probably not the right approach. But quickly:

1. IDB should not be throwing that error ever; if it is, then it's a bug in Pouch or a bug in the browser
2. even if it is throwing that error, it should be exposed to the user as an asynchronous error on the Promise (probably the first operation on the DB), not as a synchronous error. Yes, I know it makes debugging easier, but this is a limitation of browsers that they don't report these kind of errors well, and it's not Pouch's place to start randomly throwing synchronous errors around.

That said, if you can consistently reproduce this error and explain which browser/device/adapter it occurred in, that would be very helpful for debugging.

futpib:
> IDB should not be throwing that error ever; if it is, then it's a bug in Pouch or a bug in the browser

Nope, it could also be a legit indexeddb failure like `QuotaExceededError` which pouchdb to this day is hiding from user code.

dheimoz:
hello @futpib do you think there is programmatic way to test your theory? Maybe creating large documents?

Thanks

futpib:
One could write a test for this manually with chrome devtools quota override or automatically under puppeteer by calling (`Storage.overrideQuotaForOrigin`](https://chromedevtools.github.io/devtools-protocol/tot/Storage/#method-overrideQuotaForOrigin) and exceeding the overriden quota.

PR #5055, opened 2016-04-12
(#4963) Fail entire replication batch if any docs within the batch fail

alxndrsn:
A new fix for #4963.  If any docs in a replication batch fail, the successful docs are saved, but the batch itself is considered a failure.

* In contrast to #5029, this patch will still allow any successfully-downloaded docs to be saved locally.
* In contrast to #5033, this patch will consider the batch a failure.

### DIFF
--- a/src/replicate/getDocs.js
+++ b/src/replicate/getDocs.js
@@ -34,7 +34,8 @@ function createBulkGetOpts(diffs) {
 function getDocs(src, diffs, state) {
   diffs = clone(diffs); // we do not need to modify this
 
-  var resultDocs = [];
+  var resultDocs = [],
+      ok = true;
 
   function getAllDocs() {
 
@@ -53,7 +54,11 @@ function getDocs(src, diffs, state) {
         bulkGetInfo.docs.forEach(function (doc) {
           if (doc.ok) {
             resultDocs.push(doc.ok);
+          } else if (doc.error !== undefined) {
+            ok = false;
           }
+          // else: when AUTO_COMPACTION is set, docs can be returned which look
+          // like this: {"missing":"1-7c3ac256b693c462af8442f992b83696"}
         });
       });
     });
@@ -99,14 +104,14 @@ function getDocs(src, diffs, state) {
     }
   }
 
-  function returnDocs() {
-    return resultDocs;
+  function returnResult() {
+    return { ok:ok, docs:resultDocs };
   }
 
   return Promise.resolve()
     .then(getRevisionOneDocs)
     .then(getAllDocs)
-    .then(returnDocs);
+    .then(returnResult);
 }
 
 export default getDocs;
--- a/tests/integration/test.replication.js
+++ b/tests/integration/test.replication.js
@@ -2395,6 +2395,136 @@ adapters.forEach(function (adapters) {
       });
     });
 
+    it('(#4963) Ensure successful docs are saved but seq not updated if single doc fails to replicate', function (done) {
+      var db = new PouchDB(dbs.name);
+      var remote = new PouchDB(dbs.remote);
+      var Promise = PouchDB.utils.Promise;
+
+      // 10 test documents
+      var num = 10;
+      var docs = [];
+      for (var i = 0; i < num; i++) {
+        docs.push({
+          _id: 'doc_' + i,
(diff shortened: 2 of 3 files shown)


src/replicate/getDocs.js
```diff
@@ -53,7 +54,11 @@ function getDocs(src, diffs, state) {
         bulkGetInfo.docs.forEach(function (doc) {
           if (doc.ok) {
             resultDocs.push(doc.ok);
+          } else if(typeof doc.error !== 'undefined') {
```
nolanlawson: `(!doc.error)` is fine here, also typically we do `if (` instead of `if(` (but no biggie)

tests/integration/test.replication.js
```diff
@@ -2395,6 +2395,138 @@ adapters.forEach(function (adapters) {
       });
     });
 
+    it('(#4963) Ensure successful docs are saved but seq not updated if single doc fails to replicate', function (done) {
```
nolanlawson: typically we use the mocha promise style (i.e. no `done`) and just return a promise from inside this function, but this test looks great, so no worries

tests/integration/test.replication.js
```diff
@@ -2395,6 +2395,138 @@ adapters.forEach(function (adapters) {
...
+              db.get(id, function (err) {
+                should.not.exist(err);
+                check_docs(docs.shift());
+              });
+            }
+
+            check_docs(docs.shift());
+          }).catch(function(err) {
+            done(err);
+          });
```
nolanlawson: `.catch(done)` works too

nolanlawson:
This looks great to me - I am +1 on this solution and love the new test. Can you please confirm that the test fails before the fix? (I'd test it myself, but I don't have time right now.)

alxndrsn:
@nolanlawson

```
$ git checkout master -- src/ && npm test
...
    1) (#4963) Ensure successful docs are saved but seq not updated if single doc fails to replicate


  967 passing (2m)
  9 pending
  1 failing

  1) suite2 test.replication.js-local-http (#4963) Ensure successful docs are saved but seq not updated if single doc fails to replicate:
     Error: First replication should fail
      at db.replicate.from.then.catch.docs (tests/integration/test.replication.js:2445:18)
      at Array.0 (node_modules/lie/lib/index.js:88:21)
      at nextTick (node_modules/lie/node_modules/immediate/lib/index.js:61:18)
```

alxndrsn:
```
$ git checkout master -- src/ && npm test
...
    1) (#4963) Ensure successful docs are saved but seq not updated if single doc fails to replicate


  967 passing (3m)
  9 pending
  1 failing

  1) suite2 test.replication.js-local-http (#4963) Ensure successful docs are saved but seq not updated if single doc fails to replicate:
     Error: First replication should fail
      at db.replicate.from.then.catch.docs (tests/integration/test.replication.js:2445:18)
      at Array.0 (node_modules/lie/lib/index.js:88:21)
      at nextTick (node_modules/lie/node_modules/immediate/lib/index.js:61:18)
```

src/replicate/getDocs.js
alxndrsn: `doc.error !== undefined` seems to be necessary.  I don't recall why right now, but can investigate if you're interested.

garethbowen:
Nice one!

What happens with continuous replication? Does it fail or catch the error and retry silently? It'd be nice to add a test for this if it's not too much work.

Also, it'd be awesome if the existing test checked to see that only one doc was requested on the second replication as well as checking that only one doc was written. eg: spy on the bulkGet and assert that it's called as expected.

nolanlawson:
@garethbowen During continuous replication, any error will cause it to retry. I think the expected behavior is that it would retry the batch over and over again, except if there's a 40x in there (since we can expect it will continue to fail). It'd be nice to add another test for live replication, but I'm happy enough to merge this PR as-is and then add tests later. I think this bug may even explain some other replication-related errors I've seen (e.g. https://github.com/nolanlawson/local-npm/issues/79).

Thanks for the fix @alxndrsn !

nolanlawson:
29a0bcb7477cd0e6b88c00feccbca6fa1ee453ba

alxndrsn:
:+1:

PR #5056, opened 2016-04-12
Add error listener for clearer test failure

alxndrsn:


daleharvey:
:+1: Thanks good catch, looks great

Issue #5061, opened 2016-04-14
ajax in browser returns unknown error on timeout


ig3:
On timeout, in node, ajax returns error ETIMEDOUT with status 400 but in the browser ajax returns an unknown error.

request-browser returns an empty error and a response with statusCode: 0. In ajaxCore, this is resolved to an unknown error.

nolanlawson:
Yeah it would be nice to unify those error messages. I think status code 0 is normal, because the browser is explicitly _not_ trying to reveal the actual source of the message (same for CORS errors).

ig3:
Status code 400 Bad Request isn't ideal. Per HTTP 1.1 spec it is:
> The request could not be understood by the server due to malformed syntax. The client SHOULD NOT repeat the request without modifications. 

A timeout is not a situation where the client should not repeat the request and it can occur in cases where the request hasn't been sent to the server so no implication the server hasn't understood.

Status code 408 Request Timeout is a little closer:
> The client did not produce a request within the time that the server was prepared to wait. The client MAY repeat the request without modifications at any later time. 

At least this is a timeout but it is not the server timing out the request and it arises in situations where the request hasn't been sent to the server so nothing to do with how long the server is prepared to wait.

The only other specified timeout error is 504 Gateway Timeout, but this would only be appropriate if it were produced by the server.

It seems there is no HTTP status code that indicates a timeout at the client. The timeout may occur before the HTTP protocol exchange begins or regardless of the status if it has begun so an error that does not include status code or with status code 0 might be better options.

I am more concerned with the error than the status code. 'Unknown error' is particularly unhelpful. ETIMEDOUT seems appropriate. Consistency with node is good but in this case, perhaps node sets status code inappropriately and should be overridden.

For me, a more specific error from ajax is a step towards dealing with timeouts during replication which seems to be the subject of multiple issues here and is problematic in several of my applications.

nolanlawson:
fixed in #5062

PR #5062, opened 2016-04-14
(#5061) ajax returns unknown error on timeout

ig3:
In the browser, ajax was returning an unknown error on timeout. With this
change, it returns an ETIMEDOUT error with status 400, consistent with
node request.

### DIFF
--- a/src/deps/ajax/request-browser.js
+++ b/src/deps/ajax/request-browser.js
@@ -107,8 +107,14 @@ function fetchRequest(options, callback) {
 function xhRequest(options, callback) {
 
   var xhr, timer;
+  var timedout = false;
 
   var abortReq = function () {
+    xhr.abort();
+  };
+
+  var timeoutReq = function () {
+    timedout = true;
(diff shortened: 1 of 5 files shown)


src/deps/ajax/request-browser.js
```diff
@@ -156,10 +162,11 @@ function xhRequest(options, callback) {
   }
 
   if (options.timeout > 0) {
-    timer = setTimeout(abortReq, options.timeout);
+    timer = setTimeout(timeoutReq, options.timeout);
     xhr.onprogress = function () {
       clearTimeout(timer);
-      timer = setTimeout(abortReq, options.timeout);
+      if(xhr.readyState !== 4)
+        timer = setTimeout(timeoutReq, options.timeout);
```
nolanlawson: we require braces here, dunno why eslint didn't catch it

nolanlawson:
looks great, thanks! I'll follow up on eslint issues in another PR

PR #5064, opened 2016-04-15
Protect for missing doc

ssured:
I found this is my logs, on safari mac, using fruitdown adapter. 

` TypeError: undefined is not an object (evaluating 'metadata.id')` on line 1128

Clearly there is no error check, but I have no clue of further implications of my modification downstream.


nolanlawson:
Actually in that part of the code, `err` should really, really, really never be defined. If *down adapter ever throws an error, then something is wrong with that adapter. In this case, I'm guessing the bug is either in FruitDOWN or it's in Safari's implementation of IndexedDB. So if you could report your Safari/iOS version and ideally a test case to reproduce, that would be awesome. Could you please report it over here? https://github.com/nolanlawson/fruitdown/issues

PR #5069, opened 2016-04-16
(#5069) - fix eslint rules

nolanlawson:
A few changes here to better match our previous JSHint rules:

- `function () {}` not `function() {}`
- `if (condition) { ... }` (braces required)
- spaces after unary (e.g. `typeof(foo)` is wrong)

As it turned out, there were very few fixes to make, and eslint's
`--fix` feature did a great job here.


daleharvey:
Very nice, 👍

PR #5070, opened 2016-04-16
(#5070) - stop testing node 0.10 in CI

nolanlawson:
We should probably start a discussion about whether to keep
Node 0.10 support or not, but I'd be in favor of just
removing it because it's constantly failing in Travis, and it's
annoying to have to deal with.

We _could_ add it to allowed failures, but I figure why eat
up Travis CPU when we don't need to.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -141,9 +141,6 @@ matrix:
   fast_finish: true
 
   include:
-    - node_js: "0.10"
-      services: docker
-      env: CLIENT=node COMMAND=test
     - node_js: "0.11"
       services: docker
       env: CLIENT=node COMMAND=test


nickcolley:
ping @gr2m what Node versions is Hoodie targetting right now?

gr2m:
Hoodie is built on hapi which requires node v4+

nolanlawson:
Cool, I'm gonna merge this then.

yaronyg:
Nooo!!! We are completely dependent on Node 10 for Thali. Without Node 10 we cannot ever upgrade PouchDB!!!!!

yaronyg:
BTW, I don't have good data on how many people still use node 0.10 but as just a hint, https://semaphoreci.com/blog/2015/12/15/nodejs-version-usage-in-commercial-projects-2015-edition.html, it's not a small number.

nolanlawson:
We're not going to explicitly kill 0.10 support; we're just not going to test it anymore. So it's a YMMV situation.

yaronyg:
Nothing rots faster than code. If we aren't testing on 0.10 then we don't work on 0.10. We can't ship with YMMV. :(

yaronyg:
BTW, I realize this is our problem, not Pouch's but there is still a lot of 0.10 out there. :(

mohlsen:
Bummer.  We're working on node 0.10 (via jxcore) and were waiting for https://github.com/pouchdb/pouchdb/pull/5018 to land (a fix we need). Wish we could have gotten one more PouchDB release with this fix before testing was dropped on Node 0.10.

nolanlawson:
@mohlsen Again, it's not that PouchDB no longer works in Node 0.10. We're doing nothing to intentionally break it.

I would be happy to continue supporting 0.10, but currently "supporting" means a lot of manual overhead for committers, i.e. mostly me sitting there tapping "refresh" on Travis until the build restarts. If someone can figure out where these intermittent errors are coming from and fix them, then I would be more than happy to add 0.10 testing back in! :)

daleharvey:
So we have a bit of an undocumented process for this, if something is in CI and fails our build it needs an owner, I would be happy to 👍 putting .10 back into allowed failures and if it picks up an owner and becomes stable then happy to remove it from allowed failures

nickcolley:
I'd feel more comfortable if it were in the allow failures so it's easier for someone to see and fix.

nolanlawson:
I would have put it in allowed failures, but TBH couldn't figure out how. :/ I'd be happy if someone made that change, though.

nickcolley:
Yeah was going to put a PR in but I'm not sure too, @daleharvey any ideas? :)

janl:
we might have some resources to look into this later this week.

daleharvey:
btw we added `.10` to expected failures in https://github.com/pouchdb/pouchdb/commit/b2895aa57422610091061d93580e20a788a28803, its green now however PR's that break it wont be blocked from landing

yaronyg:
We appreciate this. We'll need to figure out how to monitor and fix as appropriate.

PR #5074, opened 2016-04-19
WIP - no-eval map and reduce functions with persistent views

kabalage:
I'm using PouchDB in a chrome app project but the options to be able to use query (which needs eval to be allowed) seamed either too complicated ([sandboxing](https://developer.chrome.com/apps/sandboxingEval)) or not enough ([evidenceprime/pouchdb.mapreduce.noeval](https://github.com/evidenceprime/pouchdb.mapreduce.noeval), [nolanlawson/pouchdb-find](https://github.com/nolanlawson/pouchdb-find)).

In my architecture (and possibly in others') design docs are updated by the application itself and not by replication. Views are only used on the client (for now) and a central couchdb/cloudant/etc database would only be used for replication between devices. As far as I understand in this case it would not cause much problem if I could pass map and reduce functions to `query()` along with the name of the view basically using a persistent view but passing in the javascript functions that otherwise would be evaluated from the design doc.

This implementation allows passing `viewName` along with `map` and `reduce` functions in the first argument of `query()`. The view must exist and the functions are converted to strings and matched against the ones in the design doc. (I assume updating the design doc would invalidate the index, but maybe I'm wrong here. This ensures that design docs are updated before using new map/reduce functions)

Before digging into creating tests for it, and making changes to the documentation I would like to have some feedback. What do you think?

(Also this is my first contribution, please be patient :sweat_smile: )

### DIFF
--- a/src/mapreduce/index.js
+++ b/src/mapreduce/index.js
@@ -246,6 +246,26 @@ function checkQueryParseError(options, fun) {
   });
 }
 
+function checkInjectedFunctions(injectedFun, viewFun) {
+  Object.keys(injectedFun).forEach(function (name) {
+    if (['map', 'reduce'].indexOf(name) === -1) {
+      return;
+    }
+    if (!viewFun[name]) {
+      throw new QueryParseError('Injected ' + name + ' function does not exist in the view.');
+    }
+    var normalizedInjectedFun =
+      injectedFun[name].toString()
+        .replace(/\s+/g, '')
+        .replace(/,emit\)\{/, '){');
+    var normalizedViewFun = viewFun[name].replace(/\s+/g, '');
+    if (normalizedInjectedFun !== normalizedViewFun) {
+      throw new QueryParseError('Injected ' + name + ' function does not match the stored one in the view.' +
+        ' Injected: ' + normalizedInjectedFun + ' Expected: ' + normalizedViewFun);
+    }
+  });
+}
+
 function httpQuery(db, fun, opts) {
   // List of parameters to add to the PUT request
   var params = [];
@@ -866,7 +886,15 @@ function queryPromised(db, fun, opts) {
     return customQuery(db, fun, opts);
   }
 
-  if (typeof fun !== 'string') {
+  var fullViewName;
+  if (typeof fun === 'string') {
+    fullViewName = fun;
+    fun = {};
+  } else if(typeof fun !== 'string' && fun.viewName) {
+    fullViewName = fun.viewName;
+  }
+
+  if (!fullViewName) {
     // temp_view
     checkQueryParseError(opts, fun);
 
@@ -890,18 +918,19 @@ function queryPromised(db, fun, opts) {
     return tempViewQueue.finish();
   } else {
     // persistent view
-    var fullViewName = fun;
     var parts = parseViewName(fullViewName);
     var designDocName = parts[0];
     var viewName = parts[1];
+    var injectedFun = fun;
     return db.getView(designDocName, viewName).then(function (fun) {
       checkQueryParseError(opts, fun);
+      checkInjectedFunctions(injectedFun, fun);
 
       var createViewOpts = {
         db : db,
         viewName : fullViewName,
-        map : fun.map,
-        reduce : fun.reduce
+        map : injectedFun.map || fun.map,
+        reduce : injectedFun.reduce || fun.reduce
       };
       return createView(createViewOpts).then(function (view) {
         if (opts.stale === 'ok' || opts.stale === 'update_after') {


nolanlawson:
Thanks for the contribution and thorough explanation! :)

One of the goals we've had with PouchDB is to keep the core as small as possible. Obviously this is a tall order when you are trying to run in every JavaScript environment known to man. We got lots of requests to support edge-casey environments like Chrome apps, FXWebView, Adobe Air, QtWebKit, etc., and in each case our standard solution has been to move things out of core and into plugins.

In the case of `pouchdb-find` an `pouchdb.noeval.js`, why would you say that those solutions are "not enough"? Is it because your query needs exceed that required of `pouchdb-find` and also `noeval.js` isn't sophisticated enough to run your query function?

If that's the case, then I would advise creating a plugin to do the feature you want to add. We are in the process of deprecating/rewriting mapreduce in favor of pouchdb-find, meaning it will probably permanently live in plugin-land.

Also I think this PR would complicate the API surface area pretty drastically while diverging from the CouchDB API and only for the benefit of no-eval environments such as Chrome app. For that reason, I would again either advise forking or creating a plugin.

Sorry about that, but hope my reasoning makes sense! I'll also point out that, if you are not using PouchDB's replication features, then for a Chrome app you may also find YDN-DB or Dexie.js to be great alternatives, since they have very rich querying capabilities that don't require eval.

kabalage:
Thank you for your response. I understand that my case is edge-casey and my solution hurts the API a little. I will try to maintain a fork of PouchDB (only the mapreduce plugin later when it is pulled out again from core). Currently I don't see a way to implement it as a plugin. Maybe I see another solution if evalFunc could be overridden.

I'm planning on using the replication, it's the best thing in CouchDB/PouchDB/Cloudant. :)
But thanks for the recommendations!

You asked why I think `pouchdb-find` is not enough for my queries. I don't know if this is the right place or not but here is why I don't think `pouchdb-find` will replace map/reduce for me.

I've been using `pouchdb-find` for a while and it seamed really nice until I started dealing with query performance. Of course if I want performant queries I have to think through how the query is handled, which parts will be handled by the index and which parts will be checked in-memory. In-memory checks killed performance drastically as they needed to read the whole database. For example my documents contain an array of tags and in order to filter by tags I used `$in` (as logical OR) and `$all` (as logical AND), but this read through the whole database even if I wanted just the first 10 matches. This query can be implemented with clever map functions emitting multiple complex keys for one document but sadly I can't write a `.find()` query for it that does not check all the documents in-memory. Hopefully [$text support #141](https://github.com/nolanlawson/pouchdb-find/issues/141) will solve these performance problems if I understand it right.

Sort fields also surprised me first that they act like complex keys and when a sort field was missing or I gave them in wrong order it made even basic queries to read through all the docs in the db. (It's not intuitive to add the `type` field to the `sort` array.) It surprised me because `.find()` was supposed to be much simpler than map/reduce and all I was thinking about in the end is _"What map function will this 'generate'?"_.

So basically if I have lots of documents, I can only write queries that translate to only  mapreduce-solvable queries and no in-memory checking and I can only use a very limited part of the query language (for now). Also I have to think through how my `.find()` queries map to map/reduce queries which makes the whole thing harder than just sticking to map/reduce.

Reduce is also a missing feature for me as I will need some statistics about my documents.

From my (not too much) pouchdb-find/mango/Cloudant Query and (a lot more) map/reduce experience I don't think that map/reduce can be replaced by mango queries in many cases. Yes, the query language may be more approachable, but without map/reduce in mind I don't think it's easy.

Maybe I'm misunderstanding something and if so, I apologize for that. Please help me understand why mango query is superior to map/reduce. Why would you deprecate the latter?

nolanlawson:
Map/reduce will never be fully deprecated - it'll always work (probably as a separate plugin). pouchdb-find is designed for people who are beginners or otherwise can't wrap their heads around map/reduce (doesn't sound like you're in that camp :wink:).

I agree secondary index performance is a problem, and that Mango has lots of counter-intuitive behavior, especially around in-memory vs on-disk queries (see https://github.com/nolanlawson/pouchdb-find/issues/138). We're hoping we can solve some of these issues with native secondary indexes, but it's not guaranteed.

In the case of this particular issue, the problem seems to mostly revolve about Chrome packaged apps and their insistence on disallowing `eval()`. If there were more environments where `eval()` were disallowed, we might re-think our strategy in the main repo, but I think for now we're better off trying to avoid adding complexity to core.

BTW instead of maintaining a fork, you could also fork 
pouchdb.mapreduce.noeval, since both it and map/reduce is essentially a plugin. That might result in slightly cleaner code, and then you can even distribute it as a separate plugin and hopefully help people out who are in the same boat. :)

kabalage:
Unfortunately pouchdb.mapreduce.noeval is not being maintained, the last commit was on Jun 27, 2014.
It was a fork of https://github.com/pouchdb/mapreduce, which is deprecated.
If I maintained a fork of pouchdb maybe I could create a separate build for mapreduce and distribute that as a plugin. Any other suggestions?

nolanlawson:
You can fork pouchdb.mapreduce.noeval and maintain that, or fork this repo and take the `mapreduce` directory. Map/reduce is technically a plugin, even if it is hosted in the PouchDB main repo. (`pouchdb/mapreduce` is only deprecated because its codebase was moved over here.) So using the normal `PouchDB.plugin()` API, you can create your own version of it.

kabalage:
I'm going to do that and give you an update when it's done. Thank you for your time!

PR #5081, opened 2016-04-21
(#/5080) - fix docs replication batch_size

jo:
Replication option `batch_size` is not affected by attachments, since we're not fetching attachments inline anymore.


jo:
I should have added `[skip ci]` to the commit message :(

daleharvey:
Awesome thanks, 👍

PR #5087, opened 2016-04-22
Add back the pagination gem

### LINKED ISSUE #5084: http://pouchdb.com/blog is down
So I upgraded to ElCap, which forced me to reinstall jekyll, and now apparently when I build the web site, this page is empty: http://pouchdb.com/blog.

All the blog posts are still there; it's just that the list of blogs is empty now.

yuloh:
I can't find the official source but Jekyll 3 removed the jekyll-paginate gem from the default install.  If the plugin isn't included than `{% for post in paginator.posts %}` doesn't do anything and the blog posts don't show up.

~~Ideally we should use the github pages gem and bundler to install plugins, but I thought it would be better to fix one thing at a time.~~  Since the site doesn't need to be built by github pages, the site can use any gems it wants.  Using bundler would make it easier to install the correct gems, so I will open a separate PR for that.

Closes #5084

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -3,6 +3,7 @@ description: PouchDB, the JavaScript Database that Syncs!
 url: http://pouchdb.com
 highlighter: pygments
 markdown: redcarpet
+gems: [jekyll-paginate]
 baseurl:
 version: 5.3.2
 paginate: 5


nolanlawson:
Awesome, thank you!! Does this mean we should also update `CONTRIBUTING.md` to mention that you should `gem install jekyll-paginate`? Or will it be automatically installed with the change you made?

yuloh:
Yeah, I'm pretty sure you still need to `gem install jekyll-paginate`.

PR #5089 will install the plugin automatically and the docs are updated in that one, so I would suggest waiting until that one is merged.

nolanlawson:
fixed by https://github.com/pouchdb/pouchdb/pull/5089

PR #5089, opened 2016-04-22
Use bundler to install jekyll dependencies

yuloh:
This PR uses bundler to install jekyll and all jekyll plugins.  It's like a package.json but for Ruby.

Right now you have to global install all the plugins like `sudo gem install jekyll jekyll-paginate pygments.rb redcarpet`.  If you already have one installed but it's a different version the site won't build correctly.

By using bundler, you just need to type `bundle install` and it installs the version specified in the Gemfile.lock.

The command to build jekyll is changed to `bundle exec jekyll build` which  makes sure it's built with the correct version of jekyll.  Another nice change is that the `checkJekyll` function will check for all dependencies specified in the Gemfile, including plugins.

Now that there is a Gemfile, you can run `bundle update` to update the dependency versions.  After doing that you would commit the updated `Gemfile.lock`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -97,9 +97,10 @@ Workflows can vary, but here is a very simple workflow for contributing a bug fi
 Building PouchDB Documentation
 --------------------------------------
 
-The source for the website http://pouchdb.com is stored inside the `docs` directory of the PouchDB repository, you can make changes and submit pull requests as with any other patch. To build and view the website locally you will need to install [jekyll](http://jekyllrb.com/) and a few other gems:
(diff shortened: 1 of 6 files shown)


daleharvey:
👍 the ruby dependencies have always been a nightmare for me (we tried redoing the site in js but migration was a hassle), I cant test this right now, but a bit 👍 and will merge when I can test unless someone beats me to it.

nolanlawson:
Looks good, one small fix is that we can just make `install-jekyll` a dep of `build-site`. I'll test and merge.

nolanlawson:
Actually on second thought; it's probably best not to re-run that every time we run `build-site`. Looks great as is! Thank you.

nolanlawson:
web site is fixed now btw

Issue #4199, opened 2015-08-24
Download user flow


zaccolley:
At the mo when you press the 'Download v4.0.0' button it downloads the file straight.

Would a better flow to have it link to the different set-up options and have the download link on that page? This way the user knows what options they have.

Kinda like how jQuery does their download flow: https://jquery.com/download/

Might be wrong but thought I'd do an issue anyway so what people thought.

Perhaps make it a 'Install' button rather than download?

nickcolley:
I like this a lot actually! We could have an install page which explains how to download via npm, etc all the options.

daleharvey:
(removing tag as its for contributions with clear guides written)

Yeh I am always forgetting where I can find the cdn links for pouchdb, or the various storage adapters etc, I think a download page that gave instructions to the various methods would be very useful

nolanlawson:
+1 to this idea. Would help with https://github.com/pouchdb/pouchdb/issues/4202 as well.

nolanlawson:
985f548f2cbd818ed0158ad8f97da31c79485654

PR #5090, opened 2016-04-22
(#4199) - add a download specific page

zaccolley:
This is for #4199

Rather than the download button downloading the file straight away you're taken to a page describing different download and install options.

This also moves installation instructions from the about section which was hidden away.

Trying to make this a simple as possible, which is hard so feedback is welcome!

Here's some screens:
Download page screenshot 1

Download page screenshot 2

### DIFF
--- a/docs/_includes/nav.html
+++ b/docs/_includes/nav.html
@@ -1,4 +1,5 @@
-{% include nav_item.html path="/learn.html" text="About PouchDB" %}
+{% include nav_item.html path="/learn.html" text="About" %}
+{% include nav_item.html path="/download.html" text="Download" %}
 {% include nav_item.html path="/users.html" text="Who's using PouchDB?" %}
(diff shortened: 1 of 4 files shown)


nolanlawson:
This looks great!! And it's much needed for users confused about the myriad of ways to use PouchDB.

Just two small suggestions:

- use jsdelivr instead of cdnjs in the example merely because their URLs are shorter, so no wrapping
- mention Rollup as well as browserify/webpack

I'm happy to make that change on the way in, though. This is a big improvement to the usability of the download page.

nolanlawson:
Made those minor fixups and then merged, cheers: 985f548f2cbd818ed0158ad8f97da31c79485654

zaccolley:
nice, the wrapping annoyed me too ha! :fire:

nolanlawson:
Published! https://pouchdb.com/download.html

PR #5092, opened 2016-04-24
Fix the install jekyll script

yuloh:
I just looked back at my changes from #5089 and realized the install jekyll script is wrong.

I was checking if the dependencies are satisfied when I should have been checking if bundler is in the PATH.

I tested the bundler check by doing:

```bash
PATH='' ./bin/install-jekyll.sh 
```


daleharvey:
Gave this a test and its working great, thanks

PR #5093, opened 2016-04-24
Use rouge for documentation syntax highlighting

yuloh:
Right now syntax highlighting uses pygments.  This means you need to have python installed and `pip install pygments` before the syntax highlighting will work.

This PR replaces pygments with [rouge](https://github.com/jneen/rouge) which is pure ruby and the default for Jekyll 3.  Rouge uses the same stylesheets as pygments so there shouldn't be any change in appearance and the `highlight.less` stylesheet will continue to work.

Note that testing this won't work until #5092 is merged, unless you manually `cd docs && bundle install`.

highlight-diff

### DIFF
--- a/docs/Gemfile
+++ b/docs/Gemfile
@@ -5,5 +5,5 @@ gem 'jekyll'
 group :jekyll_plugins do
   gem 'jekyll-paginate'
   gem 'redcarpet'
-  gem 'pygments.rb'
+  gem 'rouge'
 end
--- a/docs/Gemfile.lock
+++ b/docs/Gemfile.lock
@@ -23,10 +23,6 @@ GEM
       rb-fsevent (>= 0.9.3)
       rb-inotify (>= 0.9.7)
     mercenary (0.3.6)
-    posix-spawn (0.3.11)
-    pygments.rb (0.6.3)
-      posix-spawn (~> 0.3.6)
-      yajl-ruby (~> 1.2.0)
     rb-fsevent (0.9.7)
     rb-inotify (0.9.7)
       ffi (>= 0.5.0)
@@ -34,7 +30,6 @@ GEM
     rouge (1.10.1)
     safe_yaml (1.0.4)
     sass (3.4.22)
-    yajl-ruby (1.2.1)
 
 PLATFORMS
   ruby
@@ -42,8 +37,8 @@ PLATFORMS
 DEPENDENCIES
   jekyll
   jekyll-paginate
-  pygments.rb
   redcarpet
+  rouge
 
 BUNDLED WITH
-   1.10.6
+   1.11.2
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -1,7 +1,7 @@
 name: PouchDB
 description: PouchDB, the JavaScript Database that Syncs!
 url: http://pouchdb.com
-highlighter: pygments
+highlighter: rouge
 markdown: redcarpet
 baseurl:
 version: 5.3.2


daleharvey:
I tested this and looks great, will let others take a look and with no objections merge later, thanks a lot, less fragmented dependencies is a huge 👍 

One thing to note, when you make changes to documentation if you could add `[skip ci]` in the body of the commit message, we dont have automated tests for our documentation so adding that will not run a whole suite and make things a little faster to get merged. Cheers :)

nickcolley:
Might be worth doing a once over to check some more code snippets unless @yuloh can confirm they produce the same output?

I can do this soon :+1:

nolanlawson:
I looked through the code snippets and almost everything looks great to me.

One thing that has broken, unfortunately, is the JSON snippets:

screenshot 2016-04-25 08 33 42

Most of them can be fixed by changing `{% highlight js %}` to `{% highlight json %}`:

screenshot 2016-04-25 08 41 55

Unfortunately we can't fix all such cases by just using `json`, though; some of them are cases where we include a snippet of a document:

screenshot 2016-04-25 08 42 56

We can't just put JSON here, because it's not really JSON. Unfortunately I think the only fix for those cases is to switch to a full promise/callback/async switcher and provide the full `db.put()` example.

Sorry to be nitpicky about this but the "red text on black background" just really looks off to me; makes it look like there's a mistake in the docs. Maybe we can file an issue on Rouge instead.

yuloh:
Here is a [rouge highlight](http://rouge.jneen.net/pastes/2mZW) compared to a [pygments highlight](http://pygments.org/demo/4641754/).  It's definitely an issue with the lexer and not the stylesheet.

I noticed the highlight [works if you assign the JSON object to a variable](http://rouge.jneen.net/pastes/XQgR).  Than I found [this issue](https://github.com/jneen/rouge/issues/322) saying it's intended behavior.

> This is intended behavior. Curly braces in statement position are blocks, not objects. {a: 1} is a syntax error in statement position in javascript.

So I guess the only solution is to assign the object to a variable or put the entire code snippet?

@daleharvey thanks for the `[skip ci]` tip! I felt bad watching travis run all those tests...

nolanlawson:
@yuloh Yeah I considered just doing `var doc = {...}` but in the case of those examples, I think it's only more confusing. Probably better to show the full `put()` code sample.

daleharvey:
Yeh just keeping our snippets to valid js or json sounds like a decent plan

daleharvey:
So I updated the api docs, mostly switching to `highlight json` and putting in full `db.put` where appropriate, however I ran into this part which will not highlight correctly despite being valid js, seems like a bug in rouge?

screen shot 2016-04-28 at 15 42 38

nolanlawson:
That absolutely looks like a bug in Rouge.

TBH I'm a bit hesitant to switch from Pygments to Rouge unless we really go through all the docs with a fine-toothed comb and make sure it didn't break anything. The developer convenience of one less dependency is not worth breaking code samples on the entire site IMO.

daleharvey:
Yeh will definitely check through the rest of the site if we can get ^ fixed, would be nice to have a cleaner dep but certainly wont merge if it is breaking things

yuloh:
I opened an issue with rouge, https://github.com/jneen/rouge/issues/459

daleharvey:
Awesome, so we can just remove the quotes here and should be fine, if anyone wants to do some edits on top of my commits then please do, otherwise I will get to it at some point

nolanlawson:
Cool, sounds good to just remove the quotes, but keep in mind we gotta make these changes in lots of places. I'm pretty sure there are JSON examples strewn throughout the guide, blog posts, etc.

daleharvey:
So we cant actually remove the quotes there, those are file names as properties so they need to be quoted, so yeh this is blocked by rouge being fixed

daleharvey:
Closing this while rouge has that bug, we cant switch to it if it breaks rendering, thanks though

Issue #5096, opened 2016-04-25
Write a tutorial on prepopulated databases in Cordova


nolanlawson:
This is technically feasible; it's just really hard to figure out how to do it. I'd be happy to write a blog post about it unless someone else wants to take it up.

Details: https://twitter.com/nolanlawson/status/724657466605187072

nolanlawson:
I couldn't help myself, so I decided to just do this: https://github.com/pouchdb/pouchdb/pull/5098

nolanlawson:
3828467418c83ade2af0d5716cebd48139aa5013

PR #5098, opened 2016-04-26
(#5096) - write blog post on prebuilt databases

nolanlawson:
Just a short blog post explaining how to do prebuilt databases. I made
a sample app while I was writing it, to verify that it worked. There
were a few hiccups, but it's totally doable and not excessively difficult.


nickcolley:
Looks great to me, super useful :+1:

schamp:
This looks great, I was just in the process of figuring out how to do it myself.  

I see for the website dump/load version, there's a check to see if the preload has already occurred.  I don't see anything equivalent for the cordova version.  At first glance, therefore, it looks like the read-write DB will be overwritten on each load.  Would you please consider adding a check to that case, too, or, if it's not a problem, an explanation?

nolanlawson:
@schamp In fact it does have a similar check. I'll clarify that in the post.

schamp:
That comment helps a lot, thanks, I had overlooked the catch() with all those new Promise(...) calls, it didn't stand out.

nolanlawson:
3828467418c83ade2af0d5716cebd48139aa5013

Issue #5102, opened 2016-04-27
node6 is permafail


daleharvey:
All node tests, likely due to the 6x release

daleharvey:
So https://github.com/pouchdb/pouchdb/blob/master/tests/integration/test.attachments.js#L1268 is failing because the new version of node is generating different digests

nolanlawson:
very confusing since those digests match what's output by both CouchDB and spark-md5. 😕

nolanlawson:
might be related: https://github.com/nodejs/node/pull/5522

daleharvey:
Changed the title since we fixed travis putting it on 5

omsmith:
Fix for test failures in #5107 

You'll also see C/JS stack traces printed until https://github.com/Level/leveldown/pull/280 is available

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/3d7d49738afa9678a2c2231af9e1072e2d227257

PR #5103, opened 2016-04-27
(#5102) - Use LTS node.js for now

daleharvey:


nolanlawson:
:+1: when green

.travis.yml
```diff
@@ -137,10 +137,19 @@ matrix:
   - env: COMMAND=report-coverage
   - env: CLIENT=selenium:firefox:41.0.1 SERVER=couchdb-master COUCH_HOST=https://$CLOUDANT_USERNAME:$CLOUDANT_PASSWORD@pouch.cloudant.com COMMAND=test
   - env: CLIENT="saucelabs:MicrosoftEdge" COMMAND=test
+  - node_js: "stable"
+    services: docker
+    env: CLIENT=node COMMAND=test
+  - node_js: "0.10"
+    services: docker
+    env: CLIENT=node COMMAND=test
```
nolanlawson: ahh, so this is how you add 0.10 to allowed failures. LGTM

daleharvey:
Checking up on why these arent green now

Yeh I figured if I was putting stable in expected failures may as well do .10 at the same time

daleharvey:
So confused at why this hasnt broken before, but we require('body-parser') in the tests but dont list it in package.json, also this only seemed to break locally but running `test-component` wipes the lib directory then never rebuilds extras, so just switched it to full build

Issue #2580, opened 2014-08-09
Remove 'binary' buffer usage in Node


nolanlawson:
We use this in both our tests and in the running code. Per [the docs](http://nodejs.org/api/buffer.html):

> 'binary' - A way of encoding raw binary data into strings by using only the first 8 bits of each character. This encoding method is deprecated and should be avoided in favor of Buffer objects where possible. This encoding will be removed in future versions of Node.

daleharvey:
Got a pointer to where we are doing it incorrectly?

nolanlawson:
Everywhere where we are calling `new buffer(someData, 'binary')` (too many to list; there are plenty). Supposedly it's deprecated, although it still works in iojs and Node 0.12, so I don't know how important this issue really is.

calvinmetcalf:
they key is you should be specific about how you want the string to be
decoded, likely it's utf8

daleharvey:
No longer deprecated - https://github.com/nodejs/node/pull/3441

PR #5107, opened 2016-04-29
(#5102) - specify md5 data encoding

omsmith:
The default encoding was switched to utf8 in nodejs/node#5522

### DIFF
--- a/src/deps/md5.js
+++ b/src/deps/md5.js
@@ -2,7 +2,7 @@ import toPromise from './toPromise';
 import crypto from 'crypto';
 
 var res = toPromise(function (data, callback) {
-  var base64 = crypto.createHash('md5').update(data).digest('base64');
+  var base64 = crypto.createHash('md5').update(data, 'binary').digest('base64');
   callback(null, base64);
 });
 
--- a/src/mapreduce/md5.js
+++ b/src/mapreduce/md5.js
@@ -1,7 +1,7 @@
 import crypto from 'crypto';
 
 function MD5(string) {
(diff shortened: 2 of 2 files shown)


omsmith:
Fixes #5102

daleharvey:
To be honest I dont 100% understand this issue, however its certainly related https://github.com/pouchdb/pouchdb/issues/2580, would be good to have this settled, is `binary` deprecated / should we switch to `utf8`, or is being explicit about the type we want all thats needed and this PR resolves everything?

daleharvey:
@omsmith So I screwed up and forgot to actually add the travis stable run to verify that this fixes the tests, any chance you could rebase this on master to verify the fix works in node 6? (added https://github.com/pouchdb/pouchdb/commit/9ec5ceab56381dad1e295032631dd098880ca9a0)

daleharvey:
Awesome thanks

nolanlawson:
@daleharvey The difference is that `'binary'` is no longer implicit in Node 6. I actually recall seeing in the documentation some time ago that that API was subject to change, so I probably should have added `'binary'` earlier.

I'm +1 on this change because it matches the Node <6 behavior, as well as the behavior in the browser and in CouchDB (except for strings, where CouchDB gets weird because Erlang).

daleharvey:
Awesome, so this will also close out #2580 right? also 👍 when green

omsmith:
@daleharvey #2580 is closed out by https://github.com/nodejs/node/pull/3441

daleharvey:
Awesome thanks

daleharvey:
Great all passing :)

Issue #5118, opened 2016-05-03
Performance regression in leveldb adapter between 5.2.1 and 5.3.0


nolanlawson:
Reported by @omsmith in Slack:

> (5.3.0): Statistical profiling result from isolate-0x2286d90-v8.log, (34612 ticks, 21 unaccounted, 0 excluded).
> (5.2.1): Statistical profiling result from isolate-0x3392d90-v8.log, (12535 ticks, 2 unaccounted, 0 excluded).

The culprit was determined to be 0ddeae6b6c44fa12b83c5348df6947e3ce28c28c.

> My tests run ~34s with that commit, and ~9s with it reverted

I'm also suspicious of 4cccf617bce67351123a842f12ff0e694587f111. It seems to have introduced a new compaction requirement in leveldb that runs for every write operation. But according to @omsmith's tests, 0ddeae6b6c44fa12b83c5348df6947e3ce28c28c is the real cause.

nolanlawson:
Reason I'm still suspicious of that other commit is:

> FYI, doesn't seem to effect WebSQL (used https://pouchdb.com/2016/03/04/pouchdb-5.3.0-sqlite-support-in-node.html), is ~30s with or without the change. Plain LevelDB (write to disk) is effected similarly to using memdown. I'll work on a perf test

 0ddeae6 ought to affect all adapters, whereas 4cccf61 only affects the leveldb adapter.

omsmith:
Thanks for following up, sorry I haven't gotten back with a perf test yet.
Just have a lot going on right now, but I'll get it figured.

nolanlawson:
@omsmith Can you provide some idea of what your performance test did? I'm trying to reproduce with our own performance tests, and not really succeeding.

nolanlawson:
I looked into https://github.com/pouchdb/pouchdb/commit/4cccf617bce67351123a842f12ff0e694587f111 but can't determine conclusively whether or not it's a performance regression.

nolanlawson:
#5121

omsmith:
Here's the best I could come up with in short order, which is just sort of mirroring the type of stuff that is going on in our test suite. We do the cleanup between each test, maybe load in some new data before a particular test, run some queries (probably far less than 100, but it exposed it so didn't continue tweaking). We hit all indices between tests to make sure their up to date after clearing docs. Part of our application does a stale read (we tap index on write) hence why I included `stale: 'ok'` here, but that doesn't really seem to change much for this test case.

[...]

PR #5120, opened 2016-05-03
(#5118) - allow perf testing for node-websql

nolanlawson:
Before this commit, the instructions in `TESTING.md`
for testing node-websql would work for the integration
tests, but not the perf tests. Now they work for the perf
tests as well.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -172,9 +172,17 @@ Or the automated browser runner:
 
     PERF=1 CLIENT=selenium:firefox npm test
 
-You can also use `GREP` to run certain tests, or `LEVEL_ADAPTER` to use a certain *down adapter:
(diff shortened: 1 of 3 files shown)


daleharvey:
👍 I dont love the defaulting to websql @ https://github.com/pouchdb/pouchdb/pull/5120/files#diff-8783580c4d8af6826ab9c9322b2813f6R55 but this whole test setup needs some work so what works for now looks good

nolanlawson:
That `require()` call merely makes the `node-websql` adapter available; it doesn't make it the default. It gets added to the list of `preferredAdapters` after `leveldb`; you still have to do `{adapter: 'websql'}` to use it.

nolanlawson:
Thanks for the +1, merged!

PR #5121, opened 2016-05-03
(#5118) - Revert #4867 to fix performance regression

nolanlawson:
This reverts commit 0ddeae6b6c44fa12b83c5348df6947e3ce28c28c.

I tested the performance before and after reverting this commit. Using the `persisted-views` test with the iterations bumped up to 5000 and using the memdown adapter, I did 5 separate runs each and got the following numbers (median is highlighted in bold):

**Before reverting**:

23228ms / 29725ms / **25060ms** / 25461ms / 26044ms

**After reverting**:

20072ms / 22256ms / 19847ms / **21406ms** / 22568ms

By this measure, 0ddeae6b6c44fa12b83c5348df6947e3ce28c28c represents a **17.0%** performance regression.

It's worth acknowledging that @garethbowen intended #4867 as a fix for the case of a 9MB ddoc, and it does seem like 0ddeae6b6c44fa12b83c5348df6947e3ce28c28c would be a big performance increase in that case, because you wouldn't have to parse that huge ddoc for every query. In contrast, `persisted-views` tests a much smaller ddoc and measures the performance of repeatedly `query()`ing it.

However, I think it's safe to say that 9MB ddocs are the exception rather than the rule, and especially given the increased code weight and other criticisms I brought up in https://github.com/pouchdb/pouchdb/issues/4867#issuecomment-216418677, I would be in favor of reverting.

I'm opening this PR to allow others to weigh in, though.

### DIFF
--- a/src/adapter.js
+++ b/src/adapter.js
@@ -185,67 +185,6 @@ function attachmentNameError(name) {
   return false;
 }
 
-function cacheUpdateRequired(api, cache, designDocName, callback) {
-  cache.seq = cache.seq || 0;
-  var changesOpts = {
-    doc_ids: [ '_design/' + designDocName ],
-    limit: 1,
-    since: cache.seq
-  };
-  api.changes(changesOpts).then(function (res) {
-    var latestSeq = res.results && res.results.length && res.results[0].seq;
-    if (latestSeq && latestSeq > cache.seq) {
-      // invalidate the cache
-      cache.seq = latestSeq;
-      delete cache.promise;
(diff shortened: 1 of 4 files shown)


nolanlawson:
Backstory: https://github.com/pouchdb/pouchdb/issues/5118

daleharvey:
So this code always worried me a little, premature caching can often get in the way of other optimizations, it means doing more work. Given the regression for the more common case 👍 on taking it out

PR #5122, opened 2016-05-03
(#5122) - remove unnecessary code from leveldb/index.js

nolanlawson:
Noticed that `revsStemmed` is never set. Not sure what
this code was trying to do, but it doesn't seem to
make any difference.


daleharvey:
A little worrying, but if it isnt tested :) happy with all green tests here and less stuff the code is doing the better, so 👍

PR #5134, opened 2016-05-06
(#5133) - Stop posting perf results

daleharvey:
Something changed in the github api that causes the performance tests to fail while posting results, its likely easily fixed however we have not been using these results, I have a plan to improve the state of the performance tests and will work on it after the adapter PR, for now I think we can just disable the posting of results which makes sure the tests can still be ran etc


daleharvey:
This failure doesnt actually look to be permanent

Issue #5128, opened 2016-05-04
Custom build / modularization roadmap and discussion


nolanlawson:
The purpose of this issue is to lay out my thought process and progress as I work on custom PouchDB builds via separate npm modules (source issue: https://github.com/pouchdb/pouchdb/issues/4449). It's also an RFC to elicit feedback and discussion.

**High level goals:** mitigate the large-ish size of PouchDB (45MB) and difficulty of installing certain native modules (leveldown, sqlite3) by allowing people to do custom builds. There will also be presets ala Babel for common use cases – `pouchdb-browser`, `pouchdb-node`, `pouchdb-http`, whatever the community dreams up.

**Implementation:**

- Use [lerna](https://www.npmjs.com/package/lerna) to seamlessly link separate npm modules together, publish them together, and test all as one
- Monorepo, `pouchdb/pouchdb`
- Prepublish scripts for each module, using rollup and browserify+rollupify, since the only ES6 features we use are modules

Overview of modules and dependencies (very rough draft, mind you):

![cho68fsukaa5mjb jpg_large](https://cloud.githubusercontent.com/assets/283842/15028370/73983496-11fd-11e6-8d60-c58ea2476f8f.jpg)

Trickiest part is the browser/node switching for certain modules. In order to isolate the native dependencies, the websql adapter will need a Node and a browser version, whereas the leveldb adapter will come by default in pouchdb-node. It will be possible to use node-websql without installing leveldb, or even to use a purely in-memory Node-based PouchDB to avoid native deps altogether.

Custom builds will look something like this (rough sketch):

```js
var PouchCore = require('pouchdb-core');
PouchCore.add(require('pouchdb-adapter-leveldb'));
PouchCore.add(require('pouchdb-replication'));
PouchCore.add(require('pouchdb-ajax'));
PouchCore.add(require('pouchdb-mapreduce'));
module.exports = PouchCore;
```

So e.g. the "preset" modules like `pouchdb-browser` and `pouchdb-node` will simply implement the above builder pattern. People can make their own presets for various use cases, such as:

1. no-replication (local only)
2. http-only (CouchDB proxy basically)
3. no mapreduce
4. in-memory-only
5. etc.

[...]

PR #5137, opened 2016-05-07
(#5128) - WIP - monorepo with separate modules

nolanlawson:
(#5128 for backstory.) This is just a work in progress to determine what's broken and what's not. It's also to provide more visibility into what I'm working on.

Stuff I still need to fix/address:

- more APIs are exposed now on `PouchDB.utils` in order for unit/coverage tests to pass. This can be avoided potentially, e.g. via a `pouchdb-for-testing` submodule
- reduce API surface of sub-modules (if possible? it might also not be a big deal to expose things like `PouchDB.adapter()` to users, but maybe we should avoid it)
- version independently? also might not be a big deal, except I'm worried about breaking changes in `pouchdb-utils`
- need PRs to Lerna and Rollup to enable the build script, currently using a fork of Lerna to make it work
- Cordova tests broken (do we care? maybe we should remove them)
- submodules need descriptions and READMEs
- how to handle optional deps? I say just skip optional deps now that folks have an npm-based workaround
- `npm run build-node` is gone, I don't even know if it's possible to resurrect


nolanlawson:
Also here is the basic layout for the project:

- All packages are in `packages/` (Lerna convention)
- Packages define their dependencies like normal
- All packages define ES6 modules in `src` and build CJS into `lib`. Some projects are browser+node, and hence build `lib/index.js` and `lib/index-browser.js`, while others are vanilla and only build `lib/index.js`).
- `npm run build` uses Lerna to link the projects together: `lerna bootstrap && lerna run build` is all we need to build the whole project. Takes about 4 minutes on my machine for a clean install, about 3 minutes of which is `npm install` and the other 1 minute is building. (I'm also using local-npm, which helps.)
- Tests and doc scripts are at the top level. Lerna will be responsible for publishing; that's a TODO since I haven't finished the publish script yet.

daleharvey:
How long is it from a single line change for that to be rebuilt to be tested in the browser? Since we switched to the ES6 build system iterating in the browser is pretty slow / hampered, I would really like to avoid regressing that further (we can also improve the current situation where the watch script rebuilds the world).

Losing greenkeeper is a shame, its been pretty useful, but not something day to day urgent and is something that will likely be supported in the future, we can even help out with that. Being able to avoid level / websql packages is going to be really nice, awesome job

nolanlawson:
Incremental builds during development are definitely something I should add to this; thanks for bringing it up. I agree that the addition of ES6 modules slowed down development considerably due to the extra build step.

I believe during development we could avoid Rollup entirely and just use some kind of standard Browserify+Babelify+Watchify build process, which ought to give us fast increment builds because that's been supported in Browserify since forever (not supported in Rollup iirc). I'll get to work on that before removing `WIP` from this PR.

nolanlawson:
closing in favor of https://github.com/pouchdb/pouchdb/pull/5178

Issue #5138, opened 2016-05-07
Remove Cordova tests


nolanlawson:
We never got these running in CI, and they're mostly just taking up space in the codebase. A few other things have changed that make Cordova tests less attractive:

1. We currently test in Android 5.0 and iOS 8.4 in CI. These are the stock browsers, so not exactly the same as WebViews, but close enough.
2. There is now the [SQLite Plugin 2](https://github.com/nolanlawson/cordova-plugin-sqlite-2) which _does_ test in CI against various iOS and Android WebViews, and furthermore it borrows ~500 tests from the PouchDB test suite anyway.

Based on this, I don't think it's necessary to keep holding on to the Cordova tests. Without CI tests and without anyone regularly testing them, it's pretty pointless.

daleharvey:
Merged

PR #5139, opened 2016-05-07
(#5138) - remove cordova tests

nolanlawson:
I'm ready to say goodbye to these, since we never got
tests actually running in CI.


daleharvey:
Yup they get in the way when I am grepping stuff in the tests all the time and they arent being ran, I think getting edge working and maybe older androids etc will be more valuable, cheers

Issue #5145, opened 2016-05-10
Filtered replication doesn't record progress unless change is returned


garethbowen:
I use filtered replication to reduce the number of records replicated to Pouch. When a change passes the filter function the doc is replicated to Pouch and the local seq is updated which future replications use as the `since` parameter so already filtered changes can be skipped over.

However when no changes are returned the seq is not updated so subsequent replications use the old value for `since`. This means each doc can be run through the filter function multiple times until a change is made that passes the filter.

For example if the couchdb seq is 2000 and the local seq is 1000 the replication request will look something like...

```
.../_changes?style=all_docs&heartbeat=10000&filter=medic%2Fdoc_by_place&id=7E7633CB-C6CD-D85D-9ABD-1C6B7570F000&since=1000&limit=100
```

Then the server runs all the changes from 1000 to 2000 through the filter and responds with...

```json
{
  "results":[],
  "last_seq":2000
}
```

pouchdb then ignores the response as there are no `results` to update.

Modify pouchdb to store the `last_seq` from the changes response in the replication checkpoint so replication starts from the new seq next time.

nolanlawson:
fixed by https://github.com/pouchdb/pouchdb/commit/f1989ce6361193f1e213cb0b17e7dab4f2a8451c

PR #5146, opened 2016-05-10
(#5145) - Update checkpoint even when no changes

garethbowen:
Explicitly write a checkpoint when there are no current batches and no changes returned so future replications can start from the latest seq.

### DIFF
--- a/tests/integration/test.replication.js
+++ b/tests/integration/test.replication.js
@@ -1548,8 +1548,7 @@ adapters.forEach(function (adapters) {
       });
     });
 
-    it('Empty replication updates checkpoint (#5145)',
-       function (done) {
+    it('Empty replication updates checkpoint (#5145)', function () {
       var db = new PouchDB(dbs.name);
       var remote = new PouchDB(dbs.remote);
       var changes = remote.changes;
@@ -1592,10 +1591,9 @@ adapters.forEach(function (adapters) {
       }).then(function () {
         // Restore remote.changes to original
         remote.changes = changes;
-        done();
       }).catch(function (err) {
         remote.changes = changes;
-        done(err);
+        throw err;
       });
     });
 
--- a/tests/integration/test.replication.js
+++ b/tests/integration/test.replication.js
@@ -1548,6 +1548,35 @@ adapters.forEach(function (adapters) {
       });
     });
 
+    it('Empty replication updates checkpoint (#5145)',
+       function (done) {
+      var db = new PouchDB(dbs.name);
+      var remote = new PouchDB(dbs.remote);
+      var docs1 = [ {_id: '0', integer: 0} ];
+      remote.bulkDocs({ docs: docs1 }, function () {
+        var filter = function () {
+          return false;
+        };
+        db.replicate.from(remote, {
+          filter: filter
+        }, function (err, result) {
+          result.ok.should.equal(true);
+          result.docs_written.should.equal(0);
+          result.docs_read.should.equal(0);
+          result.last_seq.should.equal(1);
+          db.replicate.from(remote, {
+            filter: filter
+          }, function (err, result) {
+            result.ok.should.equal(true);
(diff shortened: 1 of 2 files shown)


tests/integration/test.replication.js
```diff
@@ -1548,6 +1548,35 @@ adapters.forEach(function (adapters) {
...
+          result.docs_written.should.equal(0);
+          result.docs_read.should.equal(0);
+          result.last_seq.should.equal(1);
+          db.replicate.from(remote, {
+            filter: filter
+          }, function (err, result) {
+            result.ok.should.equal(true);
+            result.docs_written.should.equal(0);
+            result.docs_read.should.equal(0);
+            result.last_seq.should.equal(1);
```
daleharvey: Just want to check, is this check going to be compatible with Couch 2.0? those tests are currently failing already

daleharvey: @kxepal / @willholley any input on ^?

daleharvey:
This makes sense and comes with a test, nice :) I just want to be sure `update_seq` checks are compatible with 2.0 but otherwise big 👍, thanks

tests/integration/test.replication.js
willholley: @daleharvey correct - last_seq is not deterministic in clustered CouchDB / Cloudant so we should avoid that comparison in tests.

garethbowen: @willholley Is this just a problem with the test and not the actual fix? ie: We cannot guarantee the next seq will be 1 but we do know it'll be an integer greater than the last seq?

daleharvey: Yup that is what I suspected, its purely a test problem. Looking at the test I am not sure I understand how it is actually testing this fix, it seems like it should pass either way? Having a hard time thinking of how to actually test this, you could override the ajax function and check that the second replication is triggered with the correct `since`

garethbowen: Good thinking - that's a much better assertion. I'll have a go at fixing this now.

daleharvey: Awesome, just another note, its not always possible, but in general we try sticking to promise based tests as they generally end up cleaner, https://github.com/pouchdb/pouchdb/blob/master/tests/integration/test.replication.js#L192 would be an example

daleharvey:
@garethbowen I invited you to the pouchdb org, all PR's still need to be reviewed however this means you can review and merge others PR's, thanks for your contributions. :)

tests/integration/test.replication.js
```diff
@@ -1548,6 +1548,57 @@ adapters.forEach(function (adapters) {
       });
     });
 
+    it('Empty replication updates checkpoint (#5145)',
+       function (done) {
```
daleharvey: It would be handy to put this on the same line

tests/integration/test.replication.js
```diff
@@ -1548,6 +1548,57 @@ adapters.forEach(function (adapters) {
...
+        // should be the same
+        changesSince[1].should.equal(changesSince[2]);
+      }).then(function () {
+        // Restore remote.changes to original
+        remote.changes = changes;
+        done();
+      }).catch(function (err) {
+        remote.changes = changes;
+        done(err);
+      });
```
daleharvey: if you throw here, then you wont need 'done' at all, the `.then` can just return and the test will see the throw and fail

daleharvey:
Mentioned some tiny nits but this looks great thanks a lot, 👍 from me but will let others give it a check and some time to fix the nits if you fancy. Thanks

garethbowen:
Nits fixed. Thanks!

daleharvey:
Awesome, thanks so much, looks perfect now, merged :)

nolanlawson:
Nice catch! Thanks for the fix, and welcome! :tada: 💃  :tada:

Issue #5140, opened 2016-05-07
Remove buffer.toArrayBuffer() usage


nolanlawson:
This was apparently dropped in `buffer` but we are still using it: https://github.com/feross/buffer/issues/90

See test failures in: https://github.com/pouchdb/pouchdb/pull/5135

nolanlawson:
Marking this as "help wanted" because it should be a pretty easy fix. Presumably if they dropped `to.ArrayBuffer()`, they replaced it with something else we can use. So it's just about doing the research and figuring out what the new API is.

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/8f55544440dcfa543b3f580388aefb0e35fae6e5

PR #5148, opened 2016-05-10
(#5140) - remove use of buffer.toArrayBuffer()

willholley:
buffer [dropped support for toArrayBuffer](https://github.com/feross/buffer/issues/90). However, all the
browsers we support have[ typed array support](http://caniuse.com/#feat=typedarrays) so we can use storedObject directly, I think.


willholley:
ok - this isn't quite right. More reading required...

daleharvey:
Awesome great job, doesnt need a test since this is preemtively fixing the build, 👍

Issue #5094, opened 2016-04-25
pouchdb.com home page layout broken


daleharvey:
https://pouchdb.com/ has the editable title which isnt supposed to display on the home page

@nickcolley Any chance you could point out where the bug is here? you're probably the most familiar with that area

nickcolley:
Right so the logic to show that title is here: https://github.com/pouchdb/pouchdb/blob/master/docs/_layouts/default.html#L63

Maybe the page.url has changed.

I wonder if we need to be more careful about website deploys hmm.

nolanlawson:
https://github.com/pouchdb/pouchdb/pull/5150

PR #5150, opened 2016-05-10
(#5094) - Remove editable title from home page

daleharvey:
[skip ci]


nolanlawson:
Oof, dunno how that slipped through. Good find.

PR #5170, opened 2016-05-17
(#5128) - monorepo blog post

nolanlawson:
First draft of the blog post describing the new monorepo architecture. I figured it was a big enough change that it deserved its own blog post. The rest of the release notes can still be a separate post.

This also takes a stance on some architectural decisions (prefixes vs scopes, semver, etc.), so this is an opportunity for us to consider these choices and make sure we're going down the right path. I hope I explained my reasoning well enough, but please let me know if you disagree with anything I say.

Also I'm sorry to speak on behalf of the "core PouchDB team;" if anybody feels that their view isn't being properly represented, let me know.

**Edit:** here's a [built version of the page](https://nolanlawson.s3.amazonaws.com/pouchdb/www/20160517/unbundling.htm) so that it's easier on your eyes. 😃

### DIFF
--- a/docs/_posts/2016-05-17-introducing-pouchdb-custom-builds.md
+++ b/docs/_posts/2016-05-17-introducing-pouchdb-custom-builds.md
@@ -1,0 +1,241 @@
+---
+layout: post
+
+title: Introducing PouchDB custom builds
+author: Nolan Lawson
+
+---
+
+{% include alert/start.html variant="info" %}
+
+{% markdown %}
+
+**TL;DR:** PouchDB 5.4.0 has a new [monorepo architecture](https://lernajs.io/) with multiple npm packages that you can install and build separately. Check out [Custom Builds](/custom.html) to get started.
+
+{% endmarkdown %}
+{% include alert/end.html %}
+
+> "Hold the pickles, hold the lettuce; special orders don't upset us."
+>
+> _– Burger King jingle_
+
+When you're building a large open-source project, the challenge often becomes less about "how do we add new features?" and more about "how do we let people opt-out of the features they don't want?".
+
+This is especially important in JavaScript projects, where the code is often shipped to users' browsers, meaning every byte counts. Furthermore, with modern build systems based on npm, it's painful to sit and wait for a large dependency to be `npm install`ed, especially if you're only using a small part of it.
+
+Large dependencies can mean a loading spinner in your user's browser, as well as an npm spinner in your own console.
+
+### PouchDB – not so pocket-sized anymore
+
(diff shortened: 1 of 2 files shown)


daleharvey:
👍 fully agreed this should have its own blog post and it reads great to me, few minor nits

> Plugin authors should update their code to either use the extras API, or ideally the separate sub-packages.

I think plugin authors should use the sub-packages only, we should cull extras fairly quickly and certainly not point people to start using it.

> However, open-source is not strictly a democracy, 

I think `However ultimately the core PouchDB team decided that it's too early` makes the same point without inviting a debate about the nature of OSS

nolanlawson:
> I think plugin authors should use the sub-packages only

Makes sense, will do.

> makes the same point without inviting a debate about the nature of OSS

:+1: let's not get political, it's already heated enough here in the US without it

nickcolley:
Super great. You make the great point of leaner builds, I wonder if it'd be worth saying how many bytes you can save with the new custom builds?

nolanlawson:
I'm working on calculating it all. TLDR is it's not as amazing as I thought it would be (`pouchdb-http` is 20KB), but it's something at least.

schamp:
The post mentions PouchDB 5.4.0, but this seems like a breaking change to me -- if PouchDB is using SemVer, doesn't that mean an update to 6.0.0?  Or is it truly non-breaking?

nolanlawson:
I think I'm also going to rename the article from "The great unbundling" to "Introducing PouchDB custom builds." More literal title == easier to google later when you want to understand how it works

nolanlawson:
I added a small new section showing off the various bundle sizes. @daleharvey @nickcolley Y'all may want to review these new paragraphs at least:

> Note that `pouchdb-browser` is slightly bigger than `pouchdb`, because `pouchdb`
> is heavily optimized with Rollup before publishing, whereas `pouchdb-browser` is
> not. Also, some optional plugins (like `pouchdb-adapter-memory`) are still quite large. In the future, we may consider using Rollup more aggressively, or try to reduce the size of `pouchdb-core`. Or we might just wait for the ecosystem to switch to tools that do automatic tree-shaking for ES modules,
> such as Rollup or Webpack 2. (If you use such a tool, and it supports `jsnext:main`, then you should already be able to get smaller bundle sizes.)
> 
> In the meantime, though, a major benefit of PouchDB unbundling is still in reduced `npm install` times, as well as avoiding compatibility issues with native dependencies like `leveldown` and `sqlite3`.

nolanlawson:
BTW I did a test run with Sinopia and everything looks :+1:, ready to publish in the next few days.

nolanlawson:
0f53aff644a94f0a5b80ec0b9d3b6a0af87147b1

Issue #5154, opened 2016-05-11
Deprecate db.put(doc, id, rev)


fade2g:
My way of updating a document is like this:

Make a get, then put the new document with the _rev value from the document from get call. If the get call returns no document, make a put without _rev (obvisouly).

```javascript
 function addDoc() {
      db.get(designDoc._id)
        .then(function(originalDoc) {
          return db.put(designDoc, designDoc._id, originalDoc._rev);
        })
        .catch(function(err) {
          if (err.status === 404) {
            return db.put(designDoc);
          } else {
            console.log('Error while updating design document:' + err);
          }
        });
    }
```
When executing this, I get the error message "invalid rev format". Experimenting, I realised, that I should not supply a value for ID, i.e. `return db.put(designDoc, undefined, originalDoc._rev);`.

I think that the error message is quite misleading and should be changed to indicate, that supplying an _id in the document AND in the put call is wrong. Possibliy, updating the documentation at https://pouchdb.com/api.html#create_document with hints on not using both values would be good as well.

See my plnkr http://plnkr.co/edit/QSpI1X?p=preview

daleharvey:
So yeh this is the same issue as https://github.com/pouchdb/pouchdb/issues/2344, to be honest I find any other variation on `db.put(doc)` super confusing. I think we may want to remove it 

The `put(id, rev, doc)` was done as motivation for pouch / couchdb to support values other than json, so we could `put(id, rev, new Blob())` etc, but that has never actually happened. 

For you I would suggest using `db.put(doc)`, for us I think we may want to deprecated `db.put(id, rev, doc)`

daleharvey:
ok, I would like to deprecate this, basically a long time ago we thought about the ability to store arbitrary data in Couch / PouchDB, not only json. We prematurely implemented some API changes to start supporting that including `db.put(id, rev, data)`, however we have found this to have confusing behaviour (as in this bug), the code to implement it is fairly nasty (https://github.com/pouchdb/pouchdb/blob/master/src/adapter.js#L217). Will put a PR in to put a deprecation notice in the code whenever it is used

nolanlawson:
I totally agree. I only use `db.put(doc)`, it aligns it with `db.post(doc)`, and I think it makes a lot more sense conceptually. I'm +1 for deprecating.

navarrojava:
I totally Agree Too!
[...]

PR #5173, opened 2016-05-18
(#5154) - Add deprecation notice for db.put(doc, id, rev)

daleharvey:


src/adapter.js
```diff
@@ -218,6 +218,9 @@ AbstractPouchDB.prototype.put =
     temp = args.shift();
     temptype = typeof temp;
     if (temptype === "string" && !id) {
+      console.warn('db.put(doc, id, rev) has been deprecated and will be ' +
+                   'removed in a future release, please use ' +
+                   'db.put({_id: id, _rev: rev}) instead');
       doc._id = temp;
       id = true;
     } else if (temptype === "string" && id && !('_rev' in doc)) {
```
nolanlawson: I think we also need to put a console warning here. I would like to deprecate everything but `put(doc)` or `put(doc, callback)`.

daleharvey: hrm yeh, may need another way to do this notice, as far as I can tell from the code putting another warn there would double warn for `db.put(doc, id, rev)`

nolanlawson: Double-warning is fine in my opinion; really yell at 'em for using an old API! :D

daleharvey:
Updated, I would still prefer it to only print out once, not so worried about the code being ugly since we can delete it all shortly

nolanlawson:
OK, I'm fine with that. The important thing is that it's deprecated in the API at least.

daleharvey:
Ah yeh forgot to add back to the docs when I rebased, done that, cheers

PR #5177, opened 2016-05-19
test Travis w/ building in Node 5 then npm pack and test in Node 0.10

nolanlawson:


nolanlawson:
This seems promising, but I might wait to do this until after the monorepo is merged.

Issue #5167, opened 2016-05-17
_bulk_get does not provide any noticeable performance for me. PouchDB 5.3.2 / CouchDB 2.0


dharders:
I've read great things about the new `_bulk_get` API endpoint, so I upgraded my CouchDB server from `1.6.1` to the `dev 2.0` branch to take advantage.

However, my tests so far have shown `sync() with 2.0` to be just as 'chatty' as 1.6.1.

My database has ~3000 docs (many 1st revision). When I `sync()` for the **first** time, there are approx **10,000** xhr requests **using v1.6.1** before my chrome dev network tab settles (very chatty!).

Using the new v2.0 branch, first time `sync()` takes approximately...**9,500** xhr requests before settling (not much of an improvement, and negligible time difference to v1.6.1).

Closer inspection of the network activity shows that the `_bulk_get` endpoint is reached (quite early), just after an _all_docs response returned ~3000 docs in one go. So I get my data pretty quick (~5 seconds), but then...

The next ~9500 xhr requests are a repetition of these 3 over and over:

- GET http://192.168.1.190:5984/testdb/_local/Tl02KOzWrRiwlIunoChEhw%3D%3D?
- POST http://192.168.1.190:5984/testdb/_bulk_docs
- POST http://192.168.1.190:5984/testdb/_revs_diff

... with increasing rev 0-1, rev 0-2, rev 0-3..., taking approx 2 mins to complete.

**Q: Is this expected behaviour ? Or is there a bug with sync() ?**

I seem to have my data straight away, but the repeating GET, POST, POST seems very inefficient. For a {live: true} sync, it also **prevents any changes being replicated OUT (to server)**, until the entire ~9500 GET, POST, POST dance is finished. So essentially data changes are not propagated to other connected users for the first 2 minutes, which is frustrating. Data changes IN (from server) still appear immediately, as expected, while the dance is happening! 

Strangely, I've also noticed that if I change some data then refresh my page before the 2 minute GET, POST, POST dance is complete, the changes are pushed out immediately to other connected users and there is no background G.P.P dance beginning again, and all my data seems to be synced.?

More background:

`var localDB = new PouchDB('testdb', {revs_limit: 100, adapter: 'websql', auto_compaction: true});
 localDB.sync('http://192.168.1.190:5984/testdb', {
                live: true,
                retry: true,
                batch_size: 10000
            })...
`
PouchDB 5.3.2
CouchDB 2.0@dev (and 1.6.1)
[...]

PR #5180, opened 2016-05-19
(#5167) - WIP - try recommended fix for replicateTo

nolanlawson:
Testing the proposed fix from @man4j in #5157.
The point of this PR is to see if it passes the
Travis tests. Clearly we need to add a test that captures
the bug, if this actually fixes the bug.


daleharvey:
dupe of https://github.com/pouchdb/pouchdb/pull/5199

PR #5181, opened 2016-05-19
(#5181) - add an issue template

nolanlawson:
I think this is worth trying out. I'm hoping it can result in higher-quality
issues, where we don't need to ask users to provide more details about their error.

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,25 @@
+<!--
+     Hello! 👋
+     Thank you for opening an issue on PouchDB.
+     This template is optional, but it may help us fix your issue faster.
+-->
+
+### Environment (Node.js/browser/hybrid app/etc.)
+
+(Write here.)
+
+### Browser/platform (Chrome/FF/Safari/Edge/iOS/Android/etc.)
+
+(Write here.)
+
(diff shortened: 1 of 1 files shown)


ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,25 @@
...
+
+### Adapter (IndexedDB/WebSQL/LevelDB/etc.)
+
+(Write here.)
+
+### Server (CouchDB/Cloudant/Couchbase/PouchDB Server/etc.)
+
+(Write here.)
+
+### Please describe your issue. Test cases are very helpful!
```
nickcolley: Maybe be more explicit what 'test case' means, not so sure it's obvious exactly what we're asking for here?

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,25 @@
+<!--
+     Hello! 👋
+     Thank you for filing an issue on PouchDB.
```
nickcolley: super nit: filing -> opening?

nickcolley:
This seems helpful to me! 👍

nolanlawson:
I like your changes. I'll make those changes and then merge!

PR #5184, opened 2016-05-20
Try building in Node 5 and then test in the others

nolanlawson:
Supporting Node 0.10 - Node 6 for _both_ building and testing is killing me. Lerna doesn't work with 0.10, Rollup has bugs in Node 6, and none of that matters to our users, who won't need either one.

Testing out this PR to see if it works. If so, I advocate this approach because it makes more sense moving forward.


nolanlawson:
Of course it still fails due to the prepublish. I am dumb. Maybe there is no way to do this.

PR #5186, opened 2016-05-20
Build in Node 5, `npm pack`, then test in Node 6 and Node 0.10

nolanlawson:
Forgot about `npm install --production`. Let's try this again.

Follow-up to #5184 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,10 +44,6 @@ before_script:
   - curl -X PUT http://127.0.0.1:3001/_node/node1@127.0.0.1/_config/cors/methods -d '"GET, PUT, POST, HEAD, DELETE"'
   - curl -X PUT http://127.0.0.1:3001/_node/node1@127.0.0.1/_config/cors/headers -d '"accept, authorization, content-type, origin, referer, x-csrf-token"'
 
-    # for some reason this extra install is required for leveldown, but
-    # only on npm@2
-  - "cd packages/pouchdb && npm install && cd ../.."
-
 script: npm run $COMMAND
 
 env:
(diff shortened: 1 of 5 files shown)


nolanlawson:
muahahaha, it works! No more need to use the same version of Node for both compiling and testing!

nolanlawson:
OK, this is a proposed commit on top of #5178, which explicitly builds in Node 5 and then tests in the other Nodes. 0.10 and 6 are still allowed failures, but perhaps we can remove them from allowed failures if this becomes stable enough (they passed in one Travis run, at least).

nolanlawson:
File an issue on Lerna for Node 0.10: https://github.com/lerna/lerna/issues/157

nolanlawson:
Filed an issue on Rollup for Node 6: https://github.com/rollup/rollup-plugin-node-resolve/issues/38

daleharvey:
👍 when green, when this gets merged we should open an issue tracking both ^ and remove this code when its no longer needed

nolanlawson:
Will do!

nolanlawson:
^ Rebased, also went ahead and removed 0.10 and 6.0 from allowed failures; they seem to be passing pretty consistently now, and I have more confidence they'll work due to being built in one version.

nolanlawson:
@daleharvey Also I tend to disagree that the code should be removed when Lerna/Rollup fix those issues. Reasons being:

1. Testing in Node 0.10 is one thing; building in Node is entirely another. The more Node diverges, the more difficult it will become to use the same code for building and testing.
2. This script reveals bugs that wouldn't otherwise be discovered, e.g. because of an error in `.npmignore` or `"files"` causing the wrong files to not be published. `npm pack` gives us exactly the version of `pouchdb` that would be `npm publish`ed, hence identifying those potential errors.

The code is indeed ugly, which is why I isolated it to `bin/npm-pack-in-travis.sh` so it was obvious what it was doing.

nolanlawson:
Green!

PR #5188, opened 2016-05-21
(#5188) - improve watch+rebuild speed during dev

nolanlawson:
Follow-up to https://github.com/pouchdb/pouchdb/pull/5178#issuecomment-220726259 since @daleharvey mentioned it was a bit slow.

I tried `watchify`, but it turns out it's actually exceedingly slow because 1) there are no incremental Rollup-based tools, and 2) the alternative (Babel with ES6 modules -> CJS transform) is suuuuuper slow and wonky because of needing to apply the transform globally (in order to handle our sub-packages).

Luckily I found a better solution: continue doing the `watch-glob` + rebuild as before, but:

1. do it all inside one Node process, don't spawn
2. listen to `packages/pouchdb`, `tests/integration/utils.js`, and `tests/performance/index.js` separately, and bundle each one separately, so we don't have to rebundle all three
3. use `lodash.debounce` to avoid a big backlog when re-saving a file many times in a short span of time

On my MacBook Air, I'm getting rebuild times for `dist/pouchdb.js` at around 3 seconds, which I think is acceptable. The extras take a bit longer to build (~6 seconds), but those are deprecated anyway and should be removed soon.

Also I enabled source maps on everything, which is nice when you're debugging.


nolanlawson:
OK, I also added browserify-incremental so that it's even a bit faster. Build times are printed to the console; seems subsequent rebuilds are much faster than first builds. Building `dist/pouchdb.js` can take as little as one second.

daleharvey:
Yup this is much faster, nice, especially when we kill extras

PR #5190, opened 2016-05-21
(#5190) - selectively install CouchDB on Travis

nolanlawson:
Trying to speed up our Travis builds. There's no reason to install _both_ CouchDB 1.6 and CouchDB master; we should selectively install them based on the build settings.


daleharvey:
looks good, cheers

PR #5191, opened 2016-05-21
(#5191) - disable npm progress bar in Travis

nolanlawson:
Apparently this still has a ~5% perf impact in npm@3 (https://github.com/npm/npm/issues/11283#issuecomment-218990814). See also https://github.com/jashkenas/backbone/pull/3938 for some others who are doing this. If this doesn't speed up our npm installs, we can remove it, but currently `npm install` is taking ~160 seconds in Travis, so getting that down seems like a worthy endeavor.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,6 +48,7 @@ script: npm run $COMMAND
 
 env:
   global:
+  - NPM_CONFIG_PROGRESS="false"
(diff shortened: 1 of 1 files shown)


nolanlawson:
Compared a couple runs: [before](https://travis-ci.org/pouchdb/pouchdb/builds/131923589) and [after](https://travis-ci.org/pouchdb/pouchdb/builds/131968038). Taking the median of three Node runs in each, and only measuring the `npm install`, we get

```
before: 190.98, 175.84, 157.07 -> 175.84
after: 164.14, 157.5, 162.31 -> 162.31
improvement: 7.69%
```

So this does indeed seem to make our `npm install`s faster, and probably other `npm` tasks too. Looking at the overall build times between those two builds, it seems to shave off as much as 30 seconds (some builds are slightly slower though - probably just normal variation).

Given that this is a one-line code change, it's probably worth it.

daleharvey:
Yeh this seems good, cheers

samccone:
seems like I should do another look now at the perf hit.... for npm.

Issue #5192, opened 2016-05-21
RFC: move pouchdb-server et al into PouchDB core


nolanlawson:
Here are the modules I propose we move into PouchDB core:

* [ ] pouchdb-collections
* [ ] pouchdb-collate
* [ ] pouchdb-server
* [ ] express-pouchdb
* [ ] pouchdb-express-router
* [ ] pouchdb-find

Open question is whether we should preserve the git history when we move them back in. It might be a nice gesture to contributors to those modules to include their changes, but we'd have to do a merge, which would break our single lineage.

BTW I plan on waiting until post-5.4.0 to do this, because I want to make sure we have a stable 5.4.0 release with the existing architecture.

marten-de-vries:
Preserving git history when merging also helps with finding why something was introduced. Even if the commit message is unhelpful, the time of the change and the state of the code at that point often gives hints. If the old repo can only be hidden/put into read only mode instead of removed that would work too for that use case, but not sure if Github supports that (maybe by moving to a to-be-created pouchdb-deprecated user?).

daleharvey:
`pouchdb-collections` `pouchdb-collate` `pouchdb-find` should pretty certainly be moved into monorepo, they are very much coupled to pouchdb development.

I am not so sure about the various servers, they are not a core part of the pouchdb project and I think merging them in may give the impression that they are (ie we will now be responsible for maintaining compatibility with whatever couchdb features pouchdb-server replicates). It makes working on those projects more complicated and have very little advantage. 

Did we fix the complications with testing pouchdb-server with the (current patch) version of pouch? it seems so @ https://github.com/pouchdb/pouchdb/blob/master/tests/misc/express-pouchdb-minimum-for-pouchdb.js, but if not then it should be fixed in pouchdb-server as it hasnt been a problem for pouchdb-express-router

nolanlawson:
Not sure what you mean by complications, but there's still [lots of janky scripts](https://github.com/pouchdb/pouchdb-server/blob/82ab4138b20fa3a7732f2b4c2b7d47fe4a01d564/bin/test-pouchdb.sh) in the pouchdb-server and express-pouchdb codebase. (Lerna would make those obsolete.)

[...]

PR #5193, opened 2016-05-21
(#5192) - improve test coverage

nolanlawson:
We regressed a bit on coverage (99.5%) with the monorepo stuff, mostly due to
me mistakenly including some non-Node stuff in with the Node build. This fixes
that, and also adds some `istanbul ignore`s where it makes sense.

### DIFF
--- a/tests/integration/test.sync.js
+++ b/tests/integration/test.sync.js
@@ -332,11 +332,11 @@ adapters.forEach(function (adapters) {
       replications.cancel();
     });
 
-    it.skip('Test sync cancel called twice', function (done) {
+    it('Test sync cancel called twice', function (done) {
       var db = new PouchDB(dbs.name);
       var remote = new PouchDB(dbs.remote);
       var replications = db.sync(remote).on('complete', function () {
-        setTimeout(done); // let cancel() get called twice before finishing
+        done();
       });
       should.exist(replications);
       replications.cancel();
--- a/packages/pouchdb-adapter-fruitdown/package.json
+++ b/packages/pouchdb-adapter-fruitdown/package.json
@@ -15,7 +15,6 @@
   "dependencies": {
     "js-extend": "1.0.1",
     "fruitdown": "1.0.2",
-    "pouchdb-adapter-leveldb-core": "5.4.0-prerelease",
-    "pouchdb-utils": "5.4.0-prerelease"
+    "pouchdb-adapter-leveldb-core": "5.4.0-prerelease"
   }
 }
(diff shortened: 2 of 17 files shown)


nolanlawson:
99.929% now ^

tests/integration/test.sync.js
```diff
@@ -332,11 +332,11 @@ adapters.forEach(function (adapters) {
       replications.cancel();
     });
 
-    it.skip('Test sync cancel called twice', function (done) {
+    it('Test sync cancel called twice', function (done) {
       var db = new PouchDB(dbs.name);
       var remote = new PouchDB(dbs.remote);
       var replications = db.sync(remote).on('complete', function () {
-        setTimeout(done); // let cancel() get called twice before finishing
+        done();
```
daleharvey: Can we take this out? this worries me for 2 reasons

`.cancel()` caused intermittent crashes in newer nodes so these tests had to be skipped to get back to a stable tree, the other is that comment indicates that the setTimeout shouldnt be removed (also indicates that this isnt the most stable test), the `complete` is going to be triggered by the first call to `.cancel`, if we then call test immediately then the code is almost certainly going to still be running while the test suite moves onto the next test.

We can still do this, but I would prefer it go into a new PR and we do a large node test run to ensure we dont bring back the issues that caused us to skip the test, both those previous issues causes non reliable test failures that are super hard to track down

daleharvey:
Nice work, I was meaning to get coverage back to 100% shortly, 👍 aside from the change to `test.sync.js`, I would like to split that out, theres a bunch of tests skipped for the same reason so we can enable them all with a full test run

tests/integration/test.sync.js
nolanlawson: I checked the code and can't see why we would need to do `setTimeout()`. The second `cancel()` call should synchronously do nothing, since it looks like this:

```js
Sync.prototype.cancel = function () {
  if (!this.canceled) {
    this.canceled = true;
    this.push.cancel();
    this.pull.cancel();
  }
};
```

I felt comfortable re-enabling this test because it looks exactly the same as the test that precedes it, if you consider that the second `cancel()` is necessarily a no-op. I might have mistakenly thought before that it needed a `setTimeout()`, but seems to me now that it doesn't.

daleharvey: The test preceding it is also skipped, https://github.com/pouchdb/pouchdb/issues/4679 was the issue and it was super hard to find, I dont want to reenable these tests without being sure that particular issue is fixed (the reported error rate was around 1 in 15 test runs). The setTimeout I am less worried about

nolanlawson: Whoops, my bad. OK in that case I will re-skip the test and then just let `cancel` dock us in the coverage reports for now. Will merge.

nolanlawson:
66480d6fdb001cd6154f913719607c76abcd118c, merged with the `test.sync.js` changes

PR #5195, opened 2016-05-22
(#5157) - avoid launching more requests than browser can perform

nolanlawson:
The test succeeds before the fix and also succeeds after. This is a performance improvement, but one that I think it would be difficult to measure since you'd need a remote server that took a long time to respond and therefore caused the browser to cancel the request. In any case, it feels like a pretty obvious fix to throttle our open requests based on what the browser can actually handle. See https://github.com/pouchdb/pouchdb/pull/5157#issuecomment-220810820 for my reasoning behind this.

### DIFF
--- a/packages/pouchdb-utils/src/bulkGetShim.js
+++ b/packages/pouchdb-utils/src/bulkGetShim.js
@@ -1,5 +1,20 @@
 import pick from './pick';
 
+// Most browsers throttle concurrent requests at 6, so it's silly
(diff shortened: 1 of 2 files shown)


daleharvey:
This makes sense and looks good to me

nickcolley:
Hoping this fixes a few bugs for the people running into this...

I'm not super familiar with HTTP2, but does this impact that in anyway?

Issue #5197, opened 2016-05-22
web_sql_went_bad error on Chrome in iOS


capesean:
### Chrome on iOS

When I create a new PouchDB in iOS I can't use it. If I inspect the object, I see i t has `status: 500` with `name: web_sql_went_bad`, `message: unknown` and `reason: SecurityError`. Is this a known issue?

Here's a demo: https://jsfiddle.net/98at8jt9/

nolanlawson:
Vanilla PouchDB currently does not support Chrome/Firefox/Opera on iOS due to WKWebView bugs (see https://github.com/pouchdb/pouchdb/issues/5079 for details). The workaround is to use the `fruitdown` adapter or fallback to in-memory or http.

nolanlawson:
OTOH we should probably detect WebSQL as invalid instead of throwing a runtime error when you try to use WebSQL.

nolanlawson:
See: https://github.com/pouchdb/pouchdb/pull/5198

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/f3a7a1633455bd7095d238baf50c510c0f20a768

PR #5198, opened 2016-05-22
(#5197) - detect WebSQL invalid on WKWebView

nolanlawson:
On broken WKWebView, we should detect that WebSQL is invalid in the `valid()` method, since that allows users to gracefully fall back from WebSQL to HTTP/memory/localstorage instead of getting a runtime error when they try to use WebSQL.

Luckily this particular bug is fixed in WebKit master, so we should also make sure that our validation check is resilient against future WKWebView releases. My suggestion is to create a test database called `_pouch_validate_websql` to check whether or not WebSQL is actually working. I checked manually on WKWebView in the iOS Simulator and confirmed that this check returns false (and that it also returns true on desktop Safari and iOS Safari).

### DIFF
--- a/packages/pouchdb-adapter-websql/src/valid.js
+++ b/packages/pouchdb-adapter-websql/src/valid.js
@@ -1,0 +1,64 @@
+import { hasLocalStorage } from 'pouchdb-utils';
+
+function canOpenTestDB() {
+  try {
+    openDatabase('_pouch_validate_websql', 1, '', 1);
+    return true;
+  } catch (err) {
+    return false;
+  }
+}
+
+// WKWebView had a bug where WebSQL would throw a DOM Exception 18
+// (see https://bugs.webkit.org/show_bug.cgi?id=137760 and
+// https://github.com/pouchdb/pouchdb/issues/5079)
+// This has been fixed in latest WebKit, so we try to detect it here.
+function isValidWebSQL() {
+  // WKWebView UA:
+  //   Mozilla/5.0 (iPhone; CPU iPhone OS 9_2 like Mac OS X)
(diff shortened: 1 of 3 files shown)


packages/pouchdb-adapter-websql/src/valid.js
```diff
@@ -0,0 +1,44 @@
...
+  //   (KHTML, like Gecko) FxiOS/1.0 Mobile/12F69 Safari/600.1.4
+
+  // typeof indexedDB is null on some UIWebViews and undefined in others
+  if (typeof indexedDB !== 'object' ||
+      !/iP(hone|od|ad)/.test(navigator.userAgent)) {
+    // definitely not WKWebView, avoid creating an unnecessary database
+    return false;
+  }
+  try {
+    openDatabase('_pouch_validate_websql', '1', '', 1);
```
daleharvey: Should we be cleaning up after this verification?

nolanlawson: There's no way to clean up; we can't delete WebSQL databases, right?

nolanlawson:
Hm, this is actually failing in SauceLabs on iPhone. "Invalid adapter."  That's odd.

nolanlawson:
Oh jeez, the issue is that we actually need to cache this, because if we open it too many times, the tests start failing halfway through. Love WebSQL on Safari. :P

nolanlawson:
OK, that should do it. Using `localStorage` to cache whether or not we've already checked for a valid WebSQL.

daleharvey:
ok yeh I really dont love that, but I cant see any way around it either. Ill push harder on idb-next to fix it :)

Issue #5196, opened 2016-05-22
Possible regression in live+retry replication - processPendingBatch()


nolanlawson:
Folks are saying that `processPendingBatch()` has a regression in 5.3.2 where it goes into an infinite loop (I think that's the bug?). The proposed fix is 7092284ef397e235e5d4c0b91bc7278e54bd1cee.

I'm opening a new issue because the other ones (https://github.com/pouchdb/pouchdb/pull/5157 and https://github.com/pouchdb/pouchdb/issues/5164) became very long, and I want to focus on finding a test case to reproduce the issue.

Obviously we could just merge 7092284ef397e235e5d4c0b91bc7278e54bd1cee and be done. But it's disturbing that that change doesn't affect any of the tests. We need a test to identify when and where exactly the existing code fails.

I gave this a try in  f289d50 but failed. Any help here would be appreciated.

cc @Ampakinetic @man4j @dharders

dharders:
@nolanlawson it's not an infinite loop in my testing, but close enough from the users perspective. i.e. it seems to settle after ~10000 xhr requests and several MB of data transfer (even though my db is < 0.7 MB)

I'm not too sure about writing the correct unit test, but to reproduce the behaviour locally:

- I setup up a CouchDB server locally (v1.6.1 or v2.0) with a database named **testdb**
- Created 1295 docs (like your unit test does), but with
```
for (var i = 0; i < 1295; i++) {
   localDB.put({_id: 'testdoc_'+ i}).catch(function(e) { console.log(e)})
}
```

- Blow away the browser local db though Google Chrome Cookie + Site Data delete option
- Run the below code and inspect the network tab in Chrome (sync should start automatically)

```
var localDB = new PouchDB('testdb');
localDB.sync('http://192.168.1.180:5984/testdb', {
    live: true,
    retry: true,
    batch_size: 10000,          
    batches_limit: 10
    }).on(...)
```
- Swap out the CDN for different PouchDB version
```
<script src="https://cdn.jsdelivr.net/pouchdb/5.3.2/pouchdb.min.js"></script>
<!--<script src="https://cdn.jsdelivr.net/pouchdb/5.2.0/pouchdb.min.js"></script>-->
```
- Blow away the browser local db again
- Reload and inspect the different sync network behaviour (i.e 5.2.0 = Good, 5.3.2 = Bad)

dharders:
Also,

[...]

PR #5199, opened 2016-05-22
(#5196) - fix excessively long replication loop

nolanlawson:
This just contains the fix with no additional tests. Discussion in #5196 seems to indicate that a test would be very difficult to write, so I'm just opening a PR to see if it even passes Travis as-is.

### DIFF
--- a/packages/pouchdb-replication/src/replicate.js
+++ b/packages/pouchdb-replication/src/replicate.js
@@ -273,7 +273,7 @@ function replicate(src, target, opts, returnValue, result) {
     }
     pendingBatch.seq = change.seq;
     pendingBatch.changes.push(change);
-    processPendingBatch(changesOpts.live);
+    processPendingBatch(batches.length === 0 && changesOpts.live);
   }
 
 


daleharvey:
So performance and replication are going to be my (and I am guessing most of pouch's) future concentrations so I am 👍 on merging this to solve the immediate issue, then hoping we have more comprehensive testing in place before it has a chance to be regressed, 👍 when green

nolanlawson:
Yeah, I agree performance is something we should be taking more seriously. We really need tests in place for it (tests we pay attention to; I've been guilty of ignoring them for sure), otherwise we will continue regressing on perf. But I'll merge this for now!

dharders:
Just a quick note for future reference:

- Enables the `pendingBatch.changes` buffer to fill up before the `startNextBatch()` cycle, where previously it was flushing only 1 change per batch, hence the bad performance.
- Only relevant for `{ live: true }` 2-way replication of empty localDB with non-empty remoteDB
- This fix is much better than before, but still suffers from occasional premature batching of `pendingBatch.changes.length < batch_size`, even when there are `batch_size + N`  more changes to come. 

@daleharvey Concentrating on replication performance, I'd probably start with the batching immediacy issue (i.e. calls to `processPendingBatch(immediate:boolean)` ->  3 locations) to ensure we're only batching full buffers when it's known that more changes exist. Another important improvement to look at, although out of scope of this PR, is reducing the number of live replication xhr requests involved for a single document change. At current count, there are 11 xhr back and forths for one change, which sucks for mobile users.

@nolanlawson An idea to performance test this issue and avoid future regression:

- setup a pouchdb-express-router similar to [test.read_only_replication.js](https://github.com/pouchdb/pouchdb/blob/master/tests/component/test.read_only_replication.js)
- add an app.get() route handler and count the request urls beginning with **_local/**
- a FAIL is `numDocs < count`

dharders:
@nolanlawson I added a Performance test [here](https://github.com/dharders/pouchdb-replication-perf-regres-test)

It seems to catch the behaviour quite well. Let me know what you think?

dharders:
> ... At current count, there are 11 xhr back and forths for one change...

Actually, there are only 7 xhr requests for a single document update using `.sync(remote, {live:true} ... `

I had mistakenly disabled the cache in Chrome dev tools, which lead to extra pre-flight OPTIONS requests for CORS. Thus an extra 4 OPTIONS requests were made in my testing that would not usually occur in production. My bad!

I still think there is room for improvement to not double up check-pointing using `sync()`.

nolanlawson:
@dharders can you open up a new issue with that perf test, please? We should make it a permament part of the test suite if it really susses out the error.

Issue #5203, opened 2016-05-22
npm test -> change file -> npm test flow broken


daleharvey:
Just away to go to bed, however was looking at doing a PR, I changed some code in `packages/pouchdb-adapter-leveldb-core/src/index.js` then reran npm test and got the same failure, looking at the source it didnt pick up my changes, I would expect npm test to deal with that and rebuild as necessary

nolanlawson:
I'll fix this.

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/327f677ec8a24ca477af0f4af8d58e93851e5131

PR #5205, opened 2016-05-23
(#5203) - fix rebuild before `npm t`

nolanlawson:
Dumb mistake, I did `! -z` when I meant to do `-z`. This should also speed up our Travis builds, since it skips an unnecessary extra build.


daleharvey:
Sweet this fixes it for me, thanks

PR #5206, opened 2016-05-23
(#5206) - fix Buffer being included in bundle

nolanlawson:
This removes about 5kB from the bundle. It was a regression with the monorepo changes, now fixed.


nickcolley:
Nice, good catch 👍

Issue #5207, opened 2016-05-23
RFC: Tracking idb-next release work


daleharvey:
I wanted a bug to track the discussion for getting this merged / released. I am going to open a seperate issue for the implementation of pouchdb-find just to not confuse the issue, this is to figure out what we want to do with the adapter
- [x] Merge non adapter specific changes (https://github.com/pouchdb/pouchdb/pull/5208)
- [x] Fix last 3 skipped tests, merge changes to tests
- [x] Ensure we have tests ensuring local documents are skipped in changes
- [x] Ensure persist update_seq test is working correctly
- [x] Merge to mono repo, undocumented / unreleased
- [x] Implement pouchdb-find
- [x] Make work in current Safari?

nolanlawson:
As a suggestion: maybe the adapter name should be `'indexeddb'`? This would allow the `'idb'` and `'indexeddb'` adapters to live side-by-side, while not burdening us forever with an funky name like `'idb2'` or `'idb-next'` or such.

After the monorepo release, I foresee some users have an adapter order like `'indexeddb'` `'idb'` `'websql'`, since idb-next gives performance improvements but doesn't support IE/Edge/old Android/etc..

daleharvey:
`indexeddb` sounds fine by me, I was going to need to look at that again @ https://github.com/pouchdb/pouchdb/pull/4984/files#diff-92135614f5c1d478bba08fcf500eb83dR19
 
I am not certain how we want to release this, fairly sure we dont want to ship it as an additional adapter to the current 2, thats a lot of extra code + complexity for little benefit. I am confident that we can have this adapter supported in ie (its currently working in safari albiet unreliably, with no special safari specific work). Releasing it as evergreen (ie no upgrades) is very tempting but leaves a lot of current users in the lurch, I think I will have to write a migration for this :(

Need to remember that this will also be dealing with migrations from webSQL to indexedDB, I do plan on making the migration code easy to remove from the build so we can have an indication of exactly how much it is costing us

nolanlawson:
I'm not proposing shipping it in `pouchdb` (for now), but rather as `pouchdb-adapter-indexeddb-next` or some such. Users can use it as part of a custom build, e.g.:

```js
var PouchDB = require('pouchdb-core')
  .plugin('pouchdb-adapter-indexeddb-next');
```
As for migrations, in the case of this one I think we should write a sub-package called `pouchdb-migrate` that allows users to explicitly migrate, e.g.:

```js
[...]

PR #5208, opened 2016-05-23
(#5207) - Add info to internal _getAttachment API

daleharvey:
The current _getAttachment` API doesnt provide enough information to get the attachment unless you have attachments keyed by digest alone, idb-next needs the revisions and document id data to be able to pick them up, this shouldnt impact anything as its internal only


nolanlawson:
:+1: when green. The only external libs this could potentially affect are worker-pouch and socket-pouch, but really that's on me because I'm using undocumented internal APIs. Also just checked, and looks like this won't impact me anyway.

daleharvey:
Cheers, all green

PR #5213, opened 2016-05-23
(#5213) - simplify pouchdb-errors, add better error unit tests

nolanlawson:
This reduces the pouchdb.min.js gzipped size from 45460 to 45222.

The trick was in no longer exporting every error from `pouchdb-errors`, which is unnecessary due to the new `pouchdb-for-coverage` setup, and creates a lot of un-mangleable variable names like `INVALID_REV` and `QUERY_PARSE_ERROR`.

Since I was worried about regressing, I also took the time to write a unit test (`test.errors.js`) based on the existing behavior of `getErrorTypeByProp()` and `generateErrorFromResponse()`. This ensures the new behavior is exactly the same as the old behavior.

At the same time, I think the test suite reveals some problems with our error-processing code (echoes of #5154), which I would like to address in a future PR. Looking through the suite, it's pretty obvious that we often squelch the error message and instead replace it with a generic error message like "Database encountered an unknown error," which is not very helpful.


packages/pouchdb-errors/src/index.js
```diff
@@ -306,6 +303,5 @@ export {
   INVALID_URL,
   getErrorTypeByProp,
   createError,
-  generateErrorFromResponse,
-  allErrors as errors
```
nolanlawson: ^ this is the thing we're no longer unnecessarily exporting (`allErrors`)

packages/pouchdb-for-coverage/src/errors.js
```diff
@@ -0,0 +1,53 @@
...
+  DB_MISSING: DB_MISSING,
+  WSQ_ERROR: WSQ_ERROR,
+  LDB_ERROR: LDB_ERROR,
+  FORBIDDEN: FORBIDDEN,
+  INVALID_REV: INVALID_REV,
+  FILE_EXISTS: FILE_EXISTS,
+  MISSING_STUB: MISSING_STUB,
+  IDB_ERROR: IDB_ERROR,
+  INVALID_URL: INVALID_URL
+};
```
nolanlawson: now `pouchdb-for-coverage` needs to export it, though, for testing

daleharvey:
Hah wow, yeh I have been scared to touch a lot of the error stuff because its fairly fragile. If we got rid of 500 errors it would be amazing for users, awesome job getting it started

👍 Cheers

PR #5215, opened 2016-05-23
(#5215) - move parseHex to websql, code cleanup

nolanlawson:
Really minor change - `parseHex()` is only used by websql, so no need
to have it in `pouchdb-utils` where it'll be just one more unused variable
in CJS-based builds using `pouchdb-utils` but not `pouchdb-adapter-websql`.

I also went ahead and removed `.js` from every import, since it's unnecessary, and made sure all imports are using single quotes and not double quotes.


daleharvey:
Looks good to me

PR #5218, opened 2016-05-24
(#5218) - add pouchdb-merge and pouchdb-adapter-utils

nolanlawson:
This adds two new internal (non-semver) modules called `pouchdb-merge`
and `pouchdb-adapter-utils`. The goal is to reduce the size of `pouchdb-http`
by isolating dependencies that `pouchdb-http` does not need, and this accomplishes
that.

Before, the min+gz bundle for pouchdb-http was 26389 bytes with Browserify and
21691 with Rollup. Now it's 21971 bytes with Browserify and 19134 with Rollup, a huge
improvement (mostly credited to removing the unnecessary `spark-md5` dependency).

This does increase the complexity of the project a bit, but I think if we're going to
sell the idea of a modular PouchDB, we need to make good on the claim that you can
actually reduce bytes by opting-in to the module system. And conceptually I think `pouchdb-merge`
and `pouchdb-adpater-utils` make sense (note that `pouchdb-adapter-utils` contains stuff
only used by the LevelDB, IndexedDB, and WebSQL adapters).


nolanlawson:
Just occurred to me `pouchdb-merge` is a bit unnecessary because everything that depends on `pouchdb-merge` also depends on `pouchdb-utils`. But I kinda like how self-contained `pouchdb-merge` is, and also `pouchdb-utils` has shrunk quite a bit, which is nice.

daleharvey:
I am a little confused, I thought with treeshaking we didnt need to care about isolating things at a module level? either way I think `pouchdb-merge` as its own module makes sense. so happy to merge this

nolanlawson:
So... it's complicated.

With Rollup and ES6 modules, there's no need for manual tree-shaking. Full stop.

If Rollup encounters a CJS module, then it's unable to tree-shake, so those need to be isolated. E.g. that's the main reason I isolated `pouchdb-md5` (`spark-md5`) and `pouchdb-json` (`vuvuzela`). That way, those CJS dependencies are not required for `pouchdb-http`. (Otherwise they hitchhike on top of `pouchdb-utils`.)

If a user is not using Rollup, but instead is using Webpack or Browserify (as I expect the majority of our users to do), then packages like `pouchdb-utils` are not tree-shaken at all. From the perspective of these users, the ideal solution would be to isolate each function to its own module, which I think is a bit extreme, so instead I search for the happy medium. Here I found that splitting `pouchdb-utils` into 3 modules resulted in a smaller bundle for `pouchdb-http` (at ~20kB), which is good enough for me.

In short, we're reaching the limits of how much we can tree-shake our users code when they are installing via the plugin method, since at that point we are beholden to npm and their bundler of choice.

PR #5219, opened 2016-05-24
(#5219) - fix "unary operator expected"

nolanlawson:
Made a mistake in this bash script. I'm putting `[skip ci]` on this
because it was failing anyway due to the fact that this would only affect
the couchdb-master tests.


daleharvey:
Doh I should have caught that but hadnt got back round to master tests, good call on the skip ci, we have been backed up for like 3 days now

Issue #5204, opened 2016-05-22
npm build failing


daleharvey:
After hitting https://github.com/pouchdb/pouchdb/issues/5203 I tried a manual rebuild, it started failing

```
~/src/pouchdb (master)$ npm run build

> pouchdb-monorepo@0.0.0 build /Users/daleharvey/src/pouchdb
> lerna bootstrap && npm run build-modules && npm run build-test

Lerna v2.0.0-beta.9
Linking all dependencies
Errored while running BootstrapCommand.execute
Error: Command failed: /bin/sh -c npm install js-extend@1.0.1 websql@0.4.4
npm ERR! Darwin 15.5.0
npm ERR! argv "/Users/daleharvey/.nvm/versions/node/v5.4.0/bin/node" "/Users/daleharvey/.nvm/versions/node/v5.4.0/bin/npm" "install" "js-extend@1.0.1" "websql@0.4.4"
npm ERR! node v5.4.0
npm ERR! npm  v3.3.12
npm ERR! path /Users/daleharvey/src/pouchdb/packages/pouchdb-adapter-node-websql/node_modules/sqlite3/node_modules/mkdirp/bin/cmd.js
npm ERR! code ENOENT
npm ERR! errno -2
npm ERR! syscall chmod

npm ERR! enoent ENOENT: no such file or directory, chmod '/Users/daleharvey/src/pouchdb/packages/pouchdb-adapter-node-websql/node_modules/sqlite3/node_modules/mkdirp/bin/cmd.js'
npm ERR! enoent This is most likely not a problem with npm itself
npm ERR! enoent and is related to npm not being able to find a file.
npm ERR! enoent

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/daleharvey/src/pouchdb/packages/pouchdb-adapter-node-websql/npm-debug.log

    at ChildProcess.exithandler (child_process.js:213:12)
    at emitTwo (events.js:87:13)
    at ChildProcess.emit (events.js:172:7)
    at maybeClose (internal/child_process.js:821:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:211:5)

npm ERR! Darwin 15.5.0
npm ERR! argv "/Users/daleharvey/.nvm/versions/node/v5.4.0/bin/node" "/Users/daleharvey/.nvm/versions/node/v5.4.0/bin/npm" "run" "build"
npm ERR! node v5.4.0
npm ERR! npm  v3.3.12
[...]

PR #5220, opened 2016-05-25
(#5204) - skip non-directories in packages/

nolanlawson:


daleharvey:
Sweet cheers, given what this touches I was happy enough cancelling the run after it build like 10 times fine

Issue #5224, opened 2016-05-26
Rename pouchb-adapter-indexeddb to pouchdb-adapter-idb


nolanlawson:
Based on discussion with @daleharvey I think the best move in terms of naming for the multi-package architecture would be to use `pouchdb-adapter-idb` for the current one, and `pouchdb-adapter-indexeddb` for the new one.

This would be nice because all the other adapters have a package name that matches their "adapter name" (i.e. what you see when you do `new PouchDB('foo').adapter`), except for `indexeddb` currently (which has an adapter name of `'idb'`). Since we haven't released this yet, we can still freely change the adapter name, so now is the time to make a decision.

I'm suggesting `pouchdb-adapter-idb` and `'idb'` for the old one and `pouchdb-adapter-indexeddb` and `'indexeddb'` for the new one (i.e. what we've called `idb-next`).

nolanlawson:
The alternative is to call it something like `idb-next` or `idb2` which I think is ugly if we plan on maintaining the new one for the long term.

daleharvey:
Im 👍 on that, will do the PR since I need to get used to this stuff

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/bf9ff55722b5ab6c756a9fb5ea6b8b33dae08958

nolanlawson:
post facto :+1:

PR #5226, opened 2016-05-26
(#5224) - Rename adapter-indexeddb to adapter-idb

daleharvey:


daleharvey:
I was going to merge idb-next at the same time, however I feel safer doing it in seperate parts so I can be sure we dont accidently release idb-next early

daleharvey:
@colinskow +1'd and @nolanlawson suggested the change, so merging cheers

PR #5227, opened 2016-05-26
WIP - New idb next

daleharvey:


daleharvey:
So I based this on top of the renamed `pouchdb-adapter-idb`, It seems like it should be working mostly, the end result I would like is so see a `dist/pouchdb-next.js` built so I can manually run the tests and get them fixed up / finalised, between the errors in this build and https://github.com/pouchdb/pouchdb/issues/5204 (looking into this now, its killing me, virtually every rebuild needs to be done from scratch) I am having a hard time figuring out whats wrong

daleharvey:
@nolanlawson any idea how to fix this? I get a `pouchdb-next/lib/index.js` built however not a dist (addition is @ https://github.com/pouchdb/pouchdb/pull/5227/files#diff-d97268dd585cf586db3a7cd66846c9d8R37) I figured it should be the same as localstorage etc, but it doesnt like something

daleharvey:
Ok found and fixed previous issues, this now produces a `pouchdb.next.js`, however http://127.0.0.1:8000/tests/integration/?grep=test.basics.js-local&src=../../packages/pouchdb/dist/pouchdb.next.js is not letting me test it with, also looking at dist I am seeing 

```
-rw-r--r--   1 daleharvey  staff   945K 26 May 17:24 pouchdb.js
-rw-r--r--   1 daleharvey  staff   630K 26 May 17:24 pouchdb.next.js

```

Which means there must be some bug in the build there, those sizes are rather large

daleharvey:
Last commit I tried a different way, the package sizes seem to be correct but still seeing `PouchDB is not defined` when I try to test

nolanlawson:
@daleharvey yeah there is a bug in `build-pouchdb.js` where it doesn't `console.log` any Uglify or Browserify errors. I need to fix it to handle those.

PR #5228, opened 2016-05-26
(#5224) rename pouchdb-adapter-indexeddb in docs too

nolanlawson:


daleharvey:
Sorry missed that in my grep somehow, cheers

PR #5229, opened 2016-05-26
(#5207) - Add new indexeddb adapter to repo

daleharvey:


daleharvey:
So this is just adding it to the repo, not building or anything, now that I have done it I am not even sure we need to

nolanlawson:
Let's save it until the release. I want to make sure we can publish without a hitch

PR #5230, opened 2016-05-26
(#5207) - Add new indexeddb adapter to repo

daleharvey:


daleharvey:
Less ambitious, this lets me work on the plugin by patching where `pouchdb-browser` picks up its idb adapter but shouldnt affect any current build, if its cool to merge this then we can figure out how to get a build produce to make development easier

daleharvey:
err forgot to lint, fixed now

nolanlawson:
This needs a fix to the publish script to avoid publishing it, but I'll fix that on the way in. Definitely agree with having this in the main repo so we can iterate on it.

nolanlawson:
ced745440c239b98911fd51130f56aa6e8fadce4 and 743983acceeb4c373d53b29bd0cd6abffd9273c6

PR #5232, opened 2016-05-27
(#4767) - Add ignores for 100% coverage

daleharvey:


daleharvey:
I am comfortable ignoring these, we always use `defaults()` in node to set the directory for our data so the `instanceof` always hits in `PouchAlt` and not the core one (that path is tested in the browser though) and our coverage tests dont run against couchdb-master. I would like to get back to a stable 100% because anything else makes it very hard to keep track of regressions

daleharvey:
Added another, the `catch` is purely there to catch a race so integration testing it reliably is pretty much impossible, this is like our most exercised code so not worried about it here

nolanlawson:
:+1:

daleharvey:
So there is still one or maybe 2 intermittents, will track them down seperately though, cheers

Issue #3175, opened 2014-12-14
Don't need to auto-compact new documents in leveldb


nolanlawson:
See https://github.com/pouchdb/pouchdb/pull/3090/files#r21797003, small omission I made in that PR. This issue assumes that that PR will get merged as-is.

daleharvey:
I tried out that patch and it genuinely broke tests (more than me forgetting to pass `isUpdate`), so closing out

PR #5233, opened 2016-05-27
(#3175) - Dont auto compact new writes

daleharvey:


daleharvey:
I mostly wanted to close that issue :) I dont think this needs a test as long as it passes current tests, performance tests are next up on my plate

Issue #5082, opened 2016-04-22
Remove references to IrisCouch


zaccolley:
Dunno if they still host servers but I think the references to them is confusing as you can't sign up with them.

Anyone confirm that they're dormant? [Last blog post](http://www.iriscouch.com/blog) and [tweet](https://twitter.com/iriscouch/status/411507182090219520) was 3 years ago

http://www.iriscouch.com/service

nolanlawson:
They also shut down my account, yes.

I'd be happy to change the tutorial to use Digital Ocean with CouchDB pre-installed. It was pretty painless when I tried to set it up. I believe Azure also has the same feature.

nickcolley:
DO w/ CouchDB sounds like a good shout, could we think about maybe using PouchDB-Server for the tutorial?

nolanlawson:
@nickcolley Problem with PouchDB Server is that no provider I'm aware of (DO, Azure, etc.) offers a one-click install for it. Whereas DO does offer a one-click install for CouchDB.

I would also be +1 for changing the tutorial to use a local PouchDB Server rather than DO. Benefit is that we are not giving free advertising to any cloud provider, and we are also insulated from changes to their service. Also, the benefit of using PouchDB Server vs CouchDB in this scenario is that it is typically easier for people to install.

nickcolley:
Wonder if https://zeit.co/now/ could work since we've just got npm deps with PouchDb Server?

nolanlawson:
It looks nice, but again, I'm worried about us getting IrisCouch'ed again. I put all that effort into making a video tutorial and everything, and now it's useless because they pulled the plug. Would much rather hitch our wagon to a more reliable horse.

daleharvey:
So I plan on having pouchbase.com be a hosted admin party instance of a pouch instance that gets wiped every night or week or something, exactly so people can test against / write tutorials against a simple to maintain service that I will keep up.

daleharvey:
Theres a small question about how exactly it works, worried a little about everyone using pouchbase.com/demo and each conflicting with each other, but likely easily solved, happy to hear any input on this

cdaringe:
it may be more effort than desired, by maxodgen and mafintosh have a nifty solution to this sort of thing.  essentially

- make a really basic barebones docker image,
- on connection to SOME_SITE, run an image instance
- connect running docker image terminal to webterminal, and finally
- on website close, kill the running docker image instance

[...]

PR #5234, opened 2016-05-27
(#5082) - Remove references to iriscouch

daleharvey:
[skip ci]


daleharvey:
We could have put pouchbase in there, but I wanted these parts to mostly be agnostic so we dont need to change them again, I think it was worth mentioning pouchdb-server in the setup getting started guide though since it is the easiest to get started with

garrensmith:
+1

Issue #5080, opened 2016-04-21
Is replication `batch_size` affected by attachments


jo:
> options.batch_size: Number of documents to process at a time. Defaults to 100. This affects the number of docs held in memory and the number sent at a time to the target server. You may need to adjust downward if targeting devices with low amounts of memory (e.g. phones) or **if the documents are large in size (e.g. with attachments)**. If your documents are small in size, then increasing this number will probably speed replication up.

Since attachments aren't fetched inline anymore is this still valid?

nolanlawson:
No, it's not.

jo:
OK, I'll fix the docs then.

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/1a7d7e7cf08591288128985556da20c9995b52a3

ig3:
In my experience, the number and size of attachments is still relevant.

The comment (not the altered part) is subtly incorrect: the batch_size is not the number of documents processed in a batch - it is the number of change feed items processed in a batch. Each change feed item has a single document ID, so in this sense it refers to a single document, but it may have multiple revisions and each revision will be synchronized. The distinction is significant if the document size is significant or if the document revisions have significant number or size of attachments. There may be many conflicted revisions to synchronize.

Neither the number of document revisions nor the number of attachments are considered when determining whether a batch has reached the batch_size limit - only the number of change feed items is considered. Thus attachments are not relevant to batch_size, but this is not to say that the number or size of attachments is not relevant to the network and memory resources required to process a batch.

While attachments are no longer fetched inline, they are still fetched. getAllDocs returns all the document revisions in the batch that require synchronization, including their attachments, unless I am misunderstanding the code. 

While I haven't updated the version of PouchDB I am using for a couple of weeks now, so don't have the latest changes, I am still seeing memory utilization significantly impacted by number and size of attachments and batch_size, using the http adapter. For example, when I try to replicate a database with on the order of 100 large attachments per small document, I find that memory is exhausted with the default batch size and, in fact, sometimes even with batch_size set to 1, particularly where there are unresolved conflicts in the database.

There are also issues with timeouts related to the number of documents and attachments fetched in a batch. These can be mitigated with the recent changes allowing ajax option to be passed through synchronize, but they are not trivial. The time required to process a batch, including fetching attachments, can be reduced by reducing batch_size.

While attachments don't affect "batch_size", they do affect "batch size", if by "batch size" one means the time and memory required to process the batch.

[...]

PR #5235, opened 2016-05-27
(#5080) - Clarify batch_size documentation

daleharvey:
@ig3's suggested edit makes sense to me


daleharvey:
Being that this is @ig3's suggestion, I am 👍 on it

Issue #5225, opened 2016-05-26
Should guard all usages of `console`


nolanlawson:
Android in pre-Kitkat doesn't support `console` in a web worker. We still have plenty of unguarded usages of `console`, leading to problems like https://github.com/nolanlawson/worker-pouch/issues/16.

My recommended guard is e.g. `typeof console !== 'undefined' && 'info' in console` (or `warn` or whatever).

Unfortunately we can't really test this because Saucelabs <4.4 isn't reliable, but we can at least manually fix it.

schultyy:
I don't have an Android device at hand but I would be in to work on that.

daleharvey:
Feel free to work on this without an android device, an idea to test this may be

```
if (process.env.NO_CONSOLE) { 
  delete global.console;
}
``` in https://github.com/pouchdb/pouchdb/blob/master/tests/integration/node.setup.js, then run the tests with `NO_CONSOLE=1 npm test`, if they fail then thats all the testing that should be needed

daleharvey:
hrm I just tried that and it doesnt work since mocha tries to use the console :(, either way feel free to work on it and we can discuss / figure out if there is any way to reasonably test it, something like this may not strictly require tests

schultyy:
Alright, will do.

schultyy:
Does it make sense to have some sort of `console` mock to test if the guard works?

daleharvey:
The problem is things outside PouchDB are also using the console and its hard to differentiate, thats pretty much with what I was thinking with `delete window.console`. To be honest this sound mostly like one of those issues we fix and it just stays fixed, we will just be mindful of console's in reviews and it should be fine, but if you can think of a test that works it would be even better :)

Fixed in https://github.com/pouchdb/pouchdb/commit/16030b896769abfe39798291c73078f412fbfddc

PR #5236, opened 2016-05-27
(#5225) - Guard uses of console before calling it

schultyy:
PR for #5225 

This adds guards for all usages of the `console` object. It is checked if the console object itself is present and also the actual method to be called.


packages/pouchdb/tonic-example.js
```diff
@@ -22,16 +22,21 @@ await db.put({
   title: "Sound and Vision",
 })
 
-// later you can fetch your doc
-console.log(await db.get('mydoc'))
+if(typeof console !== undefined && 'log' in console) {
+  // later you can fetch your doc
+  console.log(await db.get('mydoc'))
+}
```
daleharvey: We dont need to do this in the example code

daleharvey:
This is a great start, thanks but i think we should clean the code up a little bit, if we created a new file `console.js` in https://github.com/pouchdb/pouchdb/tree/master/packages/pouchdb-utils/src and exported 2 functions `warn` and `error` that looked something like
```
function warn() { 
  if (typeof console !== undefined && 'warn' in console) {
    console.warn.apply(console, arguments);
  }
}
```
and used those functions it would be cleaner, what do you think? 

Thanks
Dale

packages/pouchdb-core/src/setup.js
```diff
@@ -66,9 +66,11 @@ PouchDB.parseAdapter = function (name, opts) {
       if (adapterName in PouchDB.adapters) {
         /* istanbul ignore if */
         if (skipIdb && adapterName === 'idb') {
-          // log it, because this can be confusing during development
-          console.log('PouchDB is downgrading "' + name + '" to WebSQL to' +
-            ' avoid data loss, because it was already opened with WebSQL.');
+          if(typeof console !== undefined && 'log' in console) {
```
daleharvey: If we only expose `warn` and `error` in pouchdb-utils I think we should switch this to a warn

schultyy:
Sounds good. So everywhere where `console` is used now that new two functions are imported then?

schultyy:
What about the normal `console.log` statements? Do they stay as they are right now?

nolanlawson:
I would switch to using Dale's suggestion of `pouchdb-utils/src/console` as well. However, I would keep it simple and just make it one file, something like this:

```js
function console(type, msg) {
  if (typeof console !== 'undefined' && type in console) {
    console[type](message);
  }
}
```
```js
console('info', 'my info message');
```

packages/pouchdb-adapter-idb/src/index.js
```diff
@@ -883,7 +883,9 @@ function init(api, opts, callback) {
     };
 
     idb.onabort = function (e) {
-      console.error('Database has a global failure', e.target.error);
+      if(typeof console !== undefined && 'error' in console) {
```
nolanlawson: not sure if eslint enforces this, but we do `if (` not `if(`

schultyy:
@nolanlawson I implemented it as you suggested. There's this https://github.com/pouchdb/pouchdb/blob/master/packages/pouchdb-mapreduce/src/index.js#L33-L39 where I'm not sure if this should be kept as it is or also converted to the new `guardedConsole`. What do you think?

daleharvey:
That looks awesome thanks, since `pouchdb-mapreduce` already imports from `pouchdb-utils` I think its good to get all these consistent and make the switch there too, once thats done I think this is good to land, thanks :)

daleharvey:
Awesome, so one last issue, this is currently failing in CI, have you been able to install and build locally? as far as I can tell you need to 1, add pouchdb-utils to the dependencies in `packages/pouchdb/package.json` and 2. change the imports to `import { guardedConsole } .....` in the extras files to get this to build

packages/pouchdb-utils/src/guardedConsole.js
```diff
@@ -0,0 +1,8 @@
+function guardedConsole(method) {
+  if (console !== 'undefined' && method in console) {
+    var args = Array.prototype.slice.call(arguments, 1);
+    console[method].apply(null, args);
```
daleharvey: This is failing in the browser, fairly sure it needs to be 
 
    console[method].apply(console, args);

daleharvey:
Heh sorry one more issue, posted the fix in comments

schultyy:
I will have a look in the evening and fix that :)

daleharvey:
ok awesome this is all good, I went and screwed up the coverage myself I think, but your code looks great, 👍 and merged

schultyy:
Great :)

PR #5242, opened 2016-05-29
(#4767) - Ignore cancelled clause and sync events during coverage

daleharvey:


daleharvey:
Ok, so I tried to get tests working for these however they are extremely problematic (they are the same clauses I kept hitting last time getting to 100%). 

The way we handle cancelled replication would pretty much need a rewrite or huge amounts of instrumentation to be able to reliably hit each clause since its being handled in so many places, these are being hit like 90% of the time but occasionally one of them doesnt get touched. The cancel tests that are disabled would help a little but they still wouldnt be reliable (its things like replication being cancelled after the checkpointing has finished initalising). 

Similiarly with sync I think I would need to instrument a server to guarantee to wait to respond to particular requests, again these clauses get hit 90% of the time however occasionally something like the pull replication work happens faster than expected. 

If someone wanted to pick up the work to get these reliably 100% now then that would be awesome, but failing that I think we should call amnesty on coverage and get us to the point where 100% is guaranteed and anything less is a fail, then when this code is touched again we make sure to keep in mind coverage and attempt to remove them as we go along.

daleharvey:
and with these in there a 100% run 20 times in a row @ https://travis-ci.org/pouchdb/pouchdb/builds/133720999

garrensmith:
I'm +1 on this.

Issue #4444, opened 2015-10-09
Custom constructors (via .defaults) leak eventEmitters


daleharvey:
Try adding a new test to test.defaults.js with a custom constructor and the test suite fails

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/2d99a7342ce0c6d0f59ca1964862b5987d470f75

PR #5246, opened 2016-05-29
(#4444) - Dont leak event emitters

daleharvey:


nolanlawson:
Great catch, +1 when green

Issue #5161, opened 2016-05-15
Navigation is stuck over content


nickcolley:
Reported by https://twitter.com/nekrtemplar

![](https://pbs.twimg.com/media/CibUjrlUoAAAZQ7.jpg)

Full twitter thread: https://twitter.com/nekrtemplar/status/731783145461297153

My suspicion is that Chrome Beta is behaving weird with the position: sticky polyfill.

daleharvey:
Yeh I took a look at this, tried switching to another polyfill and seen the exact same problem, I can reproduce on chrome beta, I think the only option to fix is remove the polyfill

daleharvey:
So yeh, its gonna be a while for chrome to support position sticky, will just have to hope chrome beta fixes this

nolanlawson:
fixed in https://github.com/pouchdb/pouchdb/pull/5337

PR #5247, opened 2016-05-29
(#5161) - Remove sticky polyfill

daleharvey:
[skip ci]


daleharvey:
So I am pretty close to the edge on whether we should do this, on one hand we wont be broken in chrome beta / dev (and so potentially future chromes when they release), on the other chrome (only) users may have to scroll a bunch extra in the desktop view. Chrome users will eventually get fixed (`position: sticky` is currently being worked on, I think its expected in July)

nickcolley:
I'd be :+1: if Chrome is about to ship sticky...

daleharvey:
Its slated for 52, however looking like its having problems sticking https://bugs.chromium.org/p/chromium/issues/detail?id=231752#c28

daleharvey:
So yeh, its gonna be a while for chrome to support position sticky, will just have to hope chrome beta fixes this

Issue #4922, opened 2016-02-22
destroyed event fired twice


cdaringe:
hi all!  i'm seeing the `destroyed` event fired twice [in a barebones example](https://github.com/cdaringe/pouchdb-double-destroy).

```js
'use strict';
const pdbs = require('spawn-pouchdb-server');
const PouchDB = require('pouchdb');
pdbs({
    backend: false, // e.g. memdown
    config: { file: false },
    log: { file: false },
    port: 8888,
    timeout: 10000, // in ms
    verbose: true
}, function (err, srv) {
    var count = 0;
    PouchDB.on('destroyed', () => {
        ++count;
        console.log('destroy count:', count);
    });
    const p = new PouchDB('http://localhost:8888/test-db');
    p.put({ _id: 'dummy-id' })
    .then(() => p.destroy())
    .then(() => {
        srv.kill();
    })
});
```

yields ==>

```
cdieringer@Snapper-osx:~/node/double-pouch-destroy$ node index.js
writing config file at /var/folders/s6/fzmmr2s11_92ft6bz5m3qhqr0000gn/T/tmp-24120ghYVmUm4leXR
starting pouchdb-server --config /var/folders/s6/fzmmr2s11_92ft6bz5m3qhqr0000gn/T/tmp-24120ghYVmUm4leXR --in-memory
destroy count: 1
destroy count: 2
```

I've tested with CouchDB as the backend and PouchDB server.  The GH repo posted above should be a real easy thing to clone and try yourself.  thanks!

nolanlawson:
Yup, I can confirm. Also for the record, you don't need to `require('spawn-pouchdb-server')` to reproduce this.

cdaringe:
i should have mentioned that the `created` event also gets fired a of couple times.  i did obverse variable counts of calls to the event handler depending on the run.

cdaringe:
not sure how the pure http client works, but if it does local caching (e.g. has a local db, vs. purely connecting to the remote db), the local version and the remote DB would be instantiated in my example.  if you reproduced without any remote, however, then :(

daleharvey:
So I can reproduce with a `http` example, however 
```
'use strict';

const PouchDB = require('pouchdb');
var count = 0;

//const p = new PouchDB('http://localhost:5984/test-db');
const p = new PouchDB('test-db');

PouchDB.on('destroyed', () => {
  ++count;
  console.log('destroy count:', count);
});

[...]

PR #5249, opened 2016-05-30
(#4922) - Do not call 'destroyed' twice for http instances

daleharvey:
At some point (probably prior to mapreduce / dependentDb's) the http adapter was wholly responsible for its own `destroy` so it emitted its own `destroyed` events, then we switched it to use adapters.js's `destroy` which also fires the events and didnt remove the old versions


willholley:
+1

daleharvey:
Sweet thanks

PR #5252, opened 2016-05-30
(#3923) - Add ability to set url as prefix

daleharvey:


daleharvey:
Heh I was expecting more hassle from this commit, but tests seem to be going well, I could probably get rid of either opts.originalName or opts.__db_name as well, but figured I would keep changes to a minimum here

willholley:
nice! +1

daleharvey:
Awesome thanks

Issue #5257, opened 2016-05-31
Lerna not correctly installing leveldown optionalDependency


yaronyg:
The current testing instructions defined [here](https://github.com/pouchdb/pouchdb/blob/master/TESTING.md) could be made more friendly through the addition of two bits of information in the node.js section:

1. Start with npm install
2. Then 'npm install leveldown' so the tests will 'just work'

daleharvey:
Yeh I have had to do this a couple of times but not all the time which was weird. I thought it was something `lerna` should take care of however `npm install` is all that should need for the tests to just work, either leveldown needs added to the root package.json or if its supposed to be picked up already we should figure out why it isnt

yaronyg:
I just ran a clean test on my mac and I got an error that leveldown wasn't there. Leveldown is also not in the package.json. I thought of adding it as a dev dependency but I wasn't sure if that was even strictly true since there are other adapters the tests can run on. That's why I thought it would be easiest to just make it an instruction so that those who know they are doing something fancy can do that without having to deal with downloading something like leveldown.

daleharvey:
So I think it would be good to hear what @nolanlawson thinks since he made that change, but I would very much prefer to make the flow for new / general contributors as easy as possible than having optimisations for niche use cases

nolanlawson:
I have no idea why leveldown wouldn't be installed correctly. My guess is that it has to do with leveldown being an optional dependency in the `pouchdb` repo, but I'm not sure.

FWIW I have never run into this problem and I don't see why `leveldown` would need to be installed at the root level. However it is true that you need to `npm install` first; that's pretty standard with any Node project but could use documenting.

yaronyg:
Repo:
1. Have a clean directory with no node_module directories anywhere upstream of it, probably a good idea to also run 'npm list -g leveldown' to make sure leveldown isn't installed globally.
2. 'git clone https://github.com/pouchdb/pouchdb.git'
3. cd pouchdb
4. npm install
5. npm run build
6. npm run test

Expected result: Tests run
Actual result: The tests fail at 'test.all_docs.js-local Testing all docs' because 'the 'leveldown' package is not available.'

Expected Result: Tests run
Actual Result: The tests fail after running 24 of them on the test "test.all_docs.js-local Testing all docs' with the error that the leveldown package is not available.

This makes sense as the root package.json doesn't contain leveldown.

[...]

PR #5258, opened 2016-05-31
(#5257) - Added notes on test setup

yaronyg:
#5257 - A simple solution I suppose. :) But per @daleharvey comment we could decide to try to add leveldown to package.json as a dev dependency.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -12,7 +12,6 @@ Given that you have [installed a CouchDB server](#installing-a-couchdb-server).
 Set up your test environment with:
 
     $ npm install
-    $ npm install leveldown
 
 Run all tests with:
 
--- a/TESTING.md
+++ b/TESTING.md
@@ -9,6 +9,11 @@ The PouchDB test suite expects an instance of CouchDB (version 1.6.1 and above) 
 
 Given that you have [installed a CouchDB server](#installing-a-couchdb-server).
 
+Set up your test environment with:
+
+    $ npm install
+    $ npm install leveldown
+
 Run all tests with:
 
     $ npm test
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "watch-glob": "0.1.3",
     "wd": "0.4.0"
   },
-  "devDependencies": {
+  "optionalDependencies": {
     "leveldown": "1.4.6"
   }
 }
--- a/package.json
+++ b/package.json
@@ -80,5 +80,8 @@
     "uglify-js": "2.6.2",
     "watch-glob": "0.1.3",
     "wd": "0.4.0"
+  },
+  "devDependencies": {
+    "leveldown": "1.4.6"
   }
 }


colinskow:
I think it would make more sense to add leveldown as a dev dependency so people don't need to remember to install it.

yaronyg:
@colinskow As I mentioned in #5257 the reason I didn't do that is because strictly speaking one doesn't actually need leveldown to run the tests. One can set up another adapter to do it. So if we put leveldown as a dev dependency then someone who doesn't need it to test is being forced to download it and native dependencies are just no fun. But I don't have a strong feeling here one way or another, I just went with the all text route as being the least disruptive. So if you are o.k. with forcing leveldown even when it's not required I'm fine amending the PR.

colinskow:
@yaronyg OK, I get what you are saying. LevelDown can be a pain to install, so no sense in making everyone do it if not necessary.

@daleharvey would it be practical at all to default the tests to `memdown`? I think this could even speed up our builds quite a bit. It would also avoid installing LevelDown which can occasionally be a pain on some environments.

daleharvey:
Its possible, I am wary of using memdown for the same reason we test against couchdb and not pouchdb-express-router even though that would be easier, leveldown is the base test against all level* adapters and is widely used in pouch projects, it needs to work.

leveldown has fixed a lot of its cross platform + time consuming install issues recently, its not perfect but I think having it as an optionalDependency is totally reasonable, advanced pouchdb/npm users can still skip it and it gives us an easy workflow of testing our 1st class implementation. The majority of issues with the leveldown dependency were with browser users having to install it when using pouchdb as a dep and we have fixed that with monorepo

colinskow:
OK, +1

yaronyg:
Is this any better?

daleharvey:
heh sorry to be a hassle, can you put it in `optionalDependencies` https://docs.npmjs.com/files/package.json#optionaldependencies, this lets people ignore it if they want to, it will still succeed on platforms that leveldown fails but everything will just work for the 99%

yaronyg:
If I'm reading optionaldependencies correctly then it will try to install for everyone, all the time. That sounds.... bad. I thought PouchDB did a bunch of work not to require leveldown all the time? Now I realize that if optionaldependencies is doing its job then a failure should be ignored but still... it feels like pushing our luck. I'll change it if you want but I wanted to make sure that we captured the implication that it has everyone trying to install leveldown.

daleharvey:
So we do not use `/packages.json` for anything other than development, its not published in anyway (the pouchdb npm package is `/packages/pouchdb/package.json` so every dependency here is a `devDependency` in reality, it will only ever be `npm install`'d, being optional just allows users to skip it or lets it fail

daleharvey:
The work that Nolan did to avoid everyone having to install `leveldown` all the time is we publish `pouchdb-browser` or even `pouchdb-http` if you arent using level adapters, you can also build your own configuration if you want

yaronyg:
So I just deleted my depo and re-cloned it to get a real 'first time user' experience and the optional dependency did not install. :( I cancelled the build to clear the queue while I figure out what I screwed up.

yaronyg:
Repeat after me - Yaron is an idiot. Sigh.... I really need to find another line of work. This programming stuff is clearly too hard for me. Here's a hint, um.... when you are testing something in a clean depot, it can be useful to switch TO THE RIGHT BRANCH!!!!!!

yaronyg:
O.k. using the right branch turns out to make a big difference and yes it works just fine.

daleharvey:
Awesome this looks great, I actually want to hold off on this until @nolanlawson looks at it though, it works for me however I am actually still confused at how this doesnt currently work, it may be a lerna / setup bug, the tests `require('../../packages/pouchdb')` which has the leveldown dep

yaronyg:
NP. We'll wait for @nolanlawson to get back.

nolanlawson:
See https://github.com/pouchdb/pouchdb/issues/5257#issuecomment-223944053.

I don't understand what the bug is or where it's coming from. I've never run into this issue.

In principle, you shouldn't need to install `leveldown` at the root level. That ought to be a useless operation.

I made `leveldown` an optional dependency of `pouchdb`, i.e. our kitchen-sink build. Same for `websql` (i.e. `sqlite3`). This matches the current behavior of 5.3.2.

For the custom builds, neither are optional, because it's assumed you know what you're doing when you do the custom build.

My hunch is that y'all are running into some kind of Lerna bug that I've not been able to reproduce. Maybe it's an env issue? I'm using Node 5 on Mac OS X ElCap.

@colinskow Dev/regular dependency makes no difference at the top level, because the top level project is just for development. It's actually a `private` package with the totally made-up name of `pouchdb-monorepo`.

daleharvey:
Lets close this out and figure out what the right fix should be in the issue as to not split the conversation between both threads

Issue #5238, opened 2016-05-28
PouchDB throwing unhandled rejection on replicate in Node.js


yaronyg:
### Environment (Node.js/browser/hybrid app/etc.)

Node.js

### Browser/platform (Chrome/FF/Safari/Edge/iOS/Android/etc.)

N/A

### Adapter (IndexedDB/WebSQL/LevelDB/etc.)

LevelDB

### Server (CouchDB/Cloudant/Couchbase/PouchDB Server/etc.)

None (really)

### Please describe your issue. Test cases (e.g. JSBin) are very helpful!

I was testing what would happen if I tried to issue a replicate.to pointed at a remote database that wasn't even listening for connections. I successfully received an error event (which in itself was odd since live: true but that's another issue) but my global process.on('unhandledRejection') handler in node still picked up a unhandled rejection! Just for chuckles I even added a catch which did catch the reject BUT I still get an unhandledRejection event!

```javascript
var PouchDB = require('pouchdb');

var dbPath = "foo";

var DefaultPouch = PouchDB.defaults({
  prefix: dbPath
});

process.on('unhandledRejection', function (err, p) {
  console.trace(err);
});

var badUrl = 'http://127.0.0.1:2342/db/foo';

var ajaxOptions = {
  live: true,
  retry: true,
  skip_setup: true
};

var remoteDB = new DefaultPouch(badUrl, ajaxOptions);
remoteDB.replicate.to('foo')
    .on('paused', function (err) {
      console.log('Got paused with ' + err);
    })
    .on('active', function () {
      console.log('Replication resumed');
    })
    .on('denied', function (err) {
      console.log('We got denied on a PouchDB access, this really should ' +
        'not happen - ' + err);
    })
    .on('complete', function (info) {
      console.log('complete - ' + info);
    })
    .on('error', function (err) {
      console.log('Got error on replication - ' + err);
    })
    .on('change', function (info) {
      console.log('change - ' + info);
    })
    .then(function (stuff) {
      console.log('then - '  + stuff);
    })
    .catch(function (err) {
      console.log('catch - ' + err);
    });
```

yaronyg:
[...]

PR #5262, opened 2016-06-01
(#5238) - Fix unhandled rejection in changes

yaronyg:
changes - Checks to make sure someone is listening for the error event
before emitting it. See #5238 for details.

index - Registering listenerCount

listenerCount - Supports both the modern and legacy versions of
  listenerCount

test.changes.js - This test will fail if there is an unhandled rejection
so we use it to test that the fix worked.

utils - Enables adding/removing events on the either the process or
windows object depending on environment. And dealing with the fact
that the unhandledRejection event is spelled differently in node vs
the browser. Note that this is a NOOP in Firefox but works in Chrome. This
just means if there is a bug in Firefox it won't be caught.

### DIFF
--- a/tests/integration/test.changes.js
+++ b/tests/integration/test.changes.js
@@ -465,7 +465,16 @@ adapters.forEach(function (adapter) {
       });
     });
 
-    it('3356 throw inside a filter', function (done) {
+    it.only('3356 throw inside a filter', function (done) {
+      var testFor5238 = function (err) {
+        done('There was an unhandledRejection ' + err);
+      };
+      testUtils.addGlobalEventListener(
+        testUtils.getUnHandledRejectionEventName(), testFor5238);
+      after(function () {
+        testUtils.removeGlobalEventListener(
+          testUtils.getUnHandledRejectionEventName(), testFor5238);
+      });
       var db = new PouchDB(dbs.name);
       db.put({
         _id: "_design/test",
--- a/tests/integration/test.changes.js
+++ b/tests/integration/test.changes.js
@@ -465,16 +465,11 @@ adapters.forEach(function (adapter) {
       });
     });
 
-    it.only('3356 throw inside a filter', function (done) {
+    it('3356 throw inside a filter', function (done) {
       var testFor5238 = function (err) {
         done('There was an unhandledRejection ' + err);
       };
-      testUtils.addGlobalEventListener(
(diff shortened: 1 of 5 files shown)


tests/integration/test.changes.js
```diff
@@ -465,7 +465,16 @@ adapters.forEach(function (adapter) {
...
     });
 
-    it('3356 throw inside a filter', function (done) {
+    it.only('3356 throw inside a filter', function (done) {
+      var testFor5238 = function (err) {
+        done('There was an unhandledRejection ' + err);
+      };
+      testUtils.addGlobalEventListener(
+        testUtils.getUnHandledRejectionEventName(), testFor5238);
+      after(function () {
```
daleharvey: I didnt know it could work like this, nice, however for consistency any time we are removing a listener to cleanup a test we do so just before calling `done` (or returning the last promise) so if we could do that, ie

        }).catch(function () {
          testUtils.removeGlobalEventListener(testUtils.getUnHandledRejectionEventName(), testFor5238);
          done();

Its a little clearer I think. Also a tiny nit but if we could shorten the addListener line, maybe 

    testUtils.addUnhandledRejectionListener(testFor5238)

And do the name switching inside testUtils, how does that sound?

tests/integration/utils.js
```diff
@@ -281,6 +281,37 @@ testUtils.makeBlob = function (data, type) {
   }
 };
 
+testUtils.getUnHandledRejectionEventName = function () {
+  return typeof window !== 'undefined' ? 'unhandledrejection' :
+    'unhandledRejection';
+}
```
daleharvey: Needs a semi colon here

daleharvey:
So the tests failed with a minor eslint, I pointed it out in the comments, I had a few small comments about the code style but otherwise its looking great, hopefully get a green run with that semi colon and I think it will be good to land

yaronyg:
I apologize. Somehow I managed to not commit my final changes which fixed the 'only' and the semicolon. This is why I love CI. It makes my screw ups less costly to everyone else.

I made the changes you requested. Note that to safely get rid of 'after' I had to use a pattern we use a lot in Thali which is 'then'/'catch'/'then' where the final then is used for clean up regardless of what happened to the test.

daleharvey:
Awesome that looks great, thanks and merged

yaronyg:
Thanks for reviewing! And now on to #5264

Issue #5260, opened 2016-05-31
Need to detect unhandled rejections in tests


yaronyg:
# Problem
Bug #5238 wasn't detected because we do not have any general mechanism for telling when we have unhandled rejections. 

# How to solve it?
We can put a process.on('unhandledRejection') handler into node.setup.js which would then apply to all the node tests. In really simple cases this can work. The unhandledRejection handler just throws an exception, this gets picked up by Mocha who then marks the test as failed. This is what happened with the test '4049 retry while starting offfline' before we fixed #5238.

The problem is that some tests are explicitly testing exceptions and using catch. Unfortunately if two exceptions are thrown at the same time one gets hidden. So if we have an (expected) exception in the test which, in the process of handling that exception should have an unhandled rejection then the unhandledRejection handler will catch it but when it throws the new exception will be caught by the promise framework along with the old one and both will be turned into a single failure that is then caught by the test which 'passes' since it was expecting an exception. An example of this behavior is the test '3356 throw inside a filter' (again, before #5238 was fixed).

There is a fix but it's ugly. We can also add a 'uncaughtException' handler. When the unhandledRejection throwsthis will trigger uncaughtException. If uncaughtException throws that will kill everything. This approach worked to detect the issue with 3356 above. But the downside is that when this mechanism is tripped that's it, the entire test run is over. Ideally we would have a way to communicate with Mocha what is going on, mark the test as failed and then move on.

But how can we do that without requiring us to mark up each and every test so that some global variable gets set that the handlers can trip and that the test would then pick up?

daleharvey:
I believe a way we could do this is have a global uncaught exception handler, but in tests that are specifically testing uncaughtexception we remove the global handler, use our own, then restore it at the end of the tests, this could likely even be combine with registering our own handler

```
it('tests', function(done) { 
  testUtils.registerUncaughtListener(myFun);
  .... do test
  testUtils.removeUncaughtListener(myFun);
  done();
});
```

Where `registerUncaughtListener` handles the removal of the global listener and `removeUncaughtListener` readds it, would something like that work?

yaronyg:
When would we ever have a test that was intentionally trying to produce an unhandled rejection? It's like having a test showing that we reliably produce a null pointer exception. :) It sorta defeats the point. So my guess is that we can just register the global handlers and call it a day. But we still have to decide if we are o.k. with the tests completely failing in the case that an unhandled rejection is detected. Working around that would require us to hook directly into the test framework.

[...]

PR #5265, opened 2016-06-01
(#5260) Catches unhandled rejections in tests

yaronyg:
#5260 - Adds a new 'test' called test.setup_global_hooks which uses Mocha's
beforeEach and afterEach hooks to set up a listener, both in node and in
browsers like Chrome that support unhandledrejection, to catch any
unhandled Promise rejections and cause the test run to fail.


daleharvey:
This looks great to me, seems to be a bit of a problem with the .10 tests but it may just be flaky, restarts, 👍 when green (if you do merge, the button is fine but keep an eye on the commit message https://github.com/pouchdb/pouchdb/blob/master/CONTRIBUTING.md#guide-to-contributions)

PR #5267, opened 2016-06-02
(4767) - Remove now unhit clause

daleharvey:


daleharvey:
We likely regressed this with https://github.com/pouchdb/pouchdb/commit/da1859e807378cb8488b189e7483ac364bb9d873 by removing the code that hit this clause.  We should be at a point where we dont regress coverage anymore now as well and I have a few patches in progress that will make certain we dont miss it

daleharvey:
Ok so took `report-coverage` out of expected failures (the previous problems are now fixed) and also added code to fail the build if coverage drops, I dont think coveralls is particularly necessary now but left it in just in case

yaronyg:
+1

daleharvey:
woot cheers, super happy that our builds will now straight up fail if coverage drops below 100%

nolanlawson:
LGTM, nice work on the 100% code coverage

PR #5269, opened 2016-06-02
(#4575) - Fix tests against CouchDB master

willholley:
Skip assertions around last_seq. These don't hold in CouchDB 2.0 where the sequence is non-numerical.


daleharvey:
The coverage should be fixed now, so these test changes make sense, they havent made a positive change to the results of the CI run, however should be a step in the right direction so 👍

PR #5273, opened 2016-06-03
(#4372) - Always prune the stemmed revisions

sgenoud:


daleharvey:
So codewise this looks great and makes sense. Given the severity of the problem I am not certain we can live without automated tests however. We have spend a while avoiding writing tests that validate assumptions on the underlying data format because it means writing code for each specific adapter (that we want to test) which is a hassle, however I think we are going to have to bite the bullet here. 

I think we can do this without too much hassle, basically something like

```
var db = new PouchDB(name, {auto_compaction: true, revs_limit: 1});
return db.bulkDocs(docs).then(function() { 
  return testUtils.readDocsBySeq(db);
}).then(function(res) { 
   // Need to account that this wont work on all adapters
  if (res !== null) {  
    res.should.equal(1);
  }
});
```
where readDocsBySeq looks something like
```
testUtils.readDocsBySeq = function(db) {
  return db.info().then(function(info) { 
    if (info.adapter === 'idb') {
      // Do some idb code directory 
    }
    return Promise.resolve(false);
  }); 
}
```

daleharvey:
I can help out with the test here if it is too much hassle for you, I think initially having node + idb adapters tested will be good, the failing CI indicates we are missing some testing here too

packages/pouchdb-adapter-leveldb-core/src/index.js
```diff
@@ -562,11 +562,15 @@ function LevelPouch(opts, callback) {
...
 
     function finish() {
-      if (api.auto_compaction) {
-        return autoCompact(complete);
-      } else {
-        compact(stemmedRevs, complete);
-      }
+      compact(stemmedRevs, function (error) {
+        if (error) {
+          complete(error);
```
daleharvey: This probably needs an istanbul ignore, I dont like adding them but with node callbacks its the only choice really

sgenoud:
What about exposing some generic test endpoints? For instance, in that case, `adapter._testAPI.totalDocCount` that would could all the documents in all the databases it saves to. The test would be to compare that the totalDocCount does not change (avoiding document leaks). But the count itself would not interest us (and would be adapter specific).

daleharvey:
Yup I was thinking about that route too, I think that would potentially be cleaner and easier, since the adapter code then already knows about where its data is etc however I am not totally certain how we would plumb it in so that it is available to test suite but not at all exposed to the distribution. 

@nolanlawson you have any ideas about that?

daleharvey:
All of the above, we are very wary of the size of the PouchDB build so shipping extra bytes that are only used for testing isnt good and we want to reduce anything that is exposed to developers to be only the public tested and documented api. We currently do a seperate build for the tests because they use undocumented things so it should be possible, I think if you wanted to add them to the adapters for now and before merging we can figure out a way to make sure they arent in the distro would be a decent approach

nolanlawson:
@daleharvey Yeah, I agree this needs testing, and I agree with making this a `testUtils` that is not shipped with PouchDB itself. Admittedly that's kind of tricky, but it really ought to be tested for our three supported adapters at least.

As-is, we can write `testUtils` that exist independently of each adapter, i.e. the idb one opens up raw indexedDB, the LevelDB one has access to the filesystem, and the WebSQL one calls `openDatabase()`. It's a bit of extra code, but I think it's simpler and cleaner. Third-party adapters (socket-pouch, worker-pouch) can write their own separate test code or just not test it.

An alternative would be for each adapter to `export` some test utils which only get shipped with `pouchdb-for-coverage` (which we might as well rename `pouchdb-for-testing`), but that won't get tree-shaken for people using Browserify/Webpack with custom builds (maybe not the worst thing in the world). Another alternative is to have `pouchdb-adapter-idb-for-testing` etc. Still, I think this is probably far more complex than just having separate `testUtils` in a separate part of the code.

daleharvey:
Hrm actually plain versions in `testUtils` isnt going to work, because to open the database it needs the data from the adapter (POUCH_VERSION), its also super fragile, like if we ever changed `_pouch_` then its broken, I would prefer to avoid adding stuff for browserify / webpack build and keep them as close to the browser build. 

We may be able to add some hook to add prototypes at runtime so the tests can inject these test methods into the adapter

rhiokim:
Sent from [Nylas N1](https://link.nylas.com/link/3zlnb3kvj5eg8l9mhod5aom9k
/local-156eabb8-69f1/0?redirect=https%3A%2F%2Fnylas.com%2Fn1%3Fref%3Dn1), the
extensible, open source mail client.  


On Jun 3 2016, at 9:19 pm, Dale Harvey &lt;notifications@github.com&gt; wrote:  

> So codewise this looks great and makes sense. Given the severity of the
problem I am not certain we can live without automated tests however. We have
spend a while avoiding writing tests that validate assumptions on the
underlying data format because it means writing code for each specific adapter
(that we want to test) which is a hassle, however I think we are going to have
to bite the bullet here.

>

> I think we can do this without too much hassle, basically something like

>  
>  
>     var db = new PouchDB(name, {auto_compaction: true, revs_limit: 1});

>     return db.bulkDocs(docs).then(function() {

>       return testUtils.readDocsBySeq(db);

>     }).then(function(res) {

>        // Need to account that this wont work on all adapters

>       if (res !== null) {  
>         res.should.equal(1);

>       }

>     });

>  
>

> where readDocsBySeq looks something like

>  
>  
>     testUtils.readDocsBySeq = function(db) {

>       return db.info().then(function(info) {

>         if (info.adapter === 'idb') {

>           // Do some idb code directory

>         }

>         return Promise.resolve(false);

>       });

>     }

>  
>

> —  
You are receiving this because you are subscribed to this thread.  
Reply to this email directly, [view it on
GitHub](https://github.com/pouchdb/pouchdb/pull/5273#issuecomment-223564140),
or [mute the thread](https://github.com/notifications/unsubscribe/AAI5cWUFqWmF
iZG_3hFR62WvX90C05yhks5qIBu6gaJpZM4ItW6p).

daleharvey:
Heh so I tried a hack around injecting in this via the tests, its not looking very likely

```
        PouchDB.prototype.testing = function() {
          console.log('testing', this);
        };

        db.testing();
```

daleharvey:
So still trying to test this, I think its going to have to be a function inside the adapters thats removed by the build somehow, but I have manually tested this, it works good and it makes complete sense, its gairly serious so even if we cant get the test done in time I would rather have this merged and in the next release than have it miss the release because of delays testing

daleharvey:
Missed the release, but given the problems testing it I would rather have this merged than not, 👍

nolanlawson:
Yeah I agree with merging, just really wish we had a way to actually test this. :/

nolanlawson:
Maybe we can add code comments like `/* remove start */` and `/* remove end */` and then remove them at build time with a regex? only problem is we are exposing `src` now with `jsnext:main`, but maybe we shouldn't do that either.

Issue #5272, opened 2016-06-03
Fix for #5145 breaks replication against Cloudant/CouchDB 2.0


willholley:
### Environment (Node.js/browser/hybrid app/etc.)

All

### Browser/platform (Chrome/FF/Safari/Edge/iOS/Android/etc.)

All

### Adapter (IndexedDB/WebSQL/LevelDB/etc.)

HTTP

### Server (CouchDB/Cloudant/Couchbase/PouchDB Server/etc.)

Cloudant, CouchDB 2.0

### Please describe your issue. Test cases (e.g. JSBin) are very helpful!

The fix for #5145 compares last_seq values from _changes. In Cloudant and CouchDB master, sequence numbers are strings, with an integer and a base64 part. The integer can be ignored (this is for backwards compatibility with CouchDB clients that test for an integer and is meaningless). The base64 part decodes as a list of Erlang tuples of the form `{node(), range(), {integer(), uuid()}}`, the combination of which uniquely identifies a position in a shard copy. The order of those tuples can change for each call to _changes, resulting in different sequence values to represent the same underlying state.

Unfortunately, the sequence value cannot currently be unpacked to a useful structure outside of Erlang, so they must be treated as opaque values and cannot be usefully compared (we can't event test for equality).

We can rethink how to implement the fix in #5145 without depending on last_seq but, until then, I think we need to revert the change or consciously break compatibility with Cloudant/CouchDB master.

daleharvey:
I commented on your comment, I shouldnt have missed it especially since I called out couch2 compat in the test, however `last_seq` values are supposed to be able to collate right? it seems like switching `old.last_seq > new.last_seq` should be able to be changed to collage(old.last_seq, new.last_seq)` and work as intended

willholley:
@daleharvey I don't believe that fix is safe either - sequence numbers are supposed to be treated as opaque blobs. I wonder whether that test is even required - just testing a branch with an alternative...

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/176808878a116589ec68bb6476a0c75319e92c54

PR #5275, opened 2016-06-03
(#5272) fix checkpointing against Cloudant/CouchDB 2.0

willholley:
An attempt to provide a better fix for #5145. It seems like
it's safe to eagerly write a checkpoint in the event there are no
changes returned (e.g. because a filter did not generate results).

This at least allows the tests to pass - possibly this means we
need more tests.

### DIFF
--- a/tests/integration/test.replication.js
+++ b/tests/integration/test.replication.js
@@ -1569,12 +1569,11 @@ adapters.forEach(function (adapters) {
         result.docs_written.should.equal(0);
         result.docs_read.should.equal(0);
         changesSince.length.should.equal(2);
-        // the returned last_seq should match the 'since'
-        // requested from remote
-        result.last_seq.should.equal(changesSince[1]);
+
         // the 'since' parameter should be different on the
         // next request
         changesSince[0].should.not.equal(changesSince[1]);
+
         // kick off a second replication
         return db.replicate.from(remote, replicationOpts);
       }).then(function (result) {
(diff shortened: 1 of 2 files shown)


kxepal:
> It seems like it's safe to eagerly write a checkpoint in the event there are no changes returned (e.g. because a filter did not generate results).

That's true.

tests/integration/test.replication.js
```diff
@@ -1568,23 +1568,32 @@ adapters.forEach(function (adapters) {
         result.ok.should.equal(true);
         result.docs_written.should.equal(0);
         result.docs_read.should.equal(0);
-        changesSince.length.should.equal(2);
-        // the returned last_seq should match the 'since'
-        // requested from remote
-        result.last_seq.should.equal(changesSince[1]);
+        changesSince.length.should.equal(2, JSON.stringify(changesSince));
+
+        if (!testUtils.isCouchMaster()) {
```
daleharvey: I think we should take out these assertions, if they dont hold for couch master then we dont strictly need them, I am good with making as many assertions as we can but lets keep them to the shared ones that are true regardless of the server

daleharvey:
So after some clarifications on irc, some of my assumptions about couch seq's were wrong, however this looks like the only case where we make that assumption so 👍 , got a little comment but otherwise code looks good to me, thanks :)

tests/integration/test.replication.js
willholley: sounds reasonable - I'll take them out.

daleharvey:
Awesome thanks 👍

Issue #3539, opened 2015-02-14
Exception inside event handler can break main thread


nolanlawson:
From [IRC today](https://botbot.me/freenode/pouchdb/2015-02-14/).

```
Uncaught InvalidStateError: Failed to execute 'transaction' on 'IDBDatabase': 
    The database connection is closing.
```

```
MartinBroerse
hi Nolan, when I delete a record in Ember Pouch using my local CouchDB and never with the iriscouch there is a FindAll promise never resolving for this code:
11:02 am
deleteAuthor: function() {
11:02 am
this.modelFor('author').destroyRecord().then(function() {
11:02 am
this.transitionTo('authors');
11:02 am
}.bind(this));
11:02 am
},
11:03 am
How can I report something usefull to find this bug?
11:04 am
I can't reproduce it because it only seem to fail on fast systems.
11:05 am
Also the bug is not always there??
11:07 am
The error I get:
11:07 am
Uncaught Error: Assertion Failed: You can only unload a record which is not inFlight. `<myapp@model:post::ember876:B4626ECD-63F0-72EC-A223-FCB6CE81FA47>`
11:08 am
Uncaught InvalidStateError: Failed to execute 'transaction' on 'IDBDatabase': The database connection is closing.
11:08 am
On the Pouch line: var transaction = idb.transaction(stores, 'readonly');
11:09 am
function allDocsQuery
11:12 am
nolanlawson
"The database connection is closing" is one of those IndexedDB errors that seems to be getting more prevalent
11:13 am
I have seen it a million times and have no idea how to debug it. :/
11:13 am
Does it only occur in Chrome?
11:13 am
MartinBroerse
Yes
11:13 am
nolanlawson
yup
11:13 am
Please make a reproducible test case and file the issue directly on pouchdb (not ember-pouch or relational-pouch)
11:13 am
I need to make a collection of these so I can show them to the Chromium devs and get their opinion
11:14 am
what makes me suspicious is that Firefox never, ever throws these errors
11:14 am
MartinBroerse
It seems th delete resolves but when you do allDocsQuet to fast is fails
11:15 am
function allDocsQuery
11:17 am
as I said I can't make a reproducible test. I you run a delete in https://github.com/broerse/ember-cli-blog 10 times running "ember s" against your local couchdb it will fail.
11:25 am
nolanlawson
MartinBroerse: you mean you destroy your database? that's a common cause of this error that I've seen.
11:26 am
[...]

PR #5277, opened 2016-06-03
(#3539) - Ensure throw in user code does not effect pouch

daleharvey:


daleharvey:
So I tried using `process.nextTick` here, however it significantly changes how our events are exposed to code, I think its probably the ideal way to do this however its going to need a fairly large rewrite to get everything working nicely, the try works fine and maintains existing semantics

packages/pouchdb-core/src/changes.js
```diff
@@ -58,7 +59,11 @@ function Changes(db, opts, callback) {
     if (opts.isCancelled) {
       return;
     }
-    self.emit('change', change);
+    try {
+      self.emit('change', change);
+    } catch (e) {
+      guardedConsole('error', 'Error in .on("change", function):', e);
+    }
```
nolanlawson: try/catches should be isolated to avoid V8 deopts, but I can fix this on the way in

tests/integration/test.changes.js
```diff
@@ -2542,6 +2542,35 @@ adapters.forEach(function (adapter) {
...
+
+      var db = new PouchDB(dbs.name);
+      var count = 0;
+
+      var changes = db.changes({live: true});
+
+      changes.on('change', function () {
+        ++count;
+        if (count === 1) {
+          throw 'an error';
```
nolanlawson: prefer Errors to strings here, but can also change on the way in

nolanlawson:
looks great to me, glad we finally got around to fixing this

nolanlawson:
5df3d3659996d59af21c9c6f62619f7210eb44e1

Issue #3629, opened 2015-03-11
Add a more human-readable key than 404/409


nolanlawson:
Instead of

```js
if (err.status === 404) {
```

I'd like:

```js
if (err.key === 'not_found') {
```

The problem with the status codes is 1) it's confusing, because people feel like there's an HTTP request going on, when there isn't, and 2) it's hard to remember the difference between 404 and 409.

I know we got rid of `err.name` because `name` is a JavaScript prototypical property on Errors, but that doesn't mean we can't add `err.key`.

nickcolley:
How would we document this without confusing people, would prefer to keep it simple personally.

nolanlawson:
we would remove all references to 404s/409s from the documentation, and replace it with `err.key`.

The old system would still work (no reason to remove `status; 404`), but it just wouldn't be recommended because it doesn't read as well.

mainerror:
There are cases where PouchDB might throw an error with status 412 for example. Do you want those cases to be represented by the key too?

nolanlawson:
Yep, although in our case 412 is likely something like `'attachment unknown'` (i.e. you provided an attachment stub with an md5sum, but we don't know of any attachment by that md5sum).

kxepal:
Just an idea: would be cool to have these error names as properties like `err.key == Errors.not_found` to let editor help with suggestions and prevent typos like `err.key == 'not_fond'` which may eventually cause happy debug times. Make a typo with a number was quite hard, but it will be easy with long strings. Also such enum will help to structure all errors to not have duplicate synonyms of the same errors.

nolanlawson:
Enums aren't really a big convention in the JavaScript world, but I kinda like 'em. I'd be in favor of something simple like `PouchDB.ERROR_CONFLICT`. Potentially we could just make it an int (409) instead of a string anyway. :stuck_out_tongue:

kxepal:
Good point. This could preserve BC.

nolanlawson:
42502bcefe3011a83871b0eb9e4b62f07545f184

PR #5279, opened 2016-06-05
(#3629) - Use human readable error names

daleharvey:


daleharvey:
We never got rid of error names, I think we just had some fixes around the error construction, these look to be working fine

nolanlawson:
Yeah I agree; I'm not sure why we ever decided that the `status` was the most reliable way to check errors. Error names should be consistently thrown; if not, that's a bug.

nolanlawson:
42502bcefe3011a83871b0eb9e4b62f07545f184

PR #5284, opened 2016-06-06
(#5284) - make ajax test work offline

nolanlawson:
Our tests fail if you're in airplane mode, which is annoying because I'm currently on an airplane. 😅


nickcolley:
LGTM 👍

daleharvey:
Yup looks good to me

PR #5289, opened 2016-06-06
(#4575) use unique database names in tests

willholley:
CouchDB 2.x appears to get into a funk when databases with the same name are rapidly created/deleted.

To work around this, use unique database names in each test. This code be cleaned up / de-duplicated - as a first cut this is just to get the tests passing.

### DIFF
--- a/tests/fuzzy/test.fuzzy.js
+++ b/tests/fuzzy/test.fuzzy.js
@@ -112,7 +112,7 @@ describe('chaos-monkey', function () {
   beforeEach(function (done) {
     Promise = testUtils.Promise;
     var aname = testUtils.adapterUrl('local', 'testdb');
-    var bname = testUtils.adapterUrl('http', 'test_repl_remote');
+    var bname = testUtils.adapterUrl('http', 'test_repl_remote' + (new Date()).getTime());
     testUtils.cleanup([aname, bname], function () {
       a = new PouchDB(aname);
       b = new PouchDB(bname);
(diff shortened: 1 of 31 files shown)


daleharvey:
We used to do unique test database names however it causes some problems, its harder to debug test failures since right now you can just delete `after()` ad inspect the database, `webSQL` is annoying since we cant actually delete databases there just empty them (leaving us with thousands of empty databases per test). I would really prefer us to not revert this due to an upstream bug in couchdb. 

Could you clarify what the bugs you are seeing in the bug and we can look into them, it sounded like it was platform specific? (ie passed against built couch but failed in docker)

willholley:
So the errors I see are along the lines of:

```
2016-06-06 10:46:26.810 [error] node1@127.0.0.1 <0.29647.2> Could not open file /usr/src/couchdb/dev/lib/node1/data/shards/00000000-1fffffff/testdb.1465209986.couch: no such file or directory
```

(more complete log at https://gist.github.com/willholley/d200157ccdd3bc30485dbf1d9b0b9eae)

It doesn't happen at a consistent point in the tests.

When I run against CouchDB master locally (rather than using the Docker image), I occasionally see this error as well, though the test suite passes ~80% of the time.

daleharvey:
Ok cool, I hadnt seen this error before, so its possible its a new regression to couchdb, I think if we can bisect it its likely going to be the easiest way, but yeh not comfortable with this change on the test suite so copied your comment across to the appropriate bug, thanks

Issue #5293, opened 2016-06-07
SQLite Plugin 1 (aka Cordova SQLite Storage) broken in 5.4.0


nolanlawson:
This was a bug that crept in to 5.4.0 because we don't test SQLite Plugin in CI:

> WebSQL threw an error – Error: Sorry first openDatabase argument must be an object — SQLitePlugin.js:25

nolanlawson:
#5294 is the fix, or users can also switch to https://github.com/nolanlawson/cordova-plugin-sqlite-2 which is the one we officially support.

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/b600bdd249946aa9eeabefbe868d9987c9240723

PR #5294, opened 2016-06-07
(#5293) - fix SQLite Plugin usage, add test

nolanlawson:
We've regressed on this so many times, I've also added a test that mocks `window.sqlitePlugin` and ensures we are passing in the correct args.

### DIFF
--- a/tests/integration/browser.sqlite.js
+++ b/tests/integration/browser.sqlite.js
@@ -1,0 +1,58 @@
+'use strict';
+
+// mock tests for the Cordova SQLite Plugin
+
+var adapters = ['local'];
+
+adapters.forEach(function (adapter) {
+  describe('browser.sqlite.js-' + adapter, function () {
+
+    var dbs = {};
+
+    var called = false;
+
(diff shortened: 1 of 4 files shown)


daleharvey:
Looks like this breaks the sql related tests

nolanlawson:
Yup, my bad, forgot about the cache-by-name. Should be fixed now.

daleharvey:
👍 looks good, cheers

tests/integration/browser.sqlite.js
```diff
@@ -0,0 +1,58 @@
+'use strict';
+
+// mock tests for the Cordova SQLite Plugin
+
+var adapters = ['local'];
```
daleharvey: If its only designed for one adapter, it doesnt need to do the adapters.forEach dance

daleharvey:
I made that minor change on the way in, thanks - https://github.com/pouchdb/pouchdb/commit/b600bdd249946aa9eeabefbe868d9987c9240723

PR #5302, opened 2016-06-07
(#5302) - Update gitignore to remove dist files in packages/pouchdb

nickcolley:
After npm installing fresh I get dirty dist files


nolanlawson:
Good catch! Thanks

PR #5303, opened 2016-06-07
(#5303) - Add json as a gem dep to prevent build failures

nickcolley:
I was getting

```
bundler: failed to load command: jekyll (/home/nick/bin/jekyll)
LoadError: cannot load such file -- json
```

Adding json as a dependency then running bundle install let me execute `npm run build-site` with no issues.


nolanlawson:
Cool! I'm fine with merging this without a build since we have so many other rebuilds going on right now.

PR #5304, opened 2016-06-07
(#5304) - Service workers instead of Appcache for the website

nickcolley:
Continuation of https://github.com/pouchdb/pouchdb/pull/4846

### DIFF
--- a/bin/build-site.js
+++ b/bin/build-site.js
@@ -76,7 +76,7 @@ if (!process.env.BUILD) {
   watchGlob('**', buildJekyll);
   watchGlob('docs/static/less/*/*.less', buildCSS);
   http_server.createServer({root: '_site', cache: '-1'}).listen(4000);
-  console.log('Server address: http://0.0.0.0:4000');
+  console.log('Server address: http://localhost:4000');
 }
 
 buildEverything();
--- a/docs/_layouts/default.html
+++ b/docs/_layouts/default.html
@@ -1,6 +1,5 @@
 <!DOCTYPE html>
 <html lang="en" manifest="/manifest.appcache">
-
   <head>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1">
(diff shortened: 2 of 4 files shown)


nickcolley:
Okay so this is working pretty well as is but I'm trying to think of a good way to let the user update the page like we have with appcache. I'm not sure this strategy is the right one with this in mind since a user needs to exit the page entirely for the newly generated worker to take over..

nickcolley:
Okay this is ready to review, one con to this approach is that for people to see updates to pages they already have cached they need to leave the site and come back.

If we think that's important enough I can look at trying to handle that case better so let me know your thoughts. :+1:

nolanlawson:
Hm, I would really rather not lose that "content updated, reload?" message if at all possible. I know there are ways to detect that the content changed; e.g. check out SVGOMG's method: https://github.com/jakearchibald/svgomg/blob/0f2ce1ace903d569337b72f57b8eb255b65177df/src/js/page/main-controller.js#L123-L127

nickcolley:
@nolanlawson unless I'm mistaken the user needs to close the tab and then return for the service worker to become obsolete and be overridden?

nolanlawson:
@nickcolley, no, you can update a service worker on-the-fly using `clients.claim()` and `skipWaiting()`. I admit my memory is fuzzy on which one exactly does what, but using one or both of those two, you definitely can.

nickcolley:
Okay updated the PR with your feedback

daleharvey:
Woot this is looking great, so having both appcache and serviceworkers on the page doesnt seem to conflict, https://github.com/slightlyoff/ServiceWorker/issues/697 suggests so too. Updates are working and seeing it working in both firefox and chrome, 👍 :)

daleharvey:
Given the way appcache went I expect we will probably still see a few minor issues, but I think those will get found and fixed once its released, I tried checking as much as I can and all worked great for me so gonna push this then see how it goes :)

PR #5305, opened 2016-06-07
(#5305) - fix pouchdb-adapter-node-websql

nolanlawson:
Unfortunately `pouchdb-adapter-node-websql` is broken due to a typo, and it wasn't caught in CI because it turns out we _teeeeechnically_ only test `pouchdb/extras/websql`, which is slightly different from `pouchdb-adapter-node-websql`. Since I think it's silly to run an extra test for a slightly different usage, I've just fixed it manually and gonna call that good.

### DIFF
--- a/packages/pouchdb-adapter-node-websql/src/index.js
+++ b/packages/pouchdb-adapter-node-websql/src/index.js
@@ -1,4 +1,4 @@
-import WebSqlPouch from 'pouchdb-adapter-websql';
+import WebSqlPouchCore from 'pouchdb-adapter-websql-core';
 import { extend } from 'js-extend';
 import websql from 'websql';
 
@@ -7,7 +7,7 @@ function NodeWebSqlPouch(opts, callback) {
     websql: websql // pass node-websql in as our "openDatabase" function
   }, opts);
 
-  WebSqlPouch.call(this, _opts, callback);
+  WebSqlPouchCore.call(this, _opts, callback);
 }
 
 // overrides for normal WebSQL behavior in the browser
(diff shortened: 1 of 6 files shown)


packages/pouchdb-adapter-node-websql/src/index.js
```diff
@@ -7,7 +7,7 @@ function NodeWebSqlPouch(opts, callback) {
     websql: websql // pass node-websql in as our "openDatabase" function
   }, opts);
 
-  WebSqlPouch.call(this, _opts, callback);
+  WebSqlPouchCore.call(this, _opts, callback);
```
nolanlawson: When this wasn't calling `core`, it was just uselessly calling what `pouchdb-adapter-websql` exports, which is the `PouchDB.plugin()` function.

nolanlawson:
Actually I was wrong; the test was indeed using `pouchdb-adapter-node-websql`, it's just that it was falling back to regular LevelDB. I've added a test to ensure it's actually testing node-websql, and I've fixed the (unrelated) error that was causing us to fall back to LevelDB instead of `node-websql`.

The test fails before the fix but succeeds after.

tests/integration/utils.js
```diff
@@ -354,8 +354,10 @@ if (typeof process !== 'undefined' && !process.browser) {
     // (the two strings are just to fool Browserify because sqlite3 fails
     // in Node 0.11-0.12)
     require('../../packages/' + 'pouchdb/extras/websql');
-    global.PouchDB = global.PouchDB.defaults({prefix: './tmp/_pouch_'});
-    global.PouchDB.preferredAdapters = ['websql'];
+    global.PouchDB = global.PouchDB.defaults({
+      prefix: './tmp/_pouch_'
+    });
+    delete global.PouchDB.adapters.leveldb;
```
nolanlawson: ^ this was not actually working before. Instead, it was just running `leveldb` like normal.

Issue #5309, opened 2016-06-08
LevelDB adapter needs error handling on allDocs callback


stockulus:
### Environment (Node.js/browser/hybrid app/etc.)
ReactNative
### Browser/platform (Chrome/FF/Safari/Edge/iOS/Android/etc.)
iOS Simulator, During development / (Reload 
### Adapter (IndexedDB/WebSQL/LevelDB/etc.)
LevelDBAdapter AsyncDown
### Please describe your issue. Test cases (e.g. JSBin) are very helpful!
If you app offten writes to PouchDB in ReactNative during development, it can happen, that a record is not correctly stored in PouchDB (Key stored, Value not) After Reload it causes a Problem in allDocs function of the pouchdb-adapter-leveldb-core, because there is no check for Error on read. 

I'd suggest to add an Error check there, it should not harm anyone. Pull Request will follow.

Christoph

daleharvey:
So this needs to be fixed by https://github.com/pouchdb/pouchdb/issues/3224

Issue #3224, opened 2014-12-19
PouchDB on node.js using leveldb needs to be Crash Only


yaronyg:
One of the key design features of CouchDB is it's 'Crash Only' design. There is no shutdown. If you want to turn it off, you just kill it. The system is always in a safe state so data will never be lost.

It turns out that we use a soft lock in leveldb that will fail if the system is turned off unexpectedly and leave the system in an inconsistent state. (See [here](https://github.com/pouchdb/pouchdb/issues/3216#issuecomment-67639840))

This is bad in general, systems should be robust. But this is particularly deadly for mobile scenarios ranging from phones (Node.js already runs on Android) to tablets to laptops to IoT embedded systems that can be yanked out of the wall at any time. We can't rely on PouchDB in those scenarios if it's going to end up in a horked state if the process is killed unexpectedly.

So we need to re-work the leveldb code so it is crash only.

nolanlawson:
So we currently do a locking transaction queue and every bulkDocs is a single LevelDB batch operation, but you may be right that the update_seq and doc_count and attachment data are not in that same atomic operation, and they ought to be.

The code is complex because it contains a heck of a lot of perf optimizations. It could certainly be cleaned up though.

masaeedu:
I see this is marked as wontfix. So, just to be clear, for situations where the process is prone to brutal death at any time, is it only the leveldb driver one should avoid or is this an unavoidable tradeoff of pouchdb's design?

daleharvey:
It is only marked as WONTFIX because we only keep active bugs open around here, not that it will never be fixed. Its only an issue with the current leveldb implementation, both webSQL and IndexedDB have transactions to maintain data integrity. At some point as part of https://github.com/pouchdb/pouchdb/issues/5207 we will probably have to rewrite the level adapter and as part of that it will become crash safe

masaeedu:
@daleharvey Thanks. Is there some Node JS driver that isn't susceptible to the issue in the meantime? My use-case is a command line tool, so being able to keep things in a consistent state when the process (or entire system) is killed unceremoniously is critical.

daleharvey:
An possibility may be using sqlite with the websql driver, that uses transactions so should be crash resistent, otherwise best case is probably to write the leveldb-next, it shouldnt actually take that much effort

rasgo-cc:
@daleharvey any progress on this? Any pointers to things I could look at so maybe I could help?
[...]

PR #5310, opened 2016-06-08
(#5309) Error check in _allDocs / allDocsInner

stockulus:
more details and why on https://github.com/pouchdb/pouchdb/issues/5309

### DIFF
--- a/packages/pouchdb-adapter-leveldb-core/src/index.js
+++ b/packages/pouchdb-adapter-leveldb-core/src/index.js
@@ -927,7 +927,9 @@ function LevelPouch(opts, callback) {
         if (opts.include_docs) {
           var seq = metadata.rev_map[winningRev];
           stores.bySeqStore.get(formatSeq(seq), function (err, data) {
-            allDocsInner(data);
+            if (!err) {
+              allDocsInner(data);
+            }
           });
         }
         else {


daleharvey:
So the correct fix for this is https://github.com/pouchdb/pouchdb/issues/3224, having us fallback from invalid data when we read it is a nasty position to get into, This is also going to be extremely difficult to test correctly. 

I think we can do it this one time, but any further work should be on ensuring the data that is written is not corrupt, got a question about the actual implementation though

```js
+            if (!err) {
+              allDocsInner(data);
+            }
```

If there is an error, it seem from the code this would let that call hang and do nothing, is that what happens? should this not have an 

```js
} else { 
  callback(err);
}
```
as well?

stockulus:
because it is not pushed to the result (results.push(doc) in function allDocsInner) it will be ignored in the read completely. I do not see a other usefull behavior and in this adapter is no debug logging. 

without this check you get a error, because allDocsInner is called with null and options include_docs.

daleharvey:
ah ok it continues silently. Yeh I am really beginning to feel we should not paper over this, it means `db.allDocs` and `db.allDocs({include_docs: true})` are going to return different results right? 

At the very very least it needs a large warning, however even at that point the users data is corrupt, subsequent reads are going to fail and I think anything other than a fatal error is misleading the user

stockulus:
yes, db.allDocs and db.allDocs({include_docs: true}) are going to return different result.

-> So you'd prefer a callback(new Error('Database Corrupt') as result?

daleharvey:
I think so, at this point the data is corrupt and we can make no guarantees about what data is coming out of pouchdb, I dont think there is a reasonable way for pouch or the user for that matter to fallback aside from restore from backup and we should make that clear to them

stockulus:
got you and understand it; thougth it does not help me;

i can change it, that it callbacks with reasonable Error.

daleharvey:
Yeh its not a great situation for our data to become unrecoverably corrupt, however I think we do need to be honest about it. 

In good news, the structure I used for https://github.com/pouchdb/pouchdb/tree/master/packages/pouchdb-adapter-indexeddb should be directly portable to leveldb and should almost by default have us in a situation where we will not be able to write corrupt data, I will investigate the port shortly

stockulus:
Sounds great. Looks much cleaner too. I think I will live with a copy in the meanwhile.
Properly I take the indexeddb adapter as blueprint, to build a pouchdb-adapter for Async Storage without leveldb-bridge

daleharvey:
Yeh I think that is a really good plan, the adapter itself is not a huge amount of code and based on asyncStorage should likely be even less code than idb / leveldb, if you need any help then please feel free to ask about it

daleharvey:
@stockulus just checking, are you planning on following up on this or focusing on replacing this with the asyncStorage adapter? Can dupe it to https://github.com/pouchdb/pouchdb/issues/3224 if you are focusing on asyncstorage

stockulus:
I planing to focus on asyncstorage, so we can close this Pull Request. The pouchdb_react_native currently depends on a Package Copy, so it's ok.

Think it does not make sense to put a lot of afford here, if it gets rewritten anyway.

If possible, I'd like you to review the Code once the asyncstorage adapter is ready.

daleharvey:
Yup I would be happy to, its up to you but we may want to think about having it in the pouch repo once it is completed

stockulus:
I'd like it, if it gets part of the pouchdb repo.

nolanlawson:
@stockulus looking at this "fix", I get the impression that there must be a bug somewhere in `asyncstorage-down`. I mean, we have used this code in 5 different *down adapters (leveldown, memdown, localstorage-down, fruitdown, level.js), and never seen an issue. So that tells me that there's a 99% chance the bug is in `asyncstorage-down`.

Still I agree with @daleharvey that the error ought to be handled, and yeah, we should fix it in #3224.

stockulus:
I am still in the adapter learning. so i am not that sure. The Situation i have, is an write millisec before the application get killed; Not the easiest Sitiuation and not "real Word". 

Anyway I started writing an adapter for pouchdb without using the leveldown adapter.
https://github.com/stockulus/pouchdb-react-native/tree/master/packages/pouchdb-adapter-asyncstorage-pure/src
(is just the beginning, currently thinking how to get transactions without having transactions, testing with AsyncStoreage if MutiSet is good enougth)

Issue #5315, opened 2016-06-08
Leveldown as optionalDependencies in 5.4.0


KRuelY:
### Environment (Node.js/browser/hybrid app/etc.)

(Node.js.)

### Server (CouchDB/Cloudant/Couchbase/PouchDB Server/etc.)

(PouchDB.)

### Please describe your issue. Test cases (e.g. JSBin) are very helpful!

Hi, I've been using pouchdb 5.3.2 for my development. I am using it for starting up local server and testing. Currently I have a .npmrc that set optional=false, which will ignore all of the optionalDependencies(including pouchdb's optionalDependencies leveldown). I do not have leveldown as dependencies in my project.

After the 5.4.0 update, all of my tests are failing due to 
```
Error: the 'leveldown' package is not available. install it, or, specify another storage backend using the 'db' option
```
My thought is either leveldown should now be a dependency(?) for pouchdb(or pouchdb's dependencies?) instead of optionalDependency, or there is a specific new feature that I am using in pouchdb that require leveldown(but I don't have any code changes on my side). I'm having a hard time finding what changes from 5.3.2 to 5.4.0 that causes this issue. Do you guys have some hints that can help me find out what my problem is?

Thank you, and please correct me if I'm wrong because I'm still not familiar.

daleharvey:
If you are running the nodejs test suite without giving it an alternative level adapter, you need leveldown installed (ie to not ignore it as an optional dependency), it is there as an optional dependency because other people do not have to (or cannot) install leveldown and arent testing against it

daleharvey:
We also have an issue with optional dependencies not always installing correctly, you may need to manually `npm install leveldown`, that work is being tracked in https://github.com/pouchdb/pouchdb/issues/5257

Feel free to ask follow on questions, but closing as not a bug

KRuelY:
Hi,

I know the changes that break my components. It is this line

https://github.com/pouchdb/pouchdb/blob/5.4.1/lib/index.js#L5064

In the 5.3.2 version, the requireLevelDown function is optional:

https://github.com/pouchdb/pouchdb/blob/5.3.2/src/adapters/leveldb/index.js#L175

nolanlawson:
Whoops, looks like an error that accidentally snuck in during the monorepo changes.

nolanlawson:
[...]

PR #5335, opened 2016-06-11
(#5315) - fix leveldown as an optional dep

nolanlawson:
This should fix it.


daleharvey:
Sweet looks good to me, cheers

PR #5336, opened 2016-06-11
(#5336) - Website typography and refactor of CSS

nickcolley:
Reopening of the previous closed pull request, this time with the font changes reverted.

I still believe we should remove the fonts, but will make the case in another pull request so we can unblock most of the work.

See https://github.com/pouchdb/pouchdb/pull/4726 for history of discussion on the previous pull request.


nolanlawson:
Makes sense to me. I always loved the changes in the other PR (spacing, size, etc.) although I understand why the font would be controversial. I'm +1 for merging.

zaccolley:
Yeah, this is a good idea, they were two separate PRs imo anyway.

(It would be interesting to do some user testing surrounding the web fonts argument. I'd like to see if they offer a measurable improvement to the user experience balanced against the performance sacrifice.)

daleharvey:
👍 looks great

PR #5337, opened 2016-06-11
(#5161) - Force navigation to be static on smaller viewports

nickcolley:
I thought we fixed this but I think we closed the issue by mistake


nolanlawson:
Cool, thanks!

Issue #5214, opened 2016-05-23
Stop squelching error messages in CustomPouchErrors


nolanlawson:
#5213 reveals that our custom errors frequently squelch the underlying error message, leading to issues like #4803, #5104, and #4670 (which I've combined into this new issue).

Once #5213 is merged and we have solid unit tests in place, we can go back and make sure that the error message is more useful. I'd also like `generateErrorFromResponse()` to handle cases where the incoming `res` object is not JSON (running the tests, I saw both strings and `Buffers` (!)). If an HTML page is returned with an error message, then we should probably just make that whole string into the resopnse. Ditto for `Buffer`s.

Here's an example of the kind of squelching behavior that I think can be improved:

```js
var response = { "error": "bad_request", "reason": "Referer header required." };
var err = generateErrorFromResponse(response);
err.status.should.equal(400);
err.name.should.equal("bad_request");
err.message.should.equal("Something wrong with the request");
```

Or this one: 

```js
var response = "<!doctype html><html>YO THIS IS AN ERROR</html>";
var err = generateErrorFromResponse(response);
err.status.should.equal(500);
err.name.should.equal("unknown_error");
err.message.should.equal("Database encountered an unknown error");
```

nolanlawson:
For the test case, see `test.errors.js` in https://github.com/pouchdb/pouchdb/pull/5213.

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/d31bf170ccfa54ca58e6afd43fe9f496e639e1a9

PR #5339, opened 2016-06-11
(#5214) - Stop squelching errors

daleharvey:


daleharvey:
So at some point we tried to force consistent errors between couchdb and pouchdb, when we only had one version of one server to deal with, thats not really been sustainable and put us in the position where we keep overwriting errors we get back with our own assumptions. 

I have changed this method of trying to interpret errors to return errors as close as possible to how we receive them, it means less code and far clearer errors, kept a little bit of consistency on the errors, ensuring they have a `.status` + `.name`, but mostly we give them back as we get them, had to change a few tests but as you can see most of the changes to the tests are to give clearer messages.

Also `error.true` was something left over from the days before even nodes callback format, and way before promises we did 
```
db.allDocs(function(res) { 
  if (res.error) ...
});
```
Which we definitely dont need anymore, so removed it

nolanlawson:
This looks great to me; agree with every point you made. Hopefully this will reduce the number of "500 CustomPouchError - unknown error" confusing people.

packages/pouchdb-ajax/src/request-browser.js
```diff
@@ -126,8 +126,7 @@ function xhRequest(options, callback) {
   try {
     xhr.open(options.method, options.url);
   } catch (exception) {
-   /* error code hardcoded to throw INVALID_URL */
-    callback(exception, {statusCode: 413});
+    return callback(new Error(exception.name || 'Url is invalid'));
```
daleharvey: So this was a fairly big thing to miss and hopefully it may weed out some fairly confusing behaviour, its likely the source of the need for extra tests + istanbul comments needed to not regress coverage

daleharvey:
So wasnt sure if the 👍 's were on the comments or the patch, but added a test here and pointed out why we needed the test to not regress coverage

packages/pouchdb-ajax/src/request-browser.js
nickcolley: Should it be URL? (super nit sorry :speak_no_evil:)

daleharvey: Yeh it probably should, ill fix that on the way in, also should clarify that I have to create a `new Error` because the exception given is readonly, which breaks a bunch of our code

nolanlawson:
:+1: on the whole PR, looks great

PR #5344, opened 2016-06-13
(#5315) - fix non-leveldown migration for `opts.db`

nolanlawson:
I tested manually and confirmed that this works. Unfortunately nowhere in our tests do we confirm that manually setting `opts.db` avoids running the migration tests. OTOH our futureproof solution (the custom builds) will indeed fail the tests if anything is broken there (i.e. if it tries to run in the browser, it'll definitely fail).

This will probably require a 5.4.3 release.


daleharvey:
Cool cheers

PR #5349, opened 2016-06-13
remove dead link to extras source code

glynnbird:
Replaced with links to the Ajax, Checkpointer & Promise source code, respectively.


daleharvey:
Nice catch, 👍

Issue #5345, opened 2016-06-13
Missing Test for PR #5199 - fix excessive long replication loop


dharders:
PR #5199 was merged and released in 5.4.0 without a test. Discussion in #5196 seemed to indicate that a test would be difficult to write.

I've created a test, [test.replication_perf_regression.js](https://github.com/dharders/pouchdb-replication-perf-regres-test/blob/master/test.replication_perf_regression.js), that seems to catch the performance regression.

I have run the test on 5.2.x through to 5.4.x and it behaves as expected (i.e. Passes 5.2.0, Fails 5.2.1 through to 5.3.2, Passes 5.4+) as the regression was introduced in 5.2.1 and fixed in 5.4.0.

The test works by counting the number of GET requests to the remote `_local/` endpoint, used for replication/checkpointing. A failure occurs when the number of GET requests to the remote `_local/` endpoint are greater than the number of docs replicated. The bad performing code would only flush 1 change per batch and not enable the `pendingBatch.changes` buffer to fill up before the `startNextBatch()` cycle commenced.

I would create a PR, but as a first-time contributor I'm having issues building / running npm test on my dev environment (Windows 7 x64) after a simple git clone, npm install (I will create another issue for this).

It should be easy to modify [test.replication_perf_regression.js](https://github.com/dharders/pouchdb-replication-perf-regres-test/blob/master/test.replication_perf_regression.js) to fit PouchDB's way of testing as I based the test on [another PouchDB test, test.read_only_replication.js] (https://github.com/pouchdb/pouchdb/blob/master/tests/component/test.read_only_replication.js)

As requested @nolanlawson

daleharvey:
This looks great, 2 minor nits is that we use 2 space tabs, and `_local_GET_count ` would be `localGETCount` but the rest looks great. I know you have been having issues getting your local setup completely passing but if you are confident this test will pass then please open a PR and we can see what travis says

nolanlawson:
Awesome, awesome work. I looked over the test, and it looks great except that it may be a bit difficult to run in our test suite as-written. In particular our test suite doesn't normally set up a `pouchdb-express-router` and then count the number of `GET` requests that way; it just talks to the local CouchDB or other server. So you would need some other way to count the requests. (Sometimes we overload `db.get` to do this, but it's a bit wonky TBH.)

[...]

PR #5353, opened 2016-06-15
(#5345) - Added the missing test for PR #5199

dharders:
Added a missing test to catch future `replicateTo()` performance regression (fixed in #5199, discussed in #5345)

**Special Notes**

- Node test only (see [comment](https://github.com/pouchdb/pouchdb/issues/5345#issuecomment-226068960))
- Uses `pouchdb-express-router` for remote endpoint snooping
- Self-contained test. Destroys `express()` server on completion.

Also, we could probably be more aggressive with the performance regression check [on L104](https://github.com/pouchdb/pouchdb/commit/43f083e21ba6586a260cf2157ad614f43f81f99d#diff-515e8d7772217697f1de2703c26d85c1R104) but for now will take the conservative route.

Look forward to your feedback.

### DIFF
--- a/tests/component/test.replication_perf_regression.js
+++ b/tests/component/test.replication_perf_regression.js
@@ -102,6 +102,9 @@ describe('test.replication_perf_regression.js', function () {
       info.doc_count.should.equal(numDocs); // should have replicated to local db
       localGETCount.should.above(0);        // should have hit /remote/_local endpoint at least once
       localGETCount.should.below(numDocs);  // localGETCount should be significantly below numDocs (~9x)
+      return db.destroy();
+    }).then(function () {
+      remote.destroy();
     });
   });
 });
--- a/tests/component/test.replication_perf_regression.js
+++ b/tests/component/test.replication_perf_regression.js
@@ -104,7 +104,7 @@ describe('test.replication_perf_regression.js', function () {
       localGETCount.should.below(numDocs);  // localGETCount should be significantly below numDocs (~9x)
       return db.destroy();
     }).then(function () {
-      remote.destroy();
+      return remote.destroy();
     });
   });
 });
--- a/tests/integration/test.replication_perf_regression.js
+++ b/tests/integration/test.replication_perf_regression.js
@@ -1,6 +1,6 @@
 'use strict';
 
-var PouchDB = require('pouchdb');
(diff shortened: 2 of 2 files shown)


dharders:
Question: DB Cleanup ?

Do you need me to add something like:

```
  ...
  localGETCount.should.below(numDocs);
  return db.destroy();
}).then(function() {
  return remote.destroy();
});
```

after the test is complete ? Where ? After the last `should()` ?

daleharvey:
> after the test is complete ? Where ? After the last should() ?

Yup exactly

Also sorry for the confusion, I think Nolan missed the tests that this were based on but yup this should go in `tests/component`, cheers for the PR :)

dharders:
Ahhh, that might explain the failures in Travis I'm seeing currently with require('pouchdb')

I've noticed some tests use require('pouchdb') to define PouchDB and some just assume it's already defined. 

Do I need to require pouchdb for this type of test to pass Travis? (npm test is passing locally with and without)

Anyway to cancel pending Travis build? @daleharvey

daleharvey:
So inside `tests/integration` the tests generally expect a server setup for them and a pouchdb included, providing your own will work sometimes, however its fairly confusing and is likely to break somewhere, the component tests however mostly run as is, you should be fine in there

dharders:
Are files in tests/component covered by `npm test` ?

I just tried a local `COUCH_HOST=http://192.168.1.180:5984 GREP=test.replication_per* npm test` and it completed without error, but did not list the test and showed `0 passing`.

This same command had `1 passing` when the file was located at `tests/integration`

@daleharvey

daleharvey:
Oh and I cancelled the travis build, cheers

daleharvey:
You need to run `npm test-component` for those tests

dharders:
Just noticed my last commit didn't **return** the final `remote.destroy()`. 

i.e should've been `return remote.destroy();`. Looks like I have [a problem with promises](https://pouchdb.com/2015/05/18/we-have-a-problem-with-promises.html) nolan  ;)

I'll wait for Travis to finish the last build before I push the fix (to make sure everything else is good, looking good so far)

daleharvey:
Awesome, all green :) thanks for the PR, 👍

Issue #5359, opened 2016-06-16
(Docs) markdown regression due to typography overhaul


nolanlawson:
As noted on Twitter: https://twitter.com/optikfluffel/status/743462052849680384

It appears that anywhere we include an alert block, the markdown gets lost immediately afterwards. Since I didn't have time to look into this, I just reverted fdbaaf08e233e2b58de89981a596a505a256cd60. in a branch and did `npm run publish-site` as a hotfix. It's still broken in master.

/cc @nickcolley Sorry for not noticing this during the PR

nolanlawson:
You can see this on both the custom.html page and the adapters.html page (it's particularly bad there).

nickcolley:
Normally I've found this happens when there's whitespace issues. E.g. making sure there's no indentation and markup is on one line normally fixes these kinds of things.

Will try and sort this soon if someone else doesn't suss is out

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/1c4475fea51f6b8017212bf58955003c9706022e

PR #5362, opened 2016-06-17
(#5359) - Remove whitespace that breaks markdown rendering

nickcolley:
Honestly don't understand what magic Jekyll is doing... Wish I knew why whitespace is so weird here..

### DIFF
--- a/docs/_includes/anchor.html
+++ b/docs/_includes/anchor.html
@@ -3,19 +3,4 @@
 {% else %}
   {% assign class = 'h2' %}
 {% endif %}
-<div class='{{ class }}'>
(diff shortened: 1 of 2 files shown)


daleharvey:
This fixed the adapters page, but still seeing http://0.0.0.0:4000/faq.html broken, I went through pretty much every page and thats the only one left I can see, any chance you could fix that then 👍 cheers

nickcolley:
Okay so I've just made the include all on one line, painful for maintaining but less likely to have side effects...

daleharvey:
Ugh jekyll is weird, thanks for figuring this out

PR #5356, opened 2016-06-15
(#5356) - leveldb-core should use index-browser.js

nolanlawson:
Forgot to add this line to `package.json` to tell `pouchdb-adapter-leveldb-core` to use `index-browser.js` when browserified. The impact is that you see this when you try to browserify it:

```
Error: Cannot find module 'leveldown' from '/Users/nolan/workspace/pouchdb-all-dbs/node_modules/pouchdb-adapter-leveldb-core/lib'
    at /Users/nolan/workspace/pouchdb-all-dbs/node_modules/resolve/lib/async.js:46:17
    at process (/Users/nolan/workspace/pouchdb-all-dbs/node_modules/resolve/lib/async.js:173:43)
    at ondir (/Users/nolan/workspace/pouchdb-all-dbs/node_modules/resolve/lib/async.js:188:17)
    at load (/Users/nolan/workspace/pouchdb-all-dbs/node_modules/resolve/lib/async.js:69:43)
    at onex (/Users/nolan/workspace/pouchdb-all-dbs/node_modules/resolve/lib/async.js:92:31)
    at /Users/nolan/workspace/pouchdb-all-dbs/node_modules/resolve/lib/async.js:22:47
    at FSReqWrap.oncomplete (fs.js:82:15)
```

Unfortunately this will necessitate a 5.4.4 release. Otherwise projects like `pouchdb-memory` and anything that uses the non-leveldown adapters in a browser-based project will fail.

PR #5364, opened 2016-06-17
(#5364) - verify index-browser is referenced in browser builds

omsmith:
Adds a check during builds to catch errors in pkg.browser entries, such
as the one fixed by #5356


daleharvey:
Nice idea to get some of these bugs caught at build time, will let @nolanlawson review this as he understands the build far more than me

nolanlawson:
Thanks, looks great to me :+1:

PR #5368, opened 2016-06-18
(#5368) - Remove references to index.html to fix excerpts

NickColley:
Similar to the issue we had before, the new jekyll build doesn't work when referencing index.html so the post details were not getting excerpts on the homepage and blog first page.

[skip ci]


daleharvey:
Ah I missed that one, good catch, thanks

PR #5371, opened 2016-06-19
WIP - Document close() and cleanup listeners

daleharvey:


Charlotteis:
Looks good to me! ✨

nolanlawson:
Seems a bit dubious to me that we emit a destroyed event when close() is called. Is that really necessary? Surprised it doesn't affect the tests at all.

daleharvey:
So I didnt actually think about whether that event got exposed to the user, yeh I will take a look and do it another way if so because thats gonna confuse people. Our event setup is fairly complicated, the emit is there just to trigger https://github.com/pouchdb/pouchdb/blob/master/packages/pouchdb-core/src/constructor.js#L39 which cleans up the event emitters

daleharvey:
Yeh its exposed to the use which is fairly confusing

daleharvey:
Added a test for this, requires 

    $ mocha tests/component/test.memleak.js --expose-gc --trace-gc

The `emit('destroyed');` doesnt seem to be an entire fix,  @shimaore do you see anything wrong with my test? not the easiest to debug however still seems like a valid bug if memory continues to grow here

nolanlawson:
@daleharvey It's possible to count the number of listeners, and even to exceed the maximum of 10, which throws an error (thanks to [throw-max-listeners-error](https://www.npmjs.com/package/throw-max-listeners-error)), however it might be tricky to reproduce.

I'd be in favor of breaking this PR up into documentation and the fix itself since it seems like two separate concerns. I'm +1 on the docs.

shimaore:
@daleharvey I've been running more tests based on a [modified version of `test.memleak.js`](https://github.com/shimaore/pouchdb/blob/more-5371/tests/component/test.memleak.js). The outcome is highly dependent on timing, how many times one calls `gc()` , and during longer tests I can see the heap usage jumping up by chunks of 100ko or more. I instrumented with `heapdump` but deciphering what Chrome's Profile tool tells me is always a challenge.

I strongly agree with @nolanlawson that making the `close()` API public is the most important part.

daleharvey:
ok merged the `close` part without any of the listener changes - https://github.com/pouchdb/pouchdb/commit/58c79dad7aea00955b7192ebf297ebe3d25d877c

PR #5372, opened 2016-06-19
(#3415) - Fix hang in leveldb

daleharvey:


daleharvey:
So we do this check a lot in the level adapter, probably not the last time we will need but really its something that will need a rewrite to entirely fix. Fixes the hang in level and I could not make the browser do so

nolanlawson:
Why not merge this?

PR #5375, opened 2016-06-19
(#3962) - Limit concurrent attachment requests

daleharvey:


daleharvey:
5 seemed like a nice limit

This test is a little bit of a noop, it passes both before and after however getting a reliable integration test to fail is going to be very hard with our current setup, the browser kills the connections after a timeout and with us testing between localhost + over a network there isnt really a way to reliably trigger that, however I figured its a good sanity test

nolanlawson:
:+1: when green. If this promise pool works well, we should use it in the replicator as well, where I basically wrote my own version of this.

daleharvey:
Yeh that sounds like a good idea, I thought I had seen that type of code before but I couldnt find it, this is a lightweight dep that seems fairly stable, thanks :)

ssured:
It's nice to limit connections, but isn't a global option a better way to limit concurrency through the whole library? `pouchdb-ajax` seems the proper place to manage this to me

ssured:
BTW I use the code below to throttle Promises. It is quite short and used in production. The javascript style is not so nice, as it is converted from a coffeescript source. `es6-promise-pool` is bigger and has lots of extra's.

```javascript
// converted from https://github.com/ssured/rsvp-throttle/blob/master/throttle.coffee
import Ember from 'ember';

const {
  RSVP
} = Ember;
const slice = [].slice;

export default function(concurrency, fn) {
  var makeFulfilledHandler, makeRejectedHandler, numRunning, queue, startJob, startJobs;
  if (concurrency < 1) {
    throw new Error("Concurrency must be equal or higher than 1.");
  }
  if (typeof fn !== "function") {
    throw new Error("Worker must be a function.");
  }
  numRunning = 0;
  queue = [];
  startJobs = function() {
    var job, results;
    results = [];
    while (numRunning < concurrency && (job = queue.shift())) {
      results.push(startJob(job));
    }
    return results;
  };
  startJob = function(job) {
    var promise, rejectedHandler;
    rejectedHandler = makeRejectedHandler(job);
    numRunning++;
    try {
      promise = fn.apply(job.context, job["arguments"]);
      return promise.then(makeFulfilledHandler(job), rejectedHandler);
    } catch (error) {
      return rejectedHandler(error);
    }
  };
  makeFulfilledHandler = function(job) {
    return function(result) {
      numRunning--;
      job.resolve(result);
      if (queue.length) {
        return startJobs();
      }
    };
  };
  makeRejectedHandler = function(job) {
    return function(error) {
      numRunning--;
      job.reject(error);
      if (queue.length) {
        return startJobs();
      }
    };
  };
  return function() {
    var args;
    args = 1 <= arguments.length ? slice.call(arguments, 0) : [];
    return new RSVP.Promise((function(_this) {
      return function(resolve, reject) {
        queue.push({
          context: _this,
          "arguments": args,
          resolve: resolve,
          reject: reject
        });
        return startJobs();
      };
    })(this));
  };
}
```

PR #5388, opened 2016-06-22
(#4575) - assert correct error message in Couch master tests

willholley:
CouchDB 2.0 changed the error message text when an invalid document id is used. Update the tests to accept either the old or new errors.


daleharvey:
👍 thanks

PR #5398, opened 2016-06-24
(#5398) - move replicate/sync entirely to plugin

nolanlawson:
This isn't a breaking change; it's just something that I didn't bother to do when I made the monorepo.

OTOH I did notice this weird thing - `db.replicate.sync` - which was never documented and ~~wasn't in our tests~~ _(correction: it was in a single test)_. So I removed it.

### DIFF
--- a/packages/pouchdb-replication/src/index.js
+++ b/packages/pouchdb-replication/src/index.js
@@ -4,6 +4,24 @@ import sync from './sync';
 function replication(PouchDB) {
   PouchDB.replicate = replicate;
   PouchDB.sync = sync;
+
+  Object.defineProperty(PouchDB.prototype, 'replicate', {
+    get: function () {
+      var self = this;
+      return {
+        from: function (other, opts, callback) {
+          return self.constructor.replicate(other, self, opts, callback);
+        },
+        to: function (other, opts, callback) {
+          return self.constructor.replicate(self, other, opts, callback);
+        }
+      };
+    }
+  });
+
+  PouchDB.prototype.sync = function (dbName, opts, callback) {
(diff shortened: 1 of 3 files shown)


packages/pouchdb-replication/src/index.js
```diff
@@ -4,6 +4,24 @@ import sync from './sync';
 function replication(PouchDB) {
   PouchDB.replicate = replicate;
   PouchDB.sync = sync;
+
+  Object.defineProperty(PouchDB.prototype, 'replicate', {
```
nolanlawson: supported in IE9 and above: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty

daleharvey: Whats the reason for using `defineProperty` over just a plain property definition? I would lean towards the simple version if there wasnt a particular advantage, but may be missing something

nolanlawson: Since this is `object.foo.bar` instead of `object.bar`, I don't believe we can access the `this` any other way. I believe `replicate` has to be a dynamic computed property that grabs the `this` and then binds it to its children (`to` and `from`). If there's another way, though, then I'm all for that.

daleharvey: ah yeh I forgot its tricky in that case, yeh no bother I cant think of anything cleaner off the top of my head and wasnt a major concern, mostly minor code style, cheers

daleharvey:
Yeh they make more sense over there, thanks 👍

PR #5400, opened 2016-06-25
(#5400) - set peerDependencies on pouchdb-core

nolanlawson:
I think the simplest system is to just set a `peerDependency` on all semvered plugins (e.g. `pouchdb-adapter-http` but not `pouchdb-ajax`) to warn the user if they're using any version of `pouchdb-core` other than the exact same one as the plugin. Trying to do ranges here would be too complicated in my opinion; let's just advise users to update all the versions at the same time.


daleharvey:
Yeh this makes sense to me, cheers

Issue #4632, opened 2015-12-03
PouchDB Crashes on iOS after several memory warnings.


thedracle:
Doing a large sync, of maybe 12k documents, causes increasing memory usage which never seems to be cleaned up.

Prior to a sync, there will be ~200MB used, and this will increase to ~450MB. After Pouch stops, it never seems to reclaim this memory, and often during a sync iOS will stop the application.

thedracle:
There seems to be a memory leak with Pouch. After syncing a large database, the memory profile gradually climbs, and never seems to reduce.

Subsequent syncs of other collections will eventually crash the application.

We have tried websql as a driver, sqlite, and this behavior seems entirely independent of the driver we are using.

thedracle:
Platform Details:
iOS 9.1
UIWebView

Here is a screenshot of the memory leak issue:
![leakscreenshot](https://cloud.githubusercontent.com/assets/1301817/11604853/62aceb60-9aaf-11e5-952c-d476a9d5a2a7.png)

This is produced by replicating a database, and after it finishes, all of the memory it allocated seems to be indefinitely retained.

nolanlawson:
Can you reproduce this with the FruitDOWN adapter? It would be nice to figure out if this is a bug in the `websql` adapter (used by both WebSQL and SQLite Plugin) or in PouchDB itself. Also do you see it on any other browsers/platforms?

thedracle:
As far as I am aware, the FruitDown adapter will not work on UIWebView because it doesn't support IndexDB. We can use WKWebView, but it has the issue that if a location.reload() occurs, all of the permanent storage (LocalStorage/IndexDB/WebSQL) seems to get temporarily wiped until the Application is restarted.

I can test it against WKWebView though with FruitDown to determine if it's an issue with the websql adapter specifically.

thedracle:
I noticed you commented on https://github.com/axemclion/IndexedDBShim. 

Is it possible to use this with UIWebView + FruitDown in a stable way? It looks like there is an issue setting to window.indexDB, but I could work around that with the driver itself.

In the meantime I will try to figure out if the websql adapter in particular is the source of the issue (I suppose because this is the common link between using sqlite and websql directly), via WKWebView.

thedracle:
Using the IndexDB shim + Fruitdown has a similar result, only since it is much much slower, it happens over a very long duration of time.

[...]

PR #5401, opened 2016-06-26
(#4632) - fix attachment md5 memory leak

nolanlawson:
Turned out we had a nasty little bug in MD5 summing that went undiscovered for a long time. The `spark-md5` library we're using was chosen because it has an "incremental MD5" calculation feature, but unfortunately we weren't really using it as intended. Instead of taking a Blob (which is not read into memory) and then slicing it and reading each Blob slice as an ArrayBuffer in chunks (since `readAsArrayBuffer` is the thing that actually reads it into memory), we were instead reading in the entire ArrayBuffer and then slicing that, which is useless because we've already read the entire thing into memory at that point. The [spark-md5 documentation itself](https://github.com/satazor/js-spark-md5#hash-a-file-incrementally) describes how you're supposed to do it.

This patch corrects this behavior while also cleaning up some no-longer-needed code from `pouchdb-binary-utils`. I confirmed that this patch fixes the test case submitted in https://github.com/pouchdb/pouchdb/issues/4632#issuecomment-228325496 and makes it so that it doesn't crash in iOS anymore. I also confirmed in Chrome that I no longer see memory spikes when using the memory profiler (it used to jump from 5MB to 15MB and then back down; now it stays put around 4MB-5MB).

Note that this is not a true memory leak in that memory doesn't add up over time. However it does spike in large sudden increases, which is probably enough to crash the browser. Also note I tried to write a test for this (6237615e6fba438fa6aa5fbae2573577f9ba3109), but it's hard to write one that will crash the browser without also making a test that's just too slow to run as part of our test suite, so I abandoned that idea. Hopefully the fix is obvious enough without the test.


packages/pouchdb-adapter-utils/src/preprocessAttachments.js
```diff
@@ -1,72 +1,81 @@
 import {
   atob,
   btoa,
```
nolanlawson: I took this opportunity to clean up `preprocessAttachments` and move a lot of functions-within-functions into their own functions, which is better for perf and easier to read.

packages/pouchdb-adapter-utils/src/preprocessAttachments.js
```diff
@@ -1,72 +1,81 @@
...
-        att.data = binStringToBlobOrBuffer(asBinary, att.content_type);
-      } else if (blobType === 'base64') {
-        att.data = btoa(asBinary);
-      } else { // binary
-        att.data = asBinary;
-      }
-      binaryMd5(asBinary, function (result) {
-        att.digest = 'md5-' + result;
+function preprocessBlob(att, blobType, callback) {
+  binaryMd5(att.data, function (md5) {
```
nolanlawson: instead of taking a string or an ArrayBuffer as input, `binaryMd5` now takes a string or a Blob as input. It should never ever take an ArrayBuffer as input, because we should never be passing it a full ArrayBuffer.

packages/pouchdb-binary-utils/src/readAsArrayBuffer.js
```diff
@@ -1,8 +1,17 @@
-// In Node.js, just convert the Buffer to a Buffer rather than
-// convert a Blob to an ArrayBuffer. This function is just a convenience
-// function so we can easily switch Node vs browser environments.
-function readAsArrayBuffer(buffer, callback) {
-  callback(buffer);
```
nolanlawson: it was useless to have a different node/browser version of this, because we never used the node one

nolanlawson:
Chrome has an experimental API called `window.performance.memory`, but it doesn't really help us because you have to start Chrome with a special flag to get fine-grained data.

daleharvey:
Yeh I can understand how this would be extremely hard to test, looks good to me, thanks 👍

htejwani:
Hi Nolan, 

Thanks for this fix, we were eagerly waiting for this as we were facing memory leaks in our app during sync. Can you please let us know when we can expect a new release that will include this fix?

nolanlawson:
No prob, it's been released in 5.4.5. :)

htejwani:
Hey Nolan, I tried 5.4.5 and have the following observations:

1) The requests for attachments (not very big < 200KB) timeout sporadically
2) When there is a timeout, the sync API does not throw the error because of a javascript erorr at line: 7630 (err.status = xhr.status;). The error reads: 'Cannot assign to a read only property'
3) The memory still does pile up over time in our case (our database size is > 400MB)

nolanlawson:
@htejwani Can you please provide a test case to reproduce? Also, is this a bug that emerged in 5.4.5, or was it present in 5.4.4 or earlier? As for the memory usage, I'm not very familiar with iOS memory analysis tools, so any guidance would be helpful.

This also sounds like a good candidate for a new issue. #5401 definitely fixes _something_ (I identified a gross memory overusage in our code), but of course even if you fix one bug, it's always possible there are other bugs.

CatchSandeepVaid:
@nolanlawson - This issue is present in 5.4.5 as well as 5.4.4. I have created a new issue( [https://github.com/pouchdb/pouchdb/issues/4632](url) ) regarding this.

Issue #5402, opened 2016-06-26
Unbounded number of listeners added when performing back off.


seigel:
Pull Request: https://github.com/pouchdb/pouchdb/pull/5403

### Environment (Node.js/browser/hybrid app/etc.)

Found this while running in react-native

### Browser/platform (Chrome/FF/Safari/Edge/iOS/Android/etc.)

iOS

### Adapter (IndexedDB/WebSQL/LevelDB/etc.)

asyncdown

### Server (CouchDB/Cloudant/Couchbase/PouchDB Server/etc.)

CouchDB

### Please describe your issue. Test cases (e.g. JSBin) are very helpful!

Sorry I haven't been able to build a test case or a fix yet, but this call causes an unbounded number of listeners to be added to the event emitter when there is no activity going on.  To correct the issue when it happens, trigger a replication and the event emitters will go away again, and then start building again.

https://github.com/pouchdb/pouchdb/blob/master/packages/pouchdb-replication/src/backoff.js#L18

Below is a longer description of the problem.

I think I found where some of this is happening.  It might be a core pouchdb problem with replication.

Add code something like this to your pouchdb react native app:

```
var originalOn = EventEmitter.prototype.on;
EventEmitter.prototype.on = function (type, listener) {
    console.log("originalOn", type, this.listeners(type).length, this);
    originalOn(this, arguments);
}

```

You will notice that the number of listeners will grow with time.  I think there might be one more event that exhibits this behaviour.  After editing some source code and watching for a while, I found that one way to reset this behaviour is to just cause a replication to happen.  

This leak has everything to do with how the back off function is implemented.  It keeps adding listeners to the pile.

```
function backOff(opts, returnValue, error, callback) {
  if (opts.retry === false) {
    returnValue.emit('error', error);
    returnValue.removeAllListeners();
    return;
  }
  if (typeof opts.back_off_function !== 'function') {
    opts.back_off_function = pouchdbUtils.defaultBackOff;
  }
  returnValue.emit('requestError', error);
  if (returnValue.state === 'active' || returnValue.state === 'pending') {
    returnValue.emit('paused', error);
    returnValue.state = 'stopped';
    
   // THE CALL BELOW ADDS LISTENERS WITHOUT BOUND WHEN ALL IS QUIET
   
    returnValue.once('active', function () {
      opts.current_back_off = STARTING_BACK_OFF;
    });
  }

  opts.current_back_off = opts.current_back_off || STARTING_BACK_OFF;
[...]

PR #5403, opened 2016-06-26
Remove the unbounded accumulation of listeners

seigel:
Added some crude code to remove the unbounded accumulation of listeners when backing off happen.

Issue: https://github.com/pouchdb/pouchdb/issues/5402

### DIFF
--- a/packages/pouchdb-replication/src/backoff.js
+++ b/packages/pouchdb-replication/src/backoff.js
@@ -15,9 +15,14 @@ function backOff(opts, returnValue, error, callback) {
   if (returnValue.state === 'active' || returnValue.state === 'pending') {
     returnValue.emit('paused', error);
     returnValue.state = 'stopped';
-    returnValue.once('active', function () {
+    function backoffTimeSet() {
       opts.current_back_off = STARTING_BACK_OFF;
-    });
+    }
+    function removeBackOffTimeSet() {
+      returnValue.removeListener('active', backoffTimeSet);
+    }
+    returnValue.once('paused', removeBackOffTimeSet);
+    returnValue.once('active', backoffTimeSet);
   }
 
   opts.current_back_off = opts.current_back_off || STARTING_BACK_OFF;
--- a/packages/pouchdb-replication/src/backoff.js
+++ b/packages/pouchdb-replication/src/backoff.js
@@ -15,14 +15,14 @@ function backOff(opts, returnValue, error, callback) {
   if (returnValue.state === 'active' || returnValue.state === 'pending') {
     returnValue.emit('paused', error);
     returnValue.state = 'stopped';
-    function backoffTimeSet() {
+    var backOffSet = function backoffTimeSet() {
       opts.current_back_off = STARTING_BACK_OFF;
-    }
-    function removeBackOffTimeSet() {
-      returnValue.removeListener('active', backoffTimeSet);
-    }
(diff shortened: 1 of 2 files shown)


packages/pouchdb-replication/src/backoff.js
```diff
@@ -15,14 +15,19 @@ function backOff(opts, returnValue, error, callback) {
...
-    returnValue.once('active', function () {
+    function backoffTimeSet() {
       opts.current_back_off = STARTING_BACK_OFF;
-    });
+    }
+    function removeBackOffTimeSet() {
+      returnValue.removeListener('active', backoffTimeSet);
+    }
+    returnValue.once('paused', removeBackOffTimeSet);
+    returnValue.once('active', backoffTimeSet);
```
nolanlawson: Nice catch! Won't this add listeners to `paused` without bounds, though?

nolanlawson:
BTW the way you can repro this in a test is to do something to produce the "more than 10 listeners" warnings, ideally in `test.retry.js`. We have a module ([throw-max-listeners-error](https://www.npmjs.com/package/throw-max-listeners-error)) that will throw whenever that happens.

nolanlawson:
It's failing in Travis due to eslint, you can repro by running `npm run eslint`. Seems the error is `Move function declaration to function body root`. If you need any help, feel free to join the Slack or IRC chat; thanks a bunch for the PR! :)

seigel:
I'll see what I can do today. 

------
@cgul
atVenu.com




On Sat, Jun 25, 2016 at 11:44 PM -0600, "Nolan Lawson" <notifications@github.com> wrote:












It's failing in Travis due to eslint, you can repro by running npm run eslint. Seems the error is Move function declaration to function body root. If you need any help, feel free to join the Slack or IRC chat; thanks a bunch for the PR! :)

packages/pouchdb-replication/src/backoff.js
seigel: It shouldn't as there is an emit 'paused' almost immediately before.  This clears out both listeners before the new one is added.

daleharvey:
This looks great, because contributor PR's dont run in saucelabs and these tests are historically quite unpredictable I have opened another PR to run in saucelabs @ https://github.com/pouchdb/pouchdb/pull/5405 but I am 👍 on green looking at this

nolanlawson:
LGTM as well. :+1: when green in a Sauce run.

seigel:
I'm still not having a great amount of success re-creating in tests the scenario I get in the react-native world for this.  In the tests it appears that the adding of the `.once('active'....` around line 19 in the master branch for backoff.js is fairly well guarded by the `returnValue.state == 'stopped'` which on the next time through the backoff call, avoids the whole chunk of code in the if block.  

I am going to investigate the react-native version again and figure out why it appears to stay in pending or go back to pending and thus keep re-entering this block.  <-- The reason for my "fix".

This fix isn't damaging, and certainly is helpful for my code, but hopefully if I can figure out the difference we might be able to back this out in the future and fix the root cause.

nolanlawson:
If you're only seeing it in React Native, it could be a subtle timing issue
with JSCore. You could try reproducing in Safari or PhantomJS.

daleharvey:
So yeh the test is helpful, I dont expect this is the last bit of work we will need to do around here, going to merge but will leave the issue open, cheers

Issue #4967, opened 2016-03-09
Map/reduce calls map() function twice for every document


ssured:
Hello,

For debugging purposes I put a console.log inside a map function. To my surprise the function was called double the number of records in my database, so I made https://jsbin.com/totemituja/2/edit?html,js,console to test what happens.

It seems that if you query an unindexed view, then the view is triggered to build. If you query again while building the index, the index is built again. In the fiddle you can observe this behaviour as the DB contains 10 records, it's queried twice and runs the map function 20 times.

Kind regards

nolanlawson:
That's definitely a bug. Thanks for reporting!

colinskow:
By the way in Safari I'm getting `"WebSQL threw an error"` when loading the JSBin, so that is something else to look into.

ssured:
Pretty weird, I could not quickly find the cause of that error, but I'm having it too in Safari

nolanlawson:
^ This is undoubtedly a JSBin issue since they use iframes.

nolanlawson:
^ Issue fixed, thanks for the test case!

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/45156c5bcf5c6ebbfb049c4234611636bb8220d8

PR #5404, opened 2016-06-27
(#4967) - fix map function called twice

nolanlawson:
I confirmed this bug thanks to the test provided by @ssured. I adapted their example into a test case which fails before this fix but succeeds after.

The bug occurs when you call `db.query()` twice within the same turn of the event loop. Because the `db._cachedViews` cache was only being populated after the promise was resolved, it was possible to have two identical views in memory at the same time. By caching immediately without waiting for the promise to resolve, we can avoid this issue.


daleharvey:
This is awesome, thanks 👍

ssured:
Many thanks for this, this was a severe performance issue on mobile! Awesome

PR #5406, opened 2016-06-27
(#5406) - move sublevel-pouchdb into monorepo

nolanlawson:
While profiling the in-memory adapter, I noticed that there was a function that was taking up lots of self-time:

screenshot 2016-06-26 21 54 54

I tracked it down to [this function](https://github.com/rvagg/node-errno/blob/1c2b1fcbf22ef2bafbf6cda378cfed400f5163fd/custom.js#L25-L41) which is being called by sublevel-pouchdb from [this function](https://github.com/pouchdb/sublevel-pouchdb/blob/55351adfe5f6600dc8c445b794dbde54beea795d/lib/shell.js#L86), which is slow because it's constantly creating new custom error objects. This is useless, because all we need to do is return an Error that is a `NotFoundError` (i.e. `err.name === 'NotFoundError'`).

I wanted to fix this perf issue in sublevel-pouchdb, so I also took this opportunity to just move it into the monorepo. I converted it to ES6 modules, removed code we weren't using, `istanbul ignore`'d stuff I was a bit unsure about, and changed it so that it's not re-creating custom `NotFoundError`s over and over again.

I measured the perf difference by increasing the `temp-views` perf test to 10 iterations and running `PERF=1 LEVEL_ADAPTER=memdown GREP=temp npm test`. The result is:

* Before: 17861ms
* After: 16608ms

So, this represents a 7% speedup for the `temp-views` test at least.


daleharvey:
Awesome, 👍

PR #5407, opened 2016-06-28
Added test for event emitter listeners growing without bound

seigel:
This supports the PR that contained the fix. https://github.com/pouchdb/pouchdb/issues/5402

### DIFF
--- a/tests/integration/test.backoff.js
+++ b/tests/integration/test.backoff.js
@@ -1,42 +1,48 @@
 'use strict';
 
 var adapters = [
-    ['local', 'http']
+  ['local', 'http']
 ];
 
 adapters.forEach(function (adapters) {
-    var suiteName = 'test.backoff.js-' + adapters[0] + '-' + adapters[1];
-    describe(suiteName, function () {
+  var suiteName = 'test.backoff.js-' + adapters[0] + '-' + adapters[1];
+  describe(suiteName, function () {
 
-        var dbs = {};
+    var dbs = {};
 
-        beforeEach(function (done) {
-            dbs.name = testUtils.adapterUrl(adapters[0], 'testdb');
-            dbs.remote = testUtils.adapterUrl(adapters[1], 'test_repl_remote');
-            testUtils.cleanup([dbs.name, dbs.remote], done);
-        });
-
-        after(function (done) {
-            testUtils.cleanup([dbs.name, dbs.remote], done);
-        });
-
-        it('backoff event listeners', function (done) {
-            var remote = new PouchDB(dbs.remote);
-            var db = new PouchDB(dbs.name);
-            const rep = db.sync(remote, {
-                live: true,
-                retry: true,
-                heartbeat: 1,
-                timeout: 1,
-                back_off_function: function () { return 1; }
-            });
-
-            setTimeout(function () {
-              var numAListeners = rep.pull.listeners("active").length;
-              numAListeners.should.be.below(3);
-              done();
-            }, 600);
-        });
-
+    beforeEach(function (done) {
+      dbs.name = testUtils.adapterUrl(adapters[0], 'testdb');
+      dbs.remote = testUtils.adapterUrl(adapters[1], 'test_repl_remote');
+      testUtils.cleanup([dbs.name, dbs.remote], done);
     });
+
+    after(function (done) {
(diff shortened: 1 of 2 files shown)


tests/integration/test.backoff.js
```diff
@@ -0,0 +1,42 @@
...
+                timeout: 1,
+                back_off_function: function () { return 1; }
+            });
+
+            setTimeout(function () {
+              var numAListeners = rep.pull.listeners("active").length;
+              numAListeners.should.be.below(3);
+              done();
+            }, 600);
+        });
```
daleharvey: So this looks like a good start however there is a main problem in that the replication will still be active when the test finishes, we havent been using es6 features like `const` around so the test will need to look something like

```
    var rep = db.sync(remote, {
      live: true,
      retry: true,
      heartbeat: 1,
      timeout: 1,
      back_off_function: function () { return 1; }
    });

    rep.on('complete', function() { done(); });

    setTimeout(function () {
      var numAListeners = rep.pull.listeners("active").length;
      numAListeners.should.be.below(3);
      rep.cancel();
    }, 600);
```

If we could get rid of the setTimeout that would be perfect (count the backoffs?) however if that doesnt work it is not the wost thing in the world. Last tiny nit is that we use 2 space tabs, not 4

seigel: Hey the feedback is great!  I'll try and get all this stuff working.  I couldn't get the 'shoulds'
to throw the assertion in the listeners. It just halted the flow.  Any suggestions to avoid the setTimeout are appreciated. 
Thank youJames
------
@cgul
atVenu.com




On Tue, Jun 28, 2016 at 3:40 AM -0600, "Dale Harvey" <notifications@github.com> wrote:












In tests/integration/test.backoff.js:
> +            var remote = new PouchDB(dbs.remote);
> +            var db = new PouchDB(dbs.name);
> +            const rep = db.sync(remote, {
> +                live: true,
> +                retry: true,
> +                heartbeat: 1,
> +                timeout: 1,
> +                back_off_function: function () { return 1; }
> +            });
> +
> +            setTimeout(function () {
> +              var numAListeners = rep.pull.listeners("active").length;
> +              numAListeners.should.be.below(3);
> +              done();
> +            }, 600);
> +        });



So this looks like a good start however there is a main problem in that the replication will still be active when the test finishes, we havent been using es6 features like const around so the test will need to look something like

    var rep = db.sync(remote, {
      live: true,
      retry: true,
      heartbeat: 1,
      timeout: 1,
      back_off_function: function () { return 1; }
    });

    rep.on('complete', function() { done(); });

    setTimeout(function () {
      var numAListeners = rep.pull.listeners("active").length;
      numAListeners.should.be.below(3);
      rep.cancel();
    }, 600);




If we could get rid of the setTimeout that would be perfect (count the backoffs?) however if that doesnt work it is not the wost thing in the world. Last tiny nit is that we use 2 space tabs, not 4

tests/integration/test.backoff.js
```diff
@@ -0,0 +1,48 @@
...
+      var replication = db.sync(remote, {
+        live: true,
+        retry: true,
+        heartbeat: 1,
+        timeout: 1,
+        back_off_function: function () {
+          var numberOfActiveListeners = replication.pull.listeners("active").length;
+          ++backOffCount;
+          if (backOffCount > 15 || numberOfActiveListeners > 3) {
+            replication.cancel();
```
daleharvey: `replication.cancel()` is not sync, calls to that need to wait for 

```
replication.on('complete', function() { 
  if (activeListeners > 3) { 
    done(new Error(....));
  }
});
```

daleharvey:
Still one issue with the cleanup of the test, I know its something you will probably never notice when running the test, however it introduce very subtle breakages when the whole test suite is ran. Aside from that this looks great to me. Thanks

atvenu:
No worries..I'll see what I can do...I was still having a heck of a time trying to get it running everytime.  It is like the node thread that is doing the replication and emitting the events eats the assertion exceptions until a timeout or something else fires...I didn't want the test to be too crazy confusing and constant ugly timed.  Let me poke around at it more and see if I can dig out of the weirdness.

seigel:
Hopefully a cleaner version.

daleharvey:
This is great, test looks perfect now, thanks a lot, 👍

PR #5415, opened 2016-06-30
(#5406) - fix unnecessary sublevel dep

nolanlawson:
This is a dependency I forgot to clean up when I moved `sublevel-pouchdb` into the monorepo. Having this dependency here does nothing.


daleharvey:
Good catch, cheers 👍

PR #5416, opened 2016-06-30
pouchdb-collections back in the monorepo

nolanlawson:
No reason not to do this. We own it, it's a core part of the library, and putting it in the monorepo means we get Rollup collapsing benefits.


daleharvey:
Looks good but regressed coverage

nolanlawson:
Ah, glad it caught that; it means there's stuff we can remove. :)

daleharvey:
Sweet cheers

PR #5418, opened 2016-06-30
added some of my PouchDB packages

pgte:


daleharvey:
These are awesome, thanks a lot :)

Issue #5172, opened 2016-05-18
Cannot handle 401 error during live replication


sbroadhead:
I would like any exceptional errors that happen during PouchDB<->CouchDB replication to notify my app to log the user out (since the problem is likely to be caused by stale login data or expired CouchDB credentials). This includes attempting to access a non-existent database, or 401 errors indicating incorrect credentials. I can use the `paused` event from the replicator to see errors that happen during the replication process, but there doesn't seem to be a way to handle 401 Unauthorized errors that happen during replication. Try as I might, they always get reported as unhandled promise exceptions.

From the cursory glance at the source, it seems likely that the change handler's error events do not get propagated to the replication handler's listeners, and that the 401 error is happening from within the changes feed.

Is this a bug/oversight, or am I missing an event that I can handle?

nolanlawson:
It's undoubtedly an error if those exceptions don't bubble up, yes. Although it's kind of unclear what should happen in the case of `retry` replication (I'll assume you're using `retry`), because in principle PouchDB is supposed to retry on any kind of error, but in the case of 40x's it may make more sense to just throw the error and let the user handle it.

axul:
I'm having the same issue while usign pouchdb <--> sailsjs (JWT auth)<-> couchdb.
My frontend gets the 401 unauthorized from couchdb if I'm not signed in (normal behavior) but I'm not being able to handle it.

nolanlawson:
Unfortunately right now the solution is to use something other than `retry` replication to ensure the user is authorized, then do `retry` replication after that.

axul:
Could you be a little more specific for a newb here? :( my code:

         var replicateOptions = {
            live: true,
            retry: true,
            auto_compaction: true,
            ajax:{
              headers: {
                'Authorization': 'JWT ' + $LocalStorage.get('token')
              },
              withCredentials: false
            },
          };
          
          this.syncHandler = this.db.sync(remoteCouch, replicateOptions);
          this.syncHandler
          .on('error', function(err){
            console.log(err);
          })
          .on('denied', function(err){
            console.log(err);
          })
          .on('paused', function(err){
            console.log(err);
          })
          .on('complete', function(err){
            console.log('complete');
            console.log(err);
          });

         var changesOptions = { since: 'now', live: true};

          this.db.changes(changesOptions).on('change', function(){
            console.log('hi');
          });

[...]

PR #5421, opened 2016-07-01
Differentiate between validation and non-validation errors

anowak:
Up to now, all 401 and 403 error were treated the same, i.e. it was
assumed that it may be connected to a result of a validation function on
the destination database or to error saving a design as non-admin user
(there is also a case of saving checkpoint on a read-only source
database, but it is dealt with by the Checkpointer).

This change introduces storing the validation errors separately, to
handle 401s on fetching remote databases properly. I decided to signal
these as errors, keeping `denied` only for validation errors (following
the separation between the two).

### DIFF
--- a/packages/pouchdb-replication/src/replicate.js
+++ b/packages/pouchdb-replication/src/replicate.js
@@ -24,7 +24,6 @@ function replicate(src, target, opts, returnValue, result) {
   var doc_ids = opts.doc_ids;
   var repId;
   var checkpointer;
-  var allErrors = [];
   var changedDocs = [];
   // Like couchdb, every replication gets a unique session id
   var session = uuid();
@@ -64,36 +63,34 @@ function replicate(src, target, opts, returnValue, result) {
         completeReplication();
         throw new Error('cancelled');
       }
-      var errors = [];
-      var errorsById = {};
+
+      // `res` doesn't include full documents (which live in `docs`), so we create a map of 
+      // (id -> error), and check for errors while iterating over `docs`
+      var errorsById = Object.create(null);
       res.forEach(function (res) {
         if (res.error) {
-          result.doc_write_failures++;
-          errors.push(res);
           errorsById[res.id] = res;
         }
       });
-      allErrors = allErrors.concat(errors);
-      result.docs_written += currentBatch.docs.length - errors.length;
-      var non403s = errors.filter(function (error) {
-        return error.name !== 'unauthorized' && error.name !== 'forbidden';
-      });
+
+      var errorsNo = Object.keys(errorsById).length;
+      result.doc_write_failures += errorsNo;
+      result.docs_written += docs.length - errorsNo;
 
       docs.forEach(function (doc) {
         var error = errorsById[doc._id];
         if (error) {
-          returnValue.emit('denied', clone(error));
+          result.errors.push(error);
+          if (error.name === 'unauthorized' || error.name === 'forbidden') {
+            returnValue.emit('denied', clone(error));
+          } else {
+            throw error;
+          }
         } else {
           changedDocs.push(doc);
         }
       });
 
-      if (non403s.length > 0) {
-        var error = new Error('bulkDocs error');
-        error.other_errors = errors;
-        abortReplication('target.bulkDocs failed to write docs', error);
-        throw new Error('bulkWrite partial failure');
-      }
     }, function (err) {
       result.doc_write_failures += docs.length;
       throw err;
@@ -121,7 +118,10 @@ function replicate(src, target, opts, returnValue, result) {
       }
       currentBatch = undefined;
       getChanges();
-    }).catch(onCheckpointError);
+    }).catch(function (err) {
+      onCheckpointError(err);
+      throw err;
+    });
   }
 
   function getDiffs() {
@@ -220,19 +220,17 @@ function replicate(src, target, opts, returnValue, result) {
     }
     result.ok = false;
     result.status = 'aborting';
-    result.errors.push(err);
-    allErrors = allErrors.concat(err);
     batches = [];
     pendingBatch = {
       seq: 0,
       changes: [],
       docs: []
     };
-    completeReplication();
+    completeReplication(err);
   }
 
 
-  function completeReplication() {
+  function completeReplication(fatalError) {
     if (replicationCompleted) {
       return;
     }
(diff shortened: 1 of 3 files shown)


packages/pouchdb-replication/src/replicate.js
```diff
@@ -73,15 +74,17 @@ function replicate(src, target, opts, returnValue, result) {
           errorsById[res.id] = res;
```
anowak: I'm not sure if I understand the context here: for me iterating twice (first over `res`, and then `docs`) is redundant, as it makes sense only for case when `res.length !== docs.length`. On the other hand, in case when `docs.length > res.length`, adding extra docs to `changedDocs` is questionable.

daleharvey:
@anowak Thanks a lot for this, looks like a good start. However could you explain what you are fixing precisely? the thread linked has a lot of different problems and assumptions and would like to clarify them exactly when looking at this. It sounds as though we do not handle being logged out mid replication? how does your patch affect people without permissions to write a design document?

Sorry for needing caught up here, this is a fairly hairy piece of the code with quite a few assumptions that arent obvious. One thing that would be a huge help is if this patch could clean up some of the code that is currently not so clear. ie by looking at https://github.com/evidenceprime/pouchdb/blob/a1ef9b8479394a03345ee2d51fd05707811a2b2c/packages/pouchdb-replication/src/replicate.js#L257 I cant see what its needed for, you also mentioned some redundant loops in your notes. 

Will carry on looking through the issue / patch to figure out the consequences, but more details would hep, thanks

nolanlawson:
Actually the new code looks pretty clear to me, and the new test seems to capture what it's doing. Hot-replacing `db._ajax` to simulate a 40x is a perfect way to capture validation errors; I also like the idea of just keeping the two types of errors in separate arrays. That combined with the fact that we're still at 100% coverage would make this a :+1: from me. 

@daleharvey I think to sum up what's going on here is just to say that 40x errors currently don't propagate to clients when they're running in `retry` mode, because it just keeps retrying instead. This is worse than useless because it means that when a 40x occurs, it just keeps retrying for forever and doesn't even notify the users that they need to re-authenticate. This PR seems to fix that.

@anowak One thing I will say about the tests is that I'd feel a bit more comfortable if we also tested live+retry and live=false, just to ensure that we're getting the correct behavior in those code paths as well. Otherwise your tests look awesome.

daleharvey:
So some 40X errrors we ignore (or fire denied), this is because they either fail validate_doc_update or the server will reject design doc writes without admin mode, I want to be clear about what happens in these situations now. If we now fatally error during replication when we fail to write a design doc that is a breaking change, it may be a valid breaking change but it does need brought up.

Also looking though the PR'd code we have `errors` `errorsById` `non403s` `validationErrors` and `fatalErrors`, that isnt the fault of the patch, there are too many already however I would like to be removing and clarifying / simplifying these cases, not adding to them.

anowak:
@daleharvey Our main problem is that right now, if user session expires we would like to force re-authentication, but don't get the expected error from PouchDB (which we got in the past). We don't use `retry` and the current situation is that we receive a `complete` event (which, for live replication should be only emitted if the replication gets cancelled), with the error hidden inside.

I understand your concerns, since I also had problems tracing through the error handling. The gist of https://github.com/evidenceprime/pouchdb/blob/a1ef9b8479394a03345ee2d51fd05707811a2b2c/packages/pouchdb-replication/src/replicate.js#L257 is that all errors lead either directly or indirectly (through `abortReplication`) to `completeReplication`, where `fatalErrors` (formerly: `allErrors`) is checked to determine whether we arrived there along 'happy' or 'unhappy' path. I agree that the code could use some cleaning up, but my intention was to keep the changes minimal (at least in the beginning). 

For definiteness, let's call all the 401s and 403s you have mentioned (design docs rejections, results of validation functions) `denied` errors. My assumption here is that all the `denied` errors are signalled by, and only by, the result of the `bulkDocs` operation. So we can in fact easily tell them apart, emitting `denied` event for them and listing them in `errors` field of the `complete` event.

@nolanlawson Unfortunately, my PR doesn't fix the issue with retrying the replication. We don't use `retry` in our application and I wasn't sure what is the correct behaviour in this case.

As a sidenote, in our custom connection management code that was written before `retry` became available in PouchDB, we modelled this as FSM using [Machina](http://machina-js.org/). We have three 'error' states, to which we transition based on the status code of the error: `disconnected` (which works in similar fashion to `retry` in PouchDB), `serverError` (in which we give the server some rest) and `unauthorized` (in which we don't retry). So my first implementation was to not retry on 401s, but I felt that it would contradict what is written in the docs, and thus would be a breaking change.

anowak:
I've updated the PR to include your suggestions.

On a side note, why PouchDB doesn't invoke `_ensure_full_commit` after completing a batch? Is it the case that Cloudant doesn't support it?

nolanlawson:
Hm, I wasn't aware of `_ensure_full_commit`, but it looks like the kind of low-level thing that PouchDB shouldn't be responsible for. Server admins can configure their own write interval and call the API themselves etc.

anowak:
I probably shouldn't have squashed the commits, because now it is difficult to see what was updated. Here is the overview of the changes for your convenience (more details in the updated commit message):

- Refactor of error handling making failure that results in aborting the replication more explicit. Non-fatal errors can now be kept directly in the result.
- Added comment to `errorsById` to explain its purpose.
- Disable backoff for authorization **failures** per @nolanlawson suggestion. Added test to illustrate.

The behavior for handling non-authorization `writeErrors` is the same (it aborts the replication), which I think is in the contrary to how CouchDB handles it and what is documented in the CouchDB replication protocol.

daleharvey:
Awesome, no bother about the ammended diff, I prefer reading the changes from HEAD anyway. 

This is a nice cleanup, my only worry is now the removed test case, so as far as I can understand this used to throw the paused event, but will now halt replication even if `retry: true`? because the error is determined unrecoverable, which is what the code used to try to do but was broken?

Also will this now complete replication with an error if the user tries to replicate a whole database but `_design` documents are rejected where before this patch it would have carried on replication but put the design doc rejects in an error field and fired denied?

daleharvey:
And sorry for the constant questions, this behaviour has been fairly murky up until now, I think if we all agree on what the behaviour is / should be then we can document it more and shore up a few more tests, with the current test that is deleted I would prefer that it be changed to show the new behaviour and I think we need new tests based on the last question ^ I asked as well. If you dont want to write the extra tests then I can help out with that

anowak:
No problem, this is a very delicate piece of code, so I understand your concern.

The reason I deleted the test is that it demonstrates the behavior that is now considered invalid. This is at least my understanding of your closing comment in #4276 (the issue that is referenced in the deleted test), while closing it as a duplicate of the current one (#5172).

You are correct that 401s when trying to get a checkpoint on a remote will now halt replication (even with `retry: true`). Please note that it was not my initial approach, but I changed it after https://github.com/pouchdb/pouchdb/pull/5421#issuecomment-230114019.

This change does not break replication of design docs without admin privileges. The test for this case already exists and does not fail: https://github.com/pouchdb/pouchdb/blob/master/tests/integration/test.replication.js#L3605

daleharvey:
ok great, I think this is good to go now, thanks a lot for following this through :)

Issue #5417, opened 2016-06-30
[bug] WebSQL adapter not detected correctly in pouchdb-adapter-websql/src/valid.js (when SQLite Plugin 2 + Cordova used)


dharders:
### Environment 

Hybrid App

- cordova 6.2.0
- cordova-plugin-wkwebview-engine
- cordova-labs-local-webserver
- cordova-plugin-sqlite-2

### Browser/platform 

iOS 9.3.2 WKWebView via cordova plugin

### Adapter 

WebSQL - via cordova-plugin-sqlite-2

### Server CouchDB

CouchDB

### Please describe your issue.

I have a hybrid app running successfully on iOS 9.3.2 using the faster WKWebView and PouchDB 5.3.2 ( `{adapter: 'websql'}` and  **cordova-plugin-sqlite-2**. All good.

If I change nothing else but bump my `package.json` pouchdb version to `5.4.x` and then `npm install` and build/run my app on my device I get the error: **Error: Adapter is missing**

I had to dig for the error using Safari Dev tools and found it in the `taskqueue` of my local db object. i.e. 

```javascript
var db = new PouchDB('mydb', {adapter: 'websql', revs_limit: 100, auto_compaction: true});
console.log(db);    // db.taskqueue.failed : 'Error: Adapter is missing'
```

If I swap back to 5.3.2, all is good and app works perfectly.

I noticed there were some [deprecations in 5.4.0](https://github.com/pouchdb/pouchdb/releases/tag/5.4.0) but not sure if these are to blame. I'm more suspicious of bugfix #5197 / PR #5198. I get why this fix was introduced for browser validation, but with cordova + sqlite plugin 2 we can get it to work. Are we artificially disabling websql here for cordova apps too ?

**Edit:** Turns out we are: [valid.js#L56](https://github.com/pouchdb/pouchdb/blob/master/packages/pouchdb-adapter-websql/src/valid.js#L56) assumes a `SQLitePlugin` object is defined before `deviceready`.

But with **cordova-plugin-sqlite-2** the object isn't available until the `deviceready` event fires and it is different casing `sqlitePlugin` i.e. there is no `SQLitePlugin` object ever.

[...]

PR #5423, opened 2016-07-01
(#5417) - proper websql validation for cordova+ios

nolanlawson:
My proposed fix for #5417:
- if `typeof cordova !== 'undefined'`, return `true` from `valid()`
- once `new PouchDB()` is called, do a second `valid()` check if we're in a Cordova app. if so, and if `!valid()`, then show a warning to the user that they need to install a SQLite plugin:

screenshot 2016-07-01 08 34 51

I've tested manually to confirm it works; no real way I can see to add it to our test harness currently.

### DIFF
--- a/packages/pouchdb-adapter-websql/src/index.js
+++ b/packages/pouchdb-adapter-websql/src/index.js
@@ -1,6 +1,7 @@
 import WebSqlPouchCore from 'pouchdb-adapter-websql-core';
 import { extend } from 'js-extend';
-import valid from './valid';
+import { guardedConsole } from 'pouchdb-utils';
+import { valid, validWithoutCheckingCordova } from './validation';
 
 function createOpenDBFunction(opts) {
   return function (name, version, description, size) {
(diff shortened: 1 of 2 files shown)


nickcolley:
Is the assumption here that people will use IDB if they need a faster database?

(I thought in browser WebSQL was pretty fast..?)

nolanlawson:
The assumption here is that if you're trying to use either WebSQL or IndexedDB in WKWebView on Cordova, you're going to have a bad time. We don't support either one; the best escape hatch is the SQLite Plugin.

I don't even mention FruitDOWN because it was mostly a proof of concept. The performance is awful and I've been told that the reliability is not great either. Hopefully Apple will improve their IDB implementation in later releases. (Or maybe iOS10 is good; haven't tested yet.)

nickcolley:
(Sorry for all the questions)

What's the story with Cordova on Android in this case? Will that force them to use the plugin even though WebSQL works?

nolanlawson:
No, Android users will not see this error message. It will work

-----
Nolan Lawson
http://nolanlawson.com

daleharvey:
Yeh this looks good to me, without getting Cordova in CI there isnt much real testing that can be done 👍 thanks

PR #5426, opened 2016-07-02
(#5426) - remove dead code, improve coverage

nolanlawson:
Trying to take some time to clean up technical debt. Places where I added `istanbul ignore`s are where I considered it too difficult or unimportant to test. Places where I removed the `ignore`s are where I added tests, at least in Node.

### DIFF
--- a/packages/pouchdb-core/src/changes.js
+++ b/packages/pouchdb-core/src/changes.js
@@ -96,8 +96,10 @@ function Changes(db, opts, callback) {
 
 
   if (!db.taskqueue.isReady) {
-    db.taskqueue.addTask(function () {
-      if (self.isCancelled) {
+    db.taskqueue.addTask(function (failed) {
+      if (failed) {
+        opts.complete(failed);
+      } else if (self.isCancelled) {
         self.emit('cancel');
       } else {
         self.doChanges(opts);
@@ -203,6 +205,7 @@ Changes.prototype.doChanges = function (opts) {
   opts.limit = opts.limit === 0 ? 1 : opts.limit;
   opts.complete = callback;
   var newPromise = this.db._changes(opts);
+  /* istanbul ignore else */
   if (newPromise && typeof newPromise.cancel === 'function') {
     var cancel = self.cancel;
     self.cancel = getArguments(function (args) {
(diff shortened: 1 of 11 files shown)


packages/pouchdb-core/src/changes.js
```diff
@@ -96,8 +96,10 @@ function Changes(db, opts, callback) {
 
 
   if (!db.taskqueue.isReady) {
-    db.taskqueue.addTask(function () {
-      if (self.isCancelled) {
+    db.taskqueue.addTask(function (failed) {
```
nolanlawson: we weren't actually handling the case where the constructor failed and then the user did `db.changes()`. this is tested now

packages/pouchdb-core/src/constructor.js
```diff
@@ -54,18 +55,17 @@ function PouchDB(name, opts) {
...
   self.prefix = PouchDB.prefix;
 
   if (typeof name !== 'string') {
-    throw(new Error('Missing/invalid DB name'));
+    throw new Error('Missing/invalid DB name');
   }
 
   var prefixedName = (opts.prefix || '') + name;
-  var backend = PouchDB.parseAdapter(prefixedName, opts);
+  var backend = parseAdapter(prefixedName, opts);
```
nolanlawson: moved `parseAdapter` off the main PouchDB object, no reason to keep it there.

packages/pouchdb-core/src/parseAdapter.js
```diff
@@ -0,0 +1,49 @@
+import PouchDB from './constructor';
+import { guardedConsole, hasLocalStorage } from 'pouchdb-utils';
+
+function parseAdapter(name, opts) {
+  var match = name.match(/([a-z\-]*):\/\/(.*)/);
+  if (match) {
+    // the http adapter expects the fully qualified name
+    name = /http(s?)/.test(match[1]) ? match[1] + '://' + match[2] : match[2];
+    return {name: name, adapter: match[1]};
+  }
```
nolanlawson: I removed an "invalid adapter" check here that was already being handled upstream.

daleharvey:
Nice changes, but looks like the `websql` test is a valid failure

nolanlawson:
Yup, it was a test error due to `sqlite3` throwing an error when trying to open a DB on an invalid path. The test didn't make much sense for `node-websql` (and doesn't seem fixable without a PR to `sqlite3`), so I just skipped it for `node-websql`. Passing now.

daleharvey:
Awesome, thanks

PR #5429, opened 2016-07-02
(#5429) - WIP - leveldb: simplify stored metadata json

nolanlawson:
This is a perf improvement designed to reduce the amount of space metadata documents take up on disk, for the `leveldb-core` adapter.

It also has a speed benefit due to less time spent serializing/deserializing JSON. To prove it, I ran the temp-views test with 50 iterations 3 times (`PERF=1 GREP=temp-view npm t`), and came up with these numbers:

| Run | before | after | improvement |
| ---- | --- | ---- | ---- |
| 1 | 107868ms | 104111ms | 3.48% |
| 2 | 112161ms | 109656ms | 2.23% |
| 3 | 113830ms | 109341ms | 3.94% |

The basic idea is that for the most common class of documents &ndash; those with one non-deleted first-generation revision &ndash; we can simplify the structure on-disk and just store the id and rev. For more details, see the comments in  `metadataCompression.js`.

I would like to port this improvement to the idb/websql adapters as well, but for starters I'm just doing leveldb.

### DIFF
--- a/packages/pouchdb-adapter-leveldb-core/src/index.js
+++ b/packages/pouchdb-adapter-leveldb-core/src/index.js
@@ -54,6 +54,11 @@ import {
   createError
 } from 'pouchdb-errors';
 
+import {
+  inflateMetadata,
+  deflateMetadata
+} from './metadataCompression';
+
 var DOC_STORE = 'document-store';
 var BY_SEQ_STORE = 'by-sequence';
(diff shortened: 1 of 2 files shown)


nolanlawson:
Marking this as WIP because I want to spend more time thinking about the impact of this in #5433. Changing the way we store data on disk is something irreversible.

PR #5432, opened 2016-07-03
(#4967) - fix map function called twice

nolanlawson:


nolanlawson:
mistake, please ignore

PR #5433, opened 2016-07-03
(#5429) - WIP - simplify metadata json for idb/websql

nolanlawson:
Same change as #5429, but for idb and websql as well.

Needs performance numbers before I can feel confident with this, but opening up a PR to make sure it passes.


daleharvey:
So I had written but not posted this on the other PR

So in general the strategy of performance improvements by adding more things worries me, partly because in my experience large performance improvements come by doing a lot less, and doing more things gets in the way of that, also because although less a worry in node we are trading off performance for bundle size, an extra 100 lines is not a lot but we have a lot of extra 100 lines around.

If we see a huge performance boost then this could be worth it, but a 2-4% improvement on a best case benchmark may possibly not be

nolanlawson:
Yep, that's why I'm reconsidering this thing. Also it locks us in to a new storage format on-disk that we would have to deal with forever.

My gut tells me that we could do better by having a more efficient metadata format; I think the fact that I even got 2-4% by just removing most of the tree structure for a subset of documents shows that we could have bigger wins by changing the tree structure for all documents.

However, I'd also like to run more tests and ensure we didn't improve in one area only to regress in another.

nolanlawson:
I closed #5429, gonna close this as well. It ekes out some tiny perf improvements but complicates the code. Not worth it.

PR #5434, opened 2016-07-03
(#XXXX) - Skip a GET in http setup and remove skipSetup

daleharvey:


daleharvey:
So enough people find the 404 on getting the database when we open a http pouchdb confusing that we added a warning to say its all ok. This removes the need for that, the PUT will fail with a 412 if it already exists, which is fine

daleharvey:
So this still gives a 412 error which isnt much of an improvement, its less code but could have unexpected effects so gonna wait on this

PR #5435, opened 2016-07-03
(#XXX) - Remove getHost

daleharvey:


nickcolley:
ping @yaronyg

nolanlawson:
`getUrl()` is currently used by `pouchdb-authentication` ([source](https://github.com/nolanlawson/pouchdb-authentication/blob/ee445d8381bc12eb828faaa46457429aa9c8fc9f/lib/utils.js#L7)). I'll see if there's some standard API I can use instead of this.

nolanlawson:
There appears to be no way to get the original database name from the `db` object. I think this PR should be reconsidered or we should standardize an API. `pouchdb-authentication` for better or worse is the most popular plugin after `pouchdb-find`, so breaking it would be a big deal.

nolanlawson:
Probably the easiest thing to do would be to just attach `name` to `db` and make that a standard API. `name` could just be whatever was originally passed in either through the `opts` or as the sole argument to `new PouchDB()`.

daleharvey:
So I would like to reduce the holes we need in the api for plugins (hence this PR), however we already set `_db_name` so I think `db.name` as a single exception for a fairly important property seems reasonable.

nolanlawson:
Looks perfect to me! Sorry for the delay in merging.

PR #5438, opened 2016-07-04
(#XXX) - Remove MAX_URL logic

daleharvey:
Just noticed this while doing other work, a chance to clean things up a little, our cache handling should be fixed at the point, so have less code paths seems better if one is going to work every time


nolanlawson:
Yeah, the Safari 304 POST issue should be fixed by the `_nonce`, so this change should be unnecessary. Only other reason I can think to have it would be support for Couch <1.6.0, but we can just change our docs to say that we don't support any version of Couch older than that.

+1 when green.

PR #5441, opened 2016-07-05
release memory faster in ajax requests

olexme:
The main issue was that postponed (by setTimeout) 'timeoutReq' function was very often (half of requests) not released immediately after the request was completed but after 10 or 30 seconds (default timeout). On iPhone 6 it caused big memory consumption and very often crashes during sync with db with many inline attachments. This fix makes the sync process uses less memory and releases it faster. Exemplary project to observe the issue https://github.com/olexme/pouchdb-cordova.

### DIFF
--- a/packages/pouchdb-ajax/src/request-browser.js
+++ b/packages/pouchdb-ajax/src/request-browser.js
@@ -133,11 +133,16 @@ function xhRequest(options, callback) {
     }
   };
 
-  var cleanUp= function () {
+  var cleanUp = function () {
     doClearTimeout();
     ret.abort = function () {};
+    xhr.onprogress = undefined;
+    if (xhr.upload) {
+      xhr.upload.onprogress = undefined;
+    }
+    xhr.onreadystatechange = undefined;
     xhr = undefined;
-  }
+  };
 
   if (options.xhr) {
     xhr = new options.xhr();
--- a/packages/pouchdb-ajax/src/request-browser.js
+++ b/packages/pouchdb-ajax/src/request-browser.js
@@ -117,6 +117,28 @@ function xhRequest(options, callback) {
     xhr.abort();
   };
 
+  var ret = {abort: abortReq};
+
+  var doSetTimeout = function () {
+    if (timer !== undefined) {
+      doClearTimeout();
+    }
+    timer = setTimeout(timeoutReq, options.timeout);
+  };
+
+  var doClearTimeout = function () {
+    if (timer !== undefined) {
+      clearTimeout(timer);
+      timer = undefined;
+    }
+  };
+
+  var cleanUp= function () {
+    doClearTimeout();
+    ret.abort = function () {};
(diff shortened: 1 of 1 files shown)


packages/pouchdb-ajax/src/request-browser.js
```diff
@@ -126,6 +153,7 @@ function xhRequest(options, callback) {
   try {
     xhr.open(options.method, options.url);
   } catch (exception) {
+    cleanUp();
```
daleharvey: We shouldnt need to do any cleanup here, the `setTimeout` has not been called yet

packages/pouchdb-ajax/src/request-browser.js
```diff
@@ -117,6 +117,33 @@ function xhRequest(options, callback) {
     xhr.abort();
   };
 
+  var ret = {abort: abortReq};
+
+  var doSetTimeout = function () {
+    if (timer !== undefined) {
```
daleharvey: I dont think we need these `undefined` checks, we can unconditionally clear and set the timeout, `clearTimeout(undefined)` doesnt do anything bad

daleharvey:
Good catch, I think there may be some broken assumptions around that code. I commented on a few parts of the patch but in general I think it may have a bit much defensive code in there. 

The main problem I can see that may be possible is that if an xhr triggers `onreadystatechange` without also firing `onprogress` the timer can get left open, that seems like it should be fixed if we only do a `clearTimeout(timer)` inside `onreadystatechange` (after the readyState check)

The nulling out the properties / handlers should not be needed, either the xhr instance gets collected or it doesnt, we should ensure that it does, figuring out a test for this that doesnt require a full cordova setup would be ideal although if there is a simple fix that works for you we could merge that. Cheers

olexme:
Thanks for your review. I removed that defensive code regarding timeouts as well as unnecessary cleanup on exception.

Regarding nulling out properties and handlers:
- IMO replacing ret.abort with NOP function matters here cause abortReq keeps the reference to xhr and indirect to callback. Of course client could keep abortReq reference directly (not through returned object) but looks like it’s not the case here.

- other handlers nulling, in theory you should be right, but I did a lot of tests and GC on iOS just seems to work better with them, at least no crash with quite heavy load. E.g. in my recent test case I do sync the DB with 200 docs, each doc with 1.2MB inline attachment. Sync settings pull: {“batches_limit”: 1, “batch_size”: 20}; Setting batch number = 2 causes crash so there is still place for improvements .

I’ve added also the cleanUp to the abortReq function (setTimeout still can hang) and to the timeoutReq function (to release abort handler)
Thanks again

daleharvey:
So this looks cleaner thanks. I think we would be ok to merge this based on your local testing however should be warned without regression testing there is a fair likelyhood that this may get regressed at some point. If you can think of a way to test this in a normal browser it would be a great follow up, thanks

Issue #5379, opened 2016-06-20
v5.1.0 regression: Repeatedly checking for remote checkpoint.


gillez:
### Environment (Node.js/browser/hybrid app/etc.)

Browser running PouchDB with Couchbase Server via Couch Sync Gateway

### Browser/platform (Chrome/FF/Safari/Edge/iOS/Android/etc.)

Chrome and Safari

### Adapter (IndexedDB/WebSQL/LevelDB/etc.)

default

### Server (CouchDB/Cloudant/Couchbase/PouchDB Server/etc.)

Couchbase

### Please describe your issue. Test cases (e.g. JSBin) are very helpful!

When replicating a Couchbase database using PouchDB v5.0.0 and below, I get the following messages a couple of times:

pouchdb-5.0.0.min.js:9 GET https://localhost:4984/mydb/_local/LUNtUWldUuw9xS_mFgfzdg%3D%3D?_nonce=1466437383136 404 (Not Found) 
The above 404 is totally normal. PouchDB is just checking if a remote checkpoint exists.

As the message says, this is normal when initially replicating a database.

However, on PouchDB v5.1.0 and above, this message is repeated continuously every time a sync is attempted, i.e. every couple of seconds.

I believe this is a regression, but if it is deliberate behaviour, please let me know.

Thanks,
Giles

daleharvey:
This was a specific change https://github.com/pouchdb/pouchdb/issues/5145, before `5.1.0` we didnt update the checkpoint if there were no changes to process however that meant reprocessing a potentially large amount of documents repeatedly. 

However a possible improvement would be to check that the checkpoint actually needs updating before doing so

gillez:
Hi Dale, thanks for the explanation.  Do you think the possible improvement that you suggest is easy to do and would it cause any issues?

daleharvey:
It should be reasonably easy to implement that improvement, you can see in the previous patch where the code would need edited, mostly https://github.com/pouchdb/pouchdb/commit/f1989ce6361193f1e213cb0b17e7dab4f2a8451c#diff-1b1f318b599e364220cdb8ac006c238bR307 would need to check the last_seq vs the previously stored one

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/0d384e848f0ec5238c2f816393b5535ae2bb0d6e

PR #5443, opened 2016-07-06
(#5379) do not update checkpoint if last_seq has not changed

willholley:
Now that we checkpoint on every batch of changes (rather than when replication is complete), avoid updating the checkpoint when the source database has not changed. As a first pass, we determine
this based on whether _changes/last_seq progresses, though this won't work if the source database is CouchDB 2.0 or Cloudant where sequence values may change even if the underlying data doesn't. I can't think of a way around this, but it's also an edge case - it just means that a PouchDB database might grow if compaction isn't run.

### DIFF
--- a/packages/pouchdb-replication/src/replicate.js
+++ b/packages/pouchdb-replication/src/replicate.js
@@ -308,7 +308,8 @@ function replicate(src, target, opts, returnValue, result) {
       };
 
       // update the checkpoint so we start from the right seq next time
-      if (!currentBatch && changes.results.length === 0) {
+      if (!currentBatch && changes.results.length === 0 &&
+        changes.last_seq !== changesOpts.since) {
         writingCheckpoint = true;
         checkpointer.writeCheckpoint(changes.last_seq,
             session).then(function () {
--- a/tests/integration/test.replication.js
+++ b/tests/integration/test.replication.js
@@ -1593,6 +1593,37 @@ adapters.forEach(function (adapters) {
       });
     });
 
+    it('Does not update checkpoint unncessarily (#5379)', function () {
+      var db = new PouchDB(dbs.name);
+      var remote = new PouchDB(dbs.remote);
+      var bulkDocs = remote.bulkDocs;
+      var bulkDocsCalled = false;
+      remote.bulkDocs = function () {
+        bulkDocsCalled = true;
+        return bulkDocs.apply(this, arguments);
+      };
+      return remote.bulkDocs({ docs: docs }).then(function () {
+        return db.replicate.from(remote);
+      }).then(function (result) {
+        result.ok.should.equal(true);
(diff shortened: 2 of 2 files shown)


willholley:
closing this - needs a different approach..

gillez:
Hi Will, I'm using Couchbase of army source database, not CloudDB or Cloudant.  Do you think I can take this commit locally without causing any further issues, at least until it is fixed properly (via a different approach)?
I would like to move to the latest PouchDB, and I am holding off due to seeing the repeated 404 errors.  If I can just patch the latest version with this commit, that would probably work for me.

Thanks,
Giles

willholley:
@gillez - you should be good to use the solution I submitted in #5444

willholley:
@gillez I forgot to say, I don't understand how this fix will help with repeated 404 errors. What resource the 404s in response to?

gillez:
I hadn't noticed #5444.  I will use that one, or watch out for a PouchDB release with that commit.

The 404's appear repeatedly when syncing to the CouchBase database via couch-sync-gateway.  I'm not sure what resource they're a response to.  The message says "pouchdb-5.1.0.min.js:9 GET https://localhost:4984/mydb/_local/... 404 (Not Found)" followed by a message saying the above error is normal. (full error listed on #5379).  The check may be normal but not when it happens repeatedly, every two seconds....

If you like, I can try with the proper non-min library, so that we can find out what line the error is coming from.

willholley:
@gillez it sounds like checkpoints are not being written correctly to Couchbase Sync Gateway. This could be a permissions issue or possibly CSG doesn't allow _local documents to be written via _bulk_docs (I have a vague recollection that this may be the case - maybe @snej can confirm).

snej:
Um, yes — the possibility had never occurred to me, to be honest O_o Probably because local and regular documents are implemented so differently.

Could you work around this in PouchDB by updating the checkpoint in a separate PUT?
Otherwise (or also), you can [file an issue against Sync Gateway](https://github.com/couchbase/sync_gateway/issues/new), but as we're just freezing the 1.3 release, any fix wouldn't show up for a few months at least.

nolanlawson:
We made that _bulk_docs change months ago. I wasn't aware it broke anything in CSG. We made that change in order to simplify the http adapter code.

In general our policy with PouchDB has been to support CouchDB as a first-class target; with every other server we consider it their responsibility to maintain compatibility with us, not vice versa. (This even applies to PouchDB Server, which has been broken for months.) Unfortunately this means that every db but CouchDB 1.x is in the "allowed failures" and thus breakages inevitably happen.

I think we should work to remove them from the allowed failures. Either that or we should just own up in our docs that we only support CouchDB. Thoughts @daleharvey ?

snej:
It would probably be pretty easy to make SG compatible with this.

PR #5444, opened 2016-07-06
(#5379) - do not update checkpoint if last_seq has not changed

willholley:
Avoid updating the checkpoint when there is no change to the sequence
number. This prevents unnecessary updates to the checkpoint document,
generating traffic and compaction bloat.

This approach doesn't help if the source database is CouchDB 2.0 or
Cloudant because sequence values may change even if the underlying
database doesn't. I can't think of a way around this, but it's also an
edge case - it just means that a PouchDB database might grow if
compaction isn't run.

### DIFF
--- a/packages/pouchdb-checkpointer/src/index.js
+++ b/packages/pouchdb-checkpointer/src/index.js
@@ -35,6 +35,12 @@ function updateCheckpoint(db, id, checkpoint, session, returnValue) {
     if (returnValue.cancelled) {
       return;
     }
+
+    // if the checkpoint has not changed, do not update
+    if(doc.last_seq === checkpoint) {
+      return;
+    }
+
     // Filter out current entry for this replication
     doc.history = (doc.history || []).filter(function (item) {
       return item.session_id !== session;
--- a/tests/integration/test.replication.js
+++ b/tests/integration/test.replication.js
@@ -1593,6 +1593,37 @@ adapters.forEach(function (adapters) {
       });
     });
 
+    it('Does not update checkpoint unncessarily (#5379)', function () {
+      var db = new PouchDB(dbs.name);
+      var remote = new PouchDB(dbs.remote);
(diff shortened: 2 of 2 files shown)


kxepal:
You can track the numeric part of the sequence: if it remains the same as the last_seq has one, much likely there were indeed no updates.

willholley:
@kxepal good thought. It's not the end of the world if we don't checkpoint and it's a real edge case for the numerical component to not change when the database has.

My only hesitation is that we're then depending on an implementation detail of CouchDB - technically, the sequence value could be any JSON object.

I could be persuaded either way...

kxepal:
@willholley True that this is implementation details hack that will cause issues with else and in future. On second thought, it's strange that 2.0 cannot give a stable value for case when no changes happens.

willholley:
@kxepal I think it's because the order of the individual shard/seq values used to construct the database-level sequence value is not stable.

packages/pouchdb-checkpointer/src/index.js
```diff
@@ -35,6 +35,12 @@ function updateCheckpoint(db, id, checkpoint, session, returnValue) {
     if (returnValue.cancelled) {
       return;
     }
+
+    // if the checkpoint has not changed, do not update
+    if(doc.last_seq === checkpoint) {
```
daleharvey: Tiny nit, we put a space here, gonna check why eslint isnt catching it but ill just fix on the way in

daleharvey:
So yeh, I agree that if the seq values are considered opaque I would rather avoid putting implementation details into the replicator, if we could look into getting couchdb 2.0 seq's stable with no changes that would be a much better fix

daleharvey:
As for this patch, 👍 looks great, cheers

daleharvey:
Merged in https://github.com/pouchdb/pouchdb/commit/0d384e848f0ec5238c2f816393b5535ae2bb0d6e

PR #5445, opened 2016-07-06
(#5207) - build pouchdb-next.js as part of build process

nolanlawson:
Prompted by https://github.com/pouchdb/pouchdb/issues/5207#issuecomment-230528015, you can now build `pouchdb-next.js` as part of the build process. To simplify the dev process, you can do:

    NEXT=1 npm run dev

and it will test the "next" adapter. This ought to also work for the automated tests, but I haven't checked yet.

Note that if/when we publish, we will also publish `pouchdb-next.js` to npm, but this is low-risk I think. Maybe it's even better to get a beta out there early.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -59,6 +59,7 @@ or
 #### Other test options
 
 * `SKIP_MIGRATION=1` should be used to skip the migration tests.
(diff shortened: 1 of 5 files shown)


daleharvey:
Awesome thank you, 👍

PR #5446, opened 2016-07-06
Add emit parameter in map function examples

joakin:
Hi!

I was reading the docs on the site and noticed that the `emit` function is magically used after the first example (https://pouchdb.com/guides/queries.html#mappin-and-reducin), so I'm assuming it should be passed as the second parameter of the `map` function too in all the other examples.

First time contributing to the project, I've read `CONTRIBUTING.md` and checked other commits, if I'm doing something wrong please tell me and I'll fix it. Couldn't see if I should use a *Docs:* prefix in the commit title or something like that.

Thanks for maintaining this awesome project ❤️

### DIFF
--- a/docs/_guides/queries.md
+++ b/docs/_guides/queries.md
@@ -54,7 +54,7 @@ var ddoc = {
   _id: '_design/my_index',
   views: {
     by_name: {
-      map: function (doc) { emit(doc.name); }.toString()
+      map: function (doc, emit) { emit(doc.name); }.toString()
     }
   }
 };
@@ -137,7 +137,7 @@ In CouchDB, queries are called _map/reduce functions_.  This is because, like mo
 It may sound daunting at first, but in the simplest (and most common) case, you only need the _map_ function.  A basic map function might look like this:
 
 ```js
-function myMapFunction(doc) {
+function myMapFunction(doc, emit) {
   emit(doc.name);
 }
 ```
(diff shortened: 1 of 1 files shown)


daleharvey:
So I am not sure we should be doing this, as far as I remember fun(doc, emit) is a pouchdb only convention that we needed for some special case, however `fun(doc) { emit` will work across pouchdb + couchdb and is the recommended way to do things.

Thanks for the PR, the changes themselves look great, we dont use a `Docs` tag but do usually reference issue numbers in commits, but if a PR doesnt have a related issue we just add it on the way in so no bother. However I dont think these are the right changes, will leave open for a little in case I am wrong and someone else wants to chime in.

joakin:
Oh, I see. So [only for temp views](https://github.com/pouchdb/pouchdb/blob/master/packages/pouchdb-mapreduce/src/index.js#L503-L504) the `emit` param is passed in, but for the persistent ones the mapper functions are [stringified and evaled](https://github.com/pouchdb/pouchdb/blob/master/packages/pouchdb-mapreduce/src/index.js#L510-L511), passing in [variables in scope, like emit](https://github.com/pouchdb/pouchdb/blob/master/packages/pouchdb-mapreduce/src/evalfunc.js), right?

---

Would it be interesting to flesh out the [`.toString` warning](https://github.com/pouchdb/pouchdb/blame/3e00f1e34c4ea6524cd71d10c3d839ceb7f1a106/docs/_guides/queries.md#L69-L74) to explain why emit is missing in the map/reduce functions? Something like this?

```html
The <code>.toString()</code> at the end of the map function is necessary to prep the
object for becoming valid JSON.

<code>emit</code> will be available in scope when the function is run, so don't pass 
it in as a parameter of the map function.
```

If so I'll close this and open a new PR with that change. I think it would be interesting to mention at some point where the emit fn is coming from and why is not included anywhere.

daleharvey:
Yup exactly, and I think clarification in the docs to that end would be awesome if you wanted to open a new PR for that, thanks :)

PR #5449, opened 2016-07-07
Add info about the emit fn in map functions

joakin:
Clarify why the emit function is used in the examples that follow "persistent
queries" without being passed in as a parameter in the map function.

As discussed in https://github.com/pouchdb/pouchdb/pull/5446


daleharvey:
This looks great, thanks a lot 👍  :)

PR #5450, opened 2016-07-07
(#XXX) - Don't leak change listeners

anowak:
When listener was removed, it still was kept in the internal array. I
don't think there is a sane way of testing this.


daleharvey:
Yeh I can see how this is hard to test, its clear enough and a small patch so 👍, thanks

Issue #5452, opened 2016-07-08
An apparently unnecessary throw err


yaronyg:
The following [throw err](https://github.com/pouchdb/pouchdb/blob/0d384e848f0ec5238c2f816393b5535ae2bb0d6e/packages/pouchdb-checkpointer/src/index.js#L107) causes an unhandled promise exception. But the throw seems unnecessary. The catch handler will return and all the promises up the stack will know they are in 'catch' and will handle accordingly. So I think we can just delete that line. And yes, I do have a simple test for this where I try to replicate with a server that doesn't exist and with this line I get an unhandled promise and without this line everything is happy.

yaronyg:
Removing the throw causes the test '#3999-1 should not start over if last_seq mismatches' to fail. I'm investigating.

yaronyg:
And of course I'm wrong. Because, well, that's my skill in life. Now it looks like the actual problem is [here](https://github.com/pouchdb/pouchdb/blob/176808878a116589ec68bb6476a0c75319e92c54/packages/pouchdb-replication/src/replicate.js#L421) but given the way this day has been going I'm not making any bets. More to come.

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/b191d88448a6c54cfee976c770da5315ab7d0709

PR #5453, opened 2016-07-08
(#5452) - Removing unnecessary err

yaronyg:
#5452 

The test will trigger an onCheckpointError that will cause an unhandled
promise exception if the throw is included and works just fine if it is
removed.


daleharvey:
Good job, ran the test with the throw and it does indeed fail, thanks 👍

PR #5455, opened 2016-07-08
(#5443) - use PUT when checkpointing to remote dbs

willholley:
Allow PUT to be used for document writes if an adapter supports it.
Some backends (Couchbase Sync Gateway, for instance) do not support writing _local documents via bulk_docs.

I appreciate this adds a little complexity which we previously removed, but it seems the pragmatic fix to ensure maximum compatibility.

### DIFF
--- a/tests/component/test.read_only_replication.js
+++ b/tests/component/test.read_only_replication.js
@@ -16,6 +16,9 @@ function reject(req, res, next) {
   if (req.body.docs) {
     replicationDoc = req.body.docs[0];
     res.status(403).send({error: true, message: 'Unauthorized'});
+  } else if (req.body._id) {
+    replicationDoc = req.body;
+    res.status(403).send({error: true, message: 'Unauthorized'});
   } else {
     next();
   }
--- a/packages/node_modules/pouchdb-adapter-http/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-http/src/index.js
@@ -176,6 +176,7 @@ function HttpPouch(opts, callback) {
   function ajaxPromise(userOpts, opts) {
     return new Promise(function (resolve, reject) {
       ajax(userOpts, opts, function (err, res) {
+        /* istanbul ignore if */
         if (err) {
           return reject(err);
         }
--- a/packages/node_modules/pouchdb-adapter-http/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-http/src/index.js
@@ -633,6 +633,27 @@ function HttpPouch(opts, callback) {
     }).catch(callback);
   };
 
+
+  // Update/create document
+  api._put = function (doc, opts, callback) {
+    setup().then(function () {
+      return preprocessAttachments(doc);
+    }).then(function () {
+      // Update/create the document
+      ajax(opts, {
+        method: 'PUT',
+        url: genDBUrl(host, doc._id),
+        body: doc
+      }, function (err, result) {
+        if (err) {
+          return callback(err);
+        }
+        callback(null, result);
+      });
+    }).catch(callback);
+  };
+
+
   // Get a listing of the documents in the database given
   // by host and ordered by increasing id.
   api.allDocs = adapterFun('allDocs', function (opts, callback) {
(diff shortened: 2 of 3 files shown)


daleharvey:
So personally I am 👎 on this, if couchbase doesnt support writing `_local` via bulkdocs that seems like it should be a bug fixed on their side, If there was work on full support of csg and this was given alongside justifications for the lack of server support and tests that showed it fixed something then I think we could live with some client workarounds, however as far as the current tests show replication is working fine without this patch (I am fairly surprised it is still green) and there hasnt been much contributor support in sync-gateway thus far (jchris was the only one working on csg support and has left couchbase) so I would prefer to avoid adding workarounds on the client side

willholley:
I basically agree with @daleharvey so if the verdict is to close the PR without merging, I'm fine with that. It was a small amount of code to fix the user's issue - perhaps they can live with manually patching PouchDB until CSG addresses the issue server-side.

daleharvey:
ok cool if you agree will close out, cheers

willholley:
re-opening in light of #5516

packages/pouchdb-core/src/adapter.js
```diff
@@ -222,7 +222,12 @@ AbstractPouchDB.prototype.put = adapterFun('put', function (doc, opts, cb) {
       return this._putLocal(doc, cb);
     }
   }
-  this.bulkDocs({docs: [doc]}, opts, yankError(cb));
+  if (typeof this._put === 'function' && !!opts.new_edits) {
+    this._put(doc, opts, cb);
+  }
+  else {
```
daleharvey: nit: same line as previous

    } else {

daleharvey:
Yeh given the above bug, would like to see this fixed in couchdb but for now 👍 cuold you rebase and we can see what the tests are looking like now? cheers

willholley:
rebased - the tests pass but I have to restart all the flakey jobs...

nolanlawson:
Rebased, Firefox was having a rough week on Travis but should be good now.

nolanlawson:
Failing due only to code coverage dropping below 100%. You may need to throw some `/* istanbul ignore */`s in there.

nolanlawson:
I was going to rebased with the `istanbul ignore` fixes, but it looks like this line is totally unused:

```js
if (typeof this._put === 'function' && !!opts.new_edits) {
    this._put(doc, opts, cb);
}
```

Correct me if I'm wrong @willholley but doesn't this mean that your fix isn't actually run? Seems it should at least run for the http adapter.

nolanlawson:
Hm, trying to debug this, but `npm run test-component` fails on the master branch for me. Not sure why it's passing in Travis...

willholley:
@nolanlawson the intent there is that this code should run for adapters that have a `_put` function defined, when `new_edits=false` is not specified (since PUT does not support it). The http adapter is the only where we require the override today, and the only one that implements `_put`. I wonder if we could have a test with a fake adapter/mock...

daleharvey:
We shouldnt need to mock anything, the http adapter is used in the node test suite and we have tests that perform checkpointing so it should be run, can take a look but yeh we cant istanbul ignore that line, need to figure why its not being hit

tests/component/test.read_only_replication.js
```diff
@@ -16,6 +16,10 @@ function reject(req, res, next) {
   if (req.body.docs) {
     replicationDoc = req.body.docs[0];
     res.status(403).send({error: true, message: 'Unauthorized'});
+  }
```
daleharvey: nit: same line

willholley:
The styling issues are fixed and the tests pass. Coverage seems to fail on Travis, though. Anything left to do here?

nolanlawson:
If coverage is failing, then I suppose we need more `/* istanbul ignore */`s, assuming it's not possible to actually have the tests run through the ignored code.

willholley:
hmm - coverage is ok locally but fails with a seg fault on Travis (https://travis-ci.org/pouchdb/pouchdb/jobs/154678193).

daleharvey:
Yeh the coverage tests have been pretty intermittent recently, worth looking into, @willholley if there is a failure on travis then you should be able to restart it, stuff like this cant merge until green but I have given it a restart and 👍 when all green

daleharvey:
woot full green :)

nolanlawson:
@willholley Segfaults have been common w/ Istanbul since as far as I can remember... best bet is to restart.

Issue #5114, opened 2016-05-01
Uncaught ETIMEDOUT exceptions


SukantGujar:
During a local Pouch DB (5.3.2) from remote DB (CouchDB 1.6.1) replication call, I am observing repeated exceptions -
![capture](https://cloud.githubusercontent.com/assets/6153264/14940310/2158a7a8-0f8f-11e6-8242-2500f104052f.PNG)

The live replication call code is bare minimum -

`localDB.replicate.from(remoteURL, { live: true, retry:true })`

Is this expected? If so, could PochDB handle the long polling timeouts in a manner which doesn't end up with uncaught exceptions?

nolanlawson:
Retry replication is supposed to gracefully handle timeouts. If it's throwing an exception and halting replication, then that's a bug.

SukantGujar:
The exception is reported by the browser as uncaught and then bubbles up as err in the paused event.
On May 2, 2016 6:27 AM, Nolan Lawson <notifications@github.com> wrote:Retry replication is supposed to gracefully handle timeouts. If it's throwing an exception and halting replication, then that's a bug.

nolanlawson:
Can you provide a test case to reproduce?

SukantGujar:
I am not doing anything specific. Its a plain CouchDB setup in Windows 10 as the backend. From PouchDB client in Chrome, I am trying to replicate a remoteDB into a local instance. As you will notice in the screenshot, the exception is reporting after a _changes longpoll call is waited on till it times out. If you need, I can provide the Fiddler trace too.

nolanlawson:
A test case would still be very helpful. I've never seen this error before, so it might be something related to CouchDB's Windows 10 build.

SukantGujar:
Sure. I am assuming you are referring to a JS test case. Which test framework is preferable? I can try to rig something up in Jasmine. However, if this turns out to be reproduce-able on specific machines only, I am willing to share a screen recording with logs as well. If needed, I can do a live demo too.
Thanks.
 

    On Tuesday, 3 May 2016 9:20 PM, Nolan Lawson <notifications@github.com> wrote:
 
 

 A test case would still be very helpful. I've never seen this error before, so it might be something related to CouchDB's Windows 10 build.—

nolanlawson:
A live web site would probably be best, but I can understand if that's hard to set up. Otherwise instructions for a Windows VM or something would work too. Else if you can fix the error yourself and open a PR that would be even better. :D

[...]

PR #5457, opened 2016-07-08
(#5114) - Remove unneeded code

yaronyg:
The removed code doesn't appear to actually do anything. It tries to pull
in info from the remote DB but it's unclear why and the information doesn't
appear to be used to do anything.


daleharvey:
That seq comparison also looks suspect, removing code and still passing tests are my favourite patches so hoping this is green, cheers and good catch

daleharvey:
Woot all green :)

PR #5465, opened 2016-07-12
(#5207) - Binary fixes and skip .query tests

daleharvey:


daleharvey:
Oops didnt check eslint, will fixup

PR #5468, opened 2016-07-12
(#5207) - Binary fixes and skip .query tests

daleharvey:


daleharvey:
Almost passing, its pretty much just the auto_compaction tests that arent valid that are failing

nolanlawson:
So remind me - is the plan to just do auto-compaction by default? If so, we should be able to just manually put `AUTO_COMPACTION=1` or `AUTO_COMPACTION=0` and have it work.

daleharvey:
I havent changed the default, I would like to but not specific to this adapter, the tests that fail are the ones that depend on the internal implementation of md5 collisions, updated the PR and this should be fully passing now although seen some intermittents

daleharvey:
So whether we change the default of auto_compaction it can be done independently of this adapter, this uses the same defaults. 

The auto_compaction tests that are failing are the ones that deal with md5 collisions that depend on the internal implementation of attachments which dont hold to this adapter (it doesnt use md5 keyed attachments across documents)

tests/integration/test.local_docs.js
```diff
@@ -86,17 +86,14 @@ adapters.forEach(function (adapter) {
       var db = new PouchDB(dbs.name);
       var doc = {_id: '_local/quux'};
       return db.put(doc).then(function (res) {
-        res.rev.should.equal('0-1');
```
daleharvey: This is the only test I am worried about, I dont like depending on `NEXT` etc too much, and right now the `indexeddb` adapter doesnt copy the revision number generation that current pouch / couch does although it behaves the same way otherwise

nolanlawson:
Interesting, why is the revision number generation different? And the md5 generation?

daleharvey:
The revision number is different because `idb` has a special path for `_local` docs that manually handles revision generation but `indexeddb` uses the same code path as normal docs but forces a `revs_limit: 1` so it doesnt have any of the overhead of revision handling (https://github.com/pouchdb/pouchdb/blob/master/packages/pouchdb-adapter-indexeddb/src/bulkDocs.js#L33). 

As for md5 generation, `idb` has the seperate attachments table that is keyed by the md5 of the attachment, so it behaves differently from couchdb and `indexeddb` when it runs into md5 collisions across documents. 'indexeddb` stores the attachments along with the document so doesnt run into this problem

daleharvey:
Not sure I trust this being green, I hadnt tested in firefox, its unstable for me in chrome and I cant see `NEXT` in the url

nolanlawson:
Thanks, that makes sense. For the `_local` ID generation, it really doesn't matter to be exactly consistent with CouchDB because their own system is flawed IMO (e.g. deleting a document and then recreating it sets the rev back to `1-x`, which causes problems with the ETag; your system would avoid this problem). As for the md5 sum collision stuff, yeah, this puts us in line with CouchDB rather than with our previous behavior, however it's worth having the conversation about whether the extra CPU overhead of collision detection is worth the savings on-disk. (I honestly don't know; it vastly depends on the user's use case.)

daleharvey:
Yup, considered whether this would be a regression, but I think its one of those situations where currently everyone pays the cost of complexity (joined table, collisions etc) to optimise for a use case that is extremely rare (identical attachments in different docs) and that people with that use case could work around it themselves

nolanlawson:
Ah, so for the case of identical attachments in the same doc (but across different versions of the doc), `idb-next` will still collapse them? If that's so, then yeah, it really covers 99% of the bases.

daleharvey:
Yup we collapse attachments inside the same doc, wanted to avoid duplicating across revisions.

daleharvey:
Yup we collapse attachments inside the same doc, wanted to avoid duplicating across revisions.

packages/pouchdb-core/src/adapter.js
```diff
@@ -620,23 +619,22 @@ AbstractPouchDB.prototype.get =
...
+        return cb(null, doc);
       }
       Object.keys(attachments).forEach(function (key) {
         this._getAttachment(doc._id, key, attachments[key], {
           // Previously the revision handling was done in adapter.js
           // getAttachment, however since idb-next doesnt we need to
           // pass the rev through
           rev: doc._rev,
-          binary: opts.binary,
-          ctx: ctx
```
daleharvey: We attempted to pass `ctx` here, however the transaction cant survive, there is a `forEach` above that the transaction may survive over (forgot the exact rules), however https://github.com/pouchdb/pouchdb/blob/master/packages/pouchdb-adapter-idb/src/index.js#L381 will definitely break it, it chrome but firefox is stricter about that and timing will effect it, depending on it is certainly broken

daleharvey:
Actually figured out a way to do this without affecting the existing adapters, I still think this is wrong but would rather not change whats currently green on the current adapters

daleharvey:
Ok in the new version the only changes to the common code (adapter.js) is just refactoring, (changing `callback` to `cb` lets us avoid awkward line breaks). With this we are 100% in chrome and firefox although still with too many intermittents (mostly in chrome), working on those now. but this is good to merge

packages/pouchdb-core/src/adapter.js
nolanlawson: If we can't do a full `get()` inside of a single transaction then that may be a bug; it could introduce race conditions if the transitions end up being interleaved. Based on your statement, I'm confused why the tests are even passing currently in Firefox if this is broken.

Ideally I think we would do this in a single transaction and then do `readBlobData()` after the transaction has ended.

nolanlawson:
OK, I'll give this a second look then.

BTW about [passing in `ctx` for reading blobs](https://github.com/pouchdb/pouchdb/blob/master/packages/pouchdb-adapter-idb/src/index.js#L381), I think the reason this works is because we pass in `binary: true`, meaning that `readBlobData` just [passes the data back unchanged and calls the callback](https://github.com/pouchdb/pouchdb/blob/bf9ff55722b5ab6c756a9fb5ea6b8b33dae08958/packages/pouchdb-adapter-idb/src/utils.js#L105-L112), hence it doesn't mess with IDB transactions.

daleharvey:
Hm possible, although it definitely affected the indexeddb adapter, either way we are green in both with this (passing 100% on safari tech preview as well)

daleharvey:
The previous `update_seq` test was passing even though I wasnt persisting its data, since the open databases were being cached under the hood it was just using the same object for the new constructor, closing and reopening ensures it persists properly, fails before and green after

daleharvey:
So this is ~2 weeks old and doesnt effect the current adapter code in any way, just minor changes to the tests so self 👍 'ing as it unblocks a bunch of my work

nolanlawson:
Yeah sorry I am just swamped these days, please self-+1 as necessary. I'll look more into idb-next later.

Issue #5295, opened 2016-06-07
Move pouchdb-collate and pouchdb-collections into core


nolanlawson:
With the new monorepo setup, this should be a slam-dunk. There's no reason to have these as a separate repo.

BTW I found [this awesome Lerna tool](https://github.com/lerna/lerna/pull/173) that can import a separate Git repo _with version history_! Thanks @hzoo. :)

nlfiedler:
Are the functions from pouchdb-collate exposed in pouchdb core now? The other module is deprecated, but I can't find how to use this API within PouchDB 6.3.4. I see the functions in the code, but they are not exposed? Guides and API docs don't talk about it either.

daleharvey:
They are in the pouchdb repo to be used by pouchdb core, but they are not part of the pouchdb core api, if you want to use them you need to do so directly, closing since they are in the repo

PR #5476, opened 2016-07-15
(#5295) - Add pouchdb-collate to monorepo

daleharvey:
[skip ci]

### DIFF
--- a/bin/external-deps.js
+++ b/bin/external-deps.js
@@ -8,7 +8,7 @@ module.exports = [
   // main deps
   'argsarray', 'debug', 'double-ended-queue', 'es3ify', 'es6-promise-pool',
   'fruitdown', 'inherits', 'js-extend', 'level-write-stream', 'levelup', 'lie',
-  'localstorage-down', 'memdown', 'pouchdb-collate',
+  'localstorage-down', 'memdown',
   'request', 'scope-eval', 'spark-md5', 'through2',
   'vuvuzela', 'leveldown', 'websql',
   // core node deps
--- a/packages/pouchdb-checkpointer/package.json
+++ b/packages/pouchdb-checkpointer/package.json
@@ -13,7 +13,7 @@
     "src"
   ],
   "dependencies": {
-    "pouchdb-collate": "1.2.0",
+    "pouchdb-collate": "5.5.0-prerelease",
     "pouchdb-promise": "5.5.0-prerelease",
     "pouchdb-utils": "5.5.0-prerelease"
   }
--- a/packages/pouchdb-collate/LICENSE
+++ b/packages/pouchdb-collate/LICENSE
@@ -1,0 +1,202 @@
+
+                                Apache License
+                          Version 2.0, January 2004
+                       http://www.apache.org/licenses/
+
+  TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+  1. Definitions.
+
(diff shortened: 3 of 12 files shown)


daleharvey:
So working on this but cant get it to build, putting a WIP in case anyone has any pointers, error is 

```
~/src/pouchdb (pouchdb-collate)$ npm run build

> pouchdb-monorepo@0.0.0 build /Users/daleharvey/src/pouchdb
> lerna bootstrap && npm run build-modules && npm run build-test

Lerna v2.0.0-beta.9
Linking all dependencies
Successfully bootstrapped 34 packages.

> pouchdb-monorepo@0.0.0 build-modules /Users/daleharvey/src/pouchdb
> node bin/build-modules.js

Building pouchdb...
Building pouchdb-adapter-fruitdown...
Building pouchdb-adapter-http...
Building pouchdb-adapter-idb...
Building pouchdb-adapter-indexeddb...
Building pouchdb-adapter-leveldb...
Building pouchdb-adapter-leveldb-core...
Building pouchdb-adapter-localstorage...
Building pouchdb-adapter-memory...
Building pouchdb-adapter-node-websql...
Building pouchdb-adapter-utils...
Building pouchdb-adapter-websql...
Building pouchdb-adapter-websql-core...
Building pouchdb-ajax...
Building pouchdb-binary-utils...
Building pouchdb-browser...
Building pouchdb-checkpointer...
Building pouchdb-collate...
Building pouchdb-collections...
Building pouchdb-core...
Building pouchdb-errors...
Building pouchdb-for-coverage...
Building pouchdb-generate-replication-id...
Building pouchdb-http...
Building pouchdb-json...
Building pouchdb-mapreduce...
Building pouchdb-mapreduce-utils...
Building pouchdb-md5...
Building pouchdb-merge...
Building pouchdb-node...
Building pouchdb-promise...
Building pouchdb-replication...
Building pouchdb-utils...
Building sublevel-pouchdb...
build error
Error: Could not load null: path must be a string or Buffer
    at /Users/daleharvey/src/pouchdb/node_modules/rollup/dist/rollup.js:7340:10
    at Array.<anonymous> (/Users/daleharvey/src/pouchdb/node_modules/lie/lib/index.js:88:21)
    at nextTick (/Users/daleharvey/src/pouchdb/node_modules/immediate/lib/index.js:61:18)
    at _combinedTickCallback (internal/process/next_tick.js:67:7)
    at process._tickCallback (internal/process/next_tick.js:98:9)
```

Guessing its because this isnt ES5 contained in `/src`? just plain JS in `/lib`, I would prefer if we do not have to use ES6 for the more trivial things so if we can have it working as is it would be great. if thats a massive problem then will this work if I switch to es6?

daleharvey:
Just to test I added `pouchdb-collate` to external deps to see if it would avoid building

nolanlawson:
Yeah looks like you didn't even check in the `lib` files because they're .gitignored. My assumption with the Lerna setup is that every sub-project is ES6, so I think in this case what we should do is write a `src/index.js` and use ES6 exports.

One advantage of this is that when `pouchdb` is bundled, `pouchdb-collate` will be flattened out by Rollup and thus give us a smaller bundle size.

daleharvey:
ok switched this to ES6 and looks to be working, I think it would be nice to not require ES6 for some trivial code, but at this point its probably not worth the extra work.

nolanlawson:
:+1:

Issue #5486, opened 2016-07-19
Handling xhr.response===null (iOS http adapter)


ssured:
### Environment (Node.js/browser/hybrid app/etc.)

Browser

### Browser/platform (Chrome/FF/Safari/Edge/iOS/Android/etc.)

iOS Safari in Simulator

### Adapter (IndexedDB/WebSQL/LevelDB/etc.)

Fruitdown

### Server (CouchDB/Cloudant/Couchbase/PouchDB Server/etc.)

CouchDB 1.6

### Please describe your issue. Test cases (e.g. JSBin) are very helpful!

```
[Error] Failed to load resource: The request timed out. (10|png, line 0)
[Error] TypeError: Attempted to assign to readonly property.
	onreadystatechange (vendor-2144b83d0afea292420d69ea120a5a7d.js:88096)
```

![image](https://cloud.githubusercontent.com/assets/2558804/16963340/4665b18a-4df6-11e6-93e4-32da907be158.png)

which points to https://github.com/pouchdb/pouchdb/blob/ce73b62428d61225e23fa7c1515767b64bdd3eb9/packages/pouchdb-ajax/src/request-browser.js#L221

It seems it is not allowed to write to `error.status` in iOS Safari.

daleharvey:
hrm, this seems to work in normal safari, will try on an emuator, you got a specific version for the demo? it would also be useful to know if the error comes from `new Error` or JSON.parse, either way it seems very strange, a plain js object certainly should be fine and if its `new Error` then the `err.code = ` seems like it should be the one that is throwing

ssured:
All is well in iOS 9.3 / iPad2 simulator:
![image](https://cloud.githubusercontent.com/assets/2558804/16990768/938f12be-4e99-11e6-9fa6-73a3a1f54c05.png)

And then in an iOS 9.3.2 iPad Air 1 device, connected by USB:
![image](https://cloud.githubusercontent.com/assets/2558804/16990947/2f3cd00c-4e9a-11e6-8fc7-c84a12d43d7b.png)

However I do see these errors popping up in Rollbar. These events popped up while researching the sync of 200+ attachments on Safari. The highlighted line is where the error comes from. I have +- 100 occurrences, all on Mobile Safari.

nolanlawson:
[...]

PR #5491, opened 2016-07-22
(#5486) - Handle empty xhr response

ssured:
This can occur in Chrome on desktop, and probably other browsers too

### DIFF
--- a/packages/pouchdb-ajax/src/request-browser.js
+++ b/packages/pouchdb-ajax/src/request-browser.js
@@ -213,7 +213,7 @@ function xhRequest(options, callback) {
       if (timedout) {
         err = new Error('ETIMEDOUT');
         err.code = 'ETIMEDOUT';
-      } else {
+      } else if (typeof xhr.response === 'string') {
         try {
           err = JSON.parse(xhr.response);
         } catch(e) {}


nickcolley:
Thanks a lot for the PR, do you have any more information on where this can occur?

We could potentially test for this circumstance if you can help us reproduce this :)

Thanks 👍

ssured:
`xhr.response === null` for `xhr.readystate === 4` is a little weird and probably not conforming to spec, but Chrome on my machine sometimes hits this combo. It happens too little times to be able to reproduce why. I suspect an auto reload script triggers this, which reloads my page when editing files in my editor (ember-cli). The error happens just before a page reload and triggers a `debugger` statement in my error handler, which is how I can see this happening. Otherwise the page would just reload and the error would be gone.

I think the proposed change is valid, as `JSON.parse` expects a string.

daleharvey:
So ideally I think these type of cases would be best tested by unit tests in the xhr lib, I would like our xhr to be something a little more battle tested and less homebrew at some point, but worried about the potential size increase, in the meantime I think this is a good patch, got one small nit

daleharvey:
@ssured So this isnt entirely down to your patch, however its possible in https://github.com/pouchdb/pouchdb/blob/master/packages/pouchdb-ajax/src/request-browser.js#L212 that the user is going to get an empty error object in the callback, can we put a default error in there, something like `createError(BAD_REQUEST, 'Request to /url failed')` or such, will make debugging easier hopefully

daleharvey:
Going to listen to my own advice, things that can be improved in follow ups dont need to block the original patch, will land this now, thanks @ssured :)

ssured:
Cool & thanks

PR #5495, opened 2016-07-23
Fixes inline appearance of some headers

dlakata:
I noticed that some headers were aligned incorrectly, such as:
"Using db.put()" https://pouchdb.com/api.html#create_document
"PouchDB allDbs()" https://pouchdb.com/external.html#plugins
and others in similar locations.

As far as I could tell, this doesn't change the appearance of anything else on the site, but let me know if you find otherwise.

### DIFF
--- a/docs/_layouts/default.html
+++ b/docs/_layouts/default.html
@@ -64,12 +64,14 @@
 
         <div class="container">
 
-          <h1 class="margin-bottom-none">{{ page.title }}</h1>
-          {% if page.edit != false %}
-            <span class="h1 margin-bottom-none">
-              {% include edit_button.html %}
-            </span>
-          {% endif %}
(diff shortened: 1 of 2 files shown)


daleharvey:
👍 looks good to me, @nickcolley any chance you could take a look to make sure I am not missing something?

Thanks @dlakata

nickcolley:
Thanks for the spot and the PR 👍 

I think the reason it's not in the heading is because of accessibility reasons (for example a screenreader would read out the link which is not useful), so be great to fix this without putting the link inside the header.

daleharvey:
No activity in 2 weeks, would love to include this but needs issues addressed, please feel free to reopen if you want to address the issues mentioned (ie not put inside the header). Thanks

PR #5513, opened 2016-07-30
Add check for file:// URLs

stuartlangridge:
Display a warning if one attempts to use PouchDB in a browser on a file URL, because it won't work right.

Appreciate that there isn't a test for this, but the whole test suite is designed to spin up a server and then test against it; spinning up saucelabs/selenium to run the suite of browser tests against a file: URL just to see if this warning is emitted (and then have all the tests fail because Pouch doesn't work on file: URLs) seemed rather overkill.

### DIFF
--- a/packages/pouchdb-core/src/constructor.js
+++ b/packages/pouchdb-core/src/constructor.js
@@ -64,6 +64,11 @@ function PouchDB(name, opts) {
     throw new Error('Missing/invalid DB name');
   }
 
+  if (typeof window !== 'undefined' && window.location && window.location.href && window.location.href.match(/^file:/)) {
+    console.warn("Warning: PouchDB will not work correctly on file:// URLs, because the underlying" +
+      "storage (IndexedDB or WebSQL) will not. Please serve PouchDB-using pages from a web server.");
+  }
+  
   var prefixedName = (opts.prefix || '') + name;
   var backend = parseAdapter(prefixedName, opts);
 


daleharvey:
Thanks for the PR, so I just tested a basic script (copied and pasted http://paste.pouchdb.com/ and made the url http://) into my filesystem and it works in both chrome and firefox, looking through the bug tracker indexeddb does work in a local filesystem (ajax probably wont).

There are also cases where I think we almost always are used via the filesystem (cordova), I think a better approach to this would be to make sure we show as clear an error as possible in the cases that indexeddb fails to open, as opposed to pre warning that may not apply / be confusing

nolanlawson:
Code coverage check is breaking; an `/* istanbul ignore if */` should be enough to fix that, though. Another thing is that we have a `guardedConsole` from `pouchdb-utils` that we use for this kind of stuff, because `console` can break in certain environments (e.g. Web Workers in Android 4.4).

@daleharvey The point about Cordova is a good one; I forgot about that. I think Android is actually `android_asset://path_to_file` but iOS indeed should be `file://`. Maybe a better approach here would be to catch the relevant IDB/WebSQL error and add some useful information on top of it? Or surface it to the user? I'm not even sure what error gets thrown exactly, but I'll bet it's a befuddling `Error 18 Security` or something like that.

stuartlangridge:
I'd like to trap actual errors, but in my experience IDB _sorta_ works in file: URLs, it's just weird and unreliable and unconvincing. Hence the idea of an upfront warning which just says "hey, don't do this wrong thing" as early as possible. The point about Cordova is a well-taken one, though.

nolanlawson:
@stuartlangridge What happens exactly when you use IDB in file://? Haven't tested myself.

daleharvey:
In my testing it just worked fine in firefox and chrome, I can run a large amount of the test suite from a file url, there where definitely issues before, and I know ajax acted differently when hosted at a file url, but yeh things like these should be fixed by presenting errors correctly, even our current prewarns are confusing and prewarning in situations where it could be working fine is suprt confusing. Cheers for the PR but closing out, apologies

PR #5519, opened 2016-07-31
(#5519) - remove sqlite plugin support (BREAKING)

nolanlawson:
There is no more need to support the SQLite Plugin in PouchDB core, because this
can now be handled in [a separate plugin](https://github.com/nolanlawson/pouchdb-adapter-cordova-sqlite/blob/master/README.md). Furthermore, I think this system is better because

1. it makes it clear to the user whether they are using WebSQL or Cordova SQLite
2. it removes the need to do tricky `valid()` functions trying to guess whether we're in a Cordova app or not
3. it removes the need to maintain this stuff in PouchDB core

This is a breaking change.


docs/_includes/api/create_database.html
```diff
@@ -33,23 +33,14 @@
...
-* `androidDatabaseImplementation`: `1` to use the new `sqlite4java` implementation, `2` to use the older version based on the native Android APIs. Defaults to `1`. See the [SQLite Plugin documentation](https://github.com/litehelpers/Cordova-sqlite-storage#android-sqlite-implementation).
-
 **Notes:**
 
 1. In IndexedDB and WebSQL, PouchDB will use `_pouch_` to prefix the internal database names. Do not manually create databases with the same prefix.
 2. When acting as a client on Node, any other options given will be passed to [request][].
-3. When using the `'leveldb'` adapter (the default on Node), any other options given will be passed to [levelup][]. The storage layer of leveldb can be replaced by passing a level backend factory (such as [MemDOWN][]) as `options.db`. The rest of the [supported levelup options are documented here][levelup_options].
+3. When using the `'leveldb'` adapter (the default on Node), any other options given will be passed to [levelup][].
```
nolanlawson: I would like to deprecate `db` as well. It's an internal detail of `pouchdb-adpater-leveldb-core`, and furthermore there are no LevelDOWN adapters I'm aware of that do anything interesting with PouchDB other than the ones we ourselves maintain (such as LocalStorageDOWN, MemDOWN, and FruitDOWN). The e.g. RedisDOWN/RiakDOWN/etc. adapters were interesting experiments but ultimately pointless because perf sucks and they didn't solve the problem people hoped they would solve (PouchDB replication without CouchDB).

daleharvey:
Yeh I am 👍 on this, as long as the functionality can be supported via the plugin then I think its much clearer for the user to specifically use a cordova target than for us to keep working round issues supporting all the things

chino23:
Does this mean when switching to 6.x I can't use websql out of the box anymore in the browser? I've been using this for years as it's the fastest.
Are there plans to dropping support completely for that?

nickcolley:
To be clear @chino23 this is removing SQLite not WebSQL, there's no plans to drop WebSQL in core 👍  😄

chino23:
Oh alright, guess I messed that up. Thank you! :)

nolanlawson:
Yup, that is correct. Also you can still use the separate plugin for Cordova SQLite support.

PR #5529, opened 2016-08-03
(#5527) - Check if ArrayBuffer is defined before binary check

### LINKED ISSUE #5527: ReferenceError: 'ArrayBuffer' is undefined
The problem happens on IE9 when using the HTTP Adapter.
Pull request coming.

filionf:
Fixes #5527. I've unfortunately been unable to run the tests on the browser (tried Mac, Windows & BashOnWindows), but I believe the change is pretty straightforward.


daleharvey:
Yeh the change is fairly obvious and matches the subsequent condition, however without being verified by tests its entirely possible it will be regressed at some point, that doesnt need to block it landing but something to consider if you require solid ie9 support going forward.

As for running the test suite, please open an issue or catch us on irc if you are hitting unexpected issues running the test suite, its a high priority for us that people can run the tests as easily as possible, windows is a problem (although improving rapidly) however mac certainly should work

Issue #5545, opened 2016-08-08
Remove Lerna


nolanlawson:
Based on discussion with @boennemann I think we can remove our Lerna dependency and vastly speed up our build times by avoiding `lerna bootstrap`.

We already don't use Lerna for building (because `lerna run` is slow due to running multiple processes, so I wrote one big top-level build script) or for publishing (because I didn't take enough time to grok the Lerna docs on this, we're not using `version: independent`, and I found a bash loop with `npm publish` to be simpler), so `lerna bootstrap` is the only thing we're currently using it for.

@boennemann pointed out that we can achieve the same effect as `lerna bootstrap` with a few changes:

- `packages/` becomes `packages/node_modules`, and due to how `require()` works this means that inter-dependencies are automatically resolved.
- Common dependencies go at the top-level `package.json` (separate from `devDependencies` used for building). At publish time, we just substitute the version numbers in the sub-packages' `package.json` with the version number from the top level. One trick we can do here is to scan the JS files for `import`s and automatically add the correct dependencies, which means there's no more need to update the same dependency multiple times in multiple sub-packages, nor will we have bugs like we had with `js-extend` where we were accidentally using different versions in different sub-packages. If this trick doesn't work, we can also just include `*` versions in the `package.json`s and then replace those at publish time.

@boennemann wrote up his ideas here: https://github.com/boennemann/alle#readme

garrensmith:
I'm going to hold off on the pouchdb-server monorepo work until we have a good solution for pouchdb.

yaronyg:
In which case doesn't this mean that Express-PouchDB is now unworkable for the duration? The old code won't build tests because of PouchDB and the new code isn't finished?

garrensmith:
Hi Yaron, I'm really struggling to find time to work on this. You welcome to continue the work. Otherwise I'll pick it up once Nolan has come up with a better way of managing monorepo's

All misspelling thanks to my iPhone

yaronyg:
Garren, that is totally fair and I should have prefixed the message with "and this isn't a complaint". This is an open source project, if someone wants something the default is that they should bloody well do it themselves (in this case, if I want Express-PouchDB passing tests then I better get busy and fix the repo!). The purpose of my question was just to make sure I understand the situation properly. That right now Express-PouchDB can't accept PRs because it can't run tests and those tests won't run until the repo situation is fixed.

[...]

PR #5550, opened 2016-08-12
(#5545) - remove lerna

nolanlawson:
It turns out there were only a few changes needed to remove `lerna bootstrap` entirely:

- move `packages` to `packages/node_modules`
- fix `.gitignore` so that it only ignores the top-level `node_modules`
- add all dependencies from sub-packages to the top-level package

This change _only_ removes Lerna. There are still some remaining TODOs that can be tackled in a future PR, such as not duplicating dependencies between sub-packages and the main package, but for now this is the simplest PR.


nolanlawson:
Oh I should also mention that this fully re-enables GreenKeeper, although we will still need to manually update versions in the sub-packages for changed dependencies (not dev dependencies) until I can fix the TODO described above.

packages/node_modules/pouchdb/package.json
```diff
@@ -41,23 +41,6 @@
...
-    "pouchdb-ajax": "5.5.0-prerelease",
-    "pouchdb-browser": "5.5.0-prerelease",
-    "pouchdb-checkpointer": "5.5.0-prerelease",
-    "pouchdb-core": "5.5.0-prerelease",
-    "pouchdb-generate-replication-id": "5.5.0-prerelease",
-    "pouchdb-node": "5.5.0-prerelease",
-    "pouchdb-promise": "5.5.0-prerelease",
-    "pouchdb-replication": "5.5.0-prerelease",
-    "pouchdb-utils": "5.5.0-prerelease"
-  },
```
nolanlawson: devDependencies are no longer needed inside of the sub-packages, since everything is handled either at the top-level `node_modules` or in `packages/node_modules`. These devDependencies previously only existed so that `lerna bootstrap` would work correctly.

nolanlawson:
Looks like this shaves about 30 seconds off of each build. :)

nolanlawson:
Well unfortunately this broke the 0.10 build because of my too-clever-by-half "build in Node 6, test in Node 0.10" thing. Looks like Rollup expects `Map` to be there and barfs if it's not.

nolanlawson:
Hm, this is a very hard problem to solve. I can't think of a way to extract just the test suite (which doesn't require Rollup), and then install it and run it in Node 0.10. Maybe it would be easier just to open a PR on Rollup to support 0.10.

nolanlawson:
Uggg, I could solve this by just dumping all the devDeps into dependencies, because then `npm install --production` wouldn't run `npm run build`, but it's kinda ugly to slam them all together like that. We could also disable the `prepublish` so that it doesn't run after `npm install`. Not sure what to do here.

daleharvey:
'0.10' becomes entirely unsupported in 50 days, I think its fine for us to drop support

nolanlawson:
Big :+1: to that.

nolanlawson:
OK, removed 0.10 entirely and moved back to the old system of using the same Node for building and testing.

nolanlawson:
lol wow `eslint` doesn't work in either 0.11 or 0.12 because it uses `let`. this is ridiculous

nolanlawson:
I propose we move linting to a separate build entirely. There's no reason to run it for each and every build anyway.

nolanlawson:
Oh for crying out loud, forgot that eslint runs after each test. Let's just add 0.11 and 0.12 to allowed failures for now.

nolanlawson:
Now that Firefox is fixed, trying again...

nolanlawson:
0.11 and 0.12 are failing in all tests because we run eslint. I decided to drop all of 0.10, 0.11, and 0.12 entirely. There's no easy way I can see to support 0.10-0.12 if eslint has decided their minimum baseline is 4.

Maybe in the future we can figure out how to write a patch to do the "build in one version, test in another" thing again, but right now it's just getting too difficult to support anything less than 4.

OTOH I added a test for Node 4 itself, so we're currently testing in 4, 5, and stable, with an allowed failure on stable.

daleharvey:
So I am 👍 on that, these node versions are going entirely unsupported (no security upgrades) very shortly so if its blocking improvements then happy to drop support. 

Would like to ping @yaronyg though, if you absolutely need `0.X` then would like to give you the chance to do the extra work to get it working here, or at least notice to fork, hopefully you can use >=4 by now though

nolanlawson:
All tests are passing except code coverage. I will look into it.

    No coverage information was collected, exit without writing coverage information
    ERROR: No coverage files found.

nolanlawson:
OK, had to fix Istanbul to not-ignore `packages/node_modules`. Now hopefully this should all be good and I can move on to the other stuff that's blocking 6.0.0. 😛

.travis.yml
```diff
@@ -130,8 +124,17 @@ matrix:
   - env: CLIENT="saucelabs:MicrosoftEdge" COMMAND=test
   - env: CLIENT=selenium:firefox:48.0 PERF=1 COMMAND=test
   - env: CLIENT=selenium:firefox:48.0 NEXT=1 COMMAND=test
-
+  - node_js: "stable"
+    services: docker
+    env: CLIENT=node COMMAND=test
   fast_finish: true
+  include:
```
daleharvey: Tiny nit, could we move this to the top of the `matrix:` section, I was confused that allowed_failures had this section inside of it

daleharvey:
Had one tiny comment on the travis.yml but otherwise this is great, faster + lesss complicated builds is awesome. 

One other tiny question, so if we want to use a package in a submodule, we need to add it to both `package.json` and that submodule? if so it seems a note in the docs somewhere would be useful.

Would still like to hear from Yaron just in case, but 👍 maybe wait till monday before merging?

daleharvey:
lol ok just read https://github.com/pouchdb/pouchdb/pull/5583 so can ignore the question :)

.travis.yml
nolanlawson: Will do

nolanlawson:
Made the .travis.yml change on the way in, see 9b51141a4a54ec79ac774c6117b5aa4c986d785b

yaronyg:
@daleharvey Unfortunately we can't move to Node 4. The reason is that we are running on JXcore which is the only version of node that currently runs on mobile phones and it is based on a hybrid of Node 0.10 and 0.12. We are updating it to do things like support the latest security patches for OpenSSL. But until we figure out some way to get mainline Node working on Android and iOS we have no option but to stick with JXcore. We are looking to put in resources that could monitor PouchDB failures on Node 0.10/0.12 and submit fixes.

PR #5562, opened 2016-08-16
(#5561) - update travis to latest firefox

### LINKED ISSUE #5561: All Firefox jobs are failing in Travis
This started recently. Looks like somehow either 41.0.1 isn't being installed, or it isn't being picked up.

This might be a good moment to try just switching to a later version of Firefox anyway. It's already on 48.

nolanlawson:
No idea if this will fix #5561, but it's worth a shot.

Here's hoping that the new Marionette driver works well!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ sudo:
   false
 
 addons:
-  firefox: "41.0.1"
+  firefox: "latest"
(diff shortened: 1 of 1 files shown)


nolanlawson:
Nope, no dice.

Issue #5471, opened 2016-07-13
PouchDB.plugin() should throw a helpful error if you pass in the wrong type


nolanlawson:
I've done this so many times to myself, it's not even funny:

```js
var PouchDB = require('pouchdb-core')
  .plugin('pouchdb-adapter-memory')
  .plugin('pouchdb-adapter-http')
  .plugin('pouchdb-replication');
```

Of course what I meant to do was:

```js
var PouchDB = require('pouchdb-core')
  .plugin(require('pouchdb-adapter-memory'))
  .plugin(require('pouchdb-adapter-http'))
  .plugin(require('pouchdb-replication'));
```

When you make this mistake, though, you get a really unhelpful error about "Adapter is missing." I wasted a lot of time trying to hunt it down before realizing I passed in a string to `plugin()` instead of an object/function like it's supposed to take.

I believe we should throw a more helpful error inside of `PouchDB.plugin()`. In particular, if you pass in a non-function (i.e. an object) that doesn't have any `Object.keys()` (i.e. length of 0), then it's probably an invalid plugin and we should throw. This should guard against similar mistakes.

See [Plugin API](http://pouchdb.com/api.html#plugins) and [Custom Builds](https://pouchdb.com/custom.html) for more info on how this is supposed to work.

nolanlawson:
For first-timers who want to tackle this, you would need to add a test to `test.basics.js`, and also modify the `plugin` code here: https://github.com/pouchdb/pouchdb/blob/9e370d492ed1aba57db069ca98205eb50de6ad92/packages/pouchdb-core/src/setup.js#L45-L54

nolanlawson:
The test should probably only run in Node.js; you can check for this by doing `if (typeof process !== 'undefined' && !process.browser)`.

nolanlawson:
If you'd like any help with this issue, please come to the PouchDB Slack channel (http://slack.pouchdb.com/) and ask for help; otherwise the best source of information for getting started is in `CONTRIBUTING.md` and `TESTING.md`.

nolanlawson:
As for how to write a Mocha test that checks that an error is thrown, I like to use this pattern:

```js
doSomethingThatOughtToThrow().then(function () {
  throw new Error('expected an error here!')
}, function (err) {
  should.exist(err);
});
```

^ This will fail if the function in question _doesn't_ throw an error.

eraserxp:
Hi Nolan. I would like to work on this.

[...]

PR #5569, opened 2016-08-17
(#5471) PouchDB.plugin() giving more helpful errors

wholien:
background: https://github.com/pouchdb/pouchdb/issues/5471

### DIFF
--- a/packages/pouchdb-core/src/setup.js
+++ b/packages/pouchdb-core/src/setup.js
@@ -45,6 +45,8 @@ PouchDB.adapter = function (id, obj, addToPreferredAdapters) {
 PouchDB.plugin = function (obj) {
   if (typeof obj === 'function') { // function style for plugins
     obj(PouchDB);
+  } else if (Object.keys(obj).length == 0){
+    throw new Error('Invalid plugin: object passed in is empty or not an object');
   } else {
     Object.keys(obj).forEach(function (id) { // object style for plugins
       PouchDB.prototype[id] = obj[id];
--- a/packages/pouchdb-core/src/setup.js
+++ b/packages/pouchdb-core/src/setup.js
@@ -45,7 +45,7 @@ PouchDB.adapter = function (id, obj, addToPreferredAdapters) {
 PouchDB.plugin = function (obj) {
   if (typeof obj === 'function') { // function style for plugins
     obj(PouchDB);
-  } else if (Object.keys(obj).length === 0){
+  } else if (typeof obj !== 'object' || Object.keys(obj).length === 0){
     throw new Error('Invalid plugin: object passed in is empty or not an object');
   } else {
     Object.keys(obj).forEach(function (id) { // object style for plugins
--- a/packages/pouchdb-core/src/setup.js
+++ b/packages/pouchdb-core/src/setup.js
@@ -45,7 +45,7 @@ PouchDB.adapter = function (id, obj, addToPreferredAdapters) {
 PouchDB.plugin = function (obj) {
   if (typeof obj === 'function') { // function style for plugins
     obj(PouchDB);
-  } else if (Object.keys(obj).length == 0){
+  } else if (Object.keys(obj).length === 0){
     throw new Error('Invalid plugin: object passed in is empty or not an object');
   } else {
     Object.keys(obj).forEach(function (id) { // object style for plugins
--- a/tests/integration/test.basics.js
+++ b/tests/integration/test.basics.js
@@ -1106,5 +1106,21 @@ adapters.forEach(function (adapter) {
         });
       });
     }
+
+    if (typeof process !== 'undefined' && !process.browser) {
+      it('#5471 PouchDB.plugin() should throw error if passed wrong type or empty object', function () {
+        var db = new PouchDB(dbs.name);
+        PouchDB.plugin('pouchdb-adapter-memory');
+        var doc = {
+          _id: "x",
+          type: "testdoc",
+        };
+        db.put(doc).then(function () {
+          throw new Error('expected an error here!');
+        }, function (err) {
+          should.exist(err);
+        });
+      });
+    }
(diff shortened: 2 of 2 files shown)


packages/pouchdb-core/src/setup.js
```diff
@@ -45,6 +45,8 @@ PouchDB.adapter = function (id, obj, addToPreferredAdapters) {
 PouchDB.plugin = function (obj) {
   if (typeof obj === 'function') { // function style for plugins
     obj(PouchDB);
+  } else if (Object.keys(obj).length === 0){
```
willholley: I think you'll also want to check whether obj is an object here. e.g. `Object.keys("foo")` will return `["0", "1", "2"]` but `"foo"` is not a valid plugin. I'd suggest changing this to `} else if (typeof obj !== 'object' || Object.keys(obj).length === 0){`

tests/integration/test.basics.js
```diff
@@ -1106,5 +1106,21 @@ adapters.forEach(function (adapter) {
         });
       });
     }
+
+    if (typeof process !== 'undefined' && !process.browser) {
+      it('#5471 PouchDB.plugin() should throw error if passed wrong type or empty object', function () {
+        var db = new PouchDB(dbs.name);
+        PouchDB.plugin('pouchdb-adapter-memory');
```
willholley: I think the error should be thrown at this point. If the test gets past this line, it should fail.

tests/integration/test.basics.js
```diff
@@ -1106,5 +1106,21 @@ adapters.forEach(function (adapter) {
         });
       });
     }
+
+    if (typeof process !== 'undefined' && !process.browser) {
```
nolanlawson: Is there a reason this test only runs in Node? Ideally `PouchDB.plugin()` would throw in the browser too.

Also it may make it easier to test this; you can do `npm run dev` and then open `http://localhost:8000/tests/integration?grep=unique_word_in_my_test_name` and then test in your preferred browser.

wholien: I was just going by your comment here https://github.com/pouchdb/pouchdb/issues/5471#issuecomment-232240310 that said it should only run in Node. I can change it.

wholien: What was the reason you wrote that it should only run in Node?

packages/pouchdb-core/src/setup.js
wholien: good catch - I think that's why my test had not thrown errors before, as I was testing with a string.

wholien:
@willholley @nolanlawson : any clue on why the test would not be catching the error thrown? Right now my test fails because it doesn't get to `should.exist(err);` - any clue why that is? 

I tried 
`PouchDB.plugin('pouchdb-adapter-memory').catch(function (err) {
    should.exist(err);
});` 
but the test still fails. 

In both cases, the test outputs `Error: Invalid plugin: object passed in is empty or not an object`, which is the error I added and the error I want to catch.

tests/integration/test.basics.js
```diff
@@ -1106,5 +1106,15 @@ adapters.forEach(function (adapter) {
...
+      it('#5471 PouchDB.plugin() should throw error if passed wrong type or empty object', function () {
+        PouchDB.plugin('pouchdb-adapter-memory').then(function () {
```
willholley: you need to use a JavaScript try/catch rather than promises here. I think Mocha can do this for you - something like:

```
assert.throws(function() {
      PouchDB.plugin('pouchdb-adapter-memory');
}, Error, 'Invalid plugin: object passed in is empty or not an object');
```

wholien: @willholley - I will try this and report back

wholien: the `assert` code did not work (`assert` is not defined? Also do not see `assert` used anywhere else), but a try/catch did work.

wholien:
made the test a try/catch, which works. There does not seem to be a failure case, though. I tried to use `done()` to create a failure case (if err is thrown then goes into catch, where `done()` is called, else throw error), but it complains about `done()` not being defined (ReferenceError: done is not defined). That's confusing since `done()` is used elsewhere in the same file. If anyone can shed some light on why that is, it would be great!

wholien:
i.e.
```
       try {
          PouchDB.plugin('pouchdb-adapter-memory');
        }
        catch (err) {
          should.exist(err);
          done();
        }
        throw new Errr('expected an error here!');
```

tests/integration/test.basics.js
```diff
@@ -1106,5 +1106,21 @@ adapters.forEach(function (adapter) {
         });
       });
     }
+
+    if (typeof process !== 'undefined' && !process.browser) {
+      it('#5471 PouchDB.plugin() should throw error if passed wrong type or empty object', function () {
+        try {
+          PouchDB.plugin('pouchdb-adapter-memory');
```
willholley: have a look at http://chaijs.com/api/bdd/#method_throw. You want the test to fail if it doesn't throw.

```
expect(function() {
    PouchDB.plugin('pouchdb-adapter-memory');
}).to.throw(Error, 'Invalid plugin: object passed in is empty or not an object');
```

wholien:
@nolanlawson any thoughts regarding whether this test should run in all environments or just Node?

tests/integration/test.basics.js
nolanlawson: TBH I don't remember why I said that. It was a dumb thing to say. 😅 

That said I can modify your PR to remove the "node-only" line and then we'll test it and then :+1: from me.

nolanlawson: Oh whoops wait, I don't have access to this branch. But you know what, I'm happy with it; testing only in Node is actually fine because it will still catch the error if we regress, so I'm happy with this. Thanks for the PR!!

PR #5571, opened 2016-08-17
(#5561) - Temporarily use saucelabs for firefox testing

daleharvey:


daleharvey:
I dont want to merge anything with our testing so broken, this is going to make test times huge but at least passing hopefully.

We can get it green then work on getting firefox tests running back on travis only

nolanlawson:
I sent a support email to Travis (cc'ed you) but I agree that we basically can't do anything if Travis is failing, so I'm :+1: when green.

daleharvey:
Got the reply and looks like travis + selenium is broken,, merging to unblock then will get things back to selenium soon, cheers

PR #5580, opened 2016-08-19
(#5574) - encode db name if opts.prefix is a URL

### LINKED ISSUE #5574: http adapter: should PouchDB encode the database name?
Let’s say I want to create a database called `user/123` with PouchDB using the http adapter.

```js
var dbName = 'user/123'
var MyPouchDB = PouchDB.defaults({prefix: 'http://localhost:5984/'})
MyPouchDB.plugin(require('pouchdb-adapter-http'))

var db = new MyPouchDB(dbName)
db.info().catch(console.log.bind(console))
// logs not_found error
```

The problem is that it sends a request to `GET http://localhost:5984/user/123/`, it is not encoding the database name. 

I would expect this to happen within PouchDB, instead of me needing to do 

```js
var db = new MyPouchDB(encodeURIComponent(dbName))
```

What do you think?

gr2m:
fixes #5574

### DIFF
--- a/packages/pouchdb-adapter-http/src/index.js
+++ b/packages/pouchdb-adapter-http/src/index.js
@@ -71,9 +71,26 @@ function preprocessAttachments(doc) {
   }));
 }
 
+function hasUrlPrefix(opts) {
+  if (!opts.prefix) {
+    return false;
+  }
+
+  var protocol = parseUri(opts.prefix).protocol;
+
+  return protocol === 'http' || protocol === 'https';
+}
+
 // Get all the information you possibly can about the URI given by name and
 // return it as a suitable object.
-function getHost(name) {
+function getHost(name, opts) {
+
+  // encode db name if opts.prefix is a url (#5574)
+  if (hasUrlPrefix(opts)) {
+    var dbName = opts.name.substr(opts.prefix.length);
+    name = opts.prefix + encodeURIComponent(dbName);
+  }
+
   // Prase the URI into all its little bits
   var uri = parseUri(name);
 
@@ -127,6 +144,7 @@ function paramsToStr(params) {
 
 // Implements the PouchDB API for dealing with CouchDB instances over HTTP
 function HttpPouch(opts, callback) {
+
   // The functions that will be publicly available for HttpPouch
   var api = this;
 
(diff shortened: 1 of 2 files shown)


packages/pouchdb-adapter-http/src/index.js
```diff
@@ -71,9 +71,26 @@ function preprocessAttachments(doc) {
...
 // Get all the information you possibly can about the URI given by name and
 // return it as a suitable object.
-function getHost(name) {
+function getHost(name, opts) {
+
+  // encode db name if opts.prefix is a url (#5574)
+  if (hasUrlPrefix(opts)) {
+    var dbName = opts.name.substr(opts.prefix.length);
+    name = opts.prefix + encodeURIComponent(dbName);
+  }
```
gr2m: if wonder if it would be enough to only check if `opts.prefix` is truthy, as this already is the http adapter and no other prefix value makes sense than a URL?

gr2m: I’ve tried it out but the tests pass in `opts.prefix` as `/path/to/pouchdb/tmp/_pouch_` which makes little sense to me as this is an http adapter?

Either way, the current solution works :)

gr2m:
This is a false positive, right? 

screen shot 2016-08-19 at 16 23 56

packages/pouchdb-adapter-http/src/index.js
nolanlawson: Yeah that tests date back to when the http adapter didn't use `prefix`. The fact that your fix takes this into account is just a bonus IMO.

nolanlawson:
Yes, it's a lucky run when you only get 1 failure. :)

:+1: when green, I've restarted that job.

gr2m:
🙌 💚 🚀

nolanlawson:
green!

gr2m:
when will this be released? It’s currently blocking https://github.com/hoodiehq/hoodie-store-server/pull/45 for us, would be nice to have a way to install a new version from npm soon. Anything I can help with?

nolanlawson:
We are working on a huge 6.0.0 release; I still need to remove `extras` and Lerna which is basically my task for the weekend, but after that we can do 6.0.0.

The meantime you can definitely fork, `git add -f`, and use that branch; sorry for the delay!

gr2m:
no worries, thanks for the update! Good luck with 6.0.0!

PR #5583, opened 2016-08-21
(#5583) - automatically set dependency versions before publishing

nolanlawson:
As a follow-up to https://github.com/pouchdb/pouchdb/issues/5545, I'd like to solve a few different problems:

1. Greenkeeper should "just work"
2. We shouldn't have to manually manage dependency versions across packages

To solve this, @boennemann suggested just replacing every version inside of the sub-package's `package.json` with `*`, and then updating them with the versions from the main `package.json` right before publishing. This seems like a super simple solution and very reasonable. So I implemented it. :smiley:


nolanlawson:
Note that this PR will look a lot cleaner once #5550 is merged.

bin/update-dep-versions.js
```diff
@@ -0,0 +1,33 @@
...
+    pkg.dependencies, 
+    pkg.peerDependencies, 
+    pkg.optionalDependencies
+  ];
+  depLists.forEach(function (deps) {
+    if (!deps) {
+      return;
+    }
+    Object.keys(deps).forEach(function (dep) {
+      if (topPkg.dependencies[dep]) {
```
boennemann: Shouldn't this be variable between `dependencies`, `peerDependencies` and `optionalDependencies`?

bin/update-dep-versions.js
```diff
@@ -0,0 +1,33 @@
...
+        deps[dep] = topPkg.dependencies[dep];
+      } else { // core pouchdb-* module
+        deps[dep] = topPkg.version;
```
boennemann: I'd check for startsWith `"pouchdb"` and throw an error otherwise, to prevent forgotten dependencies being defined with a random version.

bin/update-dep-versions.js
```diff
@@ -0,0 +1,33 @@
...
+
+var modules = fs.readdirSync('./packages/node_modules');
+modules.forEach(function (mod) {
+  var pkgPath = './packages/node_modules/' + mod + '/package.json';
+  var pkg = JSON.parse(fs.readFileSync(pkgPath, 'utf8'));
+  var depLists = [
+    pkg.dependencies, 
+    pkg.peerDependencies, 
+    pkg.optionalDependencies
+  ];
```
boennemann: Maybe you can start copying the rest of the fields as well, to avoid duplicate metadata :)

boennemann:
Awesome to see PouchDB going this route, that has full Greenkeeper support ✨ 

I just want to mention again that this might cause problems when a package requires some 3rd party package that isn't defined in its package.json, but only in the top-level one. If no other core pouchdb package depends on that package as well, it will fail. This is only noticeable after publishing.

Special attention required; some sort of verification would be even better.

nolanlawson:
Yep, I realize it's a problem, but I'm hoping it'll be caught soon enough that we can just re-publish in that case.

bin/update-dep-versions.js
nolanlawson: That's a TODO. :) We could seriously just use a template for the LICENSE, README, and package.json.

bin/update-dep-versions.js
nolanlawson: There's also `sublevel-pouchdb` which is local. We could probably just check that it exists in `packages/node_modules`.

bin/update-dep-versions.js
nolanlawson: We basically want to do the same for all three. peerDependencies is basically always just `pouchdb-core`, though, and `optionalDependencies` will be removed entirely before we publish 6.0.0.

bin/update-dep-versions.js
boennemann: Yeah true, you even have that list from line 10 still :)

bin/update-dep-versions.js
boennemann: I see, makes sense. The code is confusing without knowing this ;)

nolanlawson:
Rebased and added @boennemann's suggestion from https://github.com/pouchdb/pouchdb/pull/5583#discussion_r75601907

daleharvey:
Awesome 👍, much clearer thanks

PR #5585, opened 2016-08-21
Support options.timeout in replication bulkDocs(); closes #5584

### LINKED ISSUE #5584: Timeout is documented but doesn't apply to push replication
### Environment (Node.js/browser/hybrid app/etc.)

Node.js; however I think this will apply to all push replications to a remote server.

### Adapter (IndexedDB/WebSQL/LevelDB/etc.)

Replicating from LevelDB to a remote Cloudant account

### Server (CouchDB/Cloudant/Couchbase/PouchDB Server/etc.)

Cloudant

### Please describe your issue. Test cases (e.g. JSBin) are very helpful!

I think I found a bug involving the timeout timers in replication. I am not yet sure if it is a documentation bug, or an implementation bug, or the interplay between the two.

In the API reference for `PouchDB.replicate(source, target, [options])` it says that `options.timeout` means "Request timeout (in milliseconds)."

I see the timeout option used in `changesOpts`, basically following the `_changes` feed during replication. However, when writing docs, no timeout is passed to `target.bulkDocs()`. For example, see [index.js, line 7,771 in v5.4.1](https://github.com/pouchdb/pouchdb/blob/5.4.5/lib/index.js#L7771)

So, in my usage, I am pushing some somewhat large attachments (Base64 encoded, but that's a different matter), and I think Cloudant is processing them slowly. In any case, after 10,000 milliseconds I get a timeout error. Replication cannot proceed beyond that batch.

I am not 100% sure if the same timeout option makes sense in the call to bulkDocs. One might argue that they are subtly different timers: one follows changes, and the other simply posts docs. However, since `timeout` is classified as an "advanced" feature, I think perhaps people use it so rarely that it is safe to use its value in both situations.

jhs:
This is the simplest possible PR to close #5584. Basically pouchdb-replication passes that timeout value to bulkDocs, and the HTTP adapter passes it to request.

This is a very small change, but please let me know if you would like for me to add a unit test.

### DIFF
--- a/packages/node_modules/pouchdb-adapter-http/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-http/src/index.js
@@ -620,6 +620,7 @@ function HttpPouch(opts, callback) {
       ajax(opts, {
         method: 'POST',
         url: genDBUrl(host, '_bulk_docs'),
+        timeout: opts.timeout,
         body: req
       }, function (err, results) {
         if (err) {
--- a/packages/node_modules/pouchdb-replication/src/replicate.js
+++ b/packages/node_modules/pouchdb-replication/src/replicate.js
@@ -57,7 +57,8 @@ function replicate(src, target, opts, returnValue, result) {
       return;
     }
     var docs = currentBatch.docs;
-    return target.bulkDocs({docs: docs, new_edits: false}).then(function (res) {
+    var bulkOpts = {timeout: opts.timeout};
+    return target.bulkDocs({docs: docs, new_edits: false}, bulkOpts).then(function (res) {
       /* istanbul ignore if */
       if (returnValue.cancelled) {
         completeReplication();


nolanlawson:
LGTM, +1 when green. Will restart the intermittent failures.

jhs:
Hi, Nolan. Thank you very much.

Is it written anywhere how one may assess the Travis results? Which sorts of failures are actually bad? Which sorts are intermittent? I was unsure if I had caused a true bug or not.

jhs:
These were green when I logged in this morning, but irreddened after I rebased to master and re-pushed.

nolanlawson:
The Travis results are totally open to the public and are in the "Details" link above. OTOH only maintainers can restart the jobs unfortunately. It's a huge pain point for us.

daleharvey:
Our tests vary on stability (mostly due to our dependent infrastructure), basically if all node tests fail or all browser tests, then its likely your problem, if its 2/3 tests with no common pattern then likely ours.

This seems reasonable to me, would never say no to a test for things like this but I dont think it needs to block when its code path is hit a lot in the tests, cheers 👍

daleharvey:
Also @jhs, thats like your 3rd commit so I have invited you to @pouchdb/owners, all PR's still go through the review process but you can now review and merge others PR's at your own will, thanks for the contributions :)

PR #5588, opened 2016-08-21
docs: live replication complete event

cdaringe:
# problem statement

`live` replication and changes listeners don't generally emit `complete` events, but do if they've been cancelled.  this is undocumented.

# solution

document it!

# discussion

closes https://github.com/pouchdb/express-pouchdb/issues/316

### DIFF
--- a/docs/_includes/api/replication.html
+++ b/docs/_includes/api/replication.html
@@ -1,4 +1,4 @@
-{% include anchor.html edit="true" title="Replicate a database" hash="replication" %} 
+{% include anchor.html edit="true" title="Replicate a database" hash="replication" %}
 
 {% highlight js %}
 PouchDB.replicate(source, target, [options])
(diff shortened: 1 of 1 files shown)


daleharvey:
@cdaringe Github is complaining about the `&`'s and yeh I think we should probably be using `and` in those places.

Getting the documentation right is really hard, however this is document within the `Replication events` section directly below where you added this and I think that is the right place for it (https://pouchdb.com/api.html#replication - https://github.com/pouchdb/pouchdb/blob/master/docs/_includes/api/replication.html#L70)

We could possibly make this more clear by clarifying the comment within the example (https://github.com/pouchdb/pouchdb/blob/master/docs/_includes/api/replication.html#L49) and there may be other ways to improve, but I think this duplication is going to make it harder for users to find what they are looking for so closing out, thanks for the PR

cdaringe:
you're right. thx.

PR #5591, opened 2016-08-22
(#5591) - remove extras API (BREAKING CHANGE)

nolanlawson:
Note that this doesn't remove the `dist` plugins (`pouchdb.memory.js` etc.). Instead, this just makes it so that `require('pouchdb/extras/*')` no longer works.

Removing the `dist` plugins may be a longer-term effort and might not even be feasible in 6.0.0. I'm opening a new issue: #5592.


nolanlawson:
Style problems, should be fixed now...

daleharvey:
Very happy to remove these, 👍 when green

gr2m:
that looks like  a proper bug I think? https://travis-ci.org/pouchdb/pouchdb/jobs/154895463#L1594

And I guess the `COMMAND=test-extras` can be removed from the test matrix with this PR?

nolanlawson:
That is probably a bug in the test itself. I'll take a look when I have a moment.

gr2m:
I’m looking into it right now

gr2m:
I pushed to commits that should fix the two failing builds

daleharvey:
Awesome thanks @gr2m, small nit in future if you could avoid the style changes that arent related, our general style would lean towards the original `}, {` however the performance tests need quite a bit of cleaning up anyway so will get addresssed later.

tests/performance/index.js
```diff
@@ -56,7 +56,7 @@ if (startNow) {
   if (!process.browser) {
     // the two strings are to fool Browserify, because this test
     // fails in Node 0.11-0.12 due to sqlite3 being incompatible
-    global.PouchDB.plugin(require('../../packages/node_modules/' +
+    PouchDB.plugin(require('../../packages/node_modules/' +
```
nolanlawson: Not sure what's going on here, but LGTM!

nolanlawson:
Thanks a lot @gr2m, been swamped and needed some help here! ❤️

tests/performance/index.js
gr2m: I thought it might be something that you missed before in the cleanup, I couldn’t find it out myself

gr2m:
You are more than welcome :) We are really looking forward to the 6.0.0 release so that’s the least I can do

Issue #5322, opened 2016-06-09
Can't change ajax parameters on-the-fly


cdaringe:
### Environment (Node.js/browser/hybrid app/etc.)

node.
### Adapter (IndexedDB/WebSQL/LevelDB/etc.)

all.
### Server (CouchDB/Cloudant/Couchbase/PouchDB Server/etc.)

n/a.
### problem statement / request

i need to be able to wrap `pouchdb-ajax`.

currently I can modify the inputs to the request module via `new PouchDB(name, { ajax: { ... } })`.  this is a onetime, static customization of the request.

i need runtime, dynamic ability to change the headers.  our authentication scheme requires headers adjustment periodically on sub-sequent requests.  I used to be able to wrap `PouchDB.utils.ajax` to achieve this.  unfortunately, this capability is now removed.

can we consider restoring the passing of a function to `{ ajax: ... }`?  thanks team!  you guys rock

daleharvey:
So it was likely that I removed this when refactoring the http adapter but surprised this is the first report about it, @yaronyg wasnt this something thali depended on?

I have seen a few use cases for this and in a step towards having the integration tests run purely against the supported API I think it would be good to reintroduce this in a way that supports the tests + users use cases: 

 1. A simple ability to count requests then pass them along as they were
 2. A way to dynamically modify requests, to add a header for example
 3. To return custom data and not make the request at all

yaronyg:
I vaguely remember that we did use this about a billion years ago when we were running in a webview against couchbase. But now we use node so all our work is there and we don't use ajax directly anymore.

nolanlawson:
This can currently be supported by overriding `db._ajax`. It's not a primary use case that we support, though, hence the leading underscore.

Another way you can accomplish this is by using libraries that override `require()` and then sub in their own implementations, e.g. [nock](https://github.com/node-nock/nock) does this. Yet another way is to override the global `XMLHttpRequest` or `fetch` variables in the browser.

Personally I would prefer to close this as a wontfix since I can't think of any easy way to support this, and I'm not sure what the value-add is.

ssured:
I used to have an indicator in my app which showed the upload and download speed of PouchDB syncing data. This was implemented by overriding utils.ajax. It provided a nice extra feedback to the end user about sync state. People seem to like having a little insight into what is happening; the syncing of PouchDB feels like magic to them. Having these counters again would be nice, is it something to add to PouchDB itself?

cdaringe:
[...]

PR #5609, opened 2016-08-29
Specifying custom ajax function

DanielMaly:
This is a very rough attempt at #5322. This is probably not the prettiest way to do it so I'm open to any suggestions as to how to improve it. For me, the most important thing is that it lets me generate OAuth signatures for every outgoing request and seems to be working fine.

I'm having headaches trying to redirect my project to my fork, as the HTTP adapter is a sub-dependency. I would really appreciate it if we could merge this, or a better version of this, into the main repo.

### DIFF
--- a/packages/node_modules/pouchdb-adapter-http/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-http/src/index.js
@@ -167,29 +167,17 @@ function HttpPouch(opts, callback) {
   api._ajax = ajaxCore;
   var ajaxFun = opts.ajaxFun;
 
-  /*function ajax(userOpts, options, callback) {
-    var reqAjax = userOpts.ajax || {};
-    var reqOpts = extend(clone(ajaxOpts), reqAjax, options);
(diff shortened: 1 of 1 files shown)


daleharvey:
Thanks for taking a look at this and sorry for taking so long to review, its a busy month. 

So theres a few issues with this, it doesnt allow the user to handle the request entirely (ie stop if from using `ajax` at all), simultaneously  firing `callback` and `userCb` seems likely to be wrong as well. 

My suggestion would be to start with the tests, this will need tests for various conditions starting with the 3 examples I listed in https://github.com/pouchdb/pouchdb/issues/5322#issuecomment-226635729 would be good, if you dont beat me to it and I get time today I will get an example of the tests done, with failing tests then its a matter of getting them to pass :)

daleharvey:
Sorry for not being able to merge however this doesnt satisfy the requirements that were discussed on the relevant bug, theres a WIP to replace this @ https://github.com/pouchdb/pouchdb/pull/5763, thanks for the PR

Issue #5612, opened 2016-08-30
Execute views in sandbox when in Node.js


KlausTrainer:
For e.g. pouchdb-server it would be cool to have at least some sense of sandbox when executing views. Node.js provides the [`vm`](https://nodejs.org/dist/latest/docs/api/vm.html) module, which we should make use of, maybe.

KlausTrainer:
This has been fixed with https://github.com/pouchdb/pouchdb/commit/5fb785da788ea6688953a2179b5f69416636ede4.

nolanlawson:
This has been fixed for filter/view in pouchdb-core (i.e. `changes()`) but not yet in map/reduce.

daleharvey:
And done https://github.com/pouchdb/pouchdb/commit/1e011ab0cbc8dd07e1d08d84a7d586f7ddea3b92

PR #5613, opened 2016-08-30
(#5612) - change view & filter function execution

KlausTrainer:
This changes both `evalView` and `evalFilter` in pouchdb-core to enforce
strict mode. When running Node.js, we're furthermore executing the
respective function in a sandbox now. We're using the `vm` module
provided by Node.js for that.

BREAKING CHANGE: this may break view or filter functions, as we're now
generally enforcing strict mode for view and filter functions.

### DIFF
--- a/packages/node_modules/pouchdb-core/package.json
+++ b/packages/node_modules/pouchdb-core/package.json
@@ -26,6 +26,8 @@
     "scope-eval": "*"
   },
   "browser": {
-    "./lib/index.js": "./lib/index-browser.js"
+    "./lib/index.js": "./lib/index-browser.js",
+    "./src/evalFilter.js": "./src/evalFilter-browser.js",
+    "./src/evalView.js": "./src/evalView-browser.js"
   }
 }
--- a/packages/node_modules/pouchdb-core/src/evalFilter-browser.js
+++ b/packages/node_modules/pouchdb-core/src/evalFilter-browser.js
@@ -1,0 +1,7 @@
+import scopedEval from 'scope-eval';
+
+function evalFilter(input) {
+  return scopedEval('"use strict";\nreturn ' + input + ';', {});
+}
+
+export default evalFilter;
--- a/packages/node_modules/pouchdb-core/src/evalFilter.js
+++ b/packages/node_modules/pouchdb-core/src/evalFilter.js
@@ -1,7 +1,9 @@
-import scopedEval from 'scope-eval';
+import vm from 'vm';
 
 function evalFilter(input) {
-  return scopedEval('return ' + input + ';', {});
+  var code = '(function() {\n"use strict";\nreturn ' + input + '\n})()';
+
+  return vm.runInNewContext(code);
 }
 
 export default evalFilter;
--- a/packages/node_modules/pouchdb-core/src/evalView-browser.js
+++ b/packages/node_modules/pouchdb-core/src/evalView-browser.js
@@ -1,0 +1,22 @@
+import scopedEval from 'scope-eval';
+
+function evalView(input) {
+  var code = [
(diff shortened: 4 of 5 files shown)


daleharvey:
This looks great, I had no idea node had a specific capability to execute JS this way. I am 👍 although I would like @nolanlawson to have a check first, I feel like we may possibly want to have an `evalView.js` and `evalView-browser.js` so we have specific files for each env, typically we try to avoid switches in the code and do that via replacing files in the preprocessor

KlausTrainer:
Until recently, I've been unaware of the [`browser` field spec](https://github.com/defunctzombie/package-browser-field-spec) and mostly didn't bother about browserification ^^

But yeah, splitting it up and making use of the `browser` field here seems like a good idea to me as well, and I've changed this accordingly.

Thanks!

nolanlawson:
So a few things:

1. Agree about the browser/node split; this is how we do things to better measure coverage and keep browser size small
2. I don't see the point in `"use strict'` for the browser TBH
3. For node, `vm` makes a huge amount of sense, it's better security-wise
4. There's a critical flaw here in moving from `scope-eval` to `Function` however, which is that because `Function` is in the top scope, we will suddenly have a terrible Uglify performance due to Rollup hoisting. This can be solved by creating an npm module that exports `Function` as a commonjs module; it's a bit of a hack but it's pretty crucial to avoid Uglify de-optimizations.

So anyway everything looks good to me except (2) which I don't understand the reasoning behind it and (4) which is a pretty severe performance decrease.

nolanlawson:
Also we're about to release 6.0.0 and it would be nice to get a breaking change in there before then, but IMO if it were just `vm` this would not be a breaking change but rather a patch release (to fix security), whereas `'use strict'` is indeed a breaking change.

KlausTrainer:
Thanks for the feedback, @nolanlawson :)

To me it sounds like (4) should be easy to fix. Like creating a module in `pouchdb-utils` that just exports `Function` and has a comment explaining this (I'd basically take your explanation). Right?

Regarding (2): I wonder why we wouldn't want to have strict mode (at least here). It improves language semantics and helps avoiding some subtle bugs. Furthermore, it makes some optimizations possible.

Maybe we're having different opinions on (2). However, I don't want to hold too strongly onto my opinion here; particularly given the fact that using `vm` in Node is anyway the more important issue ;)

I'd like to leave deciding on (2) up to you. Please tell me, and I'll change it accordingly.

nolanlawson:
I'm fine with (2), but agree it's a breaking change and just wanted to know the reasoning. :)

Unfortunately it's not enough to create a pouchdb-utils function because those will end up getting hoisted by Rollup into the same scope as everything else, defeating Uglify. Easiest is to create a totally separate module on npm that exports the `Function function`; then it will be excluded by Rollup. This is what `scope-eval` was buying us.

nolanlawson:
In fact if you just continued using `scope-eval` in the Browser, that would solve the problem.

KlausTrainer:
All right! Thanks for explaining 💯

KlausTrainer:
Just changed it accordingly. Note that unlike `evalFilter`, in the previous version of `evalView` we didn't use `scope-eval` but `Function` instead. I've now changed it accordingly, so that in `evalView-browser` we now use `scope-eval` as well.

nolanlawson:
LGTM! :+1: when green, should be able to make it into the 6.x release.

daleharvey:
Awesome job, yeh getting this fixed and in 6 is great as its been pointed out as being fairly worrying security wise, thanks @KlausTrainer + @nolanlawson

KlausTrainer:
Thanks @daleharvey and @nolanlawson! 🚀

PR #5622, opened 2016-09-03
(#5622) - update deps, fix duplicate deps

nolanlawson:
Two things:

1. During our hiatus from Greenkeeper, many deps became out-of-date.
2. When I removed Lerna, I mistakenly included `request` and `memdown` as both devDeps and regular deps.

Also, this updates `request` to the latest version which actually fixes the `tough-cookie` warning that you keep seeing if you install the old version:

    npm WARN deprecated tough-cookie@2.2.2: ReDoS vulnerability parsing Set-Cookie https://nodesecurity.io/advisories/130


nolanlawson:
(Ignore the branch name; issue 5592 has nothing to do with this; just an error on my part.)

PR #5624, opened 2016-09-03
(#5624) - add beta release script

nolanlawson:
I figured we needed a more automated way to do beta releases; it's too error-prone if you have to modify the script before doing one.


nolanlawson:
There's also a bug in the release script that this fixes, which is that `lerna.json` no longer exists and so we have to exclude it when we `git add`. I'm glad I caught this now, because this would have messed up the Bower release.

daleharvey:
This is great, thanks

PR #5627, opened 2016-09-03
(#5627) - remove pouchdb-http (BREAKING CHANGE)

nolanlawson:
There's no reason to have this as a first-party package. Nothing depends on it, and it merely bundles up some plugins. We can easily publish it third-party, as I did with [pouchdb-memory](https://github.com/nolanlawson/pouchdb-memory).

I'd also like to modify it to add support for `pouchdb-mapreduce`. This is the kind of thing that becomes very easy to do once it's no longer in the main monorepo.


nolanlawson:
Once we agree to remove this, I'll just copy the code over to `nolanlawson/pouchdb-http`, republish it with the updated repo link, and we'll be done.

daleharvey:
Yup happy enough for this to be 3rd party, cheers

PR #5628, opened 2016-09-04
(#5628) - attempt to fix couch master tests

nolanlawson:
Based on [discussion in the mailing list today](https://mail-archives.apache.org/mod_mbox/couchdb-dev/201609.mbox/%3CCA%2BvSmEaEH%3D-h6zV6uAjQ3UN4NpvanAdSVTqyZHOLhp2Fi4siwg%40mail.gmail.com%3E), we may need to set an artificial timeout in order to actually test CouchDB 2.

I tried to test RC4 today but was getting ECONNRESET. Dunno if this will fix the tests, but hopefully will.

### DIFF
--- a/tests/integration/utils.js
+++ b/tests/integration/utils.js
@@ -116,7 +116,13 @@ testUtils.cleanup = function (dbs, done) {
   var num = dbs.length;
   var finished = function () {
     if (--num === 0) {
-      done();
+      if (testUtils.isCouchMaster()) {
+        // We can't be sure when a DELETE will actually destroy
+        // a database in Couch 2, so we wait.
+        setTimeout(done, 1000);
+      } else {
+        done();
+      }
     }
   };
 


nolanlawson:
Hm, it kinda seems like a true failure. This test passes consistently against CouchDB 1.6, but not CouchDB 2.0-rc4.

```
  suite2 test.attachments.js- http:local
    ✓ Many many attachments replicate (383ms)
    ✓ Many many png attachments replicate (389ms)
    1) #3961 Many attachments on same doc


  8 passing (8s)
  1 failing

  1) suite2 test.attachments.js- http:local #3961 Many attachments on same doc:
     Error: connect ECONNRESET 127.0.0.1:6984
      at Object.exports._errnoException (util.js:1026:11)
      at exports._exceptionWithHostPort (util.js:1049:20)
      at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1081:14)
```

This is [a test that puts 100 attachments on a document](https://github.com/pouchdb/pouchdb/blob/9c46dec6717182d93f3992de8235ed60051fac3f/tests/integration/test.attachments.js#L3565-L3608).

To repro:

    COUCH_HOST=http://127.0.0.1:6984 GREP=Many npm t

nolanlawson:
Passing in Node in Travis but failing in Firefox, seems genuine:

```
{ lastPassed: '#3962 - Test many attachments',

  failures: 

   [ { stack: '',

       title: 'Reporting write failures (#942)',

       message: 'timeout of 100000ms exceeded. Ensure the done() callback is being called in this test.' } ],

  browser: 'Mozilla/5.0 (X11; Linux x86_64; rv:38.0) Gecko/20100101 Firefox/38.0',

  passed: 764,

  failed: 1,

  logs: 

   [ [ 'error', 'Error in .on("change", function):', [Object] ],

     [ 'error', 'Error in .on("change", function):', [Object] ] ] }
```

daleharvey:
@willholley was thinking about trying `q=1` (I dont know what that is, but its been mentioned it may help), also if that doesnt work I think an `generatedbname() { if (iscouchmaster) { name = name + random` would be better than timeouts (1 second to every delete, we do a huge amount of deletes)

willholley:
it needs rebasing but my previous attempt at this is at https://github.com/pouchdb/pouchdb/tree/4575-couchdb-master-unique-db-names if that's useful.

willholley:
I've also found CouchDB 2 to hit file descriptor limits earlier due to sharding defaults. Regarding the attachment error, if memory serves, this is one that went away after I increased the default file descriptor limit.

nolanlawson:
Oddly the test only failed in the browser, not in Node. I am trying to re-run Couch locally but somehow my docker installation got messed up.

nolanlawson:
Still failing, but only the browser (Firefox) test, not the Node test. I'm not sure why: https://travis-ci.org/pouchdb/pouchdb/jobs/157386420

nolanlawson:
So @daleharvey I still think this is reasonable to merge; we don't really have a better solution than "if it's Couch master, insert an artificial delay." It's a crappy test, but if Couch doesn't have support for actually telling us when a DB was destroyed, then there's not much we can do.

daleharvey:
I would be far more comfortable with merging https://github.com/pouchdb/pouchdb/commit/1f67f156062ec306a18e0b02f4935ff0a9485097 but refactoring it to only change in `adapterUrl` in the case that it is couch master, its a small difference but while its still passing with the artificial delay its still a bit more of a hassle testing etc, I can take a look at this in the next few days, now that couch is getting closed to releasing I would like us to get this green and out of allowed failures

nolanlawson:
I agree, that's a good idea.

daleharvey:
Opened a new PR that does the name change, passing in node and failing in the browser right now - https://github.com/pouchdb/pouchdb/pull/5734

PR #5629, opened 2016-09-04
(#5612) - add vm to ignored deps

nolanlawson:
Rollup ignores this anyway, but it's nice to be explicit so that we don't get an additional warning message during the build.


daleharvey:
We no longer have this due to https://github.com/pouchdb/pouchdb/commit/98e311cc2c1d2f4c497c1f038feb7553c92073d4

nolanlawson:
I'm going to auto-+1 this because it's a pretty small+obvious fix and really just removes a warning anyway.

PR #5630, opened 2016-09-04
(#5630) - use rollup-prepublish for less build config

nolanlawson:
I got tired of configuring so many different options for Rollup, so I wrote [rollup-prepublish](https://github.com/nolanlawson/rollup-prepublish) which basically just pre-configures Rollup to use it the way we use it (to bundle into CommonJS).

In particular, this means we no longer have to keep a running list of dependencies that should be treated as external CommonJS. So this reduces the maintenance burden because we don't have to add a new item to the list every time we add a new dep.


daleharvey:
This looks nice and clean, cheers 👍

Issue #4010, opened 2015-06-29
Firefox Private Browsing causes InvalidStateError / 500 Internal Error


rarkins:
PouchDB fails with a 500 Internal Error if launched with Firefox in Private Browsing mode.

From reading issues in other repositories (e.g. [here](https://github.com/mozilla/localForage/issues/70)), it sounds like indexedDB fails ungraciously when in a private window. Would it be appropriate for PouchDB to have a more advanced check for indexedDB support before attempting to use it?

daleharvey:
I have added a more more visible warning for private browsing which is about the best we can do at the moment - https://github.com/pouchdb/pouchdb/commit/9ecf016aeb55bf228d7ac77cb39ca674dcd4c955

oliviertassinari:
For anyone looking for a work-around to this issue, here is mine:
```js
  new Promise((resolve) => {
    const pouchdbOptions = {};
    let dbLocal = {};

    // ... Some iOS adapter detection.

    if (typeof window !== 'undefined' && window.indexedDB && /Firefox/.test(window.navigator.userAgent)) {
      const request = window.indexedDB.open('test_firefox_pm');

      /**
       * We are most likely under the buggy private mode.
       * IndexDB is not available.
       */
      request.onerror = () => {
        pouchdbOptions.adapter = 'localstorage';
        PouchDB.plugin(require('pouchdb-adapter-localstorage'));

        dbLocal = new PouchDB('db', pouchdbOptions);
        resolve(dbLocal);

        return true; // It's preventing `InvalidStateError` and `UnknownError` exceptions.
      };
      request.onsuccess = () => {
        dbLocal = new PouchDB('db', pouchdbOptions);
        resolve(dbLocal);
      };
    } else {
      dbLocal = new PouchDB('db', pouchdbOptions);
      resolve(dbLocal);
    }
  };
```

nolanlawson:
I opened a new issue here fwiw: #5641

PR #5632, opened 2016-09-04
Prevent from throwing in firefox private mode

oliviertassinari:
This is a continution of #4010.
I'm seeing a lot of `InvalidStateError` error from users accessing my app in the private mode of Firefox.
I don't think that throwing an error here have much value.

I'm using https://github.com/localForage/localForage/issues/195#issuecomment-185135074 as inspiration.

- [ ] We need to upgrade the pouchdb docs error section if we move forward.

### DIFF
--- a/packages/node_modules/pouchdb-adapter-idb/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-idb/src/index.js
@@ -970,6 +970,8 @@ function init(api, opts, callback) {
     var msg = 'Failed to open indexedDB, are you in private browsing mode?';
     guardedConsole('error', msg);
     callback(createError(IDB_ERROR, msg));
+    
+    return true; // It's preventing `InvalidStateError` and `UnknownError` exceptions.
   };
 }
 


nolanlawson:
@daleharvey I saw [your comment on the Firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=781982#c32) from last year, I thought this was something we already handled?

daleharvey:
We run that `onerror` code and warn the user, but there is an additional InvalidateState shown by the browser, this is what the PR removes

daleharvey:
So the current error when the user tries to open a PouchDB in private browsing mode is: 

    > Failed to open indexedDB, are you in private browsing mode?  webrunner.js:100:9
    > 12:44:18.690 InvalidStateError  pouchdb.js:7574:10

This removes the `InvalidStateError` from the message, however it also seems to have a side effect of letting the code continue (a bunch of tests pass in pb mode with the patch that fail without it).

We already give a warning that specifically calls our private browsing, not sure why we would want to hide the InvalidStateError, and would definitely need to know what side effects this `return true` introduces

daleharvey:
Also this `onerror` is not only triggered from private browsing mode and while we can do hacks to detect pb mode they are fundementally broken (pb mode is supposed to be indetectable). We absolutely do not want to be hiding other errors from the user

oliviertassinari:
I'm gonna close this PR. As you pointed out, I don't think it's going in the right direction.

One better approach could be to have a idb issue detection code (async using the user agent and a simple idb operation.
Then failing back to localstorage.
I might implement this on my userland.

daleharvey:
You should be able to do something like:

    var db = new PouchDB('name');
    db.info().catch(function(err) { 
      // Failed to open database, do 
      db = new PouchDB('name', {adapter: 'localstorage'});
    });

nolanlawson:
If there were a way for us to detect Firefox's private browsing mode in a _synchronous_ way, then we could modify idb's `valid()` function to return false in that case. That's the more standard way for adapters to fall back when they're running in an unsupported environment. `db.info().catch(...)` definitely works for now, but ideally adapters should fall back synchronously.

Issue #5635, opened 2016-09-05
Cannot find module "leveldown" in pouchdb-adapter-memory


nolanlawson:
Seems like we misconfigured something in v6. If I try to use `pouchdb-adapter-memory` in a browser, I get:

```
Error: Cannot find module 'leveldown' from '/Users/nolan/workspace/pouchdb-authentication/node_modules/pouchdb-adapter-memory/lib'
    at /Users/nolan/workspace/pouchdb-authentication/node_modules/resolve/lib/async.js:46:17
    at process (/Users/nolan/workspace/pouchdb-authentication/node_modules/resolve/lib/async.js:173:43)
    at ondir (/Users/nolan/workspace/pouchdb-authentication/node_modules/resolve/lib/async.js:188:17)
    at load (/Users/nolan/workspace/pouchdb-authentication/node_modules/resolve/lib/async.js:69:43)
    at onex (/Users/nolan/workspace/pouchdb-authentication/node_modules/resolve/lib/async.js:92:31)
    at /Users/nolan/workspace/pouchdb-authentication/node_modules/resolve/lib/async.js:22:47
    at FSReqWrap.oncomplete (fs.js:123:15)
Error: Cannot find module 'leveldown' from '/Users/nolan/workspace/pouchdb-authentication/node_modules/pouchdb-adapter-memory/lib'
```

nolanlawson:
OK, so for those following along at home, `rollup-prepublish` unconditionally bundles _all_ modules that expose a `jsnext:main`, so since `pouchdb-adapter-memory` relied on `pouchdb-adapter-leveldb-core`, and the latter had a `"browser"` switch but the former did not, then in the `rollup-prepublish` version `pouchdb-adapter-memory` bundled the _Node_ version into its `lib/index.js`, which caused Browserify to freak out because it didn't see the `"leveldown"` dependency.

I'm starting to think that there is no configuration I could make for `rollup-prepublish` to suit PouchDB's unique needs, so we may just need a custom Rollup build after all.

nolanlawson:
fixed in #5636

PR #5636, opened 2016-09-05
(#5635) - Revert "(#5630) - use rollup-prepublish for less build config"

nolanlawson:
This reverts commit 98e311cc2c1d2f4c497c1f038feb7553c92073d4.

Unfortunately `rollup-prepublish` caused some subtle changes in the build structure, e.g. in how modules are bundled and therefore how the `"browser"` key is treated. It's best to just revert for now, in my opinion.


daleharvey:
👍

PR #5638, opened 2016-09-05
Fix docs markdown

nolanlawson:
I had to revert these two commits in order to fix the markdown issues in #5637. Maybe @williambelle when you have a chance you can look into this? If not, for now I've just done a new `npm run publish-site` with the reverts.

PR #5640, opened 2016-09-06
docs update: #5154 deprecated params of put

mnewmedia:
I also noticed that the title `using db.put (h3)` is on the same line like the title `create/update a document (h2)` removing the css `display: inline-block` would help, but maybe introduces other unwanted changes..


daleharvey:
Yeh the other issue is being tracked @ https://github.com/pouchdb/pouchdb/pull/5638, was fixed but regressed other parts, this is a good catch I should have noticed it, thanks

PR #5645, opened 2016-09-07
(#5645) - prefer immediate to process.nextTick() for better perf

nolanlawson:
Based on my understanding of `process.nextTick()` vs `setTimeout()` vs microtasks, I see no reason why we wouldn't want to prefer `immediate` (which uses MutationObsevers under the hood to create microtasks) to `process.nextTick()`.

Looking at PouchDB in a JS profiler (e.g. Chrome flamegraphs), I see a lot of time being spent idle, just waiting for `drainQueue()` (i.e. the `process.nextTick()` shim). That time could be better spent executing microtasks, so that PouchDB can execute more quickly.

The only downside of microtasks is the potential to starve the DOM, but in those cases people should just be putting PouchDB into a web worker anyway. The only exception I found in our codebase where it made sense not to use `immediate` was in the batched md5 summing, because in that case if we used `immediate` then there would be no point in breaking up things into batches, because the DOM would never have a chance to breathe. Hence I kept @calvinmetcalf's original choice of `global.setImmediate || global.setTimeout`, since `setImmediate` also gives the DOM a chance to breathe but without waiting a tick.

I'm waiting to see if this passes in Travis; if so, I'll do some perf tests to see the improvement. Also note that I'm using a slightly older version of `immediate`, because it's a smaller version of `immediate` that matches our current version of `lie` and thus doesn't increase the bundle size.

### DIFF
--- a/bin/external-deps.js
+++ b/bin/external-deps.js
@@ -8,7 +8,7 @@ module.exports = [
   // main deps
   'argsarray', 'debug', 'double-ended-queue', 'es3ify', 'es6-promise-pool',
   'fruitdown', 'inherits', 'js-extend', 'level-write-stream', 'levelup', 'lie',
-  'localstorage-down', 'memdown',
+  'localstorage-down', 'memdown', 'immediate',
   'request', 'scope-eval', 'spark-md5', 'through2',
   'vuvuzela', 'leveldown', 'websql',
   // core node deps
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "double-ended-queue": "2.1.0-0",
     "es6-promise-pool": "2.4.4",
     "fruitdown": "1.0.2",
+    "immediate": "3.0.6",
     "inherits": "2.0.1",
     "js-extend": "1.0.1",
     "level-codec": "6.2.0",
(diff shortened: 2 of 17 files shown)


nolanlawson:
Chrome 52, temp-views test, 10 iterations:

- Before: 103721ms, After: 102499ms (1.1% faster)

Firefox 48, temp-views test, 10 iterations:

- Before: 99222ms, After: 99166ms (0.05% faster, no statistical difference)

Big improvement for Chrome, pretty much nothing for Firefox.

nolanlawson:
This needs a lot more discussion and research.

calvinmetcalf:
the version of immediate lie uses doesn't handle errors in the callback so any exception thrown in any callback **will break everything**, seriously immediate and lie will just stop working.  Since lie only ever runs code already wrapped in a try catch, that version of immediate has the error handling stuff removed to save space

nolanlawson:
Ah, that makes sense. Seems a bad idea to use that version of immediate, then. Although odd that the test suite still passes.

nolanlawson:
I prefer to defer to @calvinmetcalf's judgment here, at least until I have evidence that using `immediate` is better than `process.nextTick`.

calvinmetcalf:
so immediate vs nextick is not a big deal it should lower latency in some situations so go nuts (except places like md5 obvs) just use the most recent version of it as we run a lot of callbacks via nextTick so if a user uses the callback api and throws an error in the callback 'bad things' will happen in the version with lie.

nickcolley:
👍  thanks for the PR though, learnt a lot! 🙏

Issue #5637, opened 2016-09-05
Markdown is unformatted in the docs page


nolanlawson:
I'm not sure how this happened, but we've got messed-up formatting again: https://twitter.com/zaherg/status/772860396105502720

nolanlawson:
![screenshot 2016-09-05 13 11 13](https://cloud.githubusercontent.com/assets/283842/18256448/63864bf0-736a-11e6-876c-6736548423c2.png)

nolanlawson:
I'm not sure what's going on, but reverting both 88d0db396731b2e39d4efee33bc17b92904ee0aa and 5c8289bb4d1c53cd89e7ba4a8f355b2773b0c196 fixes the issue. It seems to be caused by the "alert" divs for some reason.

nolanlawson:
I tried converting the alerts from `.html` to `.md` but it didn't fix it. Unfortunately I don't have time to look at this right now.

nolanlawson:
fixed in https://github.com/pouchdb/pouchdb/pull/5648

PR #5648, opened 2016-09-07
Fix docs markdown

williambelle:
I think that this should fix all issues with markdown and docs page (#5637).


nolanlawson:
Awesome, I gave this a quick look-over, and everything looks great to me. There's one other visual bug I noticed, but it seems to predate all these PRs, so I'll file it separately. Thanks!

nolanlawson:
Opened a separate issue: https://github.com/pouchdb/pouchdb/issues/5654

PR #5661, opened 2016-09-10
(#5612) - execute map/reduce functions in a sandbox in Node

nolanlawson:
Follow-up to #5612 and #5613. All arbitrary code should be executed in a sandbox, and with this PR, we're now doing it for map/reduce functions as well as for view/filter from `changes()`. I grepped the code and didn't see any other usages of `eval()` or `Function()`, so this should be the last fix needed for sandboxing.


nolanlawson:
Crap, this breaks because of Istanbul code coverage breaking once the function is stringified.

packages/node_modules/pouchdb-mapreduce/src/evalfunc.js
```diff
@@ -1,16 +1,17 @@
...
-      toJSON: toJSON
-    }
-  );
+var evalFuncToExport;
+// The "stringify, then execute in a VM" strategy totally breaks Istanbul due
+// to missing __coverage global objects. As a solution, export different
+// code during coverage testing and during regular execution.
+// Note that this doesn't get shipped to consumers because Rollup replaces it
+// with rollup-plugin-replace, so users just get if (false) {...} else {...}
+/* istanbul ignore else */
```
nolanlawson: This ought to fix the coverage tests ^

nolanlawson:
Should be good now

nolanlawson:
Coverage still failing, ug.

nolanlawson:
OK, _finally_ got coverage passing, apparently Travis needed an extra `COVERAGE=1` for some reason. I'm confident this PR will finally be green.

daleharvey:
heh that COVERAGE issue looked pretty tricky, awesome job getting to the bottom, 👍

KlausTrainer:
Looks like I've only done the easy parts of the issue, and missed the actual challenges 😄 

But seriously, I should have checked more thoroughly, especially given the fact that this issue affects security 😞 

Anyway, I'm happy this has landed. Thanks!

I also had a look at the diff, which looks good to me as well 👍

There's just one detail I'm wondering about: Can we avoid the hack with parsing and stringifying JSON (https://github.com/pouchdb/pouchdb/commit/1e011ab0cbc8dd07e1d08d84a7d586f7ddea3b92#diff-eeac4723352c2619b7de767d7854a0b5R52) by using a plain `assert` instead of a `should` assertion for the corresponding test (and adding a respective comment that explains why we do so)?

nolanlawson:
Yes, we would be able to avoid the hack if we did that. And in fact we probably should, because as-is it's most likely a perf regression (although this would need to be tested - since in V8 structured cloning can be very expensive, sometimes it's actually better to stringify, see [my blog post on web worker messages](https://nolanlawson.com/2016/02/29/high-performance-web-worker-messages/) for some discussion of this.

So I would say: 1) do a quick perf sanity check to make sure it's worth it, and then 2) yeah, we can swap out `chai.should` in this case for any other library, such as `assert`.

KlausTrainer:
I did some very basic performance measurement, running only the `mapreduce` test group with `time`. In doing so, I compared the current version with a new version where I removed the stringify/parse hack and adjusted the respective tests.

After several runs, I came to the conclusion that there doesn't seem to be a big difference, at least not when measuring with our mapreduce tests.

Of course it would be possible to do some more precise measurement, which however doesn't seem to be worth the effort to me.

I think that it would be good to get rid of the stringify/parse hack anyway, and made a respective pull request (#5698).

KlausTrainer:
More precisely, the results of my performance measurement didn't show any significant differences at all.

PR #5663, opened 2016-09-10
(#5663) - remove unnecessary NotFoundError allocation

nolanlawson:
Did some profiling for the in-memory adapter, found that these allocations eat up a lot of time. After this fix, they don't. There's no need to wrap the error BTW because the leveldown spec only requires that the error name be `'NotFound'`.


daleharvey:
Sounds good to me, 👍

PR #5667, opened 2016-09-10
(#5612) - actually silence "vm" warnings

nolanlawson:
Turns out `'vm'` is a built-in, which is why the previous fix to silence the warning wasn't working.

Turns out we can actually just ignore all built-ins and top-level dependencies now, which simplifies our build scripts a lot.


bin/build-module.js
```diff
@@ -30,20 +30,7 @@ function buildModule(filepath) {
...
   // be skipped by Rollup. We may revisit this later.
   var depsToSkip = Object.keys(topPkg.dependencies || {});
-  depsToSkip = depsToSkip.concat([
-    'crypto', 'fs', 'events', 'path'
-  ]);
-
-  if (pkg.name === 'pouchdb-for-coverage') {
-    // special case - for the coverage reports, the whole thing is
-    // bundled into one index.js. so we don't want to externalize
-    // the pouchdb repos
```
nolanlawson: I'm not sure why we were doing this, but it turns out we don't need to. I checked it out, and we're still getting 100% code coverage, although will need to check that we're not ignoring more LoC now.

nolanlawson: Okay, `pouchdb-for-coverage/lib/index.js` and `pouchdb/lib/index.js` are both the same before and after this change, so LGTM.

nolanlawson:
Actually there may be a more fundamental issue here, will investigate.

Issue #5626, opened 2016-09-03
`pouchdb-browser` is less efficiently bundled than `pouchdb`


nolanlawson:
Do `npm install pouchdb`, and you get the nice Rollup-ed version. Yay!

Do `npm install pouchdb-browser`, and you get the unbundled version that ends up being about 3-4kB larger and takes longer to `npm install` due to the extra dependencies. Womp womp.

I think to fix this, we could just Rollup both `pouchdb-browser` and `pouchdb-node`, and then have `pouchdb` simply depend on the two. There will be a teensy bit of extra overhead because Browserify/Webpack will consider it one extra module, but on the whole the difference should not be that great.

nolanlawson:
Fixed in #5681

PR #5653, opened 2016-09-08
(#5653) - automatically determine dependencies at publish time

nolanlawson:
Problems:

1. We have to declare our dependencies in two places: top-level `package.json` and individual sub-package `package.json`s.
2. These can get out-of-date. For instance, I noticed `pull-stream` is no longer needed, and the core `pouchdb` package still had some dependencies from the legacy `extras` (e.g. `localstorage-down`).

Solution:

1. Script, `bin/update-dependencies.js` that runs during the release script
2. This script searches `lib/**/*.js` and finds all `require()` calls
3. Then it inserts `depenedencies` and `peerDependencies` with the correct verisons.

Thus we no longer need to declare dependencies inside of the sub-packages. I have manually inspected the output of `node bin/update-dependencies.js` and it looks good to me. I also removed `pull-stream` since it's no longer needed.

PR #5669, opened 2016-09-10
(#5612) - silence "vm" warnings, verify build, fix build deps

nolanlawson:
In the course of trying to silence `vm` warnings (#5612), I realized that #5653 actually introduced a subtle problem whereby sub-packages are aggressively bundled (e.g. `pouchdb-mapreduce` will bundle `pouchdb-utils` into itself), which would be a problem for custom builds because presumably in those cases people want module-deduping more than they want rollup-inlining.

In the course of fixing that issue, I also went ahead and added a new test, `npm run verify-build` (replacing and enhancing `npm run verify-bundle-size`), which checks that the various sub-packages were built correctly. This should avoid accidentally introducing a change that breaks this in the future.


TESTING.md
```diff
@@ -159,6 +159,12 @@ There's a WebSQL storage quota test available in:
 
 Run `npm run dev`, then open it in Safari or iOS.
 
+### Build tests
+
+To verify that the build was done correctly, there are some tests here:
+
+    npm run verify-build
+
```
nolanlawson: verify-bundle-size was not documented

daleharvey:
The extra test is great, sounds like this has been a hassle so good to get some assurance it wont regret, needs rebasing however but 👍

nolanlawson:
Yeah also the added test suite makes it so that if I fix #5626 I'll be able to add a test to ensure it actually worked.

nolanlawson:
Squash-merged as 1a589fdd5e0a183354e380a0eb84734b07972dc4

nolanlawson:
Oops, #5612 was the wrong issue to link this to.

nolanlawson:
Made a mistake during rebase, fixed it with 4cd2a1fa88a8b3de0f4dfe731620c89d46e1c5d8

PR #5677, opened 2016-09-13
(#5663) - use same error for NotFoundError

nolanlawson:
Occurred to me that this error would be more useful if it had a reliable stack trace (which we lose if it's only declared once). Passing through the error from the sub-level itself should accomplish this, while also reducing the overall code size.


nolanlawson:
As it turns out, this will fail, because the underlying adapter doesn't actually pass through the NotFoundError correctly. >_<

PR #5681, opened 2016-09-14
(#5626) - aggressively bundle pouchdb-browser/pouchdb-node

nolanlawson:
With this change, `pouchdb-browser` and `pouchdb-node` really give you exactly the same thing (respectively) as `pouchdb`. You no longer pay the bundler tax by switching from `pouchdb` to `pouchdb-browser`.

I also had to go in and update our `jsnext:main` for `pouchdb-browser`/`pouchdb-node`, in the odd case that someone is actually using ES6 modules and wants the ES6 version. `verify-dependencies.sh` was also updated to ensure it tests for the proper dependencies.

### DIFF
--- a/bin/build-modules.js
+++ b/bin/build-modules.js
@@ -31,15 +31,9 @@ function buildPackage(pkg) {
 }
 
 readDir('packages/node_modules').then(function (packages) {
-  // `pouchdb-for-coverage` has to be built last because it depends on
-  // `pouchdb`, which has a particular index.es.js file as its jsnext:main
-  return Promise.all(packages.filter(function (pkg) {
-    return pkg !== 'pouchdb-for-coverage';
-  }).map(buildPackage)).then(function () {
-    return buildPackage('pouchdb-for-coverage');
+  return Promise.all(packages.map(buildPackage)).catch(function (err) {
+    console.error('build error');
+    console.error(err.stack);
+    process.exit(1);
   });
-}).catch(function (err) {
-  console.error('build error');
-  console.error(err.stack);
-  process.exit(1);
 });
--- a/bin/verify-dependencies.js
+++ b/bin/verify-dependencies.js
@@ -30,9 +30,15 @@ getReqs('pouchdb/lib/index-browser.js').should.not.contain('pouchdb');
 getReqs('pouchdb/lib/index-browser.js').should.not.contain('leveldown');
 getReqs('pouchdb/lib/index-browser.js').should.not.contain('pouchdb-core');
 
-// sub-dependencies are not so aggressively bundled (for now, anyway)
-getReqs('pouchdb-node/lib/index.js').should.contain('pouchdb-core');
-getReqs('pouchdb-browser/lib/index.js').should.contain('pouchdb-core');
+// pouchdb-node and pouchdb-browser are also aggressively bundled
+getReqs('pouchdb-node/lib/index.js').should.not.contain('pouchdb-core');
+getReqs('pouchdb-node/lib/index.js').should.not.contain('pouchdb-promise');
+getReqs('pouchdb-node/lib/index.js').should.contain('leveldown');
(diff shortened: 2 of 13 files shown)


nolanlawson:
Another benefit of this approach is that `pouchdb-browser` and `pouchdb-node` will be faster to `npm install`, because there are fewer total dependencies.

nolanlawson:
The coverage tests seem to be failing, probably due to a bug in the build script. I'll look into it.

packages/node_modules/pouchdb-for-coverage/package.json
```diff
@@ -2,8 +2,5 @@
   "name": "pouchdb-for-coverage",
   "version": "6.1.0-prerelease",
   "main": "./lib/index.js",
-  "private": true,
-  "browser": {
-    "./lib/index.js": "./lib/index-browser.js"
-  }
```
nolanlawson: pouchdb-for-coverage is always only used in Node mode; this wasn't doing anything

packages/node_modules/pouchdb/src/index.js
```diff
@@ -1,2 +1,2 @@
-import PouchDB from 'pouchdb-node';
+import PouchDB from './pouchdb';
```
nolanlawson: made this two separate files because `"browser"` is funky and this is the only way I could get it to work when you're importing from subfolders within an external package.

bin/build-modules.js
```diff
@@ -31,15 +31,9 @@ function buildPackage(pkg) {
...
-  // `pouchdb-for-coverage` has to be built last because it depends on
-  // `pouchdb`, which has a particular index.es.js file as its jsnext:main
-  return Promise.all(packages.filter(function (pkg) {
-    return pkg !== 'pouchdb-for-coverage';
-  }).map(buildPackage)).then(function () {
-    return buildPackage('pouchdb-for-coverage');
+  return Promise.all(packages.map(buildPackage)).catch(function (err) {
+    console.error('build error');
+    console.error(err.stack);
+    process.exit(1);
```
nolanlawson: There was no need to do these in any particular order anymore, since all the aggressively-bundled packages read from the `src` of another package. This has the additional benefit of faster builds.

bin/verify-dependencies.js
```diff
@@ -30,9 +30,15 @@ getReqs('pouchdb/lib/index-browser.js').should.not.contain('pouchdb');
...
-getReqs('pouchdb-browser/lib/index.js').should.contain('pouchdb-core');
+// pouchdb-node and pouchdb-browser are also aggressively bundled
+getReqs('pouchdb-node/lib/index.js').should.not.contain('pouchdb-core');
+getReqs('pouchdb-node/lib/index.js').should.not.contain('pouchdb-promise');
+getReqs('pouchdb-node/lib/index.js').should.contain('leveldown');
+getReqs('pouchdb-browser/lib/index.js').should.not.contain('pouchdb-core');
+getReqs('pouchdb-browser/lib/index.js').should.not.contain('pouchdb-promise');
+getReqs('pouchdb-browser/lib/index.js').should.contain('lie');
+
+// other sub-dependencies are not so aggressively bundled (for now, anyway)
```
nolanlawson: Actual tests to ensure this PR does what I claim it's doing. ;)

nolanlawson:
Green! 🍏

nolanlawson:
I did a quick analysis. Here is the list of [slow-deps](https://www.npmjs.com/package/slow-deps) for the `pouchdb` package, and in red are the ones that are removed in `pouchdb-browser`:

screenshot 2016-09-15 21 29 08

Note that memdown, localstorage-down, and fruitdown are all mistakes; they shouldn't be in there. I'll open a separate PR to remove them.

willholley:
+1

nolanlawson:
Thanks @willholley !

PR #5682, opened 2016-09-14
(#5682) - find requires(), ignore imports

nolanlawson:
I noticed that `node bin/update-dependencies.js` was failing due to new ES6 module code in `lib/`. The parser used by `find-requires` will throw if it encounters an `import` keyword. This fixes it so that such errors are ignored. (We don't need to look at `import`s because there's always an equivalent `require()` in the CJS code.)

I added `[skip ci]` because this script never runs during build/test, only during publish.


garrensmith:
+1

nolanlawson:
Thx!

PR #5683, opened 2016-09-14
(#5322) - update pouchdb-ajax to exec preRequest and preResponse funct…

martinfridrich:
This allow execute `preRequest` and `preResponse` functions defined in replication `options.ajax`.

```
var replication = PouchDB.replicate(source, targer, {
  live: true,
  ...
  ajax: {
    preRequest: function(options) {
      return new Promise(function(resolve, reject) {
        ...
        //options are passed to request
        resolve(options)
      })
    },
    preResponse: function(err, response, data) {
      return new Promise(function(resolve, reject) {
        ....
        //err, response, data - are from request() callback and passed to next callback
        resolve({'err': err, 'response': response, 'data': data})
      }
    }
  }
})

### DIFF
--- a/packages/node_modules/pouchdb-ajax/src/ajaxCore.js
+++ b/packages/node_modules/pouchdb-ajax/src/ajaxCore.js
@@ -62,25 +62,49 @@ function ajaxCore(options, callback) {
     options.json = false;
   }
 
-  return request(options, function (err, response, body) {
+  if ((options.preRequest) && (typeof options.preRequest == 'function')) {
+    return options.preRequest(options)
+      .then(function (opts) {
+        return returnRequest(opts);
+      }).catch(function (err) {
+        return callback(generateErrorFromResponse(err));
+      });
+  }
+
+  function returnRequest(options) {
+    return request(options, function (err, response, body) {
+      var content_type = response && response.headers && response.headers['content-type'];
+      var data = body || defaultBody();
+
+      // CouchDB doesn't always return the right content-type for JSON data, so
+      // we check for ^{ and }$ (ignoring leading/trailing whitespace)
+      if (!options.binary && (options.json || !options.processData) &&
+          typeof data !== 'object' &&
+          (/json/.test(content_type) ||
+           (/^[\s]*\{/.test(data) && /\}[\s]*$/.test(data)))) {
+        try {
+          data = JSON.parse(data.toString());
+        } catch (e) {}
+      }
+
+      if (options.preResponse && (typeof options.preResponse == 'function')) {
+        return options.preResponse(err, response, data)
+          .then(function (args) {
+            return handleResponse(args.err, args.response, args.data);
+          }).catch(function (err) {     
+            return callback(generateErrorFromResponse(err));
+          });
+      } 
+
+      return handleResponse(err, response, data);
+    });
+  }
+
+  function handleResponse(err, response, data) {
+    var error;
 
     if (err) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by nolanlawson** [comment]
Hi, thank you very much for the PR! After looking over your pull request, this looks like a very reasonable implementation, but I have some feedback about it:

1. Looking back on the discussion in #5322 it seems we had mostly agreed upon a `preRequest` function, but I don't understand the need for `preResponse`? If anything it should probably be proposed as a separate PR so it can be discussed separately.
2. We would need a new test to ensure that this works correctly; as is it's not passing in Travis because we require 100% code coverage.

Another point is that @daleharvey made a good case in https://github.com/pouchdb/pouchdb/issues/5322#issuecomment-226635729 that it might not be enough to simply have a function that takes an options object as input and returns another options object (or a promise for it). Many people have asked us in the past for the ability to define their own "ajax" implementations (e.g. over whatever transport mechanism they may desire), and I think we ought to try to support that.

So in general I would prefer to go with @daleharvey's suggestion rather than the `preRequest` suggestion, but this is always open to discussion as lots of people are bound to have opinions on this – @cdaringe @yaronyg ?

**Comment by martinfridrich** [comment]
Hi,
`preResponse` is required when you make operations on received data before data are added to local pouch instace e.g data signing against MITM modification. You have to check signature and don't want to add data to pouch if data was modified. Or when you want to add transport encyption like [jose](https://tools.ietf.org/html/rfc7516) 

I see your point, that you can do whatever you want when you have ability to define your own `ajax` function. Maybe pouch can support both api based on type of `options.ajax`. Something line
```
  if ('function' == typeof options.ajax) {
    //use custom ajax function
   ... 
 } else {
   //use as settings for native pouch ajax 
   ...
 }
```

**Comment by yaronyg** [comment]
Certainly for the cases we were dealing with in Thali we would want to be able to replace the whole ajax object with our own object so that we could put in transport types and encryption functionality that aren't supported natively. But I have to say that over time we have found less and less need for this. We handle TOR via a local proxy that 'looks' right to normal TCP and for encryption we have switched to TLS (although to be fair that required extending our custom node server, JXcore).

But if we are going to do something here I think the first step should just be to make ajax pluggable and then all the other extensions can be handled outside of PouchDB.

**Comment by daleharvey** [comment]
So the reason for using a custom ajax function is that it covers all of the use cases with a single api, to satisfy the pre / post response use case it looks like 

```
var db = new PouchDB('name', {
  ajax: function(opts, ajax) {
    opts.data = encrypt(opts.data);
    return ajax(opts).then(function(data) { 
      return decrypt(data);
    });
  }
});
```

daleharvey:
This looks great but quite a few people have varying use cases for this, I still think we should go with https://github.com/pouchdb/pouchdb/issues/5322#issuecomment-226635729 thats based on the feedback we have had from various use cases, if noone else wants to take a shot at this I will, its certainly been a requested feature

nolanlawson:
Confirm, I would prefer custom ajax as well, instead of pre/post functions. It covers more use cases (e.g. Thali's).

cdaringe:
Sorry I've been slammed!  +1 to either.  Somehow it's gotta be exposed, and
max exposure sounds easiest and most flexible!

daleharvey:
Sorry for not being able to merge however this doesnt satisfy the requirements that were discussed on the relevant bug, theres a WIP to replace this @ https://github.com/pouchdb/pouchdb/pull/5763, thanks for the PR

PR #5687, opened 2016-09-15
use node implementation for react-native (FileReaderSync)

stockulus:
in react-native there is no FileReaderSync, but the Node Implementation with Buffer works fine.
This would help me, and all other PouchDb react native users.

Thanks Christoph

### DIFF
--- a/packages/node_modules/pouchdb-binary-utils/package.json
+++ b/packages/node_modules/pouchdb-binary-utils/package.json
@@ -21,5 +21,15 @@
     "./src/blobOrBufferToBase64.js": "./src/blobOrBufferToBase64-browser.js",
     "./src/blobOrBufferToBinaryString.js": "./src/blobOrBufferToBinaryString-browser.js",
     "./src/typedBuffer.js": "./src/typedBuffer-browser.js"
+  },
+  "react-native": {
+    "./lib/index.js": "./lib/index.js",
+    "./src/base64.js": "./src/base64-browser.js",
+    "./src/base64StringToBlobOrBuffer.js": "./src/base64StringToBlobOrBuffer.js",
+    "./src/blob.js": "./src/blob.js",
+    "./src/binaryStringToBlobOrBuffer.js": "./src/binaryStringToBlobOrBuffer.js",
+    "./src/blobOrBufferToBase64.js": "./src/blobOrBufferToBase64.js",
+    "./src/blobOrBufferToBinaryString.js": "./src/blobOrBufferToBinaryString.js",
+    "./src/typedBuffer.js": "./src/typedBuffer.js"
   }
 }


nickcolley:
Hey @stockulus thanks for raising a PR :)

I'm thinking if we were to accept this we'd need to have a test to check that `react-native` works so we know if anything breaks with this setup..

stockulus:
Hey @nickcolley 

it is breaking right now ;)

I real test, that everything is fine, would end up in Starting an IOS Simulator, not quite sure if you want this in your build Pipeline. The change just tells the React Native Packager to take the Node Implementation.

If you were interested in react-native support for pouchdb, properly have a look here
https://github.com/stockulus/pouchdb-react-native
(the tests are running against the node Implementation, because it they are started from node)

I would like to integrate the Package in pouchdb core, if wanted by pouchDB.

Greez
  Christoph

nolanlawson:
The reason Buffer works fine is that it's a polyfill of the Buffer API using pure JS (and is huge). The reason the browser version fails for you is that JavaScriptCore (the engine used by RN) doesn't support FileReader/FileReaderSync. Are those APIs not polyfillable for ReactNative?

This is a very interesting problem. There is code inside of `pouchdb-adapter-leveldb-core` that checks whether we're in a browser or Node and e.g. stringifies the data in one case but not in the other. I would be interested to know which one works best in the case of React Native. If you change this for `pouchdb-binary-utils` but not for `pouchdb-adapter-leveldb-core` then you may find yourself running into some very subtle bugs (especially around attachments) due to mismatched browser/Node implementations. Have you checked to see if e.g. binary image attachments are saved correctly? Strings containing `\u0000` characters/

In principle I would push back against any change that added React Native dependencies to PouchDB. Otherwise as @nickcolley says we're in a situation where we have no idea if this fix works or not unless we ourselves add tests for React Native. Then we might as well add tests for Electron, NativeScript, NW.js, etc... you get the idea. :wink:

Here's an alternative suggestion: I noticed in your codebase that you're using PouchDB as a submodule. Instead of that, could you use the sub-packages as they're currently provided, and then if you need a browser vs Node version of any particular package, you rely on prepublish scripts or other transformations to make the required changes?

Another alternative would be to offer official `pouchdb-some-package/browser` and `pouchdb-some-package/node` APIs, but I'm a bit reluctant to do that because it would add a lot of time to the build scripts due to needing to Rollup a completely separate file (in order for the `"browser"` field to not take effect). Another option would be for you to publish a package, call it `pouchdb-binary-utils-node-only`, which depends on `pouchdb-binary-utils` as a devDependency, does the prepublish, and publishes the Node version.

nolanlawson:
Actually I just realized there's an even simpler solution: `pouchdb-binary-utils-node` and `pouchdb-binary-utils-browser`. `pouchdb-binary-utils` can rely on both. Bam! :smiley:

stockulus:
Hi @nolanlawson,

just a few Comments on that:
- FileReader/FileReaderSync is currently not polifillable. (at least I did not find/manage it)
- PouchDB is currenly a Submodule, in order to use the Tests from PouchDB, not for more, do not like it, but do not know a better solution
- The pouchdb-react-native does not use any leveldown anymore, its a direct adapter
- The dependencies with pouchdb-binary-utils came from the pouchdb-http-adapter, not from any pouchdb-react-native code

my current workaround is a Post Install script that changes the package.json of pouchdb-binary-utils, not the nicest Solution but works. Just thought it would be no Problem for this minor change in the package.json.

Greez Christoph

nolanlawson:
Ahhh, I see. For the testing fix, yes, that is currently the best approach (I've also copypasta'd the tests before... not a great situation.) I also see you have a fork of `pouchdb-adapter-leveldb` on top of asyncstorage, very clever. And this `binary-utils` fix is only for `pouchdb-adapter-http`.

In that case my preference would actually be to make a `pouchdb-adapter-http-browser` and `pouchdb-adapter-http-node`. This would actually benefit PouchDB as well because currently the slowest part of `npm install pouchdb-browser` is `request` (because it has tons of deps and is very large), and this is a shame because then that Node code is never used. I'll open a PR to create `pouchdb-adapter-http-browser`/`-node` unless someone else wants to tackle it.

stockulus:
@nolanlawson that would help, thanks!

nolanlawson:
Can we close this PR then? Is #5690 enough?

nolanlawson:
Also would you be willing to write a blog post on the PouchDB blog about React Native support? Your implementation looks really great!

stockulus:
Pouchdb-react-native is now on the rigth way, get first Contributions from other. Still the Attachment handling is not 100% of the pouchDB tests, and the Transaction handling can be optimised too.

Not sure what you expect, but i can write something about the last Month of the Project, how it works, and my way to the first open Source Project.

nolanlawson:
That sounds super interesting! We're always eager to get guest blog posts, and I'm also happy to give you pointers and copy-edit the post for you when it's done. :)

When I got PouchDB's test suite passing 100% for different browsers (Edge/IE, Safari, Android) and for different storage engines (localstorage, memdown), I found that attachments were always the hardest part. If you want any advice about how to get the attachment tests passing, don't hesitate to pop in to the Slack channel and ask. :smiley:

stockulus:
Hi @nolanlawson,

just thinking... Do you think it would be possible to have a pouchdb-test package.

```js
import pouchdbTest from 'pouchdb-test'
import pouchdb from 'pouchdb-my-super-bundle'

pouchdbTest(pouchdb, {...someOpts})
  .then(result => console.log('Test Passed', result)
  .catch(error => console.log('Test Errors, error)
```

nolanlawson:
I don't think so. There are so many options for the test, that it's impractical IMO. The only solution I've found so far is copypasta.

I've been thinking about this problem a lot. I can't conceive of an elegant solution, but I really dislike the package.json solution because it makes no sense. (A RN implementation of the binary utils adapter could just as easily use a FileReader shim. Or a Webpack loader for RN could inject the shim.)

I can create a separate package that builds a custom version of the http adapter as a prepublish step. The downside is that you would have to update it for every pouch release (greenkeeper). Is that acceptable?

nolanlawson:
OK @stockulus I have set up a repo for you. You are an npm owner and a GitHub owner and can move the project to your own GitHub if you'd like. :) Hopefully this solves the problem for you! https://www.npmjs.com/package/pouchdb-adapter-http-react-native

nolanlawson:
Oh the one thing I'll note though is that you'll need to update this for every PouchDB release; it'd be a good idea to set up Greenkeeper or something.

stockulus:
Thank for all your effort, the update would be no Problem. But finally it does not help. Now the whole http package ist the node version, not only the binary-utilities (sorry for not seeing it before). That causes other Problems.

I'll check if the only Problem is the FileReaderSync. let you know.

stockulus:
Found something from the Cordova Project, that would be possible to create a FileReader (backed by Java/ObjectC). Though I do not really like. Properly the PostInstallSetup Hack, changing the package.json of pouchdb-binary-utils is not the worst; There are also other pouchdb packages that depend on it (e. g. pouchdb-adapter-utils) so the splitting up / republishing will get really complicated as well. Properly this is even the reason, getting the browser / react-native / ... in the package.json, so that dependent packages/apps have a more simple life....

Thanks again for all the effort you put in this!

Greez Christoph

nolanlawson:
@stockulus That can be fixed by changing the build script. I agree though that based on your description maybe changing the `package.json` in a postinstall is not the worst thing due to the dependency from pouchdb-adapter-utils.

Maybe you can also build a separate version of pouchdb-adapter-utils? Would that help?

stockulus:
Properly that would help... but to be honest I'd prefer the current solution and focus on other Problems. Facebook is also working on more and better Blob support, so properly this will get solved too. Complete Blob Support from the Browser, Crypto and fs from node, better Windows, macOS support -> react-native would be totally great.

Add least I learned something about republishing and keep it in mind. Properly it will help me in the future.

Issue #5035, opened 2016-04-05
Add a visual diagram to adapters.html


nolanlawson:
Somebody asked in IRC about this quote from [adapters.html](https://pouchdb.com/adapters.html):

> In the Adapter docs for Node Pouch mentions it can be used 'over SQLite3 in Node, using the WebSQL adapter and node-websql'. That kind of confuses me, are those 3 separate, interacting modules?

This made me realize that we support so many adapters, it's just far too confusing. The "browser support" diagram is nice, but I wish we had a diagram to explain the whole thing, i.e.:

- 3 adapters: idb, leveldb, websql
- http adapter (special case)
- *down adapters
- node-websql
- SQLite Plugin

I have an idea of how this might be laid out, but I think it needs a human touch (hand-drawn? styled with CSS?) and I'm a terrible artist, so I nominate this as a "good first-timer issue."

First-timer instructions: click the pencil icon on that page to start editing, and to upload any images you may have, just add them to `docs/static/img`. If you need to modify CSS, that'll be in `docs/static/less` and you can build using `npm install && npm run build-site`.

nickcolley:
I possible let's try to avoid any information in images as it's not accessible.

So if we could do this in CSS or SVG that'd be great.

We could potentially do this as a dependency graph, this could be a static SVG image that we could inline and make accessible for example.

nolanlawson:
I agree accessibility is a big concern issue. BTW I went back and actually drew some ASCII art to show what I had in mind:

```
                           +--------+
                           |PouchDB |
                           +--------+
      Browser                            Node.js


+----------+  +-----------+  +---------------------+  +-----------+
|IndexedDB |  |WebSQL     |  |HTTP                 |  |LevelDB    |
+----------+  +----+------+  +----------+----------+  +------+----+
                   |                    |                    |
                   |                    |                    |
                   |                    |                    |
              +----+------+  +----------+----------+  +------+-----------------------------+
              |SQLite     |  |CouchDB,             |  |MemDOWN                             |
              |Plugin     |  |Cloudant,            |  |LocalStorageDOWN                    |
              |(Cordova), |  |PouchDB Server       |  |FruitDOWN                           |
              |node-websql|  |Couchbase SG         |  |*DOWN (not officially unsupported)  |
              | (Node),   |  +---------------------+  |                                    |
              |WebSQL     |                           |LevelDOWN (default)                 |
              | (default) |                           |                                    |
              |           |                           +------------------------------------+
[...]

PR #5691, opened 2016-09-15
Added a visual diagram to adapters.html (#5035)

pjhampton:
ping @nolanlawson and @nickcolley

Please let me know if you guys don't like it or the info has outdated since we last chatted. I created the SVG file from draw.io. There is a png to default to if SVG isn't supported by the browser. 

Refs: #5035


nolanlawson:
This. Is. AWESOME. 💯 

Thank you so much!! Can't wait to have this live on the website. Only a few minor corrections:

- node-websql actually runs in Node, whereas HTTP runs in both browser/node, as does memdown, and fruitdown/localstoragedown run in the browser. Maybe the node/browser distinction isn't so helpful and we should just remove it?
- rather than mention *DOWN as unsupported, I would just go ahead and not mention it at all.
- nitpick: you can put `[skip ci]` in the body of your git commit and then that way the tests won't run (no need to run the tests for doc changes)

pjhampton:
Cool @nolanlawson. And no problem at all. I think I fall into 95% of github contributors and don't read the contributing.md

Anyway, is this what you had in mind? If not, let me know and I can hit up the drawing board again.
Thanks for your help ❤️

nolanlawson:
Looks perfect to me, and I am also very happy that you included the PNG as a fallback (although most browsers support SVG these days, so we're _probably_ okay). This gets a big +1 from me.

nolanlawson:
Thank you very much! Appreciate it a lot. I'll do a release and new publish of the site sometime today.

PR #5695, opened 2016-09-16
(#5695) - re-organize Downloads page

nolanlawson:
I wanted to make `pouchdb-browser` more prominent, so that more folks are aware of it. My tests show it can vastly speed up `npm install` times compared to the `pouchdb` package.

While I was at it, I made some more small changes.

/cc @nickcolley

### DIFF
--- a/docs/download.md
+++ b/docs/download.md
@@ -4,43 +4,59 @@ title: Download
 sidebar: nav.html
 ---
 
-{% include anchor.html class="h3" title="File Download" hash="file" %}
+{% include anchor.html class="h3" title="Download" hash="file" %}
 
-The PouchDB file should come before any files that use it.
-
-For use in an HTML file:
+Quick start:
 
 {% highlight html %}
-<script src="https://cdn.jsdelivr.net/pouchdb/{{ site.version }}/pouchdb.min.js"></script>
+<script src="https://cdn.jsdelivr.net/pouchdb/{{ site.version }}/pouchdb.min.js">
+</script>
 <script>
   var db = new PouchDB('my_database');
 </script>
 {% endhighlight %}
 
-PouchDB is supplied as a compressed and uncompressed download:
+PouchDB can also be directly downloaded:
 
-* [Download pouchdb-{{ site.version }}.min.js][latest-min] (compressed for production)
-* [Download pouchdb-{{ site.version }}.js][latest] (uncompressed for debugging)
-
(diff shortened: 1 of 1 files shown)


docs/download.md
```diff
@@ -4,44 +4,60 @@ title: Download
 sidebar: nav.html
 ---
 
-{% include anchor.html class="h3" title="File Download" hash="file" %}
+{% include anchor.html class="h3" title="Download" hash="file" %}
 
-The PouchDB file should come before any files that use it.
-
-For use in an HTML file:
+Quick start:
```
nolanlawson: I haven't seen many people (even beginners) get confused about this, so I don't think it's worth mentioning. Instead I just say "quick start."

docs/download.md
```diff
@@ -4,44 +4,60 @@ title: Download
...
 
 {% highlight html %}
-<script src="https://cdn.jsdelivr.net/pouchdb/{{ site.version }}/pouchdb.min.js"></script>
+<script src="https://cdn.jsdelivr.net/pouchdb/{{ site.version }}/pouchdb.min.js">
+</script>
```
nolanlawson: Adding a linebreak prevents the word `script` from getting broken up. I wish we had a shorter CDN link.

docs/download.md
```diff
@@ -4,44 +4,60 @@ title: Download
...
-* [Download pouchdb-{{ site.version }}.min.js][latest-min] (compressed for production)
-* [Download pouchdb-{{ site.version }}.js][latest] (uncompressed for debugging)
+PouchDB can also be directly downloaded:
 
-PouchDB is also hosted on Content Delivery Networks:
-
-* [PouchDB on jsdelivr](http://www.jsdelivr.com/#!pouchdb)
-* [PouchDB on cdnjs](https://cdnjs.com/libraries/pouchdb)
+* [pouchdb-{{ site.version }}.min.js][latest-min] (compressed for production)
+* [pouchdb-{{ site.version }}.js][latest] (uncompressed for debugging)
```
nolanlawson: Moved CDNs further down.

docs/download.md
nickcolley: Could change the download heading to 'Quick Start'?

daleharvey:
Looks good, I took Nicks suggestions to change the heading and change the link to start with `//` which is what we should be doing anyway and merged, cheers

PR #5696, opened 2016-09-16
(#5696) - retry each browser test once in Travis

nolanlawson:
I spend an ungodly amount of time just restarting tests in Travis. The failed tests are almost always due to Saucelabs/Selenium/webdriver/gremlins/whatever. I think it's reasonable to run the tests once again if they fail. It's better than restarting the whole test, which runs `npm install` and `npm run build` and all that all over again.

### DIFF
--- a/bin/run-test.sh
+++ b/bin/run-test.sh
@@ -80,7 +80,17 @@ elif [ "$CLIENT" == "node" ]; then
 elif [ "$CLIENT" == "dev" ]; then
     npm run launch-dev-server
 else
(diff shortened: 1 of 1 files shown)


nolanlawson:
BTW the reason it's best to run just one additional time is that I belive Travis will cut us off after ~50 minutes, and some of these browser tests take around 25 minutes.

nolanlawson:
Ironically a bunch of the tests failed because they reached the maximum time limit. Looks like we're hosed. 😢

PR #5698, opened 2016-09-16
(#5612) - remove stringify/parse hack

KlausTrainer:
When evaluating map/reduce functions in a sandbox by using node's `vm` module,
we returned a stringified result, which we then parsed outside the sandbox. We
did that in order to make chai's object extension magic with `chai.should()`
work, which doesn't work for objects that are returned from a `vm` sandbox
(https://stackoverflow.com/a/16273649/680742).

This removes the hack, and solves the issue by using chai's `assert` functions
in the particular test cases instead of the `should` property.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -24,6 +24,7 @@
     "emit": true,
     "PouchDB": true,
     "should": true,
+    "assert": true,
     "testUtils": true,
     "importScripts": true
   },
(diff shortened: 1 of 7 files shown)


nolanlawson:
Awesome, :+1: when green.

KlausTrainer:
Merged it. Thanks :)

Issue #5572, opened 2016-08-17
Safari 10 IndexedDB compatibility


nolanlawson:
Starting a new issue to keep track of this.

In order to use IDB in Safari 10, we would need to:
1. Establish that it passes our IDB test suite
2. Establish that it's not unreasonably slower than WebSQL (e.g. for Chrome it's up to ~20% slower depending on the use case, which is acceptable).

Preliminary testing shows that Safari 10 does not pass the IDB test suite, but I haven't yet nailed down the cause. It seems intermittent. I haven't had time yet to diagnose.

Steps to repro:

``` bash
git clone https://github.com/pouchdb/pouchdb --branch safari-10 --single-branch --depth 1
cd pouchdb
npm install
# install couchdb and get it running on localhost:5984, # e.g. `brew install couchdb`
npm run dev
# point Safari to http://localhost:8000/tests/integration
```

nolanlawson:
I ran the tests about a month ago and saw some intermittent hangs, but just re-ran them and happy to say it almost reached 100%, except for what appears to be an unrelated timeout:

![screenshot 2016-08-17 08 10 14](https://cloud.githubusercontent.com/assets/283842/17741712/1c1cba82-6452-11e6-8a19-7c7c41551324.png)

(That 0% is a bug in Mocha.)

OTOH Safari IDB does indeed appear to be extremely slow; many of the more elaborate `all_docs` tests take several seconds to run.

nolanlawson:
This was in Safari 10.0 (12602.1.38.2) on macOS Sierra BTW.

beidson:
Just for a sanity check, if you can easily bump the timeout and confirm WebKit passes 100%, that would be good to know.

nolanlawson:
I keep restarting the test, but it often fails due to non-IDB-related HTTP timeouts apparently. Will try to give this a thorough pass this weekend.

llin96:
@nolanlawson Is there any plan to support IndexedDB on iOS Safari 10?

nolanlawson:
I have not worked on it. I couldn't get Safari 10 to fully pass the test suite, apparently due to HTTP issues (non-IDB stuff). In general it seems to be working but also seems to be incredibly slow (~50x slower than WebSQL), so I'm not sure we'd be doing any favors to our users to upgrade.

[...]

PR #5701, opened 2016-09-17
(#5572) - add Safari 10 support for IDB (WIP)

nolanlawson:
This PR is just to start the conversation about supporting Safari 10 IndexedDB. As described in #5572 there are still some issues to resolve:
- [ ] Benchmark to determine if Safari IDB is too slow compared to Safari WebSQL
- [ ] Confirm that the Dexie-reported issues don't affect us
- [ ] Confirm that the UA parsing logic I've added works across all Safari-like browsers, e.g. UIWebView, WKWebView, Chrome iOS, Firefox iOS
- [ ] Add a SauceLabs test for Safari >=10, ensure it passes 100%

It's been a long time coming, but I think it's about time we start putting WebSQL out to pasture.

### DIFF
--- a/packages/node_modules/pouchdb-adapter-idb/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-idb/src/index.js
@@ -978,17 +978,19 @@ function init(api, opts, callback) {
 }
 
 IdbPouch.valid = function () {
-  // Issue #2533, we finally gave up on doing bug
-  // detection instead of browser sniffing. Safari brought us
-  // to our knees.
-  var isSafari = typeof openDatabase !== 'undefined' &&
+  // Issue #2533, we finally gave up on doing feature
+  // detection instead of browser sniffing. Safari <10 has IDB
+  // bugs and they're too tricky to use feature detection for
+  var isOldSafari = typeof openDatabase !== 'undefined' &&
     /(Safari|iPhone|iPad|iPod)/.test(navigator.userAgent) &&
     !/Chrome/.test(navigator.userAgent) &&
-    !/BlackBerry/.test(navigator.platform);
-
+    !/BlackBerry/.test(navigator.platform) &&
+    // Issue #5572, Safari 10 fixes most major IDB bugs
+    /Version\/(\d+)/.test(navigator.userAgent) &&
+    parseInt(navigator.userAgent.match(/Version\/(\d+)/)[1], 10) < 10;
   // some outdated implementations of IDB that appear on Samsung
   // and HTC Android devices <4.4 are missing IDBKeyRange
-  return !isSafari && typeof indexedDB !== 'undefined' &&
+  return !isOldSafari && typeof indexedDB !== 'undefined' &&
     typeof IDBKeyRange !== 'undefined';
 };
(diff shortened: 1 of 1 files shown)


nolanlawson:
For anyone who wants to test this branch:

```
brew install couchdb # if you haven't
git clone --single-branch --branch safari-10-support \
  --depth 1 \
  https://github.com/pouchdb/pouchdb.git
cd pouchdb
npm install
npm run dev
# open localhost:8000/tests/integration in Safari
```

nolanlawson:
Here's [a build](https://gist.githubusercontent.com/nolanlawson/5f182d01ee897fe0c9d344f33e679583/raw/eeec2c6920a8b34b56a76a68dda8775264836e04/pouchdb.js) of pouchdb.js for folks who would like to drop it into their app to test. (Note that we automatically fall back to WebSQL if we detect that the user has already stored data in WebSQL, so you may need to clear out all browser data.)

nolanlawson:
I'm having second thoughts about this PR. Tried out some perf tests on my 2013 MacBook Air and seeing huge perf deltas between Safari WebSQL and Safari IDB:

| | WebSQL | IDB | Slowdown |
| --- | --- | --- | ---- |
| basic-inserts | 3773 | 21743 | 5.76x |
| temp-views | 3383 | 345380 | 102.09x |

I would run more tests, but I just don't have time to wait for them to finish. The same test that passes in seconds with WebSQL, I wait for 10s of minutes with IDB and it never finishes. 

Maybe we should just wait for Safari 11 before we make IDB the default. If we ship a 100x perf regression then users are going to be very unhappy. :(

This was Safari 10.0.2 (12602.3.12) BTW.

nolanlawson:
Safari TP (Release 19 (Safari 10.1, WebKit 12603.1.14.2)) is quite a bit faster:

| | WebSQL | IDB | Slowdown |
| --- | --- | --- | ---- |
| basic-inserts | 3773 | 5862 | 1.55x |
| temp-views | 3383 | 14650 | 4.33x |

This is more similar to the IDB slowdown we see with Chrome, so seems more acceptable.

NekR:
@nolanlawson what about workers? Does Safari has WebSQL in workers now? If not, maybe makes sense to enable IDB for it only workers.

nolanlawson:
No, it does not have WebSQL in workers (http://html5workertest.com). My recommendation for that situation is to use [pseudo-worker](https://github.com/nolanlawson/pseudo-worker) for Safari.

NekR:
@nolanlawson yeah, just thought if things could benefit from using real worker and IDB there. Maybe indeed makes sense to wait till Safari 11.

nolanlawson:
The ideal perf scenario per-browser IMO is:

- Chrome/Firefox: IDB+worker (because of UI thread blocking)
- Edge/IE: IDB only (it doesn't block UI thread much, no big need for worker)
- Safari: WebSQL, no worker

In practice just using PouchDB + pseudo-worker for Safari should be fine. This is what I do for https://pokedex.org BTW

NekR:
I see, makes sense.

PR #5712, opened 2016-09-21
Address website layout bugs #5495 and #5709

### LINKED ISSUE #5709: Broken markdown on "Map/reduce queries" doc page
### Issue
Some sections on the doc page "Map/reduce queries" are not well formated.

### Info
- Environment: browser
- Platform: Chrome

### Reproduce
Visit https://pouchdb.com/guides/queries.html and look at text below first blue box.

willholley:
Reverts the previous change which added `<div>`s around anchors (which fixes #5709) and attempting to fix #5495 with CSS only.

### DIFF
--- a/docs/_includes/anchor.html
+++ b/docs/_includes/anchor.html
@@ -3,12 +3,4 @@
 {% else %}
   {% assign class = 'h2' %}
 {% endif %}
-<div>
-  <div class='{{ class }}'>
-    <div class='anchor-id' id='{{ include.hash }}'></div>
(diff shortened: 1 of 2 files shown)


willholley:
any comments on this one?

pjhampton:
This looks good to me. It's just a matter of time until this breaks again.
Maybe it would be a good idea to create a project specific style guide when putting HTML in markup documents?

daleharvey:
So as for expecting it to break again, frankly I think its a matter of moving away from jekyll, it has not been handling markdown and parsing particularly well and I dont think anyone could give a definitive list of its quirks, but for now this looks great, thanks

PR #5720, opened 2016-09-27
Remove info block that is destroying formatting

rjsteinert:
The formatting is messed up on this page https://pouchdb.com/guides/queries.html

I'm guessing there is something about that info block that is doing it. Sorry I don't have the tools installed to check right now.


daleharvey:
So this was fixed in https://github.com/pouchdb/pouchdb/commit/8fcc5d3457bc01eb5a3a4f805236899d170f7937, without having to remove the info block so going to close this out, but thanks for the PR

rjsteinert:
Great! Thanks @daleharvey

PR #5734, opened 2016-09-29
Use changing db names in couch master tests

daleharvey:


daleharvey:
So my plan for this is to get the browser tests passing, move them into non expected failures, then look at removing `iscouchmaster` specific exceptions, now that 2.0 is released it should be our main target

PR #5737, opened 2016-09-29
jsnext:main pointing to incorrect file

danbucholtz:
the jsnext:main is pointing to a file that doesn't exist in the `pouchdb` package.json file.

This has led to us having trouble with rollup, as we prefer to use `module` or `jsnext:main` over `main` whenever possible.

I also added the `module` field as lots of projects such as `@angular` are moving towards that.

Thanks,
Dan


packages/node_modules/pouchdb/package.json
```diff
@@ -3,7 +3,8 @@
   "version": "6.1.0-prerelease",
   "description": "PouchDB is a pocket-sized database",
   "main": "./lib/index.js",
-  "jsnext:main": "./lib/index-es.js",
+  "jsnext:main": "./lib/index.es.js",
+  "module": "./lib/index.es.js",
```
nolanlawson: I'm not comfortable with adding "module", since we're not doing it for any other package. Also I like "jsnext:main" because it makes it more clear that it's a "beta" option ("module" is not yet standardized and thus I'm worried about potentially breaking the ecosystem once it is standardized).

danbucholtz: Okay, I will update.  No problem. Give me 5 min.

nolanlawson:
:+1:

cubicleWar:
Is this in any way related to [this issue](https://forum.ionicframework.com/t/rc0-error-could-not-resolve-with-external-npm-dependecy/65054/2)?

nolanlawson:
Yes, it's related. For Rollup to work with CommonJS modules in the ecosystem, you will typically need plugins like `rollup-plugin-node-resolve`, `rollup-plugin-commonjs`, `rollup-plugin-json`, and `rollup-plugin-globals`. I also wrote `rollup-plugin-node-resolve-auto` which attempts to automatically ignore CommonJS dependencies.

Also @danbucholtz this is released as PouchDB 6.0.6, and FWIW for Ionic you should probably be okay with using `pouchdb-browser` instead of `pouchdb`, since you don't need the Node.js version of PouchDB.

As for Rollup stuff you may find it's easier to use Rollup as a pre-compile step before Browserify/Webpack, e.g. I started writing something to be used for this; you may find it useful: http://github.com/nolanlawson/rollup-prepublish

glaucocustodio:
I am with `PouchDB 6.0.6` on `Ionic RC 0` (uses `Rollup`) but I still cannot import Pouch, I have tried many ways like:

`import PouchDB from 'pouchdb';`
`import * as PouchDB from 'pouchdb'`
`import PouchDB from 'pouchdb-browser'`

I also ran `npm install @types/pouchdb --save-dev  --save-exact` that gave me `"@types/pouchdb": "5.4.28"`.

Am I doing something wrong?

nolanlawson:
This is being tracked over in https://github.com/pouchdb/pouchdb/pull/5746 and https://github.com/pouchdb/pouchdb/pull/5746 as well as elsewhere.

nolanlawson:
pouchdb 6.0.7 fixes this issue

Issue #5739, opened 2016-09-30
6.0.6


nolanlawson:
- (#5737) - fix jsnext:main pointing to incorrect file
- (#5730) - fix leveldown included in pouchdb-adapter-leveldb-core
- (#5612) - remove unnecessary stringify/parse from vm usage

I think these are worth doing a patch release.

nolanlawson:
done

PR #5740, opened 2016-09-30
(#5739) - 6.0.6

nolanlawson:


nolanlawson:
Annoyingly I think I actually need an official +1 before I can merge this. :confused:

daleharvey:
Hrm thats annoying, theres an override for the status checks failing if you are an admin, assumed the same would be for reviews, will look into it

nolanlawson:
Yeah I was able to force-merge without Travis/Coveralls, but seems no for the review thing. Won't even let me push to master from the command line. :confused: It's not a huge problem, though; maybe it's better than we can no longer mess this up. :)

PR #5746, opened 2016-10-02
replaced extend from js-extend library with ES2015 builtin Object.assign

ckaeslin:
I'm using PouchDB with ionic2 and was following https://github.com/driftyco/ionic/issues/8356

I could reduce a rollup problem by replacing extend from js-extend library with the ES2015 builtin function Object.assign.

Still need a custom rollup config (but no other manual change in node_modules folder). This repo https://github.com/ckaeslin/ionic2-RC0-pouchdb with replaced pouchdb-browser from this branch worked for me.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,12 +36,14 @@
     "argsarray": "0.0.1",
     "debug": "2.2.0",
     "double-ended-queue": "2.1.0-0",
+    "es6-object-assign": "^1.0.3",
     "es6-promise-pool": "2.4.4",
     "fruitdown": "1.0.2",
     "inherits": "2.0.3",
     "js-extend": "1.0.1",
     "level-codec": "6.2.0",
     "level-write-stream": "1.0.0",
+    "leveldown": "1.5.0",
     "levelup": "1.3.2",
     "lie": "3.1.0",
     "localstorage-down": "0.6.6",
@@ -53,7 +55,6 @@
     "spark-md5": "2.0.2",
     "through2": "2.0.1",
     "vuvuzela": "1.0.3",
-    "leveldown": "1.5.0",
     "websql": "0.4.4"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
     "localstorage-down": "0.6.6",
     "ltgt": "2.1.2",
     "memdown": "1.2.2",
+    "object.assign": "^4.0.4",
     "readable-stream": "1.0.33",
     "request": "2.74.0",
     "scope-eval": "0.0.3",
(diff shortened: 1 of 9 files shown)


daleharvey:
Thanks, I dont have time to look at this properly tonight, and mostly want to see the test results, however I believe IE and Node .12 sound like they are going to be problems here (http://kangax.github.io/compat-table/es6/#test-Object_static_methods_Object.assign) .12 is getting very close to eol but ie support will certainly be needed

ckaeslin:
I tried to solve it with es6-object-assign which will only be set if Object.assign is not there: https://github.com/rubennorte/es6-object-assign/blob/master/src/index.js#L33
So later this code can just be removed.

It's also not that clean but I think it's more difficult to get the additional config for js-extend into ionic-app-scripts:
https://github.com/matheusgarcez/ionic-2rc-pouchdb/blob/master/config/rollup.config.js#L57

Chances for builtin and global rollup plugin are much higher. And then with this fix it's everything needed to run pouchdb with ionic2 again out of the box.

nolanlawson:
Thanks for the PR, but I'm skeptical this is the right solution to the problem. The real issue is that vanilla Rollup does not play well with the Commonjs ecosystem, and Ionic for some reason decided to go with vanilla Rollup. Also keep in mind that Ionic 2 is a RC and I've heard they are considering dropping Rollup from the final release due to this problem.

Closing, in the meantime i would advise delaying an upgrade to Ionic 2. Also yeah, PouchDB's minimum is ES5, not ES6 right now.

nolanlawson:
Also this is being tracked in https://github.com/driftyco/ionic/issues/8356

ckaeslin:
Sorry for bringing this up again. There was just a PR https://github.com/driftyco/ionic-app-scripts/pull/79 merge into ionic which would let pouchdb work out of the box again (together with this PR).
Or would you agree on replacing js-extend by another lib?

jgw96:
@ckaeslin With that PR now merged you can simply edit your rollup.config to add this:
```
commonjs(
    {
        namedExports: {
          'node_modules/js-extend/extend.js': ['extend']
        }
    }
),
```
and everything will "just work" with stock PouchDB. There is no need to replace `js-extend` (:

ckaeslin:
Thanks @jgw96 I'm aware of this fix, it works for me. Maybe a little documentation somewhere would be helpful to also make it easy for others to use these two great projects together.

nolanlawson:
Folks, Ionic 2 is in release candidate... Presumably they will fix this at some point, since many many libraries seem to be broken; not just PouchDB.

Also there is zero benefit to using PouchDB in ES6 mode, because you will end up with exactly the same bundle file/treeshaking as with the CJS build. All of our ES6 deps are already Rollup'ed in our own prepublish script, meaning you consume one gigantic module with a few CJS deps.

I'm starting to wonder if we should just remove `jsnext:main` entirely from our `package.json`.

nolanlawson:
Interesting. I see now that the reason `js-extend` fails is because it is trying to be too clever about exporting `exports.extend`. Its code is too complex for rollup-plugin-commonjs to parse, hence this error.

In the interests of unbreaking Ionic users, and since `js-extend` is like 15 lines of code, I propose just moving it into the PouchDB source.

PR #5748, opened 2016-10-03
(#5782) - always use string representation of seed

willholley:
I noticed that tests appeared to be deterministic when I specified
a seed value but the order of operations was different to that
reported by a specific test run.

This is because the initial seed generates an number (timestamp)
whereas an environment variable passed in is a string. Casting
all seed values to a string normalises these, allowing failing test
runs to be reproduced.


nickcolley:
Nice 👍

PR #5751, opened 2016-10-04
(#5746) - move js-extend into source

nolanlawson:
Stopgap solution to fix Ionic 2 breakages. See #5746 for details.

Long story short: I don't know if Ionic will fix this themselves before they release Ionic 2. Ultimately the problem is that js-extend is too clever in its module format, and it breaks Rollup. There doesn't seem to be a Rollup plugin that can handle it correctly. It's also like 15 lines of code, so we can just put it in source.

I noticed we now have two different extend functions for some reason. We should evaluate merging that and see if there are perf/compat problems, but for 15 lines of code it's probably not even worth it.

### DIFF
--- a/tests/integration/utils.js
+++ b/tests/integration/utils.js
@@ -270,7 +270,7 @@ testUtils.ajax = PouchForCoverage.ajax;
 testUtils.uuid = pouchUtils.uuid;
 testUtils.parseUri = pouchUtils.parseUri;
 testUtils.errors = PouchForCoverage.Errors;
-testUtils.extend = require('js-extend').extend;
+testUtils.extend = pouchUtils.jsExtend;
 
 testUtils.makeBlob = function (data, type) {
   if (typeof process !== 'undefined' && !process.browser) {
--- a/package.json
+++ b/package.json
@@ -39,7 +39,6 @@
     "es6-promise-pool": "2.4.4",
     "fruitdown": "1.0.2",
     "inherits": "2.0.3",
-    "js-extend": "1.0.1",
     "level-codec": "6.2.0",
     "level-write-stream": "1.0.0",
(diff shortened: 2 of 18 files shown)


nolanlawson:
Some details on why js-extend breaks here: https://github.com/rollup/rollup-plugin-commonjs#custom-named-exports

tests/integration/utils.js
```diff
@@ -270,7 +270,7 @@ testUtils.ajax = PouchForCoverage.ajax;
 testUtils.uuid = pouchUtils.uuid;
 testUtils.parseUri = pouchUtils.parseUri;
 testUtils.errors = PouchForCoverage.Errors;
-testUtils.extend = require('js-extend').extend;
+testUtils.extend = pouchUtils.jsExtend;
```
ckaeslin: testUtils.extend = pouchUtils.extend;

because it's still exported as extend in packages/node_modules/pouchdb-for-coverage/src/utils.js  https://github.com/pouchdb/pouchdb/pull/5751/files#diff-42219049ecd23077146f457971885dc3R64

ckaeslin:
Thanks @nolanlawson for your effort. The ionic pouchdb users will be very happy about this.

nolanlawson:
should be okay now.

btw someone should really just fix https://github.com/rollup/rollup-plugin-commonjs so that it's either 1) smarter about tricky CJS exports, or 2) doesn't throw errors for unknown named exports. Maybe an option like `ignoreUnknownExports`?

daleharvey:
Merged in https://github.com/pouchdb/pouchdb/commit/26020d61b6474d4370f1838af9ec7dfde743e6a8

nolanlawson:
fixed in pouchdb 6.0.7

PR #5752, opened 2016-10-05
using extend instead of js-extend which is working with rollup-plugin-commonjs

ckaeslin:
Not sure why tests in https://github.com/pouchdb/pouchdb/pull/5751 failed.
But in order to help making progress (for ionic people), I'm providing here a version with https://www.npmjs.com/package/extend instead of https://www.npmjs.com/package/js-extend.

I tested this locally and verified that it's working with latest ionic-app-scripts without any additional custom configuration.


daleharvey:
It looks very much like #5751 is failing intermittents, its likely that they can be restarted, go green and that PR can get merged today

ckaeslin:
I think https://github.com/pouchdb/pouchdb/pull/5751 needs unit tests for included js-extend to pass the requested line coverage of 100%.

daleharvey:
Yup you are right, there is one line not being covered, could be an ignore if its trouble to test

ckaeslin:
tests are here: https://github.com/vmattos/js-extend/blob/master/test/suite.js
since the code is copy from there

ckaeslin:
closing in favor of https://github.com/pouchdb/pouchdb/pull/5754

nolanlawson:
fixed in pouchdb 6.0.7

PR #5754, opened 2016-10-05
5746 add tests

ckaeslin:
added the unit tests from https://github.com/vmattos/js-extend to https://github.com/pouchdb/pouchdb/pull/5751 for 100% line coverage

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,6 @@
     "es6-promise-pool": "2.4.4",
     "fruitdown": "1.0.2",
     "inherits": "2.0.3",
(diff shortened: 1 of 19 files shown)


daleharvey:
This looks good to me, will let @nolanlawson do the final review though since this isnt my particular area of expertise

nolanlawson:
Good find, and yeah I've been on GH vacation, just now getting back. We should definitely consolidate our extend() implementions; I just did the js-extend merge to stop the bleeding for Ionic folks.

Issue #5607, opened 2016-08-29
Cannot handle 401 error during live replication


amod-joshi:
Hi,

We have an application running  both on mobile as well on web in Ionic. I am not able to handle the 401 error on mobile when user changes the password on the web. PouchDB keeps on retrying to connect the sync gateway even though sync gateway sends 401 Authentication error. No call back function is called when retry is set true.

If I set the retry to false I can catch 401 error.  But i want retry to set it true since if any other failure conditions it should retry to connect the sync gateway.

dharders:
Currently I get 401 with `localDB.replicate.to/from`, with retry & live true. But it's emitted through the `complete` event (i.e. not the `denied` event, as you would expect).

I'm using CouchDB 2.0 so YMMV as you mentioned you're using sync gateway. Also, I had issues with `sync()`  events not firing, so best to use `replicate()` instead.

P.S. I think it's all about to change with the next release looking at this PR #5421    (which I think should then fire 401's through the `denied` event)

amod-joshi:
Hi Dale,


When are planning your next release so that we can download and check it


Thanks

Amod Joshi


________________________________

dharders:
> When are planning your next release so that we can download and check it

I'm not the maintainer, you'll have to ask these guys  @nolanlawson @daleharvey  ;)

nolanlawson:
> When are planning your next release

This week. :)

amod-joshi:
Hi, I have downloaded the new version of pouchdb(6.0.4)but 401 issues does not seems to be resolved. It is not firing the error in denied event. I am attaching the image for your reference

![errorof pouchdb](https://cloud.githubusercontent.com/assets/21310848/18432781/a02a9632-7901-11e6-8fc2-6ec4baa4d42e.png)

dharders:
Try the 'error' event (I'm getting 403s)

amod-joshi:
I tried that as well but the result is same, 401 error is not getting trap on live sync or on replication

Please find the same code we are using in the application

 syncLive: function () {
        var syncActive = false;
        sync = localDb.sync(remoteDb, {live: true, retry: true})
          .on('active', function (info) {
            console.log('Sync active');
            if (window.cordova) {
              $cordovaToast.showLongBottom('Synchronizing data').then(function (success) {
                // success
                console.log('Sync Toast message displayed');
              }, function (error) {
                // error.
                console.log("Sync toast message error");
              });
            }
            syncActive = true;
          })
[...]

PR #5763, opened 2016-10-09
WIP - Add ability to intercept/override ajax

daleharvey:
This isnt finished, but enough for feedback, so `req` probably isnt the best name for the option here but cant think of anything better of the top of my head.

Also if anyone who wants to use this, @martinfridrich, @DanielMaly or anyone else could take a look at this and see if it satisfies their requirements it would be awesome

### DIFF
--- a/docs/_includes/api/create_database.html
+++ b/docs/_includes/api/create_database.html
@@ -18,12 +18,26 @@ This method creates a database or opens an existing one. If you use a URL like `
 
 **Options for remote databases:**
 
-* `ajax`: (Remote databases only.) Ajax requester options. For instance, passing in the options `{ajax: {timeout: 10000}}` will allow you to set the max timeout for an HTTP request. These are passed verbatim to [request][] (in Node.js) or  [a request shim](https://github.com/pouchdb/pouchdb/blob/master/lib/deps/request-browser.js) (in the browser), with the exception of:
-* `ajax.cache`: Appends a random string to the end of all HTTP GET requests to avoid them being cached on IE. Set this to `true` to prevent this happening.
-* `ajax.headers`: The `ajax.headers` option allows you to customise headers that are sent to the remote HTTP Server.
-* `auth.username` + `auth.password`: You can specify HTTP auth parameters either by using a database with a name in the form `http://user:pass@host/name` or via the `auth.username` + `auth.password` options.
-* `ajax.withCredentials`: Set to `false` to disable transferring cookies or [HTTP Auth information](https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Requests_with_credentials). Defaults to `true`.
+* `ajax`: Ajax requester options. For instance, passing in the options `{ajax: {timeout: 10000}}` will allow you to set the max timeout for an HTTP request. These are passed verbatim to [request][] (in Node.js) or  [a request shim](https://github.com/pouchdb/pouchdb/blob/master/lib/deps/request-browser.js) (in the browser), with the exception of:
+  * `ajax.cache`: Appends a random string to the end of all HTTP GET requests to avoid them being cached on IE. Set this to `true` to prevent this happening.
+  * `ajax.headers`: The `ajax.headers` option allows you to customise headers that are sent to the remote HTTP Server.
+  * `auth.username` + `auth.password`: You can specify HTTP auth parameters either by using a database with a name in the form `http://user:pass@host/name` or via the `auth.username` + `auth.password` options.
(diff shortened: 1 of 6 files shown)


daleharvey:
Getting closer here, converting the tests that previously overrode `db._ajax` to use the modifier. Theres an issue I need to look at ensuring that we can cancel modified requests (and possibly that they can cancel themselves?)

cdaringe:
@daleharvey, nice start!  looks great.  pedantically, i'd prefer `(ajax, opts)` vs `(opts, ajax)`, as `ajax` seems primary `opts` seem secondary to me... but nbd.  i'm borderline ashamed for even commenting about it :)

daleharvey:
So yeh I have thought about that a bit, `(ajax, opts)` follows our style a little more because in general opts are optional extras, and in terms of an api its much better to have `(required, optional)` however that the exact reason that they are this way round. There are situations where you might use neither (`Promise.reject`), situations where you would use `opts` but not `ajax` using an alternative transport, situations where you would use both `return opts(ajax)`, but almost never a situation where you use `ajax` but not `opts`

daleharvey:
kicking the travis tyres

DanielMaly:
This is awesome! I have implemented my functionality using db._ajax in the end, will that break in the near future? Anyway, as long as the `opts` object contains the method, URL, headers, body and content-type, I will be happy.

daleharvey:
yup `db._ajax` doesnt work with this patch, but yes `opts` contains all the data that is used to make the request

daleharvey:
And for those following along, https://github.com/pouchdb/pouchdb/pull/5763/files#diff-f8ba360f868e295f81e2520b3db3986aR193 is probably the best place to get familiar with what the new functionality looks like, this is passing locally so going to see what the test run looks like but aside from that this is mostly finished aside from the docs

martinfridrich:
Hi, sorry for late response. I try to test this but I've got infinite loop when try with `pouchdb-server 1.2.0` and doc isn't synced.

```
var p = new PouchDB("local_db");

var rep = PouchDB.replicate(p, 'http://127.0.0.1:8080/test', {
  live: true,
  retry: true,
  req: function (opts, ajax) {
    return ajax(opts).then(function (res) {
      res.customfield = 'foo';
      return res;
    });
  }
}).on('change', function (info) {
  console.log("changes");
  console.log(info);
  console.log("------------------------------");
}).on('paused', function (err) {
  console.log("paused");
  if (err) {
    console.log(err);
    console.log(err.stack);
  }
  console.log("------------------------------");
}).on('active', function () {
  console.log("active");
  console.log("------------------------------");
}).on('denied', function (err) {
  console.log("denied");
  console.log("------------------------------");
}).on('complete', function (info) {
  console.log("complete");
  console.log(info);
  console.log("------------------------------");
}).on('error', function (err) {
  console.log("error");
  console.log(err);
  console.log("------------------------------");
});
```
I've got this error in browser console after add doc(both `put/post`) 
```
TypeError: Cannot read property 'length' of undefined
    at Array.filter (native)
```

log from poucdb-server console
```
[info] GET / 304 - 127.0.0.1
[info] GET /test/_local/MwZiWRzmgJa21kuH14LmCg%3D%3D? 404 - 127.0.0.1
[info] POST /test/_revs_diff 200 - 127.0.0.1
```
and this loop forever

Everything is working ok when I comment `req:function...` from db options.

thanks

martinfridrich:
ok, problem was that function also modify response with revisions and then pouch chrash in function `getRevisionOneDocs()`

daleharvey:
Yeh I have seen that before, when the modifier function crashes it can be hard to determine what the problem is, it may be an idea to add some handling in there to make sure crashes are visible

daleharvey:
Ok this should be now good to go :) I went with `opts.request` as `typeof opts.ajax === 'function'` was going to prevent people from specifying both a request and `opts.ajax.cache` for example

cdaringe:
woo hooooooo @swashcap :eyes:

jsdevel:
@daleharvey would it be possible to add a handler for `response`?  If we had that, then we could probably work around #5607.

daleharvey:
@jsdevel That is part of this, you can handle and modify responses @ https://github.com/pouchdb/pouchdb/pull/5763/files#diff-f8ba360f868e295f81e2520b3db3986aR225 or do you mean something different?

daleharvey:
Actually thinking about this I may want to hold off on this a tiny bit, the `ajax` implementation looks a lot like fetch but differs in quite a few ways, we use errors for `4XX` and some params are not identical. I think it would be far easier to understand if `ajax` was `fetch`

nolanlawson:
I agree with Dale; now that we're moving towards "prefer `fetch()`, fall back to `XMLHttpRequest()`, any ajax-related hooks should really apply to `fetch()` rather than XHR. I consider this blocked by #5835  and unfortunately needing largely a rewrite to be approved; sorry about that.

PR #5785, opened 2016-10-19
fix(memleak): correct `destroyed` event handler memory leak on `close()`

### LINKED ISSUE #4868: [nodejs] memleak when creating PouchDB instance
I am seeing the issue in production with slightly more complicated code (the memleak is obvious after 30 minutes but the databases are all different and using the `http` adapter, etc.), but here is a simple example that also exhibits the issue:

```
var PouchDB, debug, dgram, heapUsed, socket;
dgram = require('dgram');
PouchDB = require('pouchdb');
debug = (require('debug'))('test');
heapUsed = null;
socket = dgram.createSocket('udp4');
socket.on('message', function(msg, rinfo) {
  var db;
  debug("Received " + msg.length + " bytes from " + rinfo.address + ":" + rinfo.port);
  db = new PouchDB('goodluck');
  return db.info(function(info) {
    var last_heapUsed, memory;
    memory = process.memoryUsage();
    last_heapUsed = heapUsed;
    heapUsed = memory.heapUsed;
    return debug('info', last_heapUsed != null ? heapUsed - last_heapUsed : null);
  });
});
socket.bind(7124);
```

I then run `while true; do echo ping | nc -u -w 1 127.0.0.1 7124; done` in a shell to trigger the `message` handler.

Forcing a `gc()` call for each message (using `--expose_gc --expose_gc_as=gc`):
- There is no leak if I remove the `new PouchDB...` code. (The difference is '0' after about 6-7 iterations.)
- The leak is about the same size with or without the call to `info()`.
- The leak is about 2 to 3ko with nodejs 5.4.1 or 5.5.0, about 10 to 14ko with nodejs 4.2.1 or 4.2.6.

I'm obviously not sure this might be a PouchDB issue, but I'm at a loss to analyze where the issue might come from (I've done some runs with [`heapdump`](https://www.npmjs.com/package/heapdump) but it apparently doesn't work when `--expose_gc` is active).

shimaore:
Fixes #4868 
This extends `close()` so that the constructor can remove its `destroyed`-event-handling callbacks; both the local `destroyed()` callback (bound to a specific PouchDB instance) and the global `destroyed(name)` Pouch callback are removed.

This also modifies `.destroy()` in a similar fashion (I suspect `destroy` was leaking as well) ­— I'm more interested in fixing `.close()` though, so maintainers are welcome to revert that specific one-line change (Travis doesn't seem to like it too much).

The code could be prettier if the destruction-listeners were inside a more clever data store than a plain Array (a Set maybe?) — but this does the job I believe.

### DIFF
--- a/packages/node_modules/pouchdb-core/src/constructor.js
+++ b/packages/node_modules/pouchdb-core/src/constructor.js
@@ -24,11 +24,24 @@ function prepareForDestruction(self) {
   }
 
   function onConstructorDestroyed() {
-    self.removeListener('destroyed', onDestroyed);
+    onClosed();
     self.emit('destroyed', self);
   }
 
+  function onClosed() {
+    self.removeListener('destroyed', onDestroyed);
+    if (destructionListeners.has(self.name)) {
+      var old_set = destructionListeners.get(self.name);
+      var old_set_index = old_set.indexOf(onConstructorDestroyed);
+      if(old_set_index >= 0) {
+        old_set.splice(old_set_index,1);
+        destructionListeners.set(self.name, []);
+      }
+    }
+  }
+
   self.once('destroyed', onDestroyed);
+  self.once('closed', onClosed);
 
   // in setup.js, the constructor is primed to listen for destroy events
   if (!destructionListeners.has(self.name)) {
--- a/packages/node_modules/pouchdb-core/src/adapter.js
+++ b/packages/node_modules/pouchdb-core/src/adapter.js
@@ -728,6 +728,7 @@ AbstractPouchDB.prototype.changes = function (opts, callback) {
 
 AbstractPouchDB.prototype.close = adapterFun('close', function (callback) {
   this._closed = true;
+  this.emit('closed');
   return this._close(callback);
 });
 


daleharvey:
Sorry was taking a look at this, it looks good however as mentioned before we could really do with a failing test case here, otherwise if this were to get merged it would likely just regress again soon and not be noticed

packages/node_modules/pouchdb-core/src/constructor.js
```diff
@@ -24,11 +24,24 @@ function prepareForDestruction(self) {
   }
 
   function onConstructorDestroyed() {
-    self.removeListener('destroyed', onDestroyed);
+    onClosed();
     self.emit('destroyed', self);
   }
 
+  function onClosed() {
+    self.removeListener('destroyed', onDestroyed);
```
nolanlawson: why are we not also removing the listener for `'closed'`?

nolanlawson:
Agree with @daleharvey; this looks great but we would need a new test case to catch the behavior. There are lots of tests in `test.retry.js` and `test.events.js` which can detect listener memory leaks and write test cases for them, which should provide some good examples to start with.

packages/node_modules/pouchdb-core/src/constructor.js
```diff
@@ -24,11 +24,24 @@ function prepareForDestruction(self) {
...
+    if (destructionListeners.has(self.name)) {
+      var old_set = destructionListeners.get(self.name);
+      var old_set_index = old_set.indexOf(onConstructorDestroyed);
+      if(old_set_index >= 0) {
+        old_set.splice(old_set_index,1);
+        destructionListeners.set(self.name, []);
```
shimaore: It's corrected for the upcoming PR :)

packages/node_modules/pouchdb-core/src/constructor.js
shimaore: Hmmm when called directly I assumed `once` would remove the listener.. however the listener should be removed when the function is called for `destroy`. I'll correct this in the next PR.

shimaore: Conversely I don't understand why we say `self.emit('destroyed',self)` on line 28, instead of simply `self.emit('destroyed')` (the code in this master). The current only consumer is the replication code, which doesn't handle the parameter; and the parameter looks redundant in itself.

Issue #5611, opened 2016-08-30
Try to run CI tests in node@6


gr2m:
follow up for https://github.com/pouchdb/pouchdb/commit/b2895aa57422610091061d93580e20a788a28803#commitcomment-18822331

> For a long time our deps were broken in 6. That may have changed now; worth testing.

I’ll have a look into it

nolanlawson:
done in https://github.com/pouchdb/pouchdb/pull/6023

PR #5788, opened 2016-10-19
(#5788) - test: fix ajax timeout test

KlausTrainer:
This adds the error code `ESOCKETTIMEDOUT` to the ajax test assertion
match pattern in order to make the test pass when testing with Node.js
v6 against PouchDB >= v6.0.7.


daleharvey:
Looks great, can we get node 6 in CI now? https://github.com/pouchdb/pouchdb/issues/5611

KlausTrainer:
Jep, it looks like it. Just ran the tests with Node.js v6.9.1 against pouchdb-server 1.2.0, and they all passed :)

PR #5767, opened 2016-10-11
5782 - handle rev-1 conflict during sync

willholley:
Adds a test to reproduce the fuzzy test failure detected in #5782.

A relatively simple workaround fixes this: if a conflict is detected in a rev-1 document, skip the bulk_docs optimisation for that id.

PR #5800, opened 2016-10-24
5793 - fixes for sync against Couchbase Sync Gateway

willholley:
This adds a couple of tests around handling of `conflicts` in `allDocs` and `bulkDocs`, the latter of which I expect to fail against Couchbase Sync Gateway. It also tweaks the workaround added in #5767 to strip the `_conflicts` array from the `allDocs` response to prevent POSTing it to the server unnecessarily.

### DIFF
--- a/packages/node_modules/pouchdb-adapter-idb/src/allDocs.js
+++ b/packages/node_modules/pouchdb-adapter-idb/src/allDocs.js
@@ -98,7 +98,10 @@ function idbAllDocs(opts, api, idb, callback) {
       docIdRevIndex.get(key).onsuccess =  function onGetDoc(e) {
         row.doc = decodeDoc(e.target.result);
         if (opts.conflicts) {
-          row.doc._conflicts = collectConflicts(metadata);
+          var conflicts = collectConflicts(metadata);
+          if (conflicts.length) {
+            row.doc._conflicts = conflicts;
+          }
         }
         fetchAttachmentsIfNecessary(row.doc, opts, txn);
       };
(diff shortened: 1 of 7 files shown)


nolanlawson:
Wow @willholley you're a hero. Nice job tracking down the CSG bug.

Since it's also conceptually simpler to have no-conflicts rather than zero-length-conflicts, and since there's a test, I'm happy with this change. Will restart failed jobs and then it's a +1 from me.

packages/node_modules/pouchdb-core/src/adapter.js
```diff
@@ -579,6 +579,9 @@ AbstractPouchDB.prototype.get = adapterFun('get', function (id, opts, cb) {
       var conflicts = collectConflicts(metadata);
       if (conflicts.length) {
         doc._conflicts = conflicts;
+        if (!doc._conflicts.length) {
+          delete doc._conflicts;
+        }
       }
```
daleharvey: This shouldnt be needed given the lines preceeding it, I also think we should use this pattern in the other areas

    var conflicts = collectConflicts(metadata);
    if (conflicts.length) { 
      doc._conflicts = conflicts;
    }

Feels a lot cleaner than setting `doc._conflicts` then deleting if empty

daleharvey:
Yup great job, the cover failure is genuine though, I think the approach I commented may fix the coverage issue here, but if not may need an ignore or 2 in here

willholley:
@daleharvey thanks for the review - agree that the approach you suggested for stripping conflicts is neater. I've updated the code based on that so lets see what coverage looks like now.

daleharvey:
Awesome thanks, just one line there still tripping up coverage, but 👍 with that fixed

willholley:
@daleharvey good to go?

daleharvey:
Yup sorry wished github notified when PR's are updated, thats great

PR #5812, opened 2016-10-27
Fix memleak in pouchdb-core

### LINKED ISSUE #4868: [nodejs] memleak when creating PouchDB instance

shimaore:
This partially fixes #4868 by removing existing memleak in pouchdb-core.
This is a revised and expanded version of a previous PR.
This PR contains:
- the memleak testing code (which is otherwise available in [my memleak-test branch](https://github.com/shimaore/pouchdb/tree/memleak-test)), with self-test and pouchdb-core tests enabled;
- changes to gitignore and package.json to enable the memleak testing code;
- changes to pouchdb-core to remove the memleak on `close()` and `destroy()`. This changeset is different from the one I previously submitted and performs better (i.e. doesn't leak on `close()` as far as the test suite can tell).

There are other memleaks in (at least) the http adapter (still tracking this one, I suspect the leak is in a dependency) and leveldown (I noticed at least one static `LazySet` in the code which might partially explain it, since `LazySet` doesn't have a method to remove entries).


shimaore:
Found a couple more locations for corrections, based on https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#32-leaking-arguments, will add them shortly.

tests/memleak/test.memleak.js
```diff
@@ -0,0 +1,458 @@
...
+      var db = new PouchDB(db_name,opts);
+      function Finally() { return db.close(); }
+      db_name = null;
+      return db.info()
+      .then( Finally, Finally )
+      .then(function () {
+        db = null;
+        return sleep(20);
+      });
+    };
```
daleharvey: Need to remove this semi colon, failing eslint

packages/node_modules/pouchdb-adapter-indexeddb/src/index.js
```diff
@@ -33,7 +33,10 @@ function IdbPouch(dbOpts, callback) {
   // the database handle as the first argument
   var $ = function (fun) {
     return function () {
-      var args = Array.prototype.slice.call(arguments);
+      var args = new Array(arguments.length);
+      for(var i = 0; i < args.length; ++i) {
+        args[i] = arguments[i];
+      }
```
daleharvey: Whats the reason for doing this? is it part of the fix for memleak?

tests/memleak/test.memleak.js
```diff
@@ -0,0 +1,458 @@
...
+    api.info = function DummyPouchAdapterInfo(callback) {
+      if(callback) {
+        callback(null,{dummy:true});
+      } else {
+        return Promise.resolve({dummy:true});
+      }
+    };
+
+    if(callback) {
```
daleharvey: I need to put a fix in for eslint, but in general we prefer to have a space before parens

    if (callback) 
    api.info = function DummyPouchAdapterInfo (callback) {

etc, not a huge deal if you dont fix, will probably just do a batch fix when we fix eslint

daleharvey:
Overall this look great, the tests are amazing, really like seeing the comparitive tests to clean state vs broken, theres a few minor comments in there, if we could avoid the `var args = Array.prototype.slice.call(arguments);` change if possible and do that seperately it would be good, unless its actually involved in the memory leak, needs the semi colon fix for the tests though

shimaore:
@daleharvey I've reverted the changes regarding `arguments`: I had noted in the commits that they addressed memleaks in pouchdb-http-adapter but I can't reproduce those changes anymore (but I'm now testing with Node.js 7 so there might be a difference there as well — anyway).

I also checked eslint and inserted spaces after `if`, but eslint complained about `Unexpected space before function parentheses` when I tried `function ABC ()` so I left those alone.

daleharvey:
Thats awesome thanks a lot, just to clarify you expect the test-memleak to fail at the current time due to the still existing leaks right? if so then I am 👍 on getting this merged then we can work on fixing the rest and enabling the test in travis

shimaore:
Yup, at this time the test for the http adapter is still failing (there is still a memleak somewhere). All the other tests should be successful.

daleharvey:
Thats awesome, yeh I seen the same in the travis run :) sweet this is a great job, bugs like this are a nightmare to track down and test, very glad we have a set of tests in here now just to get them enabled :)

Issue #3518, opened 2015-02-09
Clean up or remove skipped tests


daleharvey:

daleharvey:
1171 passing (3m)
  24 pending

PR #5821, opened 2016-10-30
(#3518) - Remove unused test

daleharvey:
This test has never been enabled and doesnt make sense, tried reading through the history to see how it made sense before but cant find any reason


daleharvey:
Not sure why this thinks its waiting on travis, since its just a deleted test it has [skip ci]

PR #5822, opened 2016-10-30
(#3518) - Remove skipped replication tests

daleharvey:
These tests have been long skipped due to breakages using mocked interfaces (trying to override `.get`) The tests were introduced quite a while ago and we have added more tests to cover this code so we should be good


daleharvey:
Cheers Will

nolanlawson:
:+1:

Issue #5736, opened 2016-09-29
PouchDB should emit a warning if you insert an attachment without a content_type


nolanlawson:
This is nearly always a bug (e.g. see https://github.com/pouchdb/pouchdb/issues/5726).

We should do a `console.warn` if this happens. (Not an error, because people may be replicating from CouchDBs they have no control over, where the type is omitted for whatever reason.)

jasuch:
Hi! :)

Can I help with this? I really want to start in Open Source projects and I feel confortable for doing this.

daleharvey:
Yup defiitely, at a guess the code that needs changed will be in https://github.com/pouchdb/pouchdb/blob/master/packages/node_modules/pouchdb-core/src/adapter.js#L782 and https://github.com/pouchdb/pouchdb/blob/master/packages/node_modules/pouchdb-core/src/adapter.js#L232

As this is a warning we dont have a decent way to test this so manually verifying it should be fine, but make sure that both db.bulkDocs and db.putAttachment do the same, there is some more help @ https://github.com/pouchdb/pouchdb/blob/master/CONTRIBUTING.md and if you hit any questions feel free to ask, thanks :)

pecey:
Is anyone still working on this? Or I would like to take this up.

jpmllr89:
Same here. Willing to work with others, too.

willholley:
@pecey @jpmllr89 please feel free to have a go and submit a PR. If you have questions/comments then just go ahead an ask in this thread.

MBFlewelling:
Hello, I'd be interested in working this if help is still needed

nolanlawson:
@MBFlewelling Yes please feel free to take this up. :)

PavanBahuguni:
I would like to take up the issue.

nolanlawson:
fixed in https://github.com/pouchdb/pouchdb/pull/5826, sorry that someone grabbed it before you!

Issue #5726, opened 2016-09-28
Safari drops attachment type


DreamingMatt:
### Issue
Storing attachments in Safari seems to drop the type attribute/property of said attachment. I've only tested it with Files/Blobs.

### Info
- Environment: browser
- Platform: Safari 9.1.1
- Adapter: WebSQL
- Server: Custom

### Reproduce
You can reproduce running [this JSbin](https://jsbin.com/milanatigo/edit?html,output) and by following these steps:

1. select a file
2. click the newly appeared button
3. check the logs;  especially `object -> _attachments -> files -> data`

You'll see that type is `undefined` in Safari. This is not the case in Chrome or Firefox.

nolanlawson:
I feel like I may have already seen this bug filed this bug on Safari, but I don't have time to look for it now. If you're able to find a small reproducible test case, please file the issue on WebKit while you're at it (they do fix bugs!): https://bugs.webkit.org/

As for this particular bug, do you have a feeling if PouchDB would be able to work around this issue? Or does WebKit themselves need to fix it?

nolanlawson:
Nevermind, I understand what's going on here. Here's your code:

```js
db.post({
  _attachments: {
    file: {
      type: file.type,
      data: file
    }
  }
})
```

This should be:

```js
db.post({
  _attachments: {
    file: {
      content_type: file.type,
      data: file
    }
  }
})
```

PouchDB uses **content_type** rather than **type** for attachments. You are essentially omitting the type, which means that IndexedDB browsers (non-Safari) stores the raw Blob whereas Safari stores the base64 encoding without the type. Hence this bug.

In the corrected version, Safari should work. Please reopen if I've mis-diagnosed.

PR #5826, opened 2016-10-31
(#5736) - Warn about attachments without content_type

andymorris:
Handles #5736. Emit a warning when an attachment is inserted without a content_type,
which is almost always a bug (e.g. see #5726). This is not treated as an
error in PouchDB since users may be replicating from a CouchDB they have
no control over.

A few people mentioned wanting to help with this one in the issue, but some time has passed without any activity. I'm not intending to step on any toes, apologies if I am!

### DIFF
--- a/packages/node_modules/pouchdb-core/src/adapter.js
+++ b/packages/node_modules/pouchdb-core/src/adapter.js
@@ -1,4 +1,7 @@
-import { jsExtend as extend } from 'pouchdb-utils';
+import {
+    jsExtend as extend,
+    guardedConsole
+} from 'pouchdb-utils';
 import Promise from 'pouchdb-promise';
 import { Map } from 'pouchdb-collections';
 import { EventEmitter } from 'events';
@@ -244,6 +247,9 @@ AbstractPouchDB.prototype.putAttachment =
     type = blob;
     blob = rev;
     rev = null;
+  }
+  if (!type) {
+    guardedConsole('warn', 'Attachment', attachmentId, 'on document', docId, 'is missing content_type');
   }
 
   function createAttachment(doc) {
@@ -785,6 +791,9 @@ AbstractPouchDB.prototype.bulkDocs =
     if (doc._attachments) {
       Object.keys(doc._attachments).forEach(function (name) {
         attachmentError = attachmentError || attachmentNameError(name);
+        if (!doc._attachments[name].content_type) {
+          guardedConsole('warn', 'Attachment', name, 'on document', doc._id, 'is missing content_type');
+        }
       });
     }
   });
--- a/tests/integration/test.attachments.js
+++ b/tests/integration/test.attachments.js
@@ -100,6 +100,23 @@ adapters.forEach(function (adapter) {
       });
     });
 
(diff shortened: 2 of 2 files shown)


packages/node_modules/pouchdb-core/src/adapter.js
```diff
@@ -245,6 +245,9 @@ AbstractPouchDB.prototype.putAttachment =
     blob = rev;
     rev = null;
   }
+  if (!type) {
+    console.warn('Attachment', attachmentId, 'on document', docId, 'is missing content_type');
```
daleharvey: So to handle the case where `console` is not defined, we have a utility for calling console functions that we should use here https://github.com/pouchdb/pouchdb/blob/master/packages/node_modules/pouchdb-core/src/parseAdapter.js

daleharvey:
This is great, thanks :) the comment I left will need addressed and there is one other issue, this seems like it will fail coverage tests. You can test this with `npm run test-coverage`, while you dont need to test that the console warning is shown, just hitting that codepath will make sure we arent crashing or anything so copying one of the very basic tests from https://github.com/pouchdb/pouchdb/blob/master/tests/integration/test.attachments.js should be good.

Give us a shout if you have any issues, thanks a lot

nolanlawson:
Feedback has been addressed, LGTM, +1 when green although coverage does seem to be failing so I don't know if it will pass.

tests/integration/test.attachments.js
```diff
@@ -100,6 +100,27 @@ adapters.forEach(function (adapter) {
       });
     });
 
+    it('5736 warning for putAttachment without content_type', function () {
+      var db = new PouchDB(dbs.name);
+      return db.put({_id: 'foo'}).then(function (response) {
+        db.putAttachment(response.id, 'bar.txt', response.rev, testUtils.btoa('text'));
```
daleharvey: This probably needs to be `return db.putAttachment` ?

andymorris:
Yeah sorry for the delay, it's been a busy couple of weeks. I'm hoping to get this finished up and passing tonight. Thank you for the pointers and patience.

andymorris:
`npm test` and `npm run test-coverage` both pass locally, but it looks like there were some timeouts on travis, and I'm not sure about [this failure](https://travis-ci.org/pouchdb/pouchdb/jobs/177215505). Is there something else I need to change?

Also, I ended up having to change the `putAttachment` spec due to differences in how adapters handle missing args (e.g., [here](https://github.com/pouchdb/pouchdb/blob/80e75620a0947fea374539c675fd8c97ed9406da/packages/node_modules/pouchdb-core/src/adapter.js#L236-L247) vs [here](https://github.com/pouchdb/pouchdb/blob/80e75620a0947fea374539c675fd8c97ed9406da/packages/node_modules/pouchdb-adapter-http/src/index.js#L573-L578)). Leaving it out entirely causes it to fail for the HTTP adapter with `callback is not a function`.

nolanlawson:
I restarted the jobs; let's see

nolanlawson:
Awesome, thank you!

Issue #5833, opened 2016-11-02
Fix for #4868 prevents firing of `destroyed` event after `close()`


shimaore:
Thinking over the changes in #4868 I came up with new tests for test.events.js that break when `close()` and `destroy()` are interleaved on the same database.

I'm working on writing proper tests for these cases, and a clean solution for `constructor.js` and `setup.js`.

nolanlawson:
closing; let's keep issue discussion in #4868

PR #5834, opened 2016-11-02
Corrections for #4868

shimaore:
As noted in #5833, the code I submitted for #4868 introduces some variations in that using both `close()` and `destroy()` on different PouchDB instances accessing the same database won't work.

This PR introduces tests to check for these conditions, and proposes a solution by:
- moving all destruction-listeners handling code to `setup.js` to clarify the API;
- adding two new (undocumented) events, `ref` and `unref` which allows the instance constructor to cleanly reference / dereference instances from PouchDB;
- having PouchDB's destruction-listener emit `destroyed` events (on each destruction listener) with a new parameter, so that the instance's destruction listener knows when to feed `destroyed` events back to PouchDB. (The adapter's code does `self.emit('destroyed')` and won't conflict.)

There are other ways to implement this (for example `ref` and `unref` could be methods instead of events, I could have kept a separate `onConstructorDestroyed` handler, ...); I tried to keep things clean by encapsulating the handling of `destructListeners` in a single place, and to minimize the risks of leaks at the same time by simplifying the code in the constructor.


packages/node_modules/pouchdb-core/src/setup.js
```diff
@@ -22,11 +22,40 @@ function setUpEventEmitter(Pouch) {
...
+      destructListeners.set(db.name, []);
+    }
+    destructListeners.get(db.name).push(db);
+  });
+
+  Pouch.on('unref', function onConstructorUnref(db) {
+    if (!destructListeners.has(db.name)) {
+      return;
+    }
+    var db_list = destructListeners.get(db.name);
```
daleharvey: Tiny nit: We use camelCase for vars in JS in general

daleharvey:
This looks mostly good to me, tiny nits that the `snake_case` vars should be switched to `camelCase`

It is however failing coverage, you can see what lines are failing by running `npm test-coverage` and reading the resulting report. Ideally we can construct tests that hit those cases (or delete the code if its just extra unused code). If for some reason it is very hard to write tests to hit those cases then we can put an istanbul ignore in there though

Cheers

shimaore:
Can't test with the regular adapters because the missing lines in coverage handle race conditions, I'll write a custom adapter so that the conditions can be exhibited.

daleharvey:
If it comes down to race conditions then having an `/* istanbul ignore:next */` is probably best for those circumstances, I have been trying to make sure we have as little mocking as possible even if it means maybe having to ignore a few lines for coverage sake.

Updates look great, just need the lint errors fixed + istanbul then 👍 from me, thanks a lot

shimaore:
FWIW one test is failing because

    var db1 = new PouchDB('testdb');
    var db2 = new PouchDB('testdb');
    db1.close(function () { db2.close() })

fails with `NOT_OPEN` (`Database not open`) when using the local adapter.

shimaore:
OK, I confirm that calling `close()` on either local or http adapters with two different instances pointing to the same underlying database will fail on the second call (which looks like an error to me).
Since I'm only looking for coverage at this point I'll catch the error so that I can finish this PR, and open a separate issue to document the multiple-`close()` problem.

shimaore:
@daleharvey Travis fails with:
- segfault in mocha on Node.js (I'm tempted to say I'm not involved :roll_eyes: )
- [error in selenium test](https://travis-ci.org/pouchdb/pouchdb/jobs/173927311#L1838) ([test.events.js:239](https://github.com/shimaore/pouchdb/blob/test-closed-and-destroyed-2/tests/integration/test.events.js#L239)) in `db1.info().then … db1.close().then … db2.destroy()` where db1 and db2 are both instances of the same db (not sure how to go about this one).

nolanlawson:
@shimaore I've restarted the failed tests; hopefully they'll be green now. Unfortunately our test suite generates a lot of false negatives.

shimaore:
Thank you @nolanlawson that helped :)

shimaore:
@daleharvey this is good to go pending review

daleharvey:
@shimaore Apologies, been crazy at work these days, so in general if we can do things that avoid new events and listeners they have just in general been a source of inconsistency and bugs in terms of pouchdb core (as you are seeing). However this looks fairly clean and passing with all the new tests so happy to get this landed, thanks a lot

nolanlawson:
This had to be reverted in https://github.com/pouchdb/pouchdb/pull/5901 because it broke Safari somehow. Unfortunately non-core contributors don't have their code run on Saucelabs (this is fixed now but we would need to fix it ourselves because it's not automatic), which is why the bug slipped through.

PR #5836, opened 2016-11-02
pass options to ajax call of http's api.bulkGet

hermitdemschoenenleben:
all api calls except bulkGet pass the options of the api command to the ajax call. This is great for allowing plugins to modify a call, so I added it to bulkGet, too (I suppose it's a mistake that all api calls except bulkGet do this)


nolanlawson:
LGTM

nolanlawson:
Thank you for the PR! :tada:

PR #5840, opened 2016-11-03
Newline after Plugins (h2)

millette:
Not sure what's the best way to skip to the next line after the "Plugins" (h2) title. It's a tad convoluted :-)


daleharvey:
That looks good to me, hadnt noticed that before, thanks :)

Issue #5845, opened 2016-11-04
Test broken in firefox nightly


daleharvey:
Seems like a basic failure, trying to write data to a read only error object but want to see if I can get this tested in CI so it doesnt get sprung on us in future

daleharvey:
Ok, so even though we specify Firefox 48 in travis its actually testing Firefox 38, I figured out how to get travis to use a later version https://github.com/pouchdb/pouchdb/pull/5846/commits/93638432735e707c5d3edb8defb0084411c2d1a3 but its broken as Firefox 47 had some breaking change for selenium, now it seems like geckodriver is supposed to be the fix for this, however as far as I can tell https://github.com/vvo/selenium-standalone/ is installing geckodriver but still not working

@vvo is there anything we are supposed to do anything to get Firefox >= 47 work?

nolanlawson:
Last I saw geckodriver only works with Firefox 48+. But I didn't test with selenium-standalone; I tested with another Selenium implementation.

vvo:
@daleharvey Did no try geckodriver for myself using last selenium-standalone. Sorry.

nolanlawson:
travis stuff has been largely fixed, we are now testing in firefox 47 which is new enough

PR #5846, opened 2016-11-04
WIP - Use latest firefox in travis

daleharvey:


daleharvey:
Not strictly part of https://github.com/pouchdb/pouchdb/issues/5845 but cleaning up travis a little bit, this means we dont need to remember to update firefox version numbers (its currently out of date)

daleharvey:
Looking at the test run, it seems we are using firefox 38 both before and after this change

daleharvey:
actually want to see whats up with these versions before merging

nolanlawson:
I was never able to figure out how to get latest firefox running in Travis. :( In many of my projects I swear I have the same config in two projects and yet I get different firefox versions.

Issue #5832, opened 2016-11-02
How to force update a doc when there's conflict? 


simpzan:
### Issue
How to update a doc based on a non-leaf revision? namely I want to create a new conflict in the doc.
The reason I want to do this is that I don't want to resolve the conflict when writing on one client, I want to resolve the conflict later when I want to.
Thanks！

### Info
- Environment: (browser)
- Platform: (Chrome)
- Adapter: (IndexedDB)
- Server: (CouchDB)

### Reproduce
N/A

daleharvey:
We dont have a huge amount of docs on this since its generally not something people want to do (it can get dangerous pretty quickly) but you can use `new_edits: false` with `db.bulkDocs` to create conflicts, you can find examples around the tests and some on the web.

What is your use case?

simpzan:
I am developing a note web app which supports both phone and desktop. It supports syncing while editing the note. and the editing note is periodically saved to local pouchdb. So It is possible that the data in pouchdb and memory are out of sync, namely note in db has modifies from server and note in memory has modifies from user. 

At this time if I save the note in memory to db based on the latest revision in db, the modifies from server will be hidden. The ideal solution is to generate a conflict and resolve later when convenient, because it is really hard to resolve the conflict on the phone.

Though I can also save the memory note as a new doc in db, I have to get a way to know which 2 notes are in conflicts. 

`one doc with a conflict revision` vs `2 docs with some info indicating they are conflict notes`, I think the first is much better.

simpzan:
I found in `new_edits: false` with `db.bulkDocs` method, I have to generate a new revision myself. In my use case, I want pouchdb generate a new revision for me.

simpzan:
In this doc about conflict https://pouchdb.com/guides/conflicts.html, it says that there are 2 types of conflicts, immediate conflict and eventual conflict. with eventual conflict I can resolve it at any time I want. It's great to have this freedom. 

but with immediate conflict, I don't have this freedom, I have to resolve it right away. I think it should not be hard to provide this freedom for immediate conflict as well, like add a `force` option to `put` operation. What is the consideration not to do it?

daleharvey:
> I found in new_edits: false with db.bulkDocs method, I have to generate a 
> new revision myself. In my use case, I want pouchdb generate a new revision for me.

[...]

PR #5852, opened 2016-11-07
(#5832) - return correct revision after doc update

willholley:
For some adapters, document updates were returning
the winning rev in the document update metadata
rather than the revision just written. This adds
a test for the behaviour and fixes it.


nolanlawson:
Awesome, :+1: when green.

Issue #5848, opened 2016-11-06
PouchDB mutates options passed in to constructor


ronag:
Currently the following will fail since pouch seems to modify the options object.

```javascript
const p1 = new PouchDB(options)
const p2 = new PouchDB(options) // uhoh, invalid options
```

I would suggest that pouch does a deep clone of the options object if it needs to modify it.

PR #5863, opened 2016-11-11
(#5848) Fix constructor modifying options object

atgjack:
Using the same options object more than once was modifying the original
object. The options object was being cloned after the modification
rather than before. This fixes that and adds a test to make sure you can
use the same object again.

### DIFF
--- a/packages/node_modules/pouchdb-core/src/constructor.js
+++ b/packages/node_modules/pouchdb-core/src/constructor.js
@@ -2,7 +2,7 @@ import debug from 'debug';
 import inherits from 'inherits';
 import Adapter from './adapter';
 import TaskQueue from './taskqueue';
-import { clone } from 'pouchdb-utils';
+import { clone, jsExtend as extend  } from 'pouchdb-utils';
 import parseAdapter from './parseAdapter';
 
 // OK, so here's the deal. Consider this code:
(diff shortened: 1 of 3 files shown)


nolanlawson:
LGTM, my only feedback is tests may be failing genuinely (I restarted them to double-check) and I'd also say `original` instead of `virgin` since it contains less cultural baggage. Thanks for the PR!

atgJack:
I've updated the variable names, and fixed what was causing the tests to fail. 

I'm having trouble trying to run the test in Firefox on my rig, but everything else seems to pass. 
The Travis build is timing out, so I don't think it's an issue with the code, but I can't be totally sure.

Lemme know if I need to do anything else, I'm kind of a rookie in the PR department.

nolanlawson:
Our tests are flaky and often require restarting individual tests in order to get a fully green run. I've restarted them and am happy with the PR; +1 when it's fully green.

nolanlawson:
There is an apparent true failure that only occurs in the coverage tests, and only occurs for this test:

```
  1) test.views.js-local-http destroy using prototype:
     AssertionError: expected [ '_pouch_test' ] to have a length of 0 but got 1
      at tests/mapreduce/test.views.js:694:47
```

This is very weird and I'm not sure what's wrong. :confused:

nolanlawson:
Sorry, this now has merge conflicts against master and would need some tweaking in order to pass our tests; please reopen if you get a chance to look at this. (I am removing old PRs/bugs to limit noise on the repo.)

PR #5868, opened 2016-11-12

simpzan:
add a `force` option to `put` operation to allow update based on
non-leaf revison and generate a new conflict revision, instead of
reporting update conflict error.

Couchdb doesn't support this option, so the test case will skip for http adapter.

### DIFF
--- a/packages/node_modules/pouchdb-adapter-utils/src/processDocs.js
+++ b/packages/node_modules/pouchdb-adapter-utils/src/processDocs.js
@@ -93,7 +93,7 @@ function processDocs(revLimit, docInfos, api, fetchedDocs, tx, results,
 
       if (fetchedDocs.has(id)) {
         updateDoc(revLimit, fetchedDocs.get(id), currentDoc, results,
-          resultsIdx, docWritten, writeDoc, newEdits);
+          resultsIdx, docWritten, writeDoc, newEdits, opts.force);
       } else {
         // Ensure stemming applies to new writes as well
         var merged = merge([], currentDoc.metadata.rev_tree[0], revLimit);
--- a/packages/node_modules/pouchdb-adapter-utils/src/updateDoc.js
+++ b/packages/node_modules/pouchdb-adapter-utils/src/updateDoc.js
@@ -8,7 +8,7 @@ import {
 } from 'pouchdb-merge';
 
 function updateDoc(revLimit, prev, docInfo, results,
-                   i, cb, writeDoc, newEdits) {
+                   i, cb, writeDoc, newEdits, forceUpdate) {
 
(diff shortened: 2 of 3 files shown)


nolanlawson:
Hi, in general we don't add features that aren't supported in CouchDB itself. Since this feature could be implemented as a plugin using the `bulkDocs()` API (with `new_edits: false`), can I suggest you write it as a third-party plugin instead? Thanks!

daleharvey:
Sorry I was slow to review this, this is a feature that we have long talked about as being a part of CouchDB so while I would like to clarify (and maybe) hold off on merging until we know whats going to happen on the couch end, I do think we should keep this open and possibly merge preemtively

nolanlawson:
Ah okay I didn't realize it was planned for CouchDB.

-----
Nolan Lawson
http://nolanlawson.com

nolanlawson:
Moved to https://github.com/pouchdb/pouchdb/pull/6030 so that we could rebase and test again; unfortunately this got a little stale so I would like to rebase to ensure that all the tests pass.

daleharvey:
Reopening since there are changes needed here and it would be nice to have the discussion in the original PR so @simpzan can see. Again apologies for the review time, holidays and all, will make sure to follow up timely in future here

simpzan:
Sorry for the late reply. 
@daleharvey Thanks for your suggestion, I think it is a good idea. 

But I don't understand too much how `new_edits` works, so it is a little hard for me to implement this. It will be great that somebody familiar with it can implement it. 

Or anyone can explain me how `new_edits` works, I will be happy to find time and try to implement it myself. :)

daleharvey:
So on a high level, new_edits: true (the default) means you are creating a new document that will be attached to the revision you have provided for the previous document, your new document will be given a newly generated revision.

new_edits: false means pouchdb expects you to provide the revision (really it expects that it already created the revision on another node and this document is being written by the replicator), so with this you can insert documents with your own generated revision.

You can search for new_edits in the test suite to see the format (https://github.com/pouchdb/pouchdb/blob/63f039cb0c79671a0bcab7e2cf2afc7be297e23c/tests/integration/test.bulk_docs.js#L250), looking at that quickly, we would turn `PUT {"_id": "foo", "_rev": "1-x"}` into `PUT { "_id": "foo", "_rev": "2-y, "_revisions": {"start": 1, "ids": ["x"]}}`

(where y is generated as we do in https://github.com/pouchdb/pouchdb/blob/63f039cb0c79671a0bcab7e2cf2afc7be297e23c/packages/node_modules/pouchdb-adapter-utils/src/parseDoc.js#L93)

Sorry for the quick example, have to run, but any more questions please feel free
Sorry for the quick example but I have to rush

Issue #5876, opened 2016-11-14
Prebuilt Cordova SQLite blog post is out of data as of PouchDB 6.x


nolanlawson:
The instructions for https://pouchdb.com/2016/04/28/prebuilt-databases-with-pouchdb.html were valid with PouchDB 5.x but now don't work with 6.x because you need pouchdb-adapter-cordova-sqlite and the `_pouch_` prefix is gone. It needs to be updated or else it's very confusing (see #5866).

nolanlawson:
See https://github.com/pouchdb/pouchdb/pull/5878

PR #5878, opened 2016-11-15
(#5876) - update blog post for pouchdb 6.x

nolanlawson:
I also updated the pouchdb-adapter-cordova-sqlite plugin to be less confusing (it automatically registers itself as a plugin now if you include it as a `<script>` tag) and I also updated https://github.com/nolanlawson/pouchdb-prebuilt-demo and confirmed it works with PouchDB 6.0.7 and pouchdb-adapter-cordova-sqlite 2 on both iOS and Android.

I did not run through the full instructions in this blog post, though, but if anyone has time to do so that would be wonderful! I just updated it based on my recollections of what changed in PouchDB 6.x.


nolanlawson:
Gonna timebomb this.

PR #5886, opened 2016-11-18
Use 'hasOwnProperty' instead of 'in' operator

cedk:
If the request id has the same name as one of the property of the prototype of Object, it will tries to push the request to the prototype property.


nolanlawson:
Looks good! Could you add a new test to capture this case, so that we don't regress? E.g. you can use a string like `constructor` to test this.

cedk:
I can not setup a test environment. Fell free to add the test.

nolanlawson:
I wrote a test and tweaked this PR; closing in favor of https://github.com/pouchdb/pouchdb/pull/5896

Issue #5884, opened 2016-11-17
build with the process plugin


calvinmetcalf:
it seems the browser bundle doesn't have process and that systemjs doesn't shim it, might be a good idea to use the [rollup-plugin-node-globals](https://github.com/calvinmetcalf/rollup-plugin-node-globals) plugin for at least the browser build or remove SystemJS from the supported bundlers (and rollup which also doesn't support it)

nolanlawson:
I have a PR open to remove process.nextTick entirely (#5898). It's used in like 7 places and seems unnecessary when setTimeout does pretty much the same thing.

calvinmetcalf:
It's not just process, it's also global and buffer

daleharvey:
We should definitely never be using buffer in the browser and seems like global we could avoid having to shim anything in. 

Is this the problem that pouchdb-browser when consumed via jsnext:main doesnt do the replacements that our build does?

calvinmetcalf:
it's that systemjs and rollup don't add them in at all

calvinmetcalf:
this was the original issue:

```
Hey guys, here is something that my team and I were running into  that we are hoping you can help us understand. When we run integration tests on the code that uses pouchdb(using the package pouchdb-browser to be exact). We were getting an error saying that *process* was not defined specifically on a line that was calling process.nextTick. After some debugging we noticed that on the pouchdb file hosted in CDNs, it includes a shim for process. Once we added the npm process package as a dependency and made sure SystemJS imports it everything worked. We were just wondering why pouchdb-browser doesn’t mention process as a dependency.
```

nolanlawson:
Thanks for the details. It seems process should no longer be a dependency in the next release, so this will become a non-issue.

PR #5887, opened 2016-11-18
(#5884) - add rollup-plugin-node-globals for browser builds

calvinmetcalf:
Since not all build systems automatically shim process and globals
and other such globals like browserify does this adds those into
the build when not building for node.

### DIFF
--- a/bin/build-pouchdb.js
+++ b/bin/build-pouchdb.js
@@ -18,6 +18,7 @@ var browserifyIncremental = require('browserify-incremental');
 var rollup = require('rollup');
 var nodeResolve = require('rollup-plugin-node-resolve');
 var replace = require('rollup-plugin-replace');
+var globals = require('rollup-plugin-node-globals');
 var derequire = require('derequire');
 var fs = require('fs');
 var writeFileAsync = denodeify(fs.writeFile);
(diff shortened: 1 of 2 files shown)


nolanlawson:
Sorry, what does this fix? I don't understand...

Also it looks like we have genuine failures in the *down adapters that run in the browser (memdown, etc.).

calvinmetcalf:
It came up on slack, basically systemjs and rollup don't include the node globals like process or global this is a rollup plugin that adds them in

nolanlawson:
Sounds reasonable, but failures still look genuine to me:

```
{ lastPassed: 'Conflict resolution 4-c',
  failures: 
   [ { stack: '',
       title: '"before each" hook for "should emit closed even when destroyed (async #2)"',
       message: 'timeout of 100000ms exceeded. Ensure the done() callback is being called in this test.' } ],
  browser: 'Mozilla/5.0 (X11; Linux x86_64; rv:38.0) Gecko/20100101 Firefox/38.0',
  passed: 490,
  failed: 1,
  logs: 
   [ [ 'error', 'Error in .on("change", function):', [Object] ],
     [ 'error', 'Error in .on("change", function):', [Object] ],
     [ 'log',
       '{"status":412,"name":"precondition_failed","message":"Database not open"}' ],
     [ 'log',
       '{"status":412,"name":"precondition_failed","message":"Database not open"}' ],
     [ 'log',
       '{"status":412,"name":"precondition_failed","message":"Database not open"}' ] ] }
```

calvinmetcalf:
Yeah it could be the other bundles need it too, will look more onto it

nolanlawson:
Those tests use a lot more built-ins than the other ones (e.g. Buffer), so
quite possible

-----
Nolan Lawson
http://nolanlawson.com

nolanlawson:
closing in favor of #5898 which should hopefully fix the same problem

Issue #5883, opened 2016-11-17
Incorrect _revisions after replication


filionf:
### Issue
After replicating a specific document, it gets an incorrects _revisions sequence. For example, I could have a rev Id starting with 6-, but have only 5 revision ids.

I have a complex environment that automatically merges documents together when a conflict occurs. It also reuse some revhashes in the process, which I think is causing the issue. I do have the response from the server when doing the replication and I believe the answer is somewhere in there.

If you notice, there are multiple open revs and the revision hash 0a21b4bd4399b51e144a06b126031edc exists at 2 places: 5-0a21b4bd4399b51e144a06b126031edc and 6-0a21b4bd4399b51e144a06b126031edc.

Once replicated, retrieving the document using the local database give the following:

`{
  "$createdBy": "a576775bb8378240d20502932f03e14a",
  "$creationDate": "2016-11-17T03:17:32.637Z",
  "status": 1,
  "$category": "ACCIDENT",
  "$lastModifiedUser": "a576775bb8378240d20502932f03e14a",
  "$lastModifiedDate": "2016-11-17T03:17:36.553Z",
  "$timezoneOffset": 300,
  "_id": "1109B3A7-E112-845B-BB39-F530CD2E158A",
  "_rev": "6-0a21b4bd4399b51e144a06b126031edc",
  "_revisions": {
    "start": 5,
    "ids": [
      "0a21b4bd4399b51e144a06b126031edc",
      "90f7c5665484a4972e8eb3c5baee6b15",
      "aa8f5ef9544876ec652bf21b500cb890",
      "57152ea3e06e0e6580941ed69c2d0729",
      "a44a851eb78ed461b6051b09fdae0813"
    ]
  }
}`

Any hint on how I could fix this would be appreciated.

### Info
- Environment: Node.js Server / Browser Client
- Platform: Chrome
- Adapter: WebSQL
- Server: CouchDB

### Reproduce

GET $couchdb/1109B3A7-E112-845B-BB39-F530CD2E158A?revs=true&open_revs=%5B%226-0a21b4bd4399b51e144a06b126031edc%22%2C%226-b424cf4593313b08af3757097e9f79dc%22%2C%225-98bba7dd19fc74bd519f0ce8962a3609%22%2C%224-bdd85d57bc1e52e7074a7c79d97182d8%22%5D

`[{
	"ok": {
		"_id": "1109B3A7-E112-845B-BB39-F530CD2E158A",
[...]

PR #5888, opened 2016-11-18
(#5883) - Retrieves the appropriate _revisions when there are duplicate rev-hash

filionf:
See #5883 for details.

Instead of always taking the first in the paths array, I modified it so it takes the one for the revision being retrieved if it exists and keep the first in the other cases.

### DIFF
--- a/packages/node_modules/pouchdb-core/src/adapter.js
+++ b/packages/node_modules/pouchdb-core/src/adapter.js
@@ -40,15 +40,6 @@ function compare(left, right) {
   return left < right ? -1 : left > right ? 1 : 0;
 }
 
-// returns first element of arr satisfying callback predicate
-function arrayFirst(arr, callback) {
-  for (var i = 0; i < arr.length; i++) {
-    if (callback(arr[i], i) === true) {
-      return arr[i];
-    }
-  }
-}
-
 // Wrapper for functions that call the bulkdocs api with a single doc,
 // if the first result is an error, return an error
 function yankError(callback) {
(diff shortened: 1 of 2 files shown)


packages/node_modules/pouchdb-core/src/adapter.js
```diff
@@ -588,10 +579,18 @@ AbstractPouchDB.prototype.get = adapterFun('get', function (id, opts, cb) {
 
     if (opts.revs || opts.revs_info) {
       var paths = rootToLeaf(metadata.rev_tree);
-      var path = arrayFirst(paths, function (arr) {
-        return arr.ids.map(function (x) { return x.id; })
-          .indexOf(doc._rev.split('-')[1]) !== -1;
-      });
+      var path = paths.reduce(function selectRevPath(result, arr) {
+        var splittedRev = doc._rev.split('-');
+        var revNo       = parseInt(splittedRev[0]);
```
nolanlawson: `parseInt(x, 10)` would be safer for non-ES5 browsers

packages/node_modules/pouchdb-core/src/adapter.js
```diff
@@ -588,10 +579,18 @@ AbstractPouchDB.prototype.get = adapterFun('get', function (id, opts, cb) {
...
+        var revHash     = splittedRev[1];
+        var hashIndex   = arr.ids.map(function (x) { return x.id; })
+          .indexOf(revHash);
```
nolanlawson: typically I prefer regular `for` loops for stuff like this because it tends to be more performant than functions-within-functions.

nolanlawson:
b1f2b6379752c4e0e981983e575b45a76882b9e1

nolanlawson:
Thank you!

packages/node_modules/pouchdb-core/src/adapter.js
filionf: No problem. That is because I tried to follow the way code was written before with arrayFirst. Unless you want to do it yourself, I will send you the updates in another PR.

Issue #5855, opened 2016-11-07
Stack overflow in safeJsonParse when running inside web worker


michael42:
### Issue
When trying to use my PouchDB (6.0.7) powered web application, it crashes early during startup with a `RangeError: Maximum call stack size exceeded` inside [safeJsonParse](https://github.com/pouchdb/pouchdb/blob/16e231ea58b67bfc2f61dc4b61961632fc1aa8ad/src/deps/safeJsonParse.js#L23).

I have the following stack trace for this error:
````
Uncaught RangeError: Maximum call stack size exceeded
safeJsonParse	@	index-browser.js:3516
decodeMetadata	@	index-browser.js:3592
api._changes.docStore.get.onsuccess	@	index-browser.js:4806
IndexedDB (async)		
onGetCursor	@	index-browser.js:4805
onsuccess	@	index-browser.js:4818
IndexedDB (async)		
fetchChanges	@	index-browser.js:4843
api._changes	@	index-browser.js:4849
Changes$1.doChanges	@	index-browser.js:1677
Changes$1	@	index-browser.js:1591
AbstractPouchDB.changes	@	index-browser.js:2436
getChanges	@	index-browser.js:10617
(anonymous function)	@	index-browser.js:10668
Promise.resolve (async)		
(anonymous function)	@	index-browser.js:151
(anonymous function)	@	index-browser.js:120
(anonymous function)	@	index.js:14
api._getLocal.req.onsuccess	@	index-browser.js:4955
IndexedDB (async)		
api._getLocal	@	index-browser.js:4946
(anonymous function)	@	index-browser.js:2219
(anonymous function)	@	index-browser.js:220
(anonymous function)	@	index.js:14
(anonymous function)	@	index-browser.js:157
(anonymous function)	@	index-browser.js:144
(anonymous function)	@	index.js:14
````

### Info
- Environment: Browser (Web Worker)
- Platform: Chrome 54
- Adapter: IndexedDB
- Server: PouchDB Server

### Reproduce
I'm not familiar with the PouchDB internals, so I'm not sure how to reproduce it. But I do have a state in the browser that crashes consistently. I think it's trying to load a frequently modified document ('m00000016' is very small, but often modified).

The JSON string that `JSON.parse` is choking on is the following (6053 chars):
````
[...]

PR #5894, opened 2016-11-19
(#5855) - fix JSON.parse overflows

nolanlawson:
See #5855 for discussion. Apparently V8 doesn't deopt try/catches anymore.


nolanlawson:
Green!

PR #5896, opened 2016-11-20
(#5886) - fix bulkGet with reserved _id

nolanlawson:
Follow-up to #5886. Instead of using `hasOwnProperty` I opted for `pouchdb-collections`'s `Map` since this is the typical pattern we use elsewhere.

I also added a `size` property on the Map per the [Map spec](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/size) and renamed `obj.store` to `obj._store` to indicate that it's private.

I also added a new test that fails before the fix but succeeds after.

### DIFF
--- a/packages/node_modules/pouchdb-collections/src/index.js
+++ b/packages/node_modules/pouchdb-collections/src/index.js
@@ -6,39 +6,44 @@ function unmangle(key) {
   return key.substring(1);
 }
 function LazyMap() {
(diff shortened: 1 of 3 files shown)


nolanlawson:
Green!

daleharvey:
This looks good, thanks

PR #5898, opened 2016-11-20
(#5898) - process.nextTick() -> setTimeout() for browser code

nolanlawson:
A small change to improve our bundle size. Old `pouchdb.min.js` size: 141835. New size: 140201. (1.6kB reduction, 1.15% total size reduced.)

There is no need to use `process.nextTick()` in browser code. This ends up being quite a large polyfill, whereas since we're just trying to avoid releasing Zalgo, a `setTimeout()` is perfectly fine.

### DIFF
--- a/packages/node_modules/pouchdb-adapter-idb/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-idb/src/index.js
@@ -5,7 +5,8 @@ import {
   toPromise,
   hasLocalStorage,
   changesHandler as Changes,
-  uuid
+  uuid,
+  nextTick
 } from 'pouchdb-utils';
 import {
   isDeleted,
@@ -821,7 +822,7 @@ function init(api, opts, callback) {
   if (cached) {
     idb = cached.idb;
     api._meta = cached.global;
-    process.nextTick(function () {
+    nextTick(function () {
       callback(null, api);
     });
     return;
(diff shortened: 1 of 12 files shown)


nolanlawson:
Canceled build temporarily because of #5900

calvinmetcalf:
so process.browser would also be something we'd need to deal with, the plugin handles that by just inlining it to true

nolanlawson:
@calvinmetcalf we don't have any process.browser in our code now; we use `"browser"` in package.json everywhere.

daleharvey:
Yeh I took a look through the code and it looks like its only the 'extra' builds that use it (fruitdown, memory etc), would that be managed by the consumers build process?

nolanlawson:
Yeah those plugins are a "pro" feature IMO so I'm fine with shipping larger code to them.

Unfortunately though it seems this PR has a genuine failure in IE10 probably due to setTimeout() having serious perf problems in older versions of Trident. postMessage() may work as a workaround but I gotta test.

nolanlawson:
OK that should pass in all browsers

nolanlawson:
Green! Looks like `setTimeout(fn, 0)` really does work. For some reason it seems `setTimeout(fn)` wasn't working in IE10, or maybe it was just a fluke.

PR #5899, opened 2016-11-20
(#5899) - simplify PouchError code

nolanlawson:
This removes 498 bytes from `pouchdb.min.js`, by just simplifying this part of the code.


nolanlawson:
Canceled build temporarily because of #5900

Issue #5902, opened 2016-11-20
fetch-based ajax implementation is untested


nolanlawson:
Our fetch-based ajax implementation is completely untested. I am 99% sure there are bugs in Service Workers and Web Workers due to this, but we won't know until we add a test for it.

nolanlawson:
As it turns out, there are a _lot_ of bugs, and they're different bugs in Chrome vs Firefox. :confused: This commit above is just to show that I've started working on it at least.

nolanlawson:
So, Chrome fails with a CORS error for an attachment PUT, Firefox returns a 0-length `.blob()` for an attachment GET, WebKit fails with a "Preflight response is not successful," and Edge fails with "TypeMismatchError." So this is apparently not a browser bug but I'm not sure why it's failing just yet.

nolanlawson:
n/m, I found the fix

daleharvey:
So probably tonight I plan on making a patch that makes fetch our primary api for making requests, `ajax` will become purely a fetch polyfill (and fetch -> request for node) as a dependency of https://github.com/pouchdb/pouchdb/pull/5763

nolanlawson:
:+1:

nolanlawson:
For "PouchDB Lite" I am running with the assumption that users will have to polyfill Promise/fetch/Map/Set/Blob themselves, at which point I think we can get down to a sub-20KB minified size.

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/f0e0e0c86212d6e28eb37ebf76a9dc14972b3623

PR #5903, opened 2016-11-20
(#5902) - fix and test fetch() implementation

nolanlawson:
This adds a new option to our test suite, `FETCH=1 npm test`, which allows you to test the `fetch()` implementation instead of XHR.

This also corrects various bugs in the implementation, and runs the test suite against SauceLabs' version of Firefox. I use SauceLabs because Travis' Firefox 32 will not cut the mustard.

Note that the `process.env.FETCH` code is not actually shipped to users because Rollup will trim it out.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,6 +68,7 @@ env:
 
   # Testing in saucelabs
   - CLIENT=saucelabs:chrome COMMAND=test
+  - FETCH=1 CLIENT=saucelabs:firefox:49 COMMAND=test
   - CLIENT=saucelabs:safari:6 COMMAND=test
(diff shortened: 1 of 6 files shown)


nolanlawson:
I note with some sadness that this PR does _not_ pass in Chrome, apparently due to CORS issues that I've seen reported pretty frequently in pouchdb-authentication and I guess aren't fixed yet. But we can just test in Firefox for now, and recommend to our users that they avoid CORS whenever possible.

nolanlawson:
Weird, tests are failing, but I can't reproduce locally. Might be because SauceLabs is running Firefox 45. I'll pin down the version and try again.

nolanlawson:
Green!

daleharvey:
This is awesome, thanks

PR #5904, opened 2016-11-20
(#5904) - more helpful error message for invalid plugin

nolanlawson:
https://github.com/pouchdb/upsert/issues/29 indicates that the current error message is still not very helpful, since folks are still confused about what to do.


nolanlawson:
Arguably we should say something like "did you forget to call require()" but I dunno how detailed we want to go on this, since there are lots of ways to do this wrong.

nolanlawson:
I put `[skip ci]` on this because it's not worth re-running the tests.

Issue #5905, opened 2016-11-20
SauceLabs tests should be run even for non-contributors


nolanlawson:
SauceLabs has now introduced the ability to run tests even for non-contributors, but you have to change the config to get it to work. This would prevent regressions like #5900.

@halkeye showed me how to do this, but I haven't yet merged it because I'm swamped with a bazillion other things. But here's a sample PR: https://github.com/nolanlawson/html5workertest/pull/15

halkeye:
@nolanlawson travis is going to be deprecating this functionality, so its probably worth closing this again :(

gr2m:
We just make the saucelabs credentials visible to make them run for pull requests

daleharvey:
Yeh I wanna get to this one shortly

daleharvey:
Working now :)

nickcolley:
This is awesome 👏

PR #5907, opened 2016-11-20
(#5905) - remove es6-promise-pool, use smaller custom impl

nolanlawson:
| | Size before | Size after |
| --- | --- | ---- |
| min | 140K | 136K |
| min+gz | 45506 | 44779 |

This shaves off 4KB from the min size and ~1KB from the min+gz size, by replacing `es6-promise-pool` with a smaller custom implementation that does the same thing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,6 @@
     "argsarray": "0.0.1",
     "debug": "2.3.2",
     "double-ended-queue": "2.1.0-0",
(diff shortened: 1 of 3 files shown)


daleharvey:
Yeh I think using a well tested implementation then switching it out to a smaller custom if need be is a nice approach, this is looking good thanks

Issue #5906, opened 2016-11-20
Deleted documents not syncing when using filter and put()


ddolheguy:
### Issue
I am using the below setup with SYNC in my React Native app with PouchDB (pouchdb-core 6.0.7) and I found that when I delete a document (using put with _delete:true) on the client, it wasn't replicating back to the server at all.  From the server to the client works fine, but not from the client back.

BUT, if I include "**|| doc._deleted === true;**" in the _design/expense document below at the end of the current javascript line, it works fine.

Can i ask?  When PouchDB performs a sync on a _deleted document, does it only send the _ID, _REV and _DELETED fields?  I'm starting to think that my CouchDB isn't getting the field "userKey" even though I'm including it when I perform a PUT and hence the filter is ignoring it back to the server?

```
this.db.sync(remoteDb, {
   live: true,
   retry: true,
   filter: 'expense/byuserkey',
   query_params: {userKey: key}
}).on('change', change => { 
     console.log('CHANGE OCCURRED');
});

```

**CouchDB Filter** looks like this:

```
{
  "_id": "_design/expense",
  "_rev": "1-edfa576ff61f1ca52c3c0b80bc698555",
  "language": "javascript",
  "filters": {
    "byuserkey": "function (doc, req) { return doc._id === '_design/expense' || doc.userKey === req.query.userKey }"
  }
}
```

### Info
- Environment: React-Native (AsyncStorage)
- Platform: iOS (React native)
- Adapter: Async Storage
- Server: CouchDB

nolanlawson:
Your filter function will return false for your deleted document because it doesn't contain a `userKey` field. The fix is to merely add `_deleted: true` to the document and then do a `put()` instead of using `remove()`. Please see https://pouchdb.com/api.html#filtered-replication for details.

If that doesn't fix your problem, please reopen the issue. :)

ddolheguy:
Unfortunately that's exactly what I did and the local PouchDb does not replicate the updated document (with the _deleted: true property) to the server.

I performed these steps exactly:
1. Created a new document on the client as follows:
```
{
   "_id": "244af5e7-5054-4fe9-9476-ad74ececdd50",
   "data": "this is some data",
   "userKey": "91b768b4-b0f7-4850-83c1-c4a30fa87fc6"
}
```

2. This saves correctly and replicates to the server as expected with the filter settings I listed in the original part of the this issue.

[...]

PR #5918, opened 2016-11-23
test coverage for #5906

willholley:
I added tests to attempt to reproduce #5906 locally. The tests are passing locally so perhaps there is an issue with a non-standard storage engine?


nolanlawson:
@ddolheguy can you take a look and see if this test reproduces your issue? If not can you modify the test?

@willholley I'll never say no to new tests; :+1: when green.

Issue #5908, opened 2016-11-20
Reduce travis running time


daleharvey:
@nolanlawson you mentioned this on IRC, I agree we should trim some of these down

`CLIENT=node LEVEL_PREFIX=foo_ COMMAND=test` - I feel like we could remove this, code that could break this rarely changes and can be verified manually

pouchdb-server / pouchdb-express-router - So pouchdb-server has been broken for a long time (I seen its getting worked on now) pouchdb-express-router I fix whenever I see a problem so having it in the test suite is useful for me however I could see these being delegated to having to test themselves as projects

Phantomjs I have long believed should go, we test in various websql platforms and I do not think testing phantom gives us any value, we could keep it for the map reduce smoke tests or we can switch them to safari / node-websql

`CLIENT="saucelabs:internet explorer:10:Windows 8" ADAPTERS=memory COMMAND=test` this was added for a very specific bug in a non core module and its an expensive test, I think we should remove

Couchbase, I would be fine removing the test, I dont believe its actually passing

Any others we can remove or any of these ones not agreed with? if not I will PR shortly

nolanlawson:
> `CLIENT=node LEVEL_PREFIX=foo_ COMMAND=test` - I feel like we could remove this, code that could break this rarely changes and can be verified manually

Yup, I'd get rid of these. Also SERVER_ADAPTER. This hugely complicates the test code and has minimal value IMO. node-websql is worth testing, maybe pouchdb-adapter-memory but not `db: require('memdown')` which is deprecated anyway.

> Phantomjs I have long believed should go, we test in various websql platforms and I do not think testing phantom gives us any value, we could keep it for the map reduce smoke tests or we can switch them to safari / node-websql

Still gonna push back on this one, if only because it's one of the few tests that _rarely_ fails, and as soon as we replace it with Safari we're looking at longer build times and more flakiness. I vastly prefer testing inside of Travis-only where possible. If Travis ran Safari natively I'd want to use that instead though. :)

`CLIENT="saucelabs:internet explorer:10:Windows 8" ADAPTERS=memory COMMAND=test` this was added for a very specific bug in a non core module and its an expensive test, I think we should remove

We need at least one Trident-based test in there because there are _so_ many hacks in the code just for IE/Edge. Sadly Edge is failing in Sauce for some reason; I'm going to look into this and file bugs on my own team as necessary; we are in bug-fixing mode right now and can potentially fix IDB race conditions in time for Edge 15.

[...]

PR #5920, opened 2016-11-23
(#5908) - remove non-essential Travis tests

nolanlawson:
First stab at this, hopefully the configuration is correct.

Main changes:

- removed many tests, including CSG, pouchdb-server, pouchdb-express-router, Edge, LEVEL_PREFIX
- use ADAPTER=memory to test Node in-memory, no more LEVEL_ADAPTER or SERVER_ADAPTER
- Using Firefox 41.0.1 because IME it's the latest we can reliably test, better than 37 at least

I'm going to try to think of other ways to make the build faster (I bought an ngrok account; we could use that for tunneling maybe?) but this is a good start.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ sudo:
   false
 
 addons:
-  firefox: "48.0"
+  firefox: "41.0.1"
 
 before_install:
   # Because Saucelabs doesnt proxy 5984 on OSX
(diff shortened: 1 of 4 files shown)


nolanlawson:
Hm this actually fails to run Firefox 41.0.1. I don't understand why the exact same config works in worker-pouch, but I've beaten my head against the Firefox+Travis problem long enough to know to quit while I'm ahead. I'll open up a separate PR for the other changes.

PR #5921, opened 2016-11-23
(#5921) - make utils-bundle.js smaller

nolanlawson:
Before: 740K. After: 388K.

Just trying to speed up our tests a bit! :)


daleharvey:
Nice cheers

PR #5923, opened 2016-11-24
(#5923) add financier to users.md

aeharding:
PouchDB is great. It made [Financier](https://financier.io) possible, and I'd love to add it as a user to the docs!


nolanlawson:
Nice! Very happy to highlight great apps that make use of PouchDB. :)

PR #5925, opened 2016-11-24
(#5925) - remove peerDependency on pouchdb-core

nolanlawson:
Currently when you install e.g. `pouchdb-adapter-memory` along with `pouchdb`, you will get this warning because technically you have not installed `pouchdb-core`. The reason for this is that `pouchdb` is aggressively bundled (as is `pouchdb-node` and `pouchdb-browser`), so neither one contains `pouchdb-core`.

Since there's apparetly no way to have "optional" peerDeps (e.g. `pouchdb-core | pouchdb | pouchdb-browser | pouchdb-node`) I would prefer to just remove this peerDep altogether and avoid the potentially confusing warning. The downside is that people will have to just "know" that if they install multiple pouchdb-* packages they should all have the same version number or the behavior is undefined. I'm okay with this.

### DIFF
--- a/bin/update-dependencies.js
+++ b/bin/update-dependencies.js
@@ -22,14 +22,6 @@ modules.forEach(function (mod) {
   var pkgPath = path.join(pkgDir, 'package.json');
   var pkg = JSON.parse(fs.readFileSync(pkgPath, 'utf8'));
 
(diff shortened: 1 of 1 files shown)


nolanlawson:
I added `[skip ci]` to this because it wouldn't affect the tests.

PR #5926, opened 2016-11-24
(#5926) - simplify migration test code

nolanlawson:
A few changes to simplify the test code:

1. Load `pouchdb-vXXXpostfiled.js` only if we're actually going to run the migration tests
2. Delete them from the `window` object after the test is run to free memory
3. Fix a bug where we weren't actually running the tests in Firefox (only Chrome) due to a mistake in `isUsingDefaultPreferredAdapters()`

### DIFF
--- a/tests/integration/index.html
+++ b/tests/integration/index.html
@@ -18,12 +18,6 @@
       var should = chai.should();
       var assert = chai.assert;
     </script>
-    <script src='deps/pouchdb-1.1.0-postfixed.js'></script>
-    <script src='deps/pouchdb-2.0.0-postfixed.js'></script>
-    <script src='deps/pouchdb-2.2.0-postfixed.js'></script>
-    <script src='deps/pouchdb-3.0.6-postfixed.js'></script>
-    <script src='deps/pouchdb-3.2.0-postfixed.js'></script>
-    <script src='deps/pouchdb-3.6.0-postfixed.js'></script>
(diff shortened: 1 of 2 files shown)


tests/integration/index.html
```diff
@@ -18,12 +18,6 @@
       var should = chai.should();
       var assert = chai.assert;
     </script>
-    <script src='deps/pouchdb-1.1.0-postfixed.js'></script>
-    <script src='deps/pouchdb-2.0.0-postfixed.js'></script>
-    <script src='deps/pouchdb-2.2.0-postfixed.js'></script>
-    <script src='deps/pouchdb-3.0.6-postfixed.js'></script>
-    <script src='deps/pouchdb-3.2.0-postfixed.js'></script>
-    <script src='deps/pouchdb-3.6.0-postfixed.js'></script>
     <script src='utils-bundle.js'></script>
```
daleharvey: I am confused at how we are actually testing these migrations without pulling in the old versions?

nolanlawson: Yup so that's what [these lines](https://github.com/pouchdb/pouchdb/blob/741ac7ba54200548ca54c74aa7c5027786972a6c/tests/integration/browser.migration.js#L44-L57) are doing. The scripts are pulled in asynchronously only when the migration tests are actually run.

nolanlawson:
Cheers, thanks!

PR #5919, opened 2016-11-23
(#5908) - Remove a bunch of travis test configs

daleharvey:

PR #5927, opened 2016-11-24
(#5908) - test in-memory in Node, remove LEVEL_ADAPTER/LEVEL_PREFIX

nolanlawson:
Follow-up to #5919.

I think it is important to test in-memory in Node since it's used pretty frequently for e.g. unit tests and the `--in-memory` mode of pouchdb-server, but we can remove the `LEVEL_ADAPTER`/`LEVEL_PREFIX` code entirely.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,9 @@ env:
   # Test WebSQL in Node (using node-websql)
   - CLIENT=node ADAPTER=websql COMMAND=test
 
+  # Test in-memory in Node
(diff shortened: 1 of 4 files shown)


nolanlawson:
Only failing due to an intermittent in IE10.

nolanlawson:
Yup the conflict was only in `TESTING.md`, so I've gone ahead and merged as 74ffe13705560e67368e52039ced5167dac94c25

PR #5928, opened 2016-11-24
DON'T MERGE - test a bunch of MS browsers

nolanlawson:
The goal of this PR is to try to figure out which combination of IE/Edge and Windows is the least faily on Travis. I ran this experiment a few years ago and IE10+Win8 came out on top, but it's always possible something changed between now and then.

In our recent Travis runs, IE10 has been failing a *lot* due to "IDB failed, are you in private browsing mode?" so I would be eager to switch to a Trident/EdgeHTML engine that doesn't give us so many false negatives.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,9 +12,6 @@ git:
 sudo:
   false
 
-addons:
-  firefox: "48.0"
-
 before_install:
   # Because Saucelabs doesnt proxy 5984 on OSX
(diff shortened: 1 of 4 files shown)


nolanlawson:
Well, all of them failed immediately except for IE10+Win8.1 and Edge14+Win10, and those two failed with `Failed to open indexedDB, are you in private browsing mode`, which is the error I usually see when I need to restart it.

So there's probably no reason to switch away from IE10+Win8.1 as our CI tested browser, but we need to figure out why those tests are so faily. Unfortunately the actual IDB error is not being shown.

PR #5929, opened 2016-11-24
DON'T MERGE - try to test in firefox 49

nolanlawson:
We'll know if this worked, because the user agent will be listed as "49" in the test output.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ sudo:
   false
 
 addons:
-  firefox: "49"
+  firefox: "49.0.2"
 
 before_install:
   # Because Saucelabs doesnt proxy 5984 on OSX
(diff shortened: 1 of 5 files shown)


daleharvey:
So I managed to get the latest firefox running @ https://github.com/pouchdb/pouchdb/pull/5846/commits/93638432735e707c5d3edb8defb0084411c2d1a3, there are also other solutions @ https://github.com/travis-ci/travis-ci/issues/4649 

It was still broken due to the selenium gecko driver issue, but we should probably get it actually using the version that is stated then get it upgraded

nolanlawson:
Hm if that's the case then we can try 46.0.1 and 47.0.1 and 48.0, I know that geckodriver was changing a lot around those versions. But yeah I like your solution of changing the path!

nolanlawson:
But yeah I am seeing a ton of Firefox failures with "no output received," so we should definitely merge https://github.com/pouchdb/pouchdb/pull/5933 first as a stopgap

daleharvey:
Yeh I was trying to get the latest firefox working, should have split it into chunks like we have here so happy to get it working this way

PR #5931, opened 2016-11-24
(#5908) - use yarn

nolanlawson:
Travis just [announced support for yarn](https://blog.travis-ci.com/2016-11-21-travis-ci-now-supports-yarn), and it has the potential to speed up our build process. It's worth a shot.

Note that contributors are not required to use yarn themselves, but the `yarn.lock` file apparently needs to be checked in to source, so it has to be updated when dependencies are updated.


nolanlawson:
I don't see an improvement; in fact it appears to take about 30 seconds longer. Maybe it's because it has to spend extra time installing yarn itself? Either way, not worth it. https://travis-ci.org/pouchdb/pouchdb/builds/178679351

Issue #5930, opened 2016-11-24
Reduce Travis flakiness


nolanlawson:
Based on an informal audit of Travis builds (i.e. me pointing and clicking to restart tests for the past few days), our most faily tests seem to be iPhone and IE. If we can find ways to make these tests less likely to fail I would be super happy.

Some stuff I'm trying out:

1. Reducing memory usage (#5926, #5921)
2. Testing out different browser/OS combos (#5928)

I would really love if our test suite were reliably green/red rather than being nearly guaranteed to have a false negative on any given run.

nolanlawson:
Occasionally I also see `No output has been received in the last 10m0s` for the Firefox builds. I believe this is due to not being able to find the installed Firefox 48.0.

nolanlawson:
Might be related to https://github.com/pouchdb/pouchdb/pull/5898, but IE10+Win8.1 tests are _super_ flaky these days. I might have to deep-dive to see what's up.

nolanlawson:
Wow, our build is really trolling us right now. On a typical run I have to restart like 6 jobs.

nolanlawson:
OK, I'm fairly certain https://github.com/pouchdb/pouchdb/commit/4eb786b702f0db9fb2db049c71f29d7acbf18fc4 regressed us on IE10. Tests rarely seem to want to pass for that browser now.

daleharvey:
Looking into this now, firefox seems far more faily than others, however I am seeing a fair amount of fails across the board as well so doing a large travis run @ https://github.com/pouchdb/pouchdb/pull/5967 and once we know / can get node stable then ill look at firefox

nolanlawson:
I think it's reasonable to close this now. Done a lot of work here recently.

PR #5933, opened 2016-11-24
(#5930) - use default Firefox 37 in Travis

nolanlawson:
If we're not actually testing Firefox 48.0, then there's no point in installing it or referencing it in `.travis.yml`.

Additionally, I think this may be the cause of some of the intermittent `No output has been received in the last 10m0s` errors I've been seeing.

I would love to actually test the latest Firefox, but until we can figure out the right Travis+Selenium+Firefox combo that works, we should just remove this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,9 +12,6 @@ git:
 sudo:
   false
 
-addons:
-  firefox: "48.0"
-
 before_install:
   # Because Saucelabs doesnt proxy 5984 on OSX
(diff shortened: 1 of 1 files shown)


nolanlawson:
Oddly this fails for all the Firefox Travis runs, due to a timeout. So I really have no idea how to fix this.

nolanlawson:
Oh wait Firefox is also failing in other PRs. :confused:

PR #5934, opened 2016-11-24
(#5934) - simplify parseAdapter.js

nolanlawson:
Cleans up this code a bit.

pouchdb.min.js size before: 138935
pouchdb.min.js size after: 138819 (116 fewer bytes)


nolanlawson:
Failing due to an intermittent in IE10. This is green.

PR #5935, opened 2016-11-24
(#5908) - don't run minify on travis

nolanlawson:
None of our Travis tests actually use the `.min.js` files, and running uglify is expensive, so this is a waste. I clocked it on a MBA and found that `npm run build` took 42.8s whereas now `TRAVIS=1 npm run build` takes 27.27s, so this could shave off around 15 seconds from each build.


nolanlawson:
The only failures are intermittent due to IE10 and Firefox, so this is basically green.

nolanlawson:
Thanks! I agree, I want master to be stable. I'm actively working on this today, so rest assured I'm keeping an eye on the build. :)

PR #5936, opened 2016-11-24
(#5908) - bundle all browser test files

nolanlawson:
Currently our browser tests in SauceLabs are running into the 6-connection limit when fetching the JS files, which slows down our builds. This is a simple change to concatenate all of our `test.*.js` files into a bundle file rather than have 34 different scripts (I counted them).

To reduce complexity, I avoided Browserify and just did a concatenation. We don't need Browserify here, although we can add it in the future if we really want to.

I had to rename some test files from `test.*.js` to `node.*.js` to ensure they were only run in Node and not in the browser. I also renamed `test.aa.setup.js` to `aa.setup.js` to ensure it truly runs first, since both `browser` and `node` are lexicographically lesser than `test`.

To avoid bitrotting myself, I put this commit on top of two other commits that I expect will get merged soon (https://github.com/pouchdb/pouchdb/pull/5926, https://github.com/pouchdb/pouchdb/pull/5921).

Another side note: I chose to use `cat` rather than some Node equivalent because I've resorted to using Bash on Windows when I'm on Windows. I think it's simply too difficult to get our build to actually run in `cmd.exe`, and I'm the only contributor on Windows anyway.


nolanlawson:
I tested this on Chrome using 3G emulation and it's not actually any faster. I'm rescinding this PR; it's not worth the added complexity, although I did like the `browser.`/`node.` change so I might turn that into another PR later.

PR #5938, opened 2016-11-25
(#5930) - WIP - test firefox

nolanlawson:
Don't merge, just testing this out.

PR #5939, opened 2016-11-25
(#5930) - test Firefox 50 in Travis

nolanlawson:
You can check manually to confirm that this is actually running Firefox 50, but based on research in https://github.com/pouchdb/pouchdb/pull/5938 and https://github.com/pouchdb/pouchdb/pull/5846/commits/93638432735e707c5d3edb8defb0084411c2d1a3 this should work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,13 +13,15 @@ sudo:
   false
 
 addons:
-  firefox: "48.0"
+  firefox: "50.0"
 
 before_install:
   # Because Saucelabs doesnt proxy 5984 on OSX
   - "if [ -z \"$COUCH_HOST\" ]; then export COUCH_HOST=http://127.0.0.1:3000; fi"
 
 before_script:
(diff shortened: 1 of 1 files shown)


nolanlawson:
I was too nervous to try updating Selenium as well, because I'm worried it might break something somewhere else.

nolanlawson:
lol, of course it doesn't work when I do an _actual_ pull request. Man, this is frustrating.

nolanlawson:
This doesn't seem to be working any better than before. Still lots of "no output received." :( What the heck happened to Firefox on Travis? It used to work just fine, even if it was Firefox 38.

Bummed out about this.

nolanlawson:
Also it's totally inconsistent. Sometimes it works, sometimes it doesn't. WTF

nolanlawson:
testing stuff again

nolanlawson:
OK, out of 10 runs, Firefox 41.0.1 worked 9 times (the 1 failure was the same old `no output received`). So I'm happy with that. it's better than 37 at least.

Issue #5857, opened 2016-11-08
support GET /db/doc_id?open_revs=[]&latest=true


willholley:
CouchDB supports a parameter, `latest=true` when fetching a specific document revision. This has the effect of always returning the leaf revision of the document in the case that the requested revision has been updated.

When fetching a specific revision (e.g. during replication), there can be a race condition whereby the document is updated between the call to `_changes` and the document GET. The CouchDB replicator uses the `latest=true` parameter when fetching the document, ensuring that it always get the current leaf.

This ticket really has 2 parts:

 1. Support latest:true across all adapters (so requires implementation for local adapters)
 1. During replication, use `latest=true` to protect against the race condition mentioned above

willholley:
I've written a bunch of tests for this at https://github.com/pouchdb/pouchdb/tree/5857-latest-true, uncovering some [issues](
https://issues.apache.org/jira/browse/COUCHDB-3239) with [CouchDB](https://issues.apache.org/jira/browse/COUCHDB-3240) in the process. Implementing `latest=true` for local storage adapters is going to be quite thorny, particularly if we need to do it without a significant performance overhead, so I'm not entirely convinced it's worthwhile. The only time this is important is when users are likely to be updating documents mid-sync *and* compaction is occurring - that seems super unlikely for local adapters.

PR #5940, opened 2016-11-25
(#5857) - latest=true support

willholley:
* Add support for the `latest` parameter to `get()` and `bulkGet()`
 * Replicator uses `latest:true` when fetching specific revisions
 * `pouchdb-merge` now has a method to find the leaf node associated with a given rev. This isn't perfect but matches the behaviour of CouchDB 1.6. I have a slight concern about push replication performance as a result of this - it's this code which probably warrants the most review as perhaps there's possibly a more efficient solution (I'm not super-familiar with the various internal tree structures that Pouch uses).

### DIFF
--- a/packages/node_modules/pouchdb-adapter-http/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-http/src/index.js
@@ -321,7 +321,7 @@ function HttpPouch(opts, callback) {
       }
 
       for (var i = 0; i < numBatches; i++) {
-        var subOpts = pick(opts, ['revs', 'attachments']);
+        var subOpts = pick(opts, ['revs', 'attachments', 'latest']);
         subOpts.ajax = ajaxOpts;
         subOpts.docs = opts.docs.slice(i * batchSize,
           Math.min(opts.docs.length, (i + 1) * batchSize));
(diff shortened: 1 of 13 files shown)


nolanlawson:
This looks great to me, but there appears to be a true failure in the WebSQL adapter because both node-websql and phantomjs are failing.

If you want an easy way to debug in Chrome, you can add `&adapters=websql` to the URL and just test the websql code.

The merge function you wrote looks perfectly performant; we typically write it using for-loops because that part of the code is very perf-critical so it's better than using map/filter/reduce/etc which has function overhead. LGTM!

Once it's green I'm +1, note however that the current master branch is extremely faily due to Firefox and IE10 but I have separate PRs open to fix that.

willholley:
After fixing the websql adapter, tests are passing...

packages/node_modules/pouchdb-adapter-http/src/index.js
```diff
@@ -296,6 +296,10 @@ function HttpPouch(opts, callback) {
         /* istanbul ignore next */
         params.attachments = true;
       }
+      if (opts.latest) {
+        /* istanbul ignore next */
+        params.latest = true;
```
daleharvey: Any chance we can have a test that hits this so we dont need to add an istanbul ignore?

willholley: oops - I'm not sure why that ignore is there as it is covered. I'll remove it.

PR #5941, opened 2016-11-25
(#5930) - simplify setTimeout/nextTick/immediate usage

nolanlawson:
### The goal

Reduce Travis flakiness. In particular, reduce flakiness of IE10 tests.

### The problem

Our current code calls `setTimeout(fn, 0)` a _lot_. We do it for nearly every API call. From my work on the Edge team I can tell you that IE/Edge did not have a great timer implementation before Edge 14 (and in fact it's still being improved in Edge 15). Random hangs and high latency do not surprise me.

Before https://github.com/pouchdb/pouchdb/commit/4eb786b702f0db9fb2db049c71f29d7acbf18fc4 this was less of a problem because the `node-process` browserify implementation essentially queues all tasks and then executes them in a single `setTimeout()` call. But since I replaced all `process.nextTick()` with `setTimeout()` the IE10 tests are failing like ~90% of the time.

### Proposed solution

In this PR, I do a few things:

1. Remove `nextTick`/`setTimeout` entirely where it's not necessary. We only usually use it to avoid [releasing Zalgo](http://blog.izs.me/post/59142742143/designing-apis-for-asynchrony), and sometimes we are seriously overusing it.
2. Replace current `nextTick()` implementation with `immediate`. We're already shipping `immediate` in PouchDB because of `lie`, and it's a tiny codebase (1KB minified). This library has the advantage of using MutationObserver where supported, meaning that it uses microtasks and thus has very low latency. This works well even in IE10.
3. Refactor the IDB task queue so that it minifies better. This isn't strictly necessary, but I just wanted to clean up that part of the codebase.

If all goes well, we should see way fewer IE10 failures and also faster tests in general because there is a lot less `setTimeout()` indirection. I've tested locally with IE, Edge, Chrome, and FF, and everything looks good. I'm opening a PR so we can actually verify this in Travis though.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "debug": "2.3.2",
     "double-ended-queue": "2.1.0-0",
     "fruitdown": "1.0.2",
+    "immediate": "3.0.6",
     "inherits": "2.0.3",
     "level-codec": "6.2.0",
     "level-write-stream": "1.0.0",
--- a/packages/node_modules/pouchdb-adapter-http/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-http/src/index.js
@@ -3,7 +3,7 @@ var MAX_SIMULTANEOUS_REVS = 50;
 
 var supportsBulkGetMap = {};
 
-import { jsExtend as extend } from 'pouchdb-utils';
+import { jsExtend as extend, nextTick } from 'pouchdb-utils';
 import Promise from 'pouchdb-promise';
 import ajaxCore from 'pouchdb-ajax';
 import getArguments from 'argsarray';
@@ -237,7 +237,7 @@ function HttpPouch(opts, callback) {
     return setupPromise;
   }
 
-  setTimeout(function () {
+  nextTick(function () {
     callback(null, api);
   });
 
@@ -932,7 +932,7 @@ function HttpPouch(opts, callback) {
 
       if ((opts.continuous && !(limit && leftToFetch <= 0)) || !finished) {
         // Queue a call to fetch again with the newest sequence number
-        setTimeout(function () { fetch(lastFetchedSeq, fetched); }, 0);
+        nextTick(function () { fetch(lastFetchedSeq, fetched); });
       } else {
         // We're done, call the callback
         opts.complete(null, results);
--- a/packages/node_modules/pouchdb-adapter-idb/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-idb/src/index.js
@@ -39,7 +39,6 @@ import {
 } from './constants';
 
 import {
-  applyNext,
   compactRevs,
   decodeDoc,
   decodeMetadata,
@@ -48,9 +47,10 @@ import {
   idbError,
   postProcessAttachments,
   readBlobData,
-  openTransactionSafely,
-  taskQueue
+  openTransactionSafely
 } from './utils';
+
+import { enqueueTask } from './taskQueue';
 
 var cachedDBs = new Map();
 var blobSupportPromise;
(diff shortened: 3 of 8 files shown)


package.json
```diff
@@ -38,6 +38,7 @@
     "debug": "2.3.2",
     "double-ended-queue": "2.1.0-0",
     "fruitdown": "1.0.2",
+    "immediate": "3.0.6",
```
nolanlawson: We use an older version of `immediate` because it's smaller and it's what `lie` uses. This is explained in a comment in `nextTick-browser.js`.

packages/node_modules/pouchdb-adapter-http/src/index.js
```diff
@@ -237,7 +237,7 @@ function HttpPouch(opts, callback) {
     return setupPromise;
   }
 
-  setTimeout(function () {
```
nolanlawson: Anywhere where we were calling `setTimeout(fn, 0)` I replaced it with `nextTick()` just to make our code more consistent.

packages/node_modules/pouchdb-adapter-idb/src/index.js
```diff
@@ -48,10 +47,11 @@ import {
   idbError,
   postProcessAttachments,
   readBlobData,
-  openTransactionSafely,
-  taskQueue
```
nolanlawson: I moved the task queue from `utils.js` to `taskQueue.js`.

packages/node_modules/pouchdb-adapter-idb/src/taskQueue.js
```diff
@@ -0,0 +1,44 @@
+// This task queue ensures that IDB open calls are done in their own tick
+// and sequentially - i.e. we wait for the async IDB open to *fully* complete
+// before calling the next one. This works around IE/Edge race conditions in IDB.
```
nolanlawson: This code was largely copied from `utils.js` but refactored so that it would minify smaller. (e.g. `var running` instead of `taskqueue.running`).

packages/node_modules/pouchdb-utils/src/nextTick-browser.js
```diff
@@ -1,5 +1,13 @@
...
+// avoid using process.nextTick() directly because the polyfill is very large and we don't
+// need all of it (see: https://github.com/defunctzombie/node-process).
+// "immediate" 3.0.8 is used by lie, and it's a smaller version of the latest "immediate"
+// package, so it's the one we use.
+// When we use nextTick() in our codebase, we only care about not releasing Zalgo
+// (see: http://blog.izs.me/post/59142742143/designing-apis-for-asynchrony).
+// Microtask vs macrotask doesn't matter to us. So we're free to use the fastest
+// (least latency) option, which is "immediate" due to use of microtasks.
+// All of our nextTicks are isolated to this one function so we can easily swap out one
+// implementation for another.
```
nolanlawson: The above comment basically explains our entire task/microtask strategy.

packages/node_modules/pouchdb-utils/src/toPromise.js
```diff
@@ -2,27 +2,15 @@ import Promise from 'pouchdb-promise';
...
-    var usedCB;
-    if (tempCB) {
-      // if it was a callback, create a new callback which calls it,
-      // but do so async so we don't trap any errors
-      usedCB = function (err, resp) {
-        nextTick(function () {
-          tempCB(err, resp);
-        });
-      };
-    }
```
nolanlawson: This was the part of the code where we were most frequently (and most unnecessarily) calling `setTimeout()`. Note that because `tempCB` is called within a `then()`/`catch()` function, there is no reason for the extra indirection. We're already inside a microtask, we're already avoiding releasing Zalgo, so just calling the function directly is fine. Our tests still fully pass even with this change.

nolanlawson:
OK, I ran IE10 10 times and it failed 3 times. I can be happy with that for now.

screenshot 2016-11-25 13 27 15

nolanlawson:
Yeah I'm still not satisfied but I think that maybe by putting `destroy()` calls into the sequential task queue we can reduce that 3/10 to 0/10; I'll test that out right after I fix Firefox 😅

PR #5944, opened 2016-11-25
(#5930) - use Firefox 41.0.1 in Travis

nolanlawson:
Look, I have no idea why, but according to my testing 41.0.1 is the most consistent. It also works best apparently when you actually specify the version number (i.e. `selenium:firefox:41.0.1`).

See https://github.com/pouchdb/pouchdb/pull/5939#issuecomment-263025848 for details.

PR #5947, opened 2016-11-26
(#5930) - use Firefox 46.0.1 in Travis, add FIREFOX_BIN

nolanlawson:
See https://github.com/pouchdb/pouchdb/pull/5944 and https://github.com/pouchdb/pouchdb/pull/5939#issuecomment-263025848 for previous discussion.

I tried [this configuration](https://travis-ci.org/pouchdb/pouchdb/builds/179077464) 5 times, and it passed 4/5 times. I would of course prefer to use a later version of Firefox than 46 but I'm not sure if I can get them to consistently work. I couldn't get v50 to consistently work.

I don't know if this is the least faily combination, but I'm getting seriously exhausted of poking at config files and babysitting Travis, and with this configuration at least, we know there's a 0% chance that it will run the built-in Travis Firefox (v38 iirc) because I'm explicitly specifying a firefox binary when using Selenium.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,13 +13,15 @@ sudo:
   false
 
 addons:
-  firefox: "48.0"
+  firefox: "46.0.1"
 
 before_install:
   # Because Saucelabs doesnt proxy 5984 on OSX
   - "if [ -z \"$COUCH_HOST\" ]; then export COUCH_HOST=http://127.0.0.1:3000; fi"
 
 before_script:
+  - echo "using firefox $(firefox --version)"
+  - export FIREFOX_BIN=$(which firefox)
   - "export DISPLAY=:99.0"
   - "sh -e /etc/init.d/xvfb start"
     # Fail early so we dont run hours of saucelabs if we know there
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -44,22 +46,22 @@ env:
...
   # Test map/reduce
   - TYPE=mapreduce CLIENT=node COMMAND=test
-  - TYPE=mapreduce CLIENT=selenium:firefox:48.0 COMMAND=test
+  - TYPE=mapreduce CLIENT=selenium:firefox:46.0.1 COMMAND=test
   - TYPE=mapreduce CLIENT=selenium:phantomjs COMMAND=test
 
   # Testing in saucelabs
   - CLIENT=saucelabs:chrome COMMAND=test
-  - FETCH=1 CLIENT=saucelabs:firefox:49 COMMAND=test
+  - FETCH=1 CLIENT=selenium:firefox:46.0.1 COMMAND=test
```
nolanlawson: I'm really looking forward to being able to run the FETCH tests in Travis rather than SauceLabs so it can run faster.

nolanlawson:
OK, I had to restart one, but two of them are working, so this is looking very promising.

nolanlawson:
lol, of course 46.0.1 would have a bug in `fetch()` so that the FETCH test fails.

nolanlawson:
I'm going to self-+1 this because 1) all tests pass except for the FETCH one which I can fix on the way in by using `saucelabs:firefox` instead, and 2) I can't get any other PRs to pass without this, at least without restarting the build a bazillion times.

nolanlawson:
487e874e828585c170e4ecce5fd7ffe72e799367

PR #5950, opened 2016-11-27
(#5930) - run Firefox 50 in Travis

nolanlawson:
OK, so mea culpa time. It turns out 86c5694fafe592adf535ac80d0f5b84d94340dad actually does break the build, because it turns out adding FIREFOX_BIN to `saucelabs:firefox` prevents Selenium from figuring out to use Firefox in SauceLabs rather than Selenium (I guess that's what's going on?). Also unfortunately Firefox 46.0.1 legit fails our fetch tests.

I think we can just upgrade to Firefox 50. It might fail a few times but we can just restart it. Oddly it seems like these tests always fail the first time and then succeed after. So hopefully this will work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ sudo:
   false
 
 addons:
-  firefox: "46.0.1"
+  firefox: "50.0"
 
 before_install:
   # Because Saucelabs doesnt proxy 5984 on OSX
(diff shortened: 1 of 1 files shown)


nolanlawson:
Cool, Firefox 50 just straight-up does not work, even though it used to. Also PhantomJS is down because I guess Bitbucket is rate-limiting downloads. Back to the old drawing board!

PR #5953, opened 2016-11-27
transitive .defaults()

### LINKED ISSUE #5922: .defaults(options) does not inherit options
### Issue

We use `PouchDB.defaults(options)` extensively. I would expect that if I do something like this:

```js
var MyPouchDB = PouchDB.defaults({foo: 'bar'})
var MyNewPouchDB = MyPouchDB.defaults({baz: 'qux'})
```

then both options (`{foo, baz}`) would be passed by the `MyNewPouchDB` constructor implicitly, but this does not seem to be the case. Is this by design? If this is a bug, I’d be happy to look into it

### Info
- Environment: tested in latest Chrome 54 on pouchdb.com
- Platform: Chrome

### Reproduce
This tries to send a request to `https://example.com/db` with the `x-myheader` header, as expected 👍 

```js
MyPouchDB = PouchDB.defaults({
  ajax: {
    headers: {
      'x-myheader': 'foobarbaz'
    }
  }
})
var db = new MyPouchDB('db')
db.replicate.to('https://example.com/db')
```

But this does not (note the additional `.defaults({})`

```js
MyPouchDB = PouchDB.defaults({
  ajax: {
    headers: {
      'x-myheader': 'foobarbaz'
    }
  }
}).defaults({})
var db = new MyPouchDB('db')
db.replicate.to('https://example.com/db')
```

gr2m:
closes #5922

### DIFF
--- a/packages/node_modules/pouchdb-core/src/setup.js
+++ b/packages/node_modules/pouchdb-core/src/setup.js
@@ -69,7 +69,7 @@ PouchDB.defaults = function (defaultOpts) {
       delete opts.name;
     }
 
-    opts = extend({}, defaultOpts, opts);
+    opts = extend({}, PouchAlt.__defaults, opts);
     PouchDB.call(this, name, opts);
   }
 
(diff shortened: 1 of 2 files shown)


nolanlawson:
There's one failing Firefox test that I have restarted like 3 times. >_<

This PR looks great to me, and I am happy to merge even though it's red.

PR #5956, opened 2016-11-27
fix plugin-import in readme

pubkey:
In the docs there is a mistake on how to import the adapter into the PouchDB-module.


nolanlawson:
Thanks! Gonna merge this with `[skip ci]` because no need to run CI tests.

Issue #5959, opened 2016-11-28
'mainVersion' is assigned a value but never used


daleharvey:

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/9c22e28c5f4623b6d7296cef0e9c29e0ee8abcc1

PR #5960, opened 2016-11-28
(#5959) - Remove unused variable

daleharvey:


daleharvey:
lol when I merged https://github.com/pouchdb/pouchdb/commit/c50c067394b57b0d9ff6aa28f9eb5dc3bfcb0554 I was about to comment on whether or not we should be skipping ci on code changes even if we think it will have no effect but I also thought it would be ok. But yeh unless its a string / docs change then we should run the tests

daleharvey:
lol ok 2 minor PR's with no test run and both broke tests, I should have payed attention before merging but yeh lets avoid [skip ci] unless its a doc change :)

nolanlawson:
+1 when green

Issue #5973, opened 2016-12-02
Build failing: java not found


nolanlawson:
Seems this is a recent Travis failure: https://github.com/travis-ci/travis-ci/issues/6928#issuecomment-264227708

nolanlawson:
fixed by Travis folks

Issue #5962, opened 2016-11-28
`npm run verify-build` gives false positives in Travis due to no minification


nolanlawson:
Right now it says "0 is smaller than 50000" which means it's not really working.

We should just run minification for that one build so that we can measure the min+gz size. Simply running uglify js on `pouchdb.js` should be good enough.

nolanlawson:
9abef3a162f9a117497aaca9c3273d20cf7bb23c

PR #5975, opened 2016-12-02
(#5962) - fix broken verify-bundle-size.sh script

nolanlawson:
Will probably fail due to #5973 but here's the PR anyway, can rebase later...

### DIFF
--- a/bin/verify-bundle-size.sh
+++ b/bin/verify-bundle-size.sh
@@ -4,7 +4,9 @@
 # We might have to change this later.
 MAX=50000
 
-SIZE=`gzip -c packages/node_modules/pouchdb/dist/pouchdb.min.js | wc -c`
+# testing pouchdb.js instead of pouchdb.min.js because minification isn't run in Travis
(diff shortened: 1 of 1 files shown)


nolanlawson:
Also the only test that matters is the "verify build" test which doesn't require Java, so we can just ignore the other failures

nolanlawson:
Went ahead and rebased on thing to workaround travis bug

nolanlawson:
de-rebased because the Travis bug seems fixed

daleharvey:
How was this not failing when we stopped minifying in travis?

nolanlawson:
The size was 0 and therefore it passed.

daleharvey:
Ah duh yeh, could you put a check in there for >1 to make sure that doesnt happen again?

nolanlawson:
Sure thing!

nolanlawson:
9abef3a162f9a117497aaca9c3273d20cf7bb23c

PR #5981, opened 2016-12-08
Test closed and destroyed (WIP)

shimaore:
Opening this PR to test the changes I submitted for #4868 (since I can't test safari on my private branch).

So far the code is the same, but the tests have been moved to a new test file so that I can focus on why the `before` hook is failing.


shimaore:
Well, the safari test is passing now. However coverage segfaulted on an unrelated test:
```
   ✓ issue #1001 cb as 3rd argument
bin/test-node.sh: line 50:  4827 Segmentation fault      (core dumped) ./node_modules/.bin/istanbul cover --no-default-excludes -x 'tests/**' -x 'node_modules/**' ./node_modules/mocha/bin/_mocha -- $BAIL_OPT --timeout $TIMEOUT --require=./tests/integration/node.setup.js --reporter=$REPORTER --grep=$GREP $TESTS_PATH
```

shimaore:
@nolanlawson @daleharvey I'll let you review this, apparently the code has been correct all along and I can't really address the question of the segfault in istanbul.

nolanlawson:
@shimaore Istanbul randomly throws segfault errors in Travis. I've restarted it and hope it passes...

nolanlawson:
Cool, everything is green now, but I have some questions about the PRs:

1. What's with the FIXMEs? You're saying that the tests are not actually correct?
2. Why the `this.timeout()`? Seems these are making the default timeout really short (it's normally 120000)

nolanlawson:
Ah wait just remembered that if the problem is we need to test Safari, then this needs to be a PR opened by a contributor unfortunately. There's an open issue to fix that: https://github.com/pouchdb/pouchdb/issues/5905

nolanlawson:
When https://github.com/pouchdb/pouchdb/pull/5995 passes we'll know if this is truly green or not...

shimaore:
### Timeout
I set the timeout very short because the tests don't need that long to succeed since most of them basically create and destroy the database, and there's not point in waiting if the test fails. I'm happy to remove them if you feel it adds unnecessary complexity.

### Optional tests
The `FIXME` are areas where I believe PouchDB is not clear on the semantics of `close()` and `destroy()`; I believe this to be adapter-related, I also consider it a minor nuisance (who cares if a `close()` fails?), and all-in-all I wanted to point it out, but did not want it to be a "hard failure" since my interpretation might be wrong. Keep in mind that the goal of this PR is to address the memory leak in PouchDB core, and these are unrelated design questions; I'd rather remove these tests than keep trying to have this PR merged because of it, but it wouldn't be honest to uncover the issue and not report it. Hence the `FIXME` indications, which point to things that are to be fixed (or not, see below) elsewhere in PouchDB, but which also indicate that the tests should be reviewed once this is done.

For a given, single PouchDB instance, it's pretty clear that both `close` and `destroy` are _final_ on that instance — once you call either of them, no further operation can be done on that instance. I also interpret the semantics of `close` as "disconnect from the underlying 'physical' representation of the data" and those of `destroy` as "remove the underlying 'physical' representation of the data and disconnect from it", which might not be correct.

Where things get messy is if you have multiple instances of PouchDB pointing to the same underlying representation. For example if one does `info()` then `close()` on one instance, and intermixes a `destroy()` (asynchronously) on another instance so that it happens _after_ the `info()`, the results are not consistent. I would expect the database to be created then destroyed, but I would also expect the `close` and the `destroy` to be _both_ successful. However this isn't _always_ the case, and the outcome depends on the adapter.

I'm not sure those situations have been considered in the design of the adapters, so I considered those as tests that might fail for now, pending reviews of the adapters (if that's considered a worthy goal, which it might not, in which case those tests should simply be removed).

nolanlawson:
> For example if one does info() then close() on one instance, and intermixes a destroy() (asynchronously) on another instance so that it happens after the info(), the results are not consistent. 

This sounds like something that could be addressed in the `adapter.js` file so that it affects all adapters; I'm guessing it's due to us hanging the `destroyed` status onto the instance rather than some kind of global mappings of name+adapters to PouchDB objects.

> Keep in mind that the goal of this PR is to address the memory leak in PouchDB core, and these are unrelated design questions

Thanks for the explanation; I was trying to understand the totality of this PR. Now that it's clearer, I'd like to have this merged and shipped in PouchDB 6.1.0. I will remove the timeouts however as I think they will probably lead to flaky tests in the future and aren't really necessary.

Thanks a lot for your hard work on this PR; this seems like an incredibly complicated problem, but everything LGTM!

nolanlawson:
https://github.com/pouchdb/pouchdb/pull/5995 was fully green, merging

nolanlawson:
Ah dang unfortunately we need to run the tests again; see https://github.com/pouchdb/pouchdb/pull/5998

daleharvey:
Sorry, this is failing consistently, https://travis-ci.org/pouchdb/pouchdb/builds/183284723

shimaore:
Yup, saw that. I don't have access to Safari, so I can't debug it. However apparently phantomjs has an issue at the same place as Safari, in the `beforeEach` hook (although I don't remember phantomjs running into issues last time we ran those); I'll see if I can reproduce and track it down in phantomjs.

nolanlawson:
@shimaore In case you have access to a Linux machine, the Epiphany browser (`apt-get install epiphany-browser`) should exhibit this same issue.

shimaore:
@nolanlawson unless I'm mistaken this should be good to merge against master, the only failing test is coverage but there is a [core dumped in istanbul](https://travis-ci.org/pouchdb/pouchdb/jobs/230311056#L5284).

daleharvey:
Unfortunately contributors PR's dont test against safari, I am doing a full test run @ https://github.com/pouchdb/pouchdb/pull/6485

daleharvey:
Sorry for the large delay on this, and many thanks for the PR + the tests, its looking great, I am happy to merge, also you have made plenty of PR's so I will add you as a committer to the pouchdb repo :) many thanks

daleharvey:
(and just to be clear for others, https://github.com/pouchdb/pouchdb/pull/6485 is full green), Now that you are on the committers list @shimaore the full test suite will run on your PR's by default

Issue #5990, opened 2016-12-11
Use native Map/Set, fallback to polyfill


nolanlawson:
We were avoiding doing this because at one point there were bugs in Chrome's implementation, but I think it's worth testing this out again, so that we can know how easy it will be to drop these polyfills in the future.

nolanlawson:
fixed in https://github.com/pouchdb/pouchdb/commit/4e35f43c1b0b0338ef2e6bcdc9c9e20068b2fc21

PR #5991, opened 2016-12-11
(#5990) - use native Map/Set

nolanlawson:


daleharvey:
So happy with this, but when we look at these things nice to check what we need the polyfills for, taking a look at http://kangax.github.io/compat-table/es6/#test-Set it seems like the polyfill will only be needed for Android <= 5.0 right? which according to https://developer.android.com/about/dashboards/index.html is starting to get to the tail end of users but still close to 50% I would say 4.4 is a reasonable minimum at the moment but hopefully soon android users will upgrade and we can drop this

Another thing I noticed is ie11 looks to have a broken Map/Set implementation, I dont have a windows box to test yet so may be worth giving it a quick check before landing

Cheers

nolanlawson:
Thanks, I have IE11 so I can confirm manually before merging. And yes, at some point we should just remove all our polyfills or make them optional.

nolanlawson:
BTW if you need a Windows VM :wink: : http://edge.ms

nolanlawson:
Ran some quick perf tests using temp-views with in-memory at 100 iterations. Seems this change is a slight perf boost in all browsers:

| Browser | Before | After |
| --- | --- | ---- |
| Chrome 54 | 47606ms | 46589ms |
| Firefox 50 | 71126ms | 70770ms |
| Edge 15 (14994) | 52817ms | 50371ms |

nolanlawson:
IE11 indeed has a bug that causes our tests to fail. (`/tests/integration/?grep=test%5C.changes%5C.js-local%20%232569%20Non-live%20doc_ids%20filter` for those curious). Looks like we should sniff for IE and avoid its Map implementation.

Edge is fine, BTW.

nolanlawson:
This is also failing in iOS 8.4. I've restarted several times, and no dice. Seems a true failure.

nolanlawson:
According to the Kangax tables we can use this to sniff for Edge 12+ and iOS 10+ Map support:

```js
function supportsMapSet() {
  if (typeof Symbol === 'undefined' || typeof Map === 'undefined' || typeof Set === 'undefined') {
    return false;
  }
  var prop = Object.getOwnPropertyDescriptor(Map, Symbol.species);
  return 'get' in prop && Map[Symbol.species] === Map;
}
```

I've tested manually in Chrome, FF, Edge, and IE11 and this looks good.

nolanlawson:
Green! :)

PR #5992, opened 2016-12-11
(#5991) - use fork of debug to avoid process polyfill

nolanlawson:
I have a PR on debug (https://github.com/visionmedia/debug/pull/345) to avoid pulling in the entire [node-process](https://github.com/defunctzombie/node-process) polyfill, but since we're going to do a release soon, let's just use my fork for now.

PouchDB min+gz size before: **45298**
After: **44796** (502 bytes saved)

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
   },
   "dependencies": {
     "argsarray": "0.0.1",
-    "debug": "2.3.2",
+    "debug": "nolanlawson/debug#abbe8a8",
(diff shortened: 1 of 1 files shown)


nolanlawson:
Based on discussion in https://github.com/visionmedia/debug/pull/345 I think we should solve this with a custom Browserify transform or by switching to Rollup (and thus omitting rollup-plugin-node-builtins).

PR #5998, opened 2016-12-12
Test #5981 in SauceLabs yet again

nolanlawson:
Unfortunately I just realized that this new `test.close.js` actually wasn't being run in any browsers, because you have to manually add `test.close.js` to the `index.html` file in `test/integrations` (really easy mistake to make; we should automate that).

~~I'm willing to hold off on 6.1.0 until this passes; it's been long enough and I'm fairly certain it'll pass.~~ On second thought let's just do a release, there are lots of changes waiting and this one can go into a quick 6.1.1


daleharvey:
Failing, closing back to the original

Issue #5948, opened 2016-11-26
v6.1.0


nolanlawson:
Once the build has settled down a bit, I think it's time to do a patch release. There are enough critical fixes in the current build (`process.nextTick` fix, `fetch()` fix, etc.), and we're close to the start of December.

daleharvey:
Sounds good, lets make sure we get the current PR's merged though, I wouldnt like for someone to have submitted a PR and have it sat idle for a while and have to wait on the next release, that also gives us time to be confident that the tests are back to being stable then (I should be able to start helping out more shortly)

nolanlawson:
All of the current open PRs have true failures AFAICT and need work: https://github.com/pouchdb/pouchdb/pulls

I'd be happy to do a release this weekend. I have some free time because I'm traveling, and would be happy to do a quick 6.0.8 without a new blog post.

nolanlawson:
Looks like this will actually be 6.1.0 due to https://github.com/pouchdb/pouchdb/issues/5857.

We should probably write a full blog post for this, since it contains lots of stuff.

PR #5999, opened 2016-12-12
(#5948) - update 6.1.0 blog post slightly

nolanlawson:
Looks like something increased the size slightly from the last time I measured, lol. Still smaller tahn 6.0.7 though, and once the `debug` fix is in we'll lose 500 bytes more.


nolanlawson:
already published btw

PR #6002, opened 2016-12-13
(#6002) - remove excessive debug dep

nolanlawson:
We have two dependencies on `debug` (a regular dep and a devDep); we only need one.


nolanlawson:
Doing `[skip ci]` because this change is fairly straightforward.

nickcolley:
👍  When green 😄

PR #6003, opened 2016-12-13
use global-replaceify to avoid process polyfill

nolanlawson:
I wrote [a simple Browserify transform](https://github.com/nolanlawson/global-replaceify) to replace global variables at build time. Using it to rename `debug`'s usage of `process` to `__process` and thus avoid Browserify pulling in [node-process](https://github.com/defunctzombie/node-process) as a dependency, we get these savings:

- min+gz before: 45297
- min+gz after: 44780 (517 bytes saved)

### DIFF
--- a/bin/build-pouchdb.js
+++ b/bin/build-pouchdb.js
@@ -129,8 +129,19 @@ function doBrowserify(filepath, opts, exclude) {
             path.dirname(filepath) + '/' + path.basename(filepath));
         });
     } else {
-      bundler = browserify(addPath(filepath), opts)
-        .transform('es3ify')
+      bundler = browserify(addPath(filepath), opts);
+      if (opts.standalone === 'PouchDB') {
+        // Force Browserify to ignore global `process` references, since `debug` pulls this in and
+        // it's large. Only do it for the main PouchDB package, though; for the plugin adapters, they
+        // actually require `process`.
+        bundler.transform('global-replaceify', {
+          replacements: {
+            process: '__process'
+          },
+          global: true
+        });
+      }
+      bundler.transform('es3ify')
         .plugin('bundle-collapser/plugin');
     }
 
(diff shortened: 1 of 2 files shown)


nolanlawson:
BTW without counting gzip, we go from 140561 to 138930 bytes, i.e. 1631 bytes saved (1.1% total)

nolanlawson:
Whoops, needed to ignore it for fruitdown/memory/localstorage...

daleharvey:
Honestly I was excited about debug() when we landed it, but in reality I very rarely use it, when I try to it is always missing the information I need to debug and its a hassle to filter the things I dont need so I generally add and remove my own logging anyway, its been a source of bugs in the past and adds a non negligible amount of code to our build

This smells like code that is going to be problematic between the various platforms and build configurations we use, how about we just remove debug?

nolanlawson:
If we remove `debug` then it's a breaking change. I know that some folks at least have found it useful and despite its warts it seems to be what the Node community has settled on (e.g. Babel uses it).

I never use `debug` either, though. I think maybe what we could do is expose a generic API that emits events from PouchDB, document that, and then make `debug` a plugin. Again this would be a breaking change but maybe it's worth it for one less dep. We could also take this opportunity to somehow turn our polyfills into a separate plugin (included by default).

As-is, this particular change is actually very non-breaking to the ecosystem because it will only affect our `dist/pouchdb.js` file. In fact it won't affect Browserify/Webpack/Rollup consumers who will still get the same large dependency. I proposed this PR mostly as a way to advertise a smaller bundle size, even if it's only available to `script` tag users.

nolanlawson:
On second thought I enjoy this idea if making `pouchdb-debug` a plugin (excluded by default) and then shipping a 7.0.0. I'll also look into a `pouchdb-polyfills` which would probably be included by default even in a 7.0.0 (Promise, Map, Set, fetch, Blob, etc.).

PR #6013, opened 2016-12-14
(#6013) - use rollup-plugin-inline-typeof

nolanlawson:
This adds a new Rollup plugin I wrote that inlines `typeof` invocations (see [the explanation](https://github.com/nolanlawson/rollup-plugin-inline-typeof#why)) for details). This reduces our min+gz size from 45297 to 45261 and our min size from 140561 to 138467 (0.14% smaller, but hey, at least we didn't grow!).

I haven't yet run performance tests to see if this moves the needle on any of our perf tests, but it reduces the bundle size and doesn't add much weight to the build, and furthermore I want to double-check that our tests still pass in Travis. Consider this experimental until it's green and I've run some more perf tests.

### DIFF
--- a/bin/build-pouchdb.js
+++ b/bin/build-pouchdb.js
@@ -18,6 +18,7 @@ var browserify = require('browserify');
 var browserifyIncremental = require('browserify-incremental');
 var rollup = require('rollup');
 var nodeResolve = require('rollup-plugin-node-resolve');
+var inlineTypeof = require('rollup-plugin-inline-typeof');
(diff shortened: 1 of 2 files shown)


nolanlawson:
Nevermind, this actually introduces bugs due to `isUndefined(chrome)` and doesn't improve execution time of `temp-views` using the memory adapter: 10 iterations went from 43683ms to 46337ms in Chrome. Interesting experiment, though!

PR #6014, opened 2016-12-16
(#6014) - update memdown to 1.2.4

nolanlawson:
Based on some quick performance testing, this seems like it should be a big boost thanks to the nextTick -> immediate change. I can run some benchmarks once we see that it's green.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "lie": "3.1.0",
     "localstorage-down": "0.6.7",
     "ltgt": "2.1.2",
-    "memdown": "1.2.2",
+    "memdown": "1.2.4",
     "readable-stream": "1.0.33",
     "request": "2.78.0",
     "scope-eval": "0.0.3",
--- a/tests/integration/test.issue3179.js
+++ b/tests/integration/test.issue3179.js
@@ -134,7 +134,11 @@ adapters.forEach(function (adapters) {
             var conflictsEqual = JSON.stringify(localDoc._conflicts || []) ===
               JSON.stringify(remoteDoc._conflicts || []);
             if (!revsEqual || !conflictsEqual) {
-              return waitForUptodate();
+              // we can get caught in an infinite loop here when using adapters based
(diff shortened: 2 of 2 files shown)


nolanlawson:
This is a true failure. Firefox is throwing up the "a script is causing this page to run slowly" dialog which AFAIK is caused by `immediate` getting caught in an infinite loop.

Subbing in `window.setTimeout` for `immediate` fixes the issue but also slows it down a ton due to increased latency. I think we should either try to fix `immediate` in memdown or use another `nextTick` implementation.

nolanlawson:
Can be repro'd in Chrome as well

nolanlawson:
I tried immediate 3.0.6 and 3.2.3 and both exhibit this problem. This may take quite a bit of effort to reduce the error and fix it in memdown.

@calvinmetcalf you may be interested in this issue. To repro:

```
hub clone pouchdb/pouchdb --single-branch --branch memdown-1.2.4 --depth 1
cd pouchdb
npm i && npm run dev
# open localhost:8000/tests/integration?grep=local in a browser
```

calvinmetcalf:
Hm have you checked out microtask? That might fix your issue assuming it's that recursive usage of immediate is pegging the event loop

nolanlawson:
OK, I've determined that `microtask` exhibits this behavior, and that a simple `var resolved = Promise.resolve(); function nextTick(fn) { resolved.then(fn) }` also exhibits it. So it's either a bug in memdown, in pouchdb, or in our tests apparently. I'll keep digging.

nolanlawson:
Figured it out. The problem was that a couple of our tests weren't really prepared for the underlying adapter to be based on microtasks, and could run into an infinite loop.

So basically memdown is so fast now, it revealed a bug in our tests. :smiley:

calvinmetcalf:
> Hm have you checked out microtask?

I meant macrotask but glad you were able to figure it out 😄

PR #6015, opened 2016-12-16
(#6015) - use Map and Set in mapreduce

nolanlawson:
Been trying to find ways to improve perf for secondary indexes, this seemed like an obvious change given #5991 showed that the native Map/Set implementations are faster than our shims. Citation needed, but should be a perf boost.  

Rebased on #5991 to avoid bitrotting myself.

### DIFF
--- a/packages/node_modules/pouchdb-collections/src/Set.js
+++ b/packages/node_modules/pouchdb-collections/src/Set.js
@@ -16,5 +16,15 @@ Set.prototype.add = function (key) {
 Set.prototype.has = function (key) {
   return this._store.has(key);
 };
+Set.prototype.forEach = function (cb) {
+  this._store.forEach(function (value, key) {
+    cb(key);
+  });
+};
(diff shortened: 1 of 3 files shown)


nolanlawson:
BTW one major reason I'm doing this is that during my perf analyses I see lots of time spent in garbage collection, so I'm hunting down unnecessary allocations. This `uniq()` function was a major offender, but after this fix it barely registers.

nolanlawson:
True failure; PhantomJS is not fond with how we're using Map and Set.

nolanlawson:
Hm, I can repro the Phantom failure in Safari 5, but not in Safari 6. And I'm having a heck of a time running PhantomJS locally.

nolanlawson:
Awesome, this actually revealed a bug in our Map/Set implementation that wouldn't have been revealed otherwise. Also it turns out Safari 5 running on Windows 7 in VirtualBox is a good way to debug PhantomJS errors!

This should be green after the next run.

PR #6023, opened 2016-12-16
(#6023) - travis: fail if Node 6 fails, test in Node 7

nolanlawson:
Node 6 has been passing for awhile, so we can remove it from the "allowed failures." However it may be a good idea to start look at Node 7 and see if we are passing in that version or not. Node 4 is LTS and so I left it in.

I avoided terms like `stable` because I'm not a fan of rolling versions; I prefer to nail these down as we currently do for most browsers we test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 
 node_js:
-  - "5"
+  - "6"
 
 services:
   - docker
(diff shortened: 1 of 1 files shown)


nolanlawson:
Node 7 is passing FWIW, but we can leave it an allowed failure until it's stable (or until 8 comes out or whichever is LTS)

Issue #6012, opened 2016-12-14
Replace js-extend with Object.assign (with polyfill)


andrew-aladev:
Hello. We are trying to use [forever agent](https://github.com/thaliproject/forever-agent/blob/master/index.js) as an option for `ajaxCore -> request.js`. `js-extend` tries to make a deep copy of this agent. Sometimes it is possible, but sometimes it creates a recursion overflow, because existing `freeSocket` is a circular object.

I am sure that we should not use deep copy of `request.js` options. `http.Agent` is just an simple example.

I've replaced `packages/node_modules/pouchdb-utils/src/jsExtend.js` with [`Object.assign`](https://github.com/sindresorhus/object-assign/blob/master/index.js) and `npm run test-node` passed.

I see failed unit tests: `Test two levels merging`, `Test multilevel merging`, `Test if non object argument raises exception`. I've tried to find a reason for these tests but I failed. It looks like all `extend` calls does not require multilevel merging and non object exception.

Can you replace `js-extend` with `object-assign`?

nolanlawson:
Can you provide a failing test case so we can understand what's failing about jsExtend? And yes, we are interested in removing our polyfills, so replacing jsExtend with Object.assign isn't out of the question.

andrew-aladev:
```
var jsExtend = require('js-extend');

var circular = {};
circular.circular = circular;

var defaults = { a: null };
var options  = { a: circular };
console.log(Object.assign({}, defaults, options));
console.log(jsExtend.extend({}, defaults, options));

defaults = { a: circular };
console.log(Object.assign({}, defaults, options));
console.log(jsExtend.extend({}, defaults, options));
```

>{ a: { circular: [Circular] } }
{ a: { circular: [Circular] } }
{ a: { circular: [Circular] } }
Maximum call stack size exceeded

Please look [here](https://github.com/pouchdb/pouchdb/blob/master/packages/node_modules/pouchdb-adapter-http/src/index.js#L171).

`var reqOpts = extend(clone(ajaxOpts), userOpts.ajax, options);`

We are passing our custom agent as an `ajax` option. But `ajaxOpts` already has another agent. If both agents are circular - it will create a hang and resursion overflow. This exception appears randomly because both http agents can be circular or not circular depends on it's usage. But pouchdb shouldn't ever try to merge these objects.

[...]

PR #6024, opened 2016-12-16
(#6012) - replace js-extend with Object.assign

nolanlawson:
Goal is to start moving to the official `Object.assign` which can eventually be moved into `pouchdb-polyfills`. Also should fix errors described in #6012.

I fully expect this PR to fail in some browser/Node or another, but let's test...


nolanlawson:
This appears to be a true failure. We will need to investigate and see what we need to do to replace js-extend.

nolanlawson:
OK, the bug was in the test code itself. Hopefully this should pass.

Incidentally I've removed `test.jsExtend.js` since we are now just using `Object.assign` so there's no reason to unit test it. I'm not sure what it was adding if all our integration tests are passing, and I don't believe any of our options objects really _need_ to be deeply cloned (`ajax` on the constructor? maybe?) so I'm happy to just use Object.assign in that case.

As for cloning we are still using a separate algorithm for that, and we still have `test.clone.js` which ensures we don't regress on e.g. https://github.com/pouchdb/pouchdb/pull/5201 so I'm fairly confident this is a safe change.

daleharvey:
This is awesome, thanks

PR #6030, opened 2016-12-18

nolanlawson:
Rebased version of https://github.com/pouchdb/pouchdb/pull/5868 in order to fully test.


nickcolley:
Do we know when CouchDB plans to ship this feature? I feel like we shouldn't add a feature unless we know it's going to land in CouchDB?

KlausTrainer:
@nickcolley This already has been shipped with CouchDB 2.0.0. See http://docs.couchdb.org/en/2.0.0/api/document/common.html#put--db-docid

nickcolley:
Nice, does https://github.com/pouchdb/pouchdb/pull/6030/files#diff-6d47f22e410ef9ff2b2424a38dca5004R47 need to be reviewed or should it still be skipped?

daleharvey:
It would need changes to the way it is implemented to avoid being skipped, I have repoened the conversation in the original PR @ https://github.com/pouchdb/pouchdb/pull/5868

PR #6033, opened 2016-12-20
(#6031) - faster IDB changes() with batched cursor

nolanlawson:
This is a rewrite of #6031. Instead of having two separate implementations of `changes()`, I've implemented a new `runBatchedCursor()` routine that abstracts away two implementations:

1. `getAll()`/`getAllKeys()`-based, for browsers that support it
2. `IDBCursor`-based, for browsers that don't support `getAll()`/`getAllKeys()` or for queries where we need cursors (e.g. descending)

**Mode 1** essentially works by fetching `n` changes for every batch, where `n` corresponds to `limit` in the `changes()` API. In the case of `auto_compact`ed databases with no conflicts, and when we're not using `opts.filter` or `opts.doc_ids`, this means that we can accomplish `changes()` with a single batch (notably this applies to secondary indexes, giving them quite a perf boost).

In other cases, we may slightly overfetch, but I'd argue this is still better than fetching one-at-a-time, and that we're still honoring the user's preferences for memory consumption (communicated via `limit`). In cases where `limit` is -1/undefined we do not use Mode 1 since it would mean reading the whole DB into memory.

**Mode 2** works like a classic IDBCursor, as we've done in the past. It's abstracted into Mode 1 by acting like a batched cursor where the batch size is 1.

I ran a benchmark using Firefox 50 and Chrome 55 on a MacBook Air, and got the following results for `temp-views` with 10 iterations:

| --- | ---- | ---- | ---- |
| Chrome | 104826ms | 51566ms | 50.8% |
| Firefox | 97959ms | 55441ms | 43.4% |

So roughly, this makes secondary index creation in IDB about 2x faster in Chrome and Firefox. It also has the potential to make replication faster (when pulling from IDB databases) but I need to write a new benchmark to confirm it.

### DIFF
--- a/packages/node_modules/pouchdb-adapter-idb/src/changes.js
+++ b/packages/node_modules/pouchdb-adapter-idb/src/changes.js
@@ -1,0 +1,212 @@
+import changesHandler from './changesHandler';
+import {
+  clone,
+  filterChange,
+  uuid
+} from 'pouchdb-utils';
+import {
+  Map,
+  Set
+} from 'pouchdb-collections';
+import {
+  ATTACH_STORE,
+  BY_SEQ_STORE,
+  DOC_STORE
+} from './constants';
+import {
+  decodeDoc,
+  decodeMetadata,
+  fetchAttachmentsIfNecessary,
+  idbError,
+  postProcessAttachments,
+  openTransactionSafely
+} from './utils';
+import runBatchedCursor from './runBatchedCursor';
+
+function changes(opts, api, dbName, idb) {
+  opts = clone(opts);
+
+  if (opts.continuous) {
+    var id = dbName + ':' + uuid();
+    changesHandler.addListener(dbName, id, api, opts);
+    changesHandler.notify(dbName);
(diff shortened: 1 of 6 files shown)


packages/node_modules/pouchdb-adapter-idb/src/changes.js
```diff
@@ -0,0 +1,212 @@
...
+  txn.oncomplete = onTxnComplete;
+
+  bySeqStore = txn.objectStore(BY_SEQ_STORE);
+  docStore = txn.objectStore(DOC_STORE);
+  docIdRevIndex = bySeqStore.index('_doc_id_rev');
+
+  runBatchedCursor(bySeqStore, opts.since, opts.descending, limit, onBatch);
+}
+
+export default changes;
```
nolanlawson: I was uncomfortable with having so much `changes` logic in `index.js` so I moved it to its own file.

daleharvey:
That is a huge speedup, awesome job 👍

PR #6040, opened 2016-12-20
(#6040) - fix coveralls badge issue in README

nolanlawson:
Apparently Coveralls has a bug where it's not showing 100% even though when you click through clearly everything is at 100%, and the bug has been open for a few years ( https://github.com/lemurheavy/coveralls-public/issues/414#issuecomment-268376344 ). Since our build straight-up fails if we're at <100 I feel comfortable pulling a Volkswagen here and just cheating.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 [PouchDB](http://pouchdb.com/) – The Database that Syncs!
 =========
 
-[![Build Status](https://travis-ci.org/pouchdb/pouchdb.svg)](https://travis-ci.org/pouchdb/pouchdb) [![Coverage Status](https://coveralls.io/repos/pouchdb/pouchdb/badge.svg?branch=master&service=github)](https://coveralls.io/github/pouchdb/pouchdb?branch=master)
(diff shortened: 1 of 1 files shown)


sbts:
@nolanlawson @daleharvey 
Hi, don't know if you are aware, but coveralls has a nice easy to use API that allows you to retrieve the current coverage data.

Rather than faking things to get your build to complete, you may want to try the API instead.

I do have some perl code around to retrieve it if you're interested

NOTE: I have no ties to coveralls nor github, however I have had to work around this issue myself.
I also have [github-badge-cache-buster](https://github.com/sbts/github-badge-cache-buster) which can be added to your test or build infrastructure to attempt a forced cache PURGE on the camo.githubusercontent.com resource for your badges

sbts:
Actually, even easier, here is a `curl` example that does the job of retrieving the coverage, and uses `jq` to only return the desired value
```bash
curl -s -o /dev/stdout https://coveralls.io/github/ledgersmb/LedgerSMB.json | jq .covered_percent
```

Issue #6039, opened 2016-12-20
Include id with bulkDocs() results even in case of conflicts?


nolanlawson:
I'm not sure if this matches the CouchDB behavior or not, but [Ben Nadel lays out a pretty good case](https://www.bennadel.com/blog/3200-retrying-bulk-updates-in-pouchdb-using-a-recursive-promise-chain.htm) for why it's annoying that the response to `bulkDocs()` doesn't include the original ID for conflicted documents. You essentially have to keep track of the request as well as the response object, which is a lot of bookkeeping.

If we were to implement this, and if CouchDB didn't support it, then we'd have to do the bookkeeping ourselves, but it may be worth it. Unless folks feel this is fine being solved in plugin-land.

willholley:
I'm pretty sure CouchDB [does return id alongside errors](http://docs.couchdb.org/en/2.0.0/api/database/bulk-api.html#bulk-documents-transaction-semantics) so hopefully this is straightforward.

nolanlawson:
@willholley Yep it does, the new test confirms same behavior between Couch and Pouch modulo some weirdo subtle differences that we may want to fix (e.g. `ok: true` even for errors, `error: "conflict"` instead of `error: true`).

PR #6047, opened 2016-12-21
(#6039) - include id in all bulk_docs results

willholley:
CouchDB already includes an id field for each result in a `_bulk_docs` response. For non-http adapters, ensure we do the same.

It seems simpler to do this in the core adapter code rather than patch up each individual storage engine.

### DIFF
--- a/tests/integration/test.bulk_docs.js
+++ b/tests/integration/test.bulk_docs.js
@@ -67,6 +67,8 @@ adapters.forEach(function (adapter) {
               results[0].name.should.equal(
                 'conflict', 'First doc should be in conflict');
               should.not.exist(results[0].rev, 'no rev in conflict');
+              should.exist(results[0].id);
+              results[0].id.should.equal("0");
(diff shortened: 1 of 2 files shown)


nolanlawson:
Very nice fix! Normally I would like to put this in each individual adapter (since that's usually where such logic goes), but this is much simpler

tests/integration/test.bulk_docs.js
```diff
@@ -67,6 +67,7 @@ adapters.forEach(function (adapter) {
               results[0].name.should.equal(
                 'conflict', 'First doc should be in conflict');
               should.not.exist(results[0].rev, 'no rev in conflict');
+              should.exist(results[0].id);
               for (i = 1; i < 5; i++) {
                 results[i].id.should.equal(i.toString());
```
nolanlawson: How in the heck was this even passing before? It looks like it's checking that each result has an id and it equals the document id.

PR #6052, opened 2016-12-22
(#5930) - use travis_retry

nolanlawson:
Our maximum build time has been increased to 90 minutes, so this should help out a lot if it works. It retries each job up to 3 times.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,10 @@ before_script:
   # is a lint failure
   - npm run eslint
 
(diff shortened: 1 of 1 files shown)


nolanlawson:
Wow, looks like it passed first time. This is huge! And looking at our test times, some of them definitely restarted (e.g. took ~45 minutes to run).

daleharvey:
Will be so happy to not be doing this manually

Issue #6055, opened 2016-12-22
total_rows is not consistent between worker threads and main thread


nolanlawson:
There's an in-memory cache for `total_rows` (aka `doc_count`) in the IDB adapter. This is cute and has some perf benefits, but it also introduces inconsistencies between the worker thread and main thread because we're caching stuff in-memory, outside of the IDB transaction.

Given the increasing usage of web workers and service workers, I don't think this is a safe tradeoff to make anymore. We should just remove this optimization and look for other, safer optimizations.

nolanlawson:
Technically you can also reproduce this by just using the same PouchDB in multiple tabs, so it's not safe for that case either.

nolanlawson:
Fixed now

PR #6056, opened 2016-12-22
(#6055) - remove docCount optimization from IDB

nolanlawson:
Adds a new test to `browser.worker.js`. The test fails before the fix but succeeds after, demonstrating the issue with `docCount`.

I'll need to run some perf tests to see what kind of impact this has, but I believe that since we can now use `getAll()`-based cursors for `allDocs()`, we can offset whatever perf regression this causes. Also it's the right thing to do because it's more correct.

I will have to implement the same fix for WebSQL (due to the cross-tab issue) but I'll do that in another PR. The worker test does not run for WebSQL because no modern browser supports WebSQL in a worker.

### DIFF
--- a/packages/node_modules/pouchdb-adapter-idb/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-idb/src/index.js
@@ -17,10 +17,10 @@ import { Map } from 'pouchdb-collections';
 import idbBulkDocs from './bulkDocs';
 import idbAllDocs from './allDocs';
 import checkBlobSupport from './blobSupport';
+import countDocs from './countDocs';
 import {
   MISSING_DOC,
   REV_CONFLICT,
-  NOT_OPEN,
   IDB_ERROR,
   createError
 } from 'pouchdb-errors';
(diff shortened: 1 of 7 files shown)


packages/node_modules/pouchdb-adapter-idb/src/index.js
```diff
@@ -387,26 +387,21 @@ function init(api, opts, callback) {
   };
 
   api._info = function idb_info(callback) {
-
-    if (idb === null || !cachedDBs.has(dbName)) {
-      var error = new Error('db isn\'t open');
-      error.id = 'idbNull';
-      return callback(error);
-    }
```
nolanlawson: Setting `idb` to `null` was not necessary; we can just check `cachedDBs`.

Also, there is no need to manually check for closed state because `adapter.js` already does that.

tests/integration/test.basics.js
```diff
@@ -1062,6 +1062,33 @@ adapters.forEach(function (adapter) {
...
+
+    it('test close() after db close', function () {
+      var db = new PouchDB(dbs.name);
+      return db.close().then(function () {
+        return db.close().catch(function (err) {
+          err.message.should.equal('database is closed');
+        });
+      });
+    });
+
```
nolanlawson: I ran into a quasi-unrelated bug with `closed` so I wanted more tests to confirm it was working correctly.

tests/integration/browser.worker.js
```diff
@@ -24,18 +24,25 @@ if (typeof window.Worker === 'function' &&
...
+  var worker;
+
+  before(function () {
+    worker = new Worker('worker.js');
+    worker.postMessage(['source', sourceFile]);
+  });
+
+  after(function () {
+    worker.terminate();
+  });
```
nolanlawson: creating and terminating a new worker for every test here was really not necessary

nolanlawson:
Ran some measurements using #6067. This is a regression on the `all-docs-startkey-endkey` test in Chrome but not in Firefox:

Chrome: 631ms -> 1081ms
Firefox: 411ms -> 414ms

My guess is that Firefox is caching the count somehow under the hood (or SQLite is just very smart about this), whereas Chrome isn't. Regardless, this PR is the _correct_ behavior, and it's better to be correct than to be fast. My tests in #6060 indicate that we can still get some wins, although I don't know if we can get Chrome to be as fast as it used to be.

nolanlawson:
It's also a modest perf regression in Edge 15 (beta): 1717ms -> 1517ms

daleharvey:
Being correct > Being fast

PR #6057, opened 2016-12-22
don't merge - test in Microsoft Edge

nolanlawson:
Just testing Sauce Labs to see if this works... based on my local testing all our tests in Edge should be fine.

There were some fairly big improvements to WebDriver support in Edge 14 so now that Sauce Labs supports 14 I'm willing to give this another go.


nolanlawson:
Weird, it's failing with `the object {\n  "message": [undefined]\n  "name": "unknown"\n  "status": 500\n} was thrown, throw an Error ` but I can't reproduce locally. Guess we should just wait for Edge 15?

PR #6058, opened 2016-12-22
(#6055) - remove docCount optimization from WebSQL

nolanlawson:
Unlike #6056 I don't believe there's any way to test this, since modern browsers don't support WebSQL in a worker. Or rather, to test it we'd have to open up two browser tabs and see the result between the two, which is way too complex given our current test infrastructure, so I'm just proposing the change without tests.

### DIFF
--- a/packages/node_modules/pouchdb-adapter-websql-core/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-websql-core/src/index.js
@@ -130,7 +130,6 @@ function WebSqlPouch(opts, callback) {
   var idRequests = [];
   var encoding;
 
-  api._docCount = -1; // cache sqlite count(*) for performance
   api._name = opts.name;
 
   // extend the options here, because sqlite plugin has a ton of options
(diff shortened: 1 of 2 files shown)


packages/node_modules/pouchdb-adapter-websql-core/src/index.js
```diff
@@ -697,59 +701,58 @@ function WebSqlPouch(opts, callback) {
     }
 
     db.readTransaction(function (tx) {
+      // count the docs in parallel to other operations
+      countDocs(tx, function (docCount) {
+        totalRows = docCount;
+      });
```
nolanlawson: Incidentally in the previous version we were stupidly waiting for the `countDocs()` to callback before running the other queries. But since the result of `countDocs()` is only needed when the transaction was complete, this was a waste. We can just do it in parallel.

daleharvey:
So its not too complicated to test across tabs, we previously had tests for this to detect changes properly, its a class of tests we need to reinstate but I dont think they need to block this

nolanlawson:
Yeah I'm not too concerned about maintaining tests for WebSQL though, and currently we can test IDB cross-thread issues using the `browser.worker.js` tests. It may be advantageous to test cross-tab because of things like our `localStorage` messaging though.

Issue #6059, opened 2016-12-22
Use getAll/getAllKeys for allDocs() as well


nolanlawson:
We're already using it for `changes()` (#6033) but it should be a perf boost for `allDocs()` as well.

PR #6060, opened 2016-12-22
(#6059) - use batch cursor in allDocs()

nolanlawson:
I rebased this on #6056 to avoid bitrotting myself.

This extends our use of `getAll()`/`getAllKeys()` to `allDocs()`, in addition to the work already done to use it in `changes()`. ~~This was a bit trickier because the two have slightly different use cases, but this is documented in the code comments for `runBatchedCursor.js`.~~

~~There is an additional optimization we can do here where we avoid `getAllKeys()` entirely for `allDocs()` (it's not necessary, since unlike the `by-seq-store` for the `doc-store` we can derive the key from the value; it's not an autoincrementing key). However I'll save that for a future PR.~~

_**edit:** I have a new implementation using batched cursors or a regular `getAll` depending on the scenario._

Benchmarks are TBD but in principle this is faster with no increased memory costs since we're still keeping the batch size limited based on `limit`.

### DIFF
--- a/packages/node_modules/pouchdb-adapter-idb/src/allDocs.js
+++ b/packages/node_modules/pouchdb-adapter-idb/src/allDocs.js
@@ -1,20 +1,20 @@
 import { createError, IDB_ERROR } from 'pouchdb-errors';
 import { collectConflicts } from 'pouchdb-merge';
-
 import {
   ATTACH_STORE,
   BY_SEQ_STORE,
   DOC_STORE
 } from './constants';
-
 import {
   decodeDoc,
   decodeMetadata,
   fetchAttachmentsIfNecessary,
   postProcessAttachments,
-  openTransactionSafely
+  openTransactionSafely,
+  idbError
 } from './utils';
-
+import runBatchedCursor from './runBatchedCursor';
+import getAll from './getAll';
 import countDocs from './countDocs';
 
 function createKeyRange(start, end, inclusiveEnd, key, descending) {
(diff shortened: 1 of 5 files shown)


nolanlawson:
Interestingly this is a perf boost relative to master in Firefox, but a regression in Chrome (chrome 119ms -> 868ms, firefox 496ms -> 157ms for an `allDocs()` test I'm running). However, that's counting the removal of the `docCount` optimization as well.

nolanlawson:
Hm, based on these recent results I'm starting to suspect that `objectStore.count()` is just expensive in Chrome whereas it's cheap in Firefox (something to do with LevelDB vs SQLite?). In any case we may want to bring the docCount optimization back, but have it inside the store instead of in-memory so that it works across threads.

I am still happy with this PR but I'm sad that the docCount optimization helped so much in Chrome.

nolanlawson:
OK, in order to compare apples to apples, I compared using the new all-docs-startkey-endkey using _just_ the commits in this PR (and using #6056 as the baseline, bumped up to 50 iterations in order to be more sure):

- Chrome: 6740ms -> 7583ms
- Firefox: 2933ms -> 1092ms

So oddly it is a perf regression in Chrome. Apparently it's faster to fetch 10 documents using a cursor than to fetch them all at once using a single `getAll()`.

This is a weird result, but I still think this is a good PR. We can improve Chrome perf by doing #6068. Also even if it's a bit slower, I can see from Dev Tools timelines that it spends less time on the UI thread because it has fewer callbacks. So a different benchmark may be fairer.

nolanlawson:
Small bug in `descending=true` mode that was only being tested in the mapreduce tests... so I added integration tests for it and fixed it. Should be green now.

nolanlawson:
Collapsed commits so it's easier to read.

nolanlawson:
Rebased, will wait for a green before merging.

nolanlawson:
Green!

PR #6062, opened 2016-12-22
(#6039) - fix ids included in bulkDocs() responses

nolanlawson:
~~This reverts commit 2553fd77bf341976f205a5e7ae8352775dee7409.~~ _already reverted, added a true fix to this PR_

I don't understand what's failing exactly, but I don't have time right now to look into it. I'd suggest we just merge because otherwise everything is failing.

### DIFF
--- a/packages/node_modules/pouchdb-core/src/adapter.js
+++ b/packages/node_modules/pouchdb-core/src/adapter.js
@@ -826,13 +826,21 @@ AbstractPouchDB.prototype.bulkDocs =
     }
   }
 
-  if (!opts.new_edits && this.type() !== 'http') {
+  var adapter = this;
+  if (!opts.new_edits && adapter.type() !== 'http') {
     // ensure revisions of the same doc are sorted, so that
     // the local adapter processes them correctly (#2935)
     req.docs.sort(compareByIdThenRev);
   }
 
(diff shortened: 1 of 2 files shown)


nolanlawson:
On second thought, I found a fix, so I'll just add a new commit here.

nolanlawson:
OK, so there's a race condition where the array passed in to the adapter's `_bulkDocs` function destroys the underlying documents, so it's not safe to check `doc._id` _after_ calling `_bulkDocs`. The fix is to copy an array of `_id`s before calling `_bulkDocs`. I also added a new test to check that the `id`s are returned correctly.

nolanlawson:
I think it's important to have master green, so I'm going to make an executive decision and merge https://github.com/pouchdb/pouchdb/pull/6062/commits/095a9a28f4d5f697a55e22714cc8741966ba32b6 to master without a +1. this PR will now be the fix to re-enable `id`s on `bulkDocs` responses.

willholley:
eek! Thanks for sorting this out @nolanlawson

nolanlawson:
@willholley No prob, with our tests as flakey as they are, it's unfortunately kinda easy for race conditions to slip through. 😅

PR #6067, opened 2016-12-24
(#6067) - better all-docs perf tests

nolanlawson:
This is a better version of the `all-docs` perf tests than the previous one.

Previously, we were running 10 concurrent `allDocs()` requests using `Promise.all()`. In the new version, they are chained together and thus not concurrent.

I believe it's a more accurate test to chain these together, because that's how people would actually use these APIs in the real world. For a pagination scenario, they would not do 10 `allDocs()` all at once but would rather do them one-at-a-time. Furthermore this actually has an impact on the measurements, because if we run all tasks together then the IDB engine can batch everything concurrently under the hood and make the results appear better than they are.

### DIFF
--- a/tests/performance/perf.basics.js
+++ b/tests/performance/perf.basics.js
@@ -77,7 +77,7 @@ module.exports = function (PouchDB, opts) {
     {
       name: 'all-docs-skip-limit',
       assertions: 1,
-      iterations: 50,
+      iterations: 10,
       setup: function (db, callback) {
         var docs = [];
         for (var i = 0; i < 1000; i++) {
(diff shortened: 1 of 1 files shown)


tests/performance/perf.basics.js
```diff
@@ -77,7 +77,7 @@ module.exports = function (PouchDB, opts) {
     {
       name: 'all-docs-skip-limit',
       assertions: 1,
-      iterations: 50,
+      iterations: 10,
```
nolanlawson: reduced iterations because the new tests is necessarily slower

Issue #6068, opened 2016-12-24
in-database total_rows cache


nolanlawson:
Once https://github.com/pouchdb/pouchdb/issues/6055 is fixed, we can replace the previous in-memory optimization with a proper in-database optimization that actually works across threads.

nolanlawson:
My current thinking is to just use the meta store and roughly use the same code add before but attaching the docCount to the database instead of in-memory

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/4a1b4e4b6ac7a46b306a7df6870b9e346ba2fea6

PR #6070, opened 2016-12-24
(#6068) - safely re-implement IDB docCount optimization

nolanlawson:
This re-implements the docCount optimization for IDB, but this time doing it in-database rather than in-memory, so it actually works across tabs/workers.

I rebased this on https://github.com/pouchdb/pouchdb/pull/6060 (which itself was rebased on #6056) because I really wanted to avoid bitrot. Sorry for the confusion.

The implementation is essentially do the same thing we were doing before, but instead of caching the `docCount` in-memory, we cache it in the `META_STORE`. It's always re-calculated when the database is initialized, and then after that the cached value is used during `allDocs()`, and updated based on a delta calculation system in `bulkDocs()`.

I measured the perf gain using FF50, Chrome 55, and Edge 14 (2013 MBA, Edge running in Parallels). The test was `all-docs-startkey-endkey` from #6067, bumped up to 50 iterations for better accuracy. I measured only the difference between #6060 and this PR:

| --- | --- | --- | ---- |
| Firefox | 982ms | 1045ms | -6.44% |
| Chrome | 6132ms | 2795ms | +55.4% |
| Edge | 9486ms | 5353ms | +43.5% |

So it's a mild perf regression for Firefox, but a huge win for Chrome and Edge. And given that Firefox is already faster than Chrome, I think this is a huge improvement.

### DIFF
--- a/packages/node_modules/pouchdb-adapter-idb/src/allDocs.js
+++ b/packages/node_modules/pouchdb-adapter-idb/src/allDocs.js
@@ -3,7 +3,8 @@ import { collectConflicts } from 'pouchdb-merge';
 import {
   ATTACH_STORE,
   BY_SEQ_STORE,
-  DOC_STORE
+  DOC_STORE,
+  META_STORE
 } from './constants';
 import {
   decodeDoc,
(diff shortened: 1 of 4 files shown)


nolanlawson:
Legit failure in Android 5, Chromium 37. I will investigate

nolanlawson:
Failures are still genuine in Android. This needs work.

nolanlawson:
Green!

PR #6071, opened 2016-12-24
(#6071) - update selenium to 3.0.1

nolanlawson:
My testing suggests we can upgrade to Firefox 50.0 if we do this, but needs confirmation.


nolanlawson:
Needs #6072 to pass. Old version of Firefox doesn't work with new version of selenium apparently

PR #6073, opened 2016-12-24
fix document removal test: actually check success

jaredly:
before the `res` being checked was for insertion of the attachment, not for the removal.

### DIFF
--- a/tests/integration/test.attachments.js
+++ b/tests/integration/test.attachments.js
@@ -1840,7 +1840,8 @@ adapters.forEach(function (adapter) {
       }).then(function (res) {
         var doc = res.rows[0].doc;
         doc._attachments['foo.txt'].data.should.equal('dG90bw==');
-        return db.remove(doc);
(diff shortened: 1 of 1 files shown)


tests/integration/test.attachments.js
```diff
@@ -2587,7 +2587,7 @@ adapters.forEach(function (adapter) {
         db.putAttachment('mydoc', 'mytext', resp.rev, blob, 'text/plain',
                          function (err, res) {
```
nolanlawson: jshint error because we should be checking this res too...

nolanlawson:
I've updated the PR with my recommended changes; waiting for a green from Travis...

nolanlawson:
Weird, eslint error and I can't reproduce locally :(

daleharvey:
Merged in https://github.com/pouchdb/pouchdb/commit/7f489da68a662d601c78ab2b17a1793b8b954570

PR #6086, opened 2016-12-30
6.1.0 debug fix

jwulf:
This is a fix for the 6.1.0 package in npm.


nolanlawson:
Sorry, what is this fixing? Could you please explain?

nolanlawson:
OK I see, you have an explanation in the first commit:

> TLDR; The 2.3.2 version of debug wipes localStorage on load when built using browserify.
> 
> Extended remix: Including 6.1.0 of pouchdb in a browserified build causes debug to lose its localStorage setting on load.
> 
> Cause: The load() method of the 2.3.2 version of debug doesn't return immediately after getting localStorage. It then checks for `process.env`. This evaluates `true` in a browserified environment, causing the localStorage value to be dropped. This is then persisted to localStorage.
> 
> The 2.6.0 version returns immediately on getting the localStorage value, so the existing value is re-persisted.

I think our usual approach to this problem would be to upgrade `debug` to the latest and then do a new release. We usually do new releases at the beginning of the month.

nolanlawson:
We've already updated [our debug dep to 2.6.0](https://github.com/pouchdb/pouchdb/blob/109f3dfdf835121adce31c80133a444cd7d053c3/package.json#L40), so this will be fixed when we do our next release.

Unfortunately building and pushing to the 6.1.0 git tag doesn't actually fix the npm 6.1.0 version. We can't overwrite npm versions so we'll just have to do a 6.1.1 as usual. Closing for now, let me know if this explanation makes sense. :)

jwulf:
I figured you're well on the way to 6.2.0, and I see that the debug version in there fixes it. Looking forward to it.

Love your work Nolan!

PR #6096, opened 2017-01-02
(#6096) - clean up and reduce memory usage of mapreduce

nolanlawson:
Testing indicates that Map/Set use less memory than `{}` so we should be preferring those.

I also took one large function-within-a-function and refactored it into smaller functions one level higher.

### DIFF
--- a/packages/node_modules/pouchdb-collate/src/index.js
+++ b/packages/node_modules/pouchdb-collate/src/index.js
@@ -20,14 +20,11 @@ function collate(a, b) {
   if ((ai - bi) !== 0) {
     return ai - bi;
   }
-  if (a === null) {
-    return 0;
-  }
   switch (typeof a) {
     case 'number':
       return a - b;
     case 'boolean':
-      return a === b ? 0 : (a < b ? -1 : 1);
+      return a < b ? -1 : 1;
     case 'string':
       return stringCollate(a, b);
   }
--- a/packages/node_modules/pouchdb-for-coverage/src/index.js
+++ b/packages/node_modules/pouchdb-for-coverage/src/index.js
@@ -2,9 +2,11 @@ import PouchDB from './pouchdb';
 import ajax from 'pouchdb-ajax';
 import utils from './utils';
 import errors from './errors';
+import * as collate from 'pouchdb-collate';
 
 PouchDB.ajax = ajax;
 PouchDB.utils = utils;
 PouchDB.Errors = errors;
+PouchDB.collate = collate;
 
 export default PouchDB;
--- a/packages/node_modules/pouchdb-mapreduce/src/index.js
+++ b/packages/node_modules/pouchdb-mapreduce/src/index.js
@@ -31,7 +31,7 @@ import {
   NotFoundError,
   BuiltInError
 } from './errors';
-import { Map } from 'pouchdb-collections';
+import { Map, Set } from 'pouchdb-collections';
 import Promise from 'pouchdb-promise';
 import sum from './sum';
 
@@ -356,8 +356,8 @@ function getDocsToPersist(docId, view, docIdsToChangesAndEmits) {
   var metaDocId = '_local/doc_' + docId;
   var defaultMetaDoc = {_id: metaDocId, keys: []};
   var docData = docIdsToChangesAndEmits.get(docId);
-  var indexableKeysToKeyValues = docData.indexableKeysToKeyValues;
-  var changes = docData.changes;
+  var indexableKeysToKeyValues = docData[0];
+  var changes = docData[1];
 
   function getMetaDoc() {
     if (isGenOne(changes)) {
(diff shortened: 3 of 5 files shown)


nolanlawson:
Failing on coverage, needs investigation.

nolanlawson:
Seems `objectCollate` and boolean comparison within `collate()` is not being used... very strange outcome. I'll investigate when I have time.

nolanlawson:
This needed a bit more work because an old code path inside `pouchdb-collate` was not being hit anymore once I implemented my optimizations. So I decided we needed some new tests to make sure that the old code paths were being tested, so I added some new mapreduce tests.

Some of those code paths were still not being hit, though, and then I realized that we forgot to port over the [pouchdb-collate tests](https://github.com/pouchdb/collate/tree/e7ee03bae73e356cdc0903fde20eb4f7e06ae2fc/test) when we moved them into PouchDB core. So I did that...

Then I found some code paths that were _still_ not being hit (inside of `pouchdb-collate`), so after careful analysis I removed them.

So now this PR adds some fairly modest code simplifications to mapreduce, but mostly adds a lot of new tests. :smiley:

packages/node_modules/pouchdb-collate/src/index.js
```diff
@@ -20,14 +20,11 @@ function collate(a, b) {
   if ((ai - bi) !== 0) {
     return ai - bi;
   }
-  if (a === null) {
-    return 0;
-  }
```
nolanlawson: this code path was only being hit if you did `collate(undefined, null)` and it turned out this was pointless after I fixed `lastKey` because we always coerce `undefined` to `null` everywhere else when emitted by map/reduce (as CouchDB itself does)

packages/node_modules/pouchdb-collate/src/index.js
```diff
@@ -20,14 +20,11 @@ function collate(a, b) {
...
   switch (typeof a) {
     case 'number':
       return a - b;
     case 'boolean':
-      return a === b ? 0 : (a < b ? -1 : 1);
+      return a < b ? -1 : 1;
```
nolanlawson: this code path was impossible to hit because if two booleans are identical, then collate will return 0 above during the `a === b` test

packages/node_modules/pouchdb-for-coverage/src/index.js
```diff
@@ -2,9 +2,11 @@ import PouchDB from './pouchdb';
 import ajax from 'pouchdb-ajax';
 import utils from './utils';
 import errors from './errors';
+import * as collate from 'pouchdb-collate';
 
 PouchDB.ajax = ajax;
 PouchDB.utils = utils;
 PouchDB.Errors = errors;
+PouchDB.collate = collate;
```
nolanlawson: need to export this in order to test in `tests/unit/test.collate.js`

tests/unit/test.collate.js
```diff
@@ -0,0 +1,469 @@
+'use strict';
```
nolanlawson: this is ported over almost verbatim from `pouchdb-collate/test`

packages/node_modules/pouchdb-mapreduce/src/index.js
```diff
@@ -356,8 +356,8 @@ function getDocsToPersist(docId, view, docIdsToChangesAndEmits) {
   var metaDocId = '_local/doc_' + docId;
   var defaultMetaDoc = {_id: metaDocId, keys: []};
   var docData = docIdsToChangesAndEmits.get(docId);
-  var indexableKeysToKeyValues = docData.indexableKeysToKeyValues;
-  var changes = docData.changes;
+  var indexableKeysToKeyValues = docData[0];
+  var changes = docData[1];
```
nolanlawson: using an array here is more efficient than `{changes: ..., indexableKeysToKeyValues: ...}`. the keys don't need to be spelled out explicitly because it's all internal code

packages/node_modules/pouchdb-mapreduce/src/index.js
```diff
@@ -505,48 +504,59 @@ function updateViewInQueue(view) {
...
-      if (results.length < CHANGES_BATCH_SIZE) {
-        return;
+      currentSeq = change.seq;
+    }
+    return docIdsToChangesAndEmits;
+  }
+
+  function createIndexableKeysToKeyValues(mapResults) {
+    var indexableKeysToKeyValues = new Map();
+    var lastKey;
```
nolanlawson: it turned out I had forgotten about JavaScript variable hoisting, and this `lastKey` was non-undefined between doc comparisons, meaning that we were doing needless comparisons here. it's only necessary to uniqify the indexable key for an emission if it's 1) the same document 2) emitting the exact same value. So by moving this to a separate function, we re-initialize `lastKey` to undefined every time, and furthermore by checking `i > 0`, this is what led to those old `pouchdb-collate` code paths not being hit anymore.

Furthermore this means that from now on our indexable keys should be quite a bit simpler now; no need to add the `i` value unless we truly are emitting the exact same key for the exact same document.

PR #6099, opened 2017-01-03
(#6099) - use Map/Set in viewCleanup()

nolanlawson:
One more place where we were unsafely mapping docIds using `{}` instead of `Map`/`Set`...

I should probably add a test to do map/reduce with a doc ID like `"constructor"` but haven't gotten around to it yet.


nolanlawson:
OK, added a new test.

PR #6104, opened 2017-01-04
socket adapter issue fix

pradiprv:
Fix of socket pouch filter replication issue
[Filtered replication #6](https://github.com/nolanlawson/socket-pouch/issues/6)

### DIFF
--- a/packages/node_modules/pouchdb-core/src/adapter.js
+++ b/packages/node_modules/pouchdb-core/src/adapter.js
@@ -732,7 +732,7 @@ AbstractPouchDB.prototype.allDocs =
       ));
       return;
     }
-    if (this.type() !== 'http') {
+    if (this.type() !== 'http' && this.type() !== 'socket') {
       return allDocsKeysQuery(this, opts, callback);
     }
   }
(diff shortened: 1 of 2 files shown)


nolanlawson:
Hmmm I don't believe this is the ideal solution to this problem. I think the core issue here is that we've been special-casing the `http` adapter for years, and now we have an ecosystem of adapters that also have some http-ish behavior (socket pouch, worker pouch, eventually WebRTC pouch if I get around to writing it). We should expose a better API to those adaptesr rather than have this special-casing.

I would be in favor of simply modifying `socket-pouch` to return `http` as its type. The `type()` API is undocumented and is currently only used internally, so it shouldn't have a negative impact on the ecosystem.

OTOH I also think it's high time we audit all usages of `db.type() === 'adapter'` to see if all of them really make sense for socket/worker/webrtc (my hunch is yes... since it's usually predicated on the idea that there is some kind of expensive boundary that you don't want to cross too often – http, websockets, worker postMessage, etc.) and if so, maybe we should create a brand-new API to capture that.

PR #6114, opened 2017-01-06
(#6114) - Remove passing tests from expected failures

daleharvey:
If anyone breaks NEXT I will fix it, otherwise lets get rid of our expected failures so we can hopefully get MASTER back to perma green


nolanlawson:
+1 when green. 

Also what's the deal with CouchDB master? We should really fix that.

PR #6120, opened 2017-01-07
WIP - PouchDB 6.1.1 performance post

nolanlawson:
Planning a parallel post alongside a general one by @daleharvey on 6.1.1. I should be able to get this done by Monday so we can publish both posts at the same time.

### DIFF
--- a/docs/_posts/2017-01-04-how-we-performance-tune-pouchdb.md
+++ b/docs/_posts/2017-01-04-how-we-performance-tune-pouchdb.md
@@ -1,0 +1,141 @@
+---
+layout: post
+
+title: How we performance-tune PouchDB
+author: Nolan Lawson
+
+---
+
+In PouchDB 6.1.1, there were substantial improvements to the performance of secondary index creation, as well as overall improvements to the IndexedDB and in-memory adapters. In this post I'll talk about how I analyzed PouchDB and implemented these performance boosts.
+
+### Measure twice, cut once
+
+There's an old English idiom that says ["measure twice, cut once"](https://en.wiktionary.org/wiki/measure_twice_and_cut_once#English). Apparently it's a reference to carpentry, but I find it to be excellent advice for performance-tuning as well.  
+
+When trying to improve the performance of any piece of code, proper measurements are your most valuable asset. You can't get
+any faster if you don't know what "fast" means.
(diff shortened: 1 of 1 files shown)


daleharvey:
This is awesome, my only tiny nit would be about mentioning the indexeddb spec by version, not sure getAll technically counts as v2 of the spec but either way I find referring to web specs by version confusing and just treat them all as rolling documentation. But as said that is a tiny detail.

Even though I followed along with this work in reviews etc I found the blog post super interesting so very sure others will, would love to do / see a bit more of this.

nolanlawson:
Yeah I've heard it referred to colloquially as "v2" but it's definitely kind of a fudge since as you said it's a living document. I can remove that bit!

nolanlawson:
Well, I tested the `temp-views` benchmark and compared 6.1.0 to 6.1.1 (median of 3 runs, Windows RS1 RTM, SurfaceBook i5):

```

Edge 14 - 6.1.0
10808.759716414242ms

Chrome 55 - 6.1.0
54982.59500000001ms

Firefox 50 - 6.1.0
6425.75ms

Edge - 6.1.1
11146.67000889026ms

Chrome - 6.1.1
4031.5350000000003ms

Firefox - 6.1.1
4531.77ms
```

So Firefox and Chrome are both improved (Chrome drastically so) but Edge has regressed slightly. I ran the test a few times and confirmed there's definitely a regression in Edge.

I need to do a `git bisect` to figure out what's going on here. Either it's an Edge-specific thing, or it's that the `getAll()` improves in Chrome/Firefox are obscuring some other regression.

In any case I'd like to hold off on this post possibly until PouchDB 6.1.2. Maybe I can write something about "how we improved between 6.1.0 and 6.1.2" instead. :smiley:

nolanlawson:
I think I will revive this post when I have a chance to do more research. So far what I've uncovered is:

- native Map/Set is not the culprit; it does make things faster
- If you switch Chrome to use IDBCursor instead of getAll/getAllKeys you see a large regression as well, which is surprising
- My hunch is that it's the in-database IDBCount optimization vs the old in-memory optimization, but I haven't confirmed yet.

If the last point is true then it's understandable why Edge would regress a bit.

PR #6143, opened 2017-01-16
(#6143) - perf tests: use marky, measure median, add iterations param

nolanlawson:
A few changes I find make our perf tests more trustworthy/usable:

- use [marky](http://github.com/nolanlawson/marky) to simplify performance mark/measure usage
- report the median of `n` iterations rather than just the total
- allow user to pass in `?iterations=n` or `ITERATIONS=n` to change default iterations

This will break any dashboard tool that might be written with our current output, but I'm not sure if anyone has created one or is using it.


daleharvey:
Look like some nice changes, thanks

PR #6144, opened 2017-01-16
(#6144) - make stemming use up less memory

nolanlawson:
`temp-views` reports stemming as being a fairly high consumer of memory. This PR ensures we only do the stemming logic if we actually _need_ to stem (which we usually don't).

Before (~800KB, note large block for anonymous function inside `stem()`):

screenshot 2017-01-16 14 34 47

After (~600KB):

screenshot 2017-01-16 14 34 55


daleharvey:
Looks good, thanks

Issue #6157, opened 2017-01-20
Move debug() to separate plugin


nolanlawson:
Follow-up from https://github.com/pouchdb/pouchdb/issues/6019. We will remove this in PouchDB v.7.0.0 but for the time being it can at least be refactored as a separate module.

PR #6158, opened 2017-01-20
(#6157) - move debug() to separate module

nolanlawson:
With this change, `pouchdb-debug` is now a new module, although for the time being it's included as pouchdb-core so end-users don't need to do anything with it. In PouchDB 7.0.0 we can remove it from core.

### DIFF
--- a/packages/node_modules/pouchdb-adapter-http/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-http/src/index.js
@@ -29,9 +29,6 @@ import {
 
 import pool from './promise-pool';
 import { createError, BAD_ARG } from 'pouchdb-errors';
-import debug from 'debug';
-
-var log = debug('pouchdb:http');
 
 function readAttachmentsAsBlobOrBuffer(row) {
   var atts = row.doc && row.doc._attachments;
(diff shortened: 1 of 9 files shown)


packages/node_modules/pouchdb-utils/src/adapterFun.js
```diff
@@ -1,32 +1,30 @@
...
-        var responseArgs = [self.name, name];
-        responseArgs = responseArgs.concat(
-          err ? ['error', err] : ['success', res]
-        );
-        log.apply(null, responseArgs);
-        origCallback(err, res);
-      };
+function logApiCall(self, name, args) {
+  /* istanbul ignore if */
+  if (self.constructor.listeners('debug').length) {
```
nolanlawson: this `if` check is to avoid a lot of heavy lifting if nobody is actually listening in on the `'debug'` event.

packages/node_modules/pouchdb-adapter-http/src/index.js
```diff
@@ -169,7 +166,10 @@ function HttpPouch(opts, callback) {
   function ajax(userOpts, options, callback) {
     var reqAjax = userOpts.ajax || {};
     var reqOpts = assign(clone(ajaxOpts), reqAjax, options);
-    log(reqOpts.method + ' ' + reqOpts.url);
+    /* istanbul ignore if */
+    if (api.constructor.listeners('debug').length) {
```
nolanlawson: this `if` check is to avoid even the array allocation in the case where nobody is actually listening in on the `'debug'` event.

nolanlawson:
Technically this adds a new API surface because we now have a `'debug'` event on the `PouchDB` object... but I'm comfortable with not documenting it because I believe it will only ever be used by the `pouchdb-debug` module.

nolanlawson:
I calculated how much removing `debug` will save us. Not a lot, but a decent amount:

| | with debug | without debug |
| --- | ----- | ----- |
| min | 142290 | 136758 |
| min+gz | 45759 | 43709 |

nolanlawson:
Yeah I'm just very cautious about breaking changes without a major version;
even the PouchDB.utils stuff ended up being kind of a mess. But now that
it's a separate module it should be easy to remove in a major version.

-----
Nolan Lawson
http://nolanlawson.com

Issue #6162, opened 2017-01-21
Make polyfills extractable


nolanlawson:
I propose eventually moving to a build system that outputs two versions of the library (one with polyfills and one without) or create a `pouchdb.polyfills.js` or something similar.

Either way we will need some way to easily build that from a source containing only e.g. `Promise` and not `import Promise from 'pouchdb-promise'`.

PR #6163, opened 2017-01-21
(#6162) - make polyfills extractable

nolanlawson:
This doesn't actually have any functional changes. This simply migrates all our code to use base `Object.assign`, `Map`, `Set`, and `Promise` and then uses [rollup-plugin-inject](https://github.com/rollup/rollup-plugin-inject) to automatically transform those into the corresponding PouchDB polyfills at build time.

This puts us in a position where, if we _wanted_ to, we could create a build of PouchDB without these polyfills. We can also add polyfills later (`fetch()`?) as needed.

I inspected the browser output and confirmed that it looks exactly the same, except for some cosmetic differences.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -14,13 +14,14 @@
   },
 
   "globals": {
+    "Map": true,
+    "Set": true,
     "chrome": true,
     "Promise": true,
     "Uint8Array": true,
     "ArrayBuffer": true,
     "FileReaderSync": true,
     "sqlitePlugin": true,
-
     "emit": true,
     "PouchDB": true,
     "should": true,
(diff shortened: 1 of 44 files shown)


nolanlawson:
BTW just to show how much the polyfills actually add:

| | with polyfills | without polyfills |
| --- | ----- | ----- |
| min | 142290 | 138285 |
| min+gz | 45759 | 44548 |

I double-checked the numbers. We really don't gain much by dropping polyfills. :confused:

nolanlawson:
I think in PouchDB v7 we can consider dropping the polyfills and also shipping a `dist/pouchdb.polyfills.js` or `dist/pouchdb.with-polyfills.js`, but it is worth keeping in mind that this doesn't save us many bytes.

In any case I still like this PR because it converts our code into canonical JavaScript and gives us the flexibility to decide what we want to do with polyfills.

Issue #6020, opened 2016-12-16
Integrate pouchdb-find into PouchDB


nolanlawson:
I've put this off for way too long, so here's a quick RFC on how I plan to accomplish this.

In my mind this would be a multi-stage process:

1. Split `pouchdb-mapreduce` into two modules: `pouchdb-mapreduce` and `pouchdb-abstract-mapreduce`. This should be fairly easy now that we have the monorepo. `pouchdb-find` current relies on a fork of `pouchdb-mapreduce` that has this exact same split, but unfortunately the fork has diverged enough that I will probably have to rewrite `pouchdb-abstract-mapreduce` from scratch. But in the end, both `pouchdb-find` and `pouchdb-mapreduce` will live inside of the monorepo and both can be tested there.
2. Remove `dependentDbs` logic from `pouchdb-core` in favor of a generic hook on database deletion that plugins like `pouchdb-mapreduce` and `pouchdb-find` can register for. This will pave the way for `idb-next` and the built-in `pouchdb-find`.
3. Start publishing `pouchdb-find` with the same version numbers as the other core modules, test it using e.g. `TYPE=find npm test`

Obviously we would also need to move documentation into `pouchdb.com` and probably add a new section to the guide, and in some future major release of PouchDB we can drop `pouchdb-mapreduce` from the default set of plugins, but those aren't strictly necessary in order to get `pouchdb-find` integrated.

nolanlawson:
Changing this from an RFC to a task because I think we should just do this.

nolanlawson:
fixed in #6201

PR #6165, opened 2017-01-21
(#6020) - separate mapreduce into mapreduce & abstract-mapreduce

nolanlawson:
This is step 1 in fixing #6020, i.e. moving `pouchdb-find` into core.

This breaks `pouchdb-mapreduce` into two modules: `pouchdb-mapreduce` and `pouchdb-abstract-mapreduce`. This will make it compatible with `pouchdb-find` (which uses `pouchdb-abstract-mapreduce` under the hood) and therefore we can move it into core and allow `pouchdb-find` and `pouchdb-mapreduce` to share the same core implementation.

The idea was laid out a long time ago in https://github.com/pouchdb/mapreduce/pull/250, but basically the delegation of responsibilities is:

- `pouchdb-abstract-mapreduce`: contains core secondary index logic, including handling of `startkey`/`endkey`/etc., building the index itself, managing the `_local/mrviews` document, registering dependent dbs, etc.
- `pouchdb-mapreduce` contains all the map/reduce-specific logic, such as `eval`ing map/reduce functions.
- `pouchdb-find` contains the Mango-specific logic, such as how to build a "map-like" function based on a query definition (fields, sort, etc.).

Both `pouchdb-mapreduce` and `pouchdb-find` depend on `pouchdb-abstract-mapreduce`. There is no `eval()` in `pouchdb-abstract-mapreduce`, meaning `pouchdb-find` allows us to do secondary indexes without any arbitrary code execution.


packages/node_modules/pouchdb-mapreduce-utils/README.md
```diff
@@ -1,21 +1,20 @@
-pouchdb
```
nolanlawson: this README accidentally had a lot of typos in it. I corrected it as part of this PR.

packages/node_modules/pouchdb-mapreduce/README.md
```diff
@@ -1,19 +1,20 @@
-pouchdb-mapreduce-utils ![semver non-compliant](https://img.shields.io/badge/semver-non--compliant-red.svg)
```
nolanlawson: this readme also had a bunch of typos

packages/node_modules/pouchdb-mapreduce-utils/src/index.js
```diff
@@ -1,6 +1,11 @@
 import argsarray from 'argsarray';
 import { nextTick } from 'pouchdb-utils';
 import { Set } from 'pouchdb-collections';
+import {
+  QueryParseError,
+  NotFoundError,
+  BuiltInError
+} from './errors';
```
nolanlawson: I moved these error objects into `pouchdb-mapreduce-utils` because they are needed by both `pouchdb-mapreduce` and `pouchdb-abstract-mapreduce`.

tests/mapreduce/test.persisted.js
```diff
@@ -243,6 +243,21 @@ function tests(suiteName, dbName, dbType) {
...
+        _id: '_design/barbar',
+        views: {
+          scores: {
+            map: 1
+          }
+        }
+      };
+      return db.post(doc).then(function () {
+        return db.query('barbar/scores', {key: 'bar'});
+      }).should.be.rejected;
```
willholley: This looks like it would incorrectly pass if `db.post()` failed (e.g. if we add validation to reject invalid design docs similar to CouchDB). Should the test be more explicit about which promise is rejected?

nolanlawson: Yes, you're right. I'll update it and the other one I copypasted from.

PR #6171, opened 2017-01-23
(#xxx) - update docs re: attachment blobs

nolanlawson:


nolanlawson:
Gonna self-+1 because this is a doc change.

Issue #6170, opened 2017-01-23
Stop exposing src directory on npm


nolanlawson:
Our src is starting to contain more and more custom code that Rollup alters on the fly. I'd be more comfortable with building explicit ES module bundles on the same way we build CJS bundles.

nolanlawson:
This would also be a good time to simplify the browser/node switching in our codebase using the trick in https://nolanlawson.com/2017/01/09/how-to-write-a-javascript-package-for-both-node-and-the-browser/

PR #6172, opened 2017-01-24
(#6170) - don't expose src directory to npm

nolanlawson:
This ensures that we don't expose any `src` directories to npm, and that we build standard ES module bundles as well as CJS bundles to the `lib` directory, which are then exposed.

### Changes in this PR

- build `lib/index.es.js` for all packages, `lib/index-browser.es.js` where there's a `"browser"` field
- when we publish to npm, automatically overwrite the `"jsnext:main"` and `"browser"` fields
- remove the warning for missing `lib/index-browser.js` in `"browser"`, since we just generate this automatically when we `npm publish` now
 - remove `"src"` from package.json `"files"` so the files don't even get published

### Notes

I didn't notice build times slowing down at all due to the extra `es` module generation target, probably due to the fact that the Rollup build script is re-using the existing "bundle" object.

One slightly weird thing is that we are using `"jsnext:main":"src/index.js"` during development and then `"jsnext:main":"lib/index.es.js"` in the published package, but I couldn't think of any more elegant solution since we need Rollup to operate on the `src` files.

### Remaining TODOs

1. Remove `*-browser.js` files entirely and use the trick from [my blog post](https://nolanlawson.com/2017/01/09/how-to-write-a-javascript-package-for-both-node-and-the-browser/) instead, which will make for much cleaner code. We can still keep an empty `"browser"` field in `package.json` to signal that the package requires building two versions, though.
2. Expose `"module"` as well as `"jsnext:main"`, seems it's becoming more commonplace although I'm worried that it's too early to squat on the `"module"` field name.

### DIFF
--- a/bin/build-module.js
+++ b/bin/build-module.js
@@ -44,11 +44,6 @@ function buildModule(filepath) {
     depsToSkip = depsToSkip.concat(pouchdbPackages);
   }
 
-  if (pkg.browser && pkg.browser['./lib/index.js'] !== './lib/index-browser.js') {
-    return Promise.reject(new Error(pkg.name +
-      ' is missing a "lib/index.js" entry in the browser field'));
-  }
-
   // browser & node vs one single vanilla version
   var versions = pkg.browser ? [false, true] : [false];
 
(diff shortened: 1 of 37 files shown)


nolanlawson:
Failures are genuine, need to diagnose...

nolanlawson:
Updated so that the ["files" field](https://docs.npmjs.com/files/package.json#files) is also generated at publish time, so we don't need to manage it manually. Also fixed a small bug in `test-webpack`.

PR #6173, opened 2017-01-25
(#5572) - support IDB in Safari 10.1+

nolanlawson:
Safari 10.1 [has shipped](https://developer.apple.com/library/prerelease/content/releasenotes/General/WhatsNewInSafari/Articles/Safari_10_1.html), with full and performant support for IndexedDB. There's no reason I'm aware of that we should continue supporting WebSQL.

I think it would be prudent, though, to ensure that our tests are passing in SauceLabs. This PR may fail; I'm not sure if SauceLabs supports 10.1 yet.

WebSQL will still be tested in PhantomJS, which should be good enough until we can gracefully deprecate the WebSQL adapter entirely.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,7 +83,7 @@ env:
   # Testing in saucelabs
   - CLIENT=saucelabs:chrome COMMAND=test
   - FETCH=1 CLIENT=selenium:firefox:47.0.2 COMMAND=test
-  - CLIENT=saucelabs:safari:10.1 COMMAND=test
+  - CLIENT=saucelabs:safari:10 COMMAND=test
   - CLIENT="saucelabs:internet explorer:10:Windows 8" COMMAND=test
 
   # split up the android+iphone tests as it goes over time
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,7 +83,7 @@ env:
   # Testing in saucelabs
   - CLIENT=saucelabs:chrome COMMAND=test
   - FETCH=1 CLIENT=selenium:firefox:47.0.2 COMMAND=test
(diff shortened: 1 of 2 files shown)


nolanlawson:
Safari 10.1 is not available in Sauce Labs yet. Marking as WIP

nolanlawson:
Just checked again, still can't get a later Safari version in Sauce Labs than 10.0. :confused:

nolanlawson:
Just tested locally in Safari 10.1, and tons of tests are failing, both http and local. Looks like there's still something funky going on in Safari.

nolanlawson:
Nope, I can consistently reproduce. Looks like something to do with Blobs/ArrayBuffers is broken in Safari, because the attachment tests for both http and local fail.

nolanlawson:
re-running Travis to see if saucelabs has added support by now

nolanlawson:
Looks like SauceLabs still doesn't support Safari 10.1. 😞

janl:
Are we in contact with them? I’d be happy to run point on this.

nolanlawson:
Yeah I know Gavin on their team; he has asked for Edge features in the
past. :) I'll email him and bug him

-----
Nolan Lawson
http://nolanlawson.com

nolanlawson:
ah wait it may just be that saucelabs calls it "10" but they give you 10.1 or later. this needs a retry

nolanlawson:
we'll have to check the user agent on that test to see if it's actually 10.1+ ^

janl:
@nolanlawson like so? https://github.com/pouchdb/pouchdb/pull/6527

daleharvey:
Would love to get back to this but closing out inactive PR's (easier to focus)

PR #6178, opened 2017-01-26
Add missing space to an error message.

flimzy:
To avoid malformed error messages such as:

    _image.pngis not a valid attachment name, attachment names cannot start with '_'


flimzy:
It would probably also be a good idea to add this limitation (and any others?) to the API documentation.

nolanlawson:
Thanks for the fix! And yes, it ought to be documented somewhere. Unfortunately some things are under-documented because "that's just how CouchDB does it," so it may be duplicated in the Couch documentation. But I'd happily accept a PR to note this in our own docs. :)

Issue #3864, opened 2015-05-20
Forcing `put()` when it conflicts


prust:
I was surprised to find that Pouch & Couch throw an error when a conflict happens and that there isn't a `force` option to force the conflicted record to save. I would like to force the `put()` so that a conflict is created and replicated to all nodes, so it can be resolved later (either in order to give the user more time to deal with the conflict or in order to allow the conflict to be resolved by someone else, such as a team leader).

I've looked at #2486 and [`upsert()`](https://github.com/pouchdb/upsert), but `upsert()` is only helpful when it's practical to programmatically merge doc revisions. In many cases, (differing text fields, for instance) it is safer and saner to present the conflicts to a human and allow the human to merge the conflict.

Even if I am constantly updating the UI with changes as they are synced, there is still the situation where someone is editing a property in a text input or textarea when a remote change for that property gets synced.

In order to force the conflict, I'm planning on using the `new_edits: false` option, but [the docs say](http://pouchdb.com/api.html#batch_create):

> Normally only the replication algorithm needs to do this.

This makes me nervous. Also, I will need to make up my own `_rev` (by incrementing the number at the beginning & appending a random uuid()), which also makes me nervous.

Is it safe to use the `new_edits: false` option like this? Am I missing a use-case or a best practice or another way to solve this problem? (one of my primary motivations for switching from our custom sync engine to Pouch was Pouch's ability to sync conflicts & resolve them anywhere... needless to say, I was a bit unpleasantly surprised to discover that "immediate" conflicts are treated differently than "eventual" conflicts)

nolanlawson:
It's not safe to force-put a conflict with `new_edits: false` unless you are comfortable `get`ing the doc with `open_revs: 'all', revs: true` and modifying the `_revisions` object.

I can understand where you're coming from, though; sometimes it makes sense to just force-put a conflicting revision. Somebody ought to write a plugin to do this.

I am honestly coming up blank right now trying to find any documentation for PouchDB or CouchDB that describes how the `_revisions` object works. There are a lot of examples in [test.conflicts.js](https://github.com/pouchdb/pouchdb/blob/master/tests/integration/test.conflicts.js) though, if that helps.

nolanlawson:
By way of example, here's a doc you can force-put with `new_edits: false` (from that test file):

```js
{
          _id: 'fubar',
          _deleted: true,
          _rev: '3-a3',
[...]

PR #6187, opened 2017-01-29

simpzan:
add a `force` option to `put` operation to allow update based on
non-leaf revison and generate a new conflict revision, instead of
reporting update conflict error.
it is implemented by generating `new_edits=false` option.


simpzan:
I just followed the example here https://github.com/pouchdb/pouchdb/issues/3864 provided by @nolanlawson. 
@daleharvey please help me review, thanks!

packages/node_modules/pouchdb-core/src/adapter.js
```diff
@@ -217,10 +218,43 @@ AbstractPouchDB.prototype.put = adapterFun('put', function (doc, opts, cb) {
...
+    self.get(doc._id, {revs: true, rev: doc._rev}, function (err, obj) {
+      /* istanbul ignore if */
+      if (err) {
+        return next(err);
+      }
+      const revisions = obj._revisions;
+      const nRevNum = revisions.start + 1;
+      const newRevId = uuid(32, 16).toLowerCase();
+      revisions.start = nRevNum;
+      revisions.ids.unshift(newRevId);
```
simpzan: Sorry, I am not familiar with the codebase.
you mean it is ok to have only partial revisions history in `doc._revisions.ids` when `bulkDocs` with `new_edits=false`? 
If so, your proposed implementation is definitely better, and I will adopt it.

simpzan: @ronag I tried your method, the test case passed. It seems to be working ok. 👍  :)

packages/node_modules/pouchdb-core/src/adapter.js
```diff
@@ -220,34 +220,29 @@ AbstractPouchDB.prototype.put = adapterFun('put', function (doc, opts, cb) {
...
+  function transformForceOptionToNewEditsOption() {
+    const parts = doc._rev.split('-');
+    const oldRevId = parts[1];
+    const oldRevNum = parseInt(parts[0]);
 
-      doc._revisions = revisions;
-      doc._rev = nRevNum + '-' + newRevId;
-      opts.new_edits = false;
+    const newRevNum = oldRevNum + 1;
+    const newRevId = uuid(32, 16).toLowerCase();
```
ronag: This should probably be a hash of the document? How does pouchdb generate revisions?

simpzan: No, it is not. pouchdb generate new rev id just like this when `new_edits=true`. see here https://github.com/pouchdb/pouchdb/blob/63f039cb0c79671a0bcab7e2cf2afc7be297e23c/packages/node_modules/pouchdb-adapter-utils/src/parseDoc.js#L93

packages/node_modules/pouchdb-core/src/adapter.js
```diff
@@ -217,10 +218,38 @@ AbstractPouchDB.prototype.put = adapterFun('put', function (doc, opts, cb) {
...
+    const oldRevNum = parseInt(parts[0]);
+
+    const newRevNum = oldRevNum + 1;
+    const newRevId = uuid(32, 16).toLowerCase();
+
+    doc._revisions = {
+      start: newRevNum,
+      ids: [newRevId, oldRevId]
+    };
+    doc._rev = newRevNum + '-' + newRevId;
```
ronag: @daleharvey: Is `_rev` necessary if `_revisions` exist?

nolanlawson:
Unfortunately it looks like this branch is conflicted now...

packages/node_modules/pouchdb-core/src/adapter.js
nolanlawson: yes

simpzan:
@nolanlawson I have resolved the conflict. Please review again. Thanks.

simpzan:
@nolanlawson I just fixed coverage drop down issue. but the other 3 issues related to phantomjs, I have no idea how to resolve. Could you help me on this?

nolanlawson:
I would test it out in Safari or Epiphany Browser or Chrome with the flag `?adapters=websql` and see if the issue is a WebSQL-related issue. If it's not, then it's something to do with PhantomJS itself and unfortunately it's a bit difficult to debug, although you can try [weinre](http://muellerware.org/papers/weinre/manual.html)

simpzan:
@nolanlawson I finally tracked down the issue using PhantomJS debugger. the cause is that I use const to define variable while apparently the installed PhantomJS does not support const yet.

simpzan:
@nolanlawson All checks have passed, please merge :)

tests/integration/test.conflicts.js
```diff
@@ -116,6 +116,37 @@ adapters.forEach(function (adapter) {
...
+      // then
+      .then(function (result) {
+        rev2_ = result.rev;
+        rev2_.should.not.equal(rev3);
+        rev2_.substring(0, 2).should.equal('2-');
+        should.exist(result.ok, 'update based on nonleaf revision');
+        return db.get(docId, {conflicts: true});
+      }).then(function (doc) {
+        doc._rev.should.equal(rev3);
+        doc._conflicts.should.eql([rev2_]);
```
nolanlawson: I'd like to check the whole rev tree here to confirm that it looks the same as CouchDB... also I wonder how this would work if the `force` were done at the 3rd level rather than the 2nd. Since your `force`-to-`new_edits` conversion only takes into consideration the parent revision, I'd like to confirm that we don't mess up the rev tree for a more complicated tree.

simpzan: 1, How can I get the whole rev tree? I can't find the public api to achieve that.

2, I assume the `_rev` user provide when doing force put is valid. Conceptually the only thing need to do is create a new child node for the `_rev` node, so it doesn't matter the new child node is on the 2nd level or 3rd level. I will add a new test case to cover that.

simpzan:
@nolanlawson 
* I will update the doc.
* It is a valid concern, but I don't familiar with Couchdb, so I don't how to resolve the concern.

simpzan:
@nolanlawson I have added a new test case of force putting on parent node at second level. It works fine.

In the meanwhile, I found another issue about the `_conflicts` info of the new generated revision by force put.

for the http version, the `_conflicts` is missing, and for the local version, the `_conflicts` is same as the `_rev`. 
I think the expected `_conflicts` should never be the current revision, right?

```
  test.conflicts.js-http
the force doc { _id: 'docId',
  _rev: '2-724cc833fcbee3a84e39d11d5328ec86',
  update: 2.2,
  _revisions:
   { start: 2,
     ids:
      [ '724cc833fcbee3a84e39d11d5328ec86',
        '6d471da436b9bb7c2c36863a70261c54' ] } }
    ✓ force put ok on 1st level (80ms)
the force put doc { _id: 'docId',
  _rev: '3-f110c119c0f4da465039711bef0f8138',
  update: 3.2,
  _revisions:
   { start: 3,
     ids:
      [ 'f110c119c0f4da465039711bef0f8138',
        'a50935e9a87d508f529d2cec75c9d260',
        '6d471da436b9bb7c2c36863a70261c54' ] } }
    ✓ force put ok on 2nd level (64ms)

  test.conflicts.js-local
the force doc { update: 2.2,
  _id: 'docId',
  _rev: '2-39237b84e06f2686833dc58149e4bfa9',
  _conflicts: [ '2-39237b84e06f2686833dc58149e4bfa9' ],
  _revisions:
   { start: 2,
     ids:
      [ '39237b84e06f2686833dc58149e4bfa9',
        'ed890063f7beda5627e649a435b22169' ] } }
    ✓ force put ok on 1st level
the force put doc { update: 3.2,
  _id: 'docId',
  _rev: '3-7a14e7c638e6f2f64d760b00e76b01a3',
  _conflicts: [ '3-7a14e7c638e6f2f64d760b00e76b01a3' ],
  _revisions:
   { start: 3,
     ids:
      [ '7a14e7c638e6f2f64d760b00e76b01a3',
        '78e80fdf542cc4c250e9ef741dbee8dd',
        '445472415210017d382238f489a708e4' ] } }
    ✓ force put ok on 2nd level

```

simpzan:
@nolanlawson I find that the `_conflicts` field of conflict revision will not have the expected value, it's an existing behavior. 

Is there any other issues about this patch? Thanks!

nolanlawson:
LGTM. Thanks for following up on this!

Issue #6198, opened 2017-02-03
pouchdb-dump-cli does not support attachments


filionf:
### Issue
When exporting a database with attachments, pouchdb-dump-cli result in the following error
`this._get is not a function`
The issue can also be found here: https://github.com/nolanlawson/pouchdb-dump-cli/issues/14

### Info
- Environment: Node.js
- Platform: Windows
- Adapter: writableStream/http
- Server: CouchDB

### Reproduce
Run the following command on a database that has attachments.
`pouchdb-dump http://localhost:5984/mydb > dump.txt`

daleharvey:
pouchdb-dump-cli is not a part of pouch core, needs filed on https://github.com/pouchdb-community/pouchdb-dump-cli

filionf:
That is fine. This issue has been resolved differently since then (with the _remote flag)

Issue #6106, opened 2017-01-04
Remove special-casing of http adapter


nolanlawson:
Right now the HTTP adapter has a lot of funky custom behavior and is not truly a standalone adapter in the same way that idb/websql/leveldb are. We should audit our use of `db.type() === 'http'` and see if these can be abstracted out into a more generic API that adapters like worker-pouch/socket-pouch/webrtc-pouch could also take advantage of.

I also wonder if it's a good time to create a hook for the parsing of adapters, since the whole magical "if it starts with `http` then it's http" thing is also an instance of special-casing.

Related: #6104

nolanlawson:
https://github.com/pouchdb/pouchdb/pull/6199 demonstrates that the adapter in `pouchdb-replication-stream` also needs this feature. Perhaps for now all three adapters should just return `http` for their type... :confused:

nolanlawson:
Proposal:

- Remove `db.type()` entirely (in a major version release; let's be safe and consider it breaking)
- Add `db._remote`, which is a boolean that if `true` indicates that the adapter crosses some expensive boundary (http, web socket, worker boundary, etc.) and thus deserves some special treatment.

At this point, the three adapters in question (worker-pouch, socket-pouch, and pouchdb-replication-stream) can just add `._remote = true` and that will handle the rest.

cellog:
FYI, PouchDB is throwing a db.type deprecation warning inside index-browser.js with latest stable pouchdb

PR #6199, opened 2017-02-03
(#6198) pouchdb-dump-cli does not support attachments

filionf:


nolanlawson:
Thanks for the PR! I see what the bug is, but I'd rather not do one-off fixes for the various plugins. This is covered by https://github.com/pouchdb/pouchdb/issues/6106 which has yet to be resolved. Thanks for bringing this up!

PR #6201, opened 2017-02-05
Merge pouchdb-find into the pouchdb repo

nolanlawson:
Work in progress. A few notes:

- I've preserved all Git commit history. It doesn't match our commit format, but credit where credit is due. :smiley: In the end I plan to make one single commit on top of all previous ones, or we can do a merge.
- `TYPE=find PLUGINS=pouchdb-find npm t` runs the pouchdb-find tests. `PLUGINS=pouchdb-find` is necessary to dynamically load the plugin itself after `pouchdb.js`; it's complicated because we made it so that `pouchdb.js` can be swapped out with env variables (e.g. for the webpack tests or perf comparisons). This `PLUGINS` trick can be re-used for `pouchdb-mapreduce` when it's moved out of core.
- Build scripts changed because `pouchdb-find` needs to build its own `dist/` folder. Other packages may want to use this eventually, e.g. `pouchdb-mapreduce` once it's split off from core
- `pouchdb-find`'s version will be bumped up to the same version scheme as pouchdb. This will be a major version bump, which makes sense because there is a breaking change – the `dist/` folder file has been renamed to `pouchdb.pouchdb-find.js` for consistency in the build script.

Biggest issue right now is that I can't get a working version of CouchDB master that will actually pass the tests. I considered using `pouchdb-server` because it passes the pouchdb-find tests, but unfortunately it doesn't pass the mapreduce tests, meaning we can't get proper coverage numbers. 

I'm thinking of fixing that mapreduce issue and using pouchdb-server, because for the life of me I can't find a working version of CouchDB 2 or CouchDB master. (Hard to build on Ubuntu, klaemo/docker version fails with "no DB shards could be opened"... I dunno.)


nolanlawson:
min+gz size for `pouchdb.pouchdb-find.js` has increased slightly from 17883 to 23212. Could probably be reduced with some aggressive bundling, but I think an extra 5KB is acceptable.

garrensmith:
If you still having the CouchDB docker issue. Use this sha for the docker image `sha256:6e80fd7797fdad6c6ebc0d94f8727174b33e3fb14691c7e0d7f1357fff535535` I know it works.

nolanlawson:
Yeah I saw what you did [here[(https://github.com/nolanlawson/pouchdb-find/blob/fa4d40100107cb26b4ab053c3b4c3a3158dbd879/bin/run-couch-master-on-travis.sh) but I'd prefer a solution that doesn't require sudo since we're using the Travis container arch here.

Tests are failing, seems I will need to solve the coverage problem because otherwise it's around 99%. I'll need to find a server that can run our full test suite. I'll keep digging and I'll try that docker image.

nolanlawson:
Hm, I've tried a few different options:

```
docker run -d -p 3000:5984 --name couchdb2 klaemo/couchdb@sha256:6e80fd7797fdad6c6ebc0d94f8727174b33e3fb14691c7e0d7f1357fff535535 --with-haproxy --with-admin-party-please -n 1
docker run -d -p 3001:5984 --name couchdb3 klaemo/couchdb@sha256:6e80fd7797fdad6c6ebc0d94f8727174b33e3fb14691c7e0d7f1357fff535535 --with-haproxy --with-admin-party-please
```

But I still get a failing test with `No DB shards could be opened.` I think I'll still have to go with pouchdb-server because that's at least something I know how to fix...

nolanlawson:
I'm moving ahead with pouchdb-server but unfortunately I'm blocked by https://github.com/pouchdb/express-pouchdb/pull/424 which is blocked by https://github.com/pouchdb/pouchdb/issues/6203

garrensmith:
@nolanlawson, I've just realised you are using the wrong docker image for CouchDB. You need to specify the dev image. The one you are using requires that you configure the node first. Try this rather:

`docker run -d -p 3000:5984 --name couchdb2 klaemo/couchdb:2.0-dev --with-haproxy --with-admin-party-please -n 1`

That is what we are using in Fauxton and it should work great.

garrensmith:
@nolanlawson I've spent the morning looking at this. It seems the issue is that we are deleting and recreating the same db and that is causing an issue. I've got a PR to fix this here #6206 

I've submitted as a PR so I can run the tests (here)[https://travis-ci.org/garrensmith/pouchdb/builds/198814639] to prove it works. The tests passed consistently locally. I did also change the docker instance to this:

`docker run -p 5984:5984 klaemo/couchdb:2.0-dev --with-admin-party --with-haproxy -n 1`

nolanlawson:
Excellent, looks like your fix is working for Node! As for Firefox and coverage, we can switch to pouchdb-server for that once pouchdb-server has all its necessary fixes in. (I would prefer to test CouchDB master _somewhere_ since otherwise we run the risk of pouchdb-find becoming incompatible with Mango.)

garrensmith:
That makes sense. Once we have merged this in I will take a look at the Firefox tests again. 

All misspelling thanks to my iPhone.




On Tue, Feb 7, 2017 at 3:36 AM +0200, "Nolan Lawson" <notifications@github.com> wrote:












Excellent, looks like your fix is working for Node! As for Firefox and coverage, we can switch to pouchdb-server for that once pouchdb-server has all its necessary fixes in. (I would prefer to test CouchDB master somewhere since otherwise we run the risk of pouchdb-find becoming incompatible with Mango.)

nolanlawson:
This is very close. Apparently coverage is just failing because pouchdb-server doesn't throw `denied` events correctly.

nolanlawson:
OK actually the code path wasn't hit because of the `isCouchDB()` test. Should be fixed now...

nolanlawson:
OK, both the Firefox and the coverage tests are failing because of bugs in pouchdb-server. And we can't use CouchDB 2.0 because it doesn't have _temp views. So sadly this is blocked on fixing bugs in express-pouchdb. :confused:

garrensmith:
Once I got the express server mono repo merged I can help you with this

All misspelling thanks to my iPhone.




On Thu, Feb 16, 2017 at 7:13 PM +0200, "Nolan Lawson" <notifications@github.com> wrote:












OK, both the Firefox and the coverage tests are failing because of bugs in pouchdb-server. And we can't use CouchDB 2.0 because it doesn't have _temp views. So sadly this is blocked on fixing bugs in express-pouchdb. 😕

nolanlawson:
I can't seem to reproduce any of the failures locally. I fixed some Istanbul coverage issues that only affect PouchDB Server (namely `compact_running` and `bulkGetShim`) and got 100% coverage locally, and Firefox also passes the find tests. Hopefully this is green...

daleharvey:
woot awesome job, sorry for not being able to get anything fixed on friday, huge 👍 and I will avoid merging anything else until this gets merged, but will let you do the actual merging

nolanlawson:
Thanks! :) And yeah question for you: would you prefer a rebased merge or a true merge? This is a lot of commits (essentially the raw commit history from `pouchdb-find`), so maybe in this one case we would prefer an exception to our "always rebase" rule? Your call; I don't have a strong feeling either way.

daleharvey:
Yeh I think an exception makes the most sense here and do a plain merge

daleharvey:
btw you may need to temporarily change permissions to do that, I have had to a few times when github hasnt liked my commits

nolanlawson:
OK, will do. I'll clean this up as much as possible and then merge.

nolanlawson:
OK, I re-rebased all of `pouchdb-find` on top of master, and then added this PR as one big commit on top of that. Look ma, no merging! :smiley:

The downside is we lost some attribution for @garrensmith since he helped out with the PR but I merged all PR commits into one. Sorry about that @garrensmith, I tried to figure out how to do the rebase while still giving you credit for that one commit, but couldn't get it to work. 😕

This seemed like the least-bad solution; in any case the contributor graph has now changed a lot because all the original `pouchdb-find` commits are preserved, and you are heavily represented because of that. :)

50d6cdd297e396ec366186595301ecdeca5f8e72

PR #6204, opened 2017-02-05
(#6203) - fix conflict error

### LINKED ISSUE #6203: PouchDB throws conflict error for delete-then-update-then-undelete
This actually caused a race condition in the mapreduce tests for pouchdb-server. Took me a long time to track it down. I have a test case and a fix.

nolanlawson:
Fixes #6203.

### DIFF
--- a/packages/node_modules/pouchdb-adapter-utils/src/updateDoc.js
+++ b/packages/node_modules/pouchdb-adapter-utils/src/updateDoc.js
@@ -32,7 +32,8 @@ function updateDoc(revLimit, prev, docInfo, results,
 
   var merged = merge(prev.rev_tree, docInfo.metadata.rev_tree[0], revLimit);
 
-  var inConflict = newEdits && (((previouslyDeleted && deleted) ||
+  var inConflict = newEdits && ((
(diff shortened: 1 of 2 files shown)


packages/node_modules/pouchdb-adapter-utils/src/updateDoc.js
```diff
@@ -32,7 +32,8 @@ function updateDoc(revLimit, prev, docInfo, results,
 
   var merged = merge(prev.rev_tree, docInfo.metadata.rev_tree[0], revLimit);
 
-  var inConflict = newEdits && (((previouslyDeleted && deleted) ||
+  var inConflict = newEdits && ((
+    (previouslyDeleted && deleted && merged.conflicts !== 'new_leaf') ||
     (!previouslyDeleted && merged.conflicts !== 'new_leaf') ||
     (previouslyDeleted && !deleted && merged.conflicts === 'new_branch')));
```
nolanlawson: This logic is starting to get really hairy, and I'm not sure it covers all cases. But on the bright side, it covers everything in our unit tests.

PR #6206, opened 2017-02-06
make couchdb server pass

garrensmith:


nolanlawson:
Awesome, thank you! Let's see what Travis says...

Also if this works we will need to update the coverage tests to use CouchDB master as well. Unfortunately I believe it may fail the full test suite due to the temp-views API not working so we may still have to go with pouchdb-server for those tests.

Issue #6234, opened 2017-02-13
Function eval()s inconsistently enforce strict mode


nolanlawson:
Currently we have [three][1] [places][2] [where][3] we eval code, and in 1 of them (mapreduce) we don't enforce strict mode whereas in the other 2 we do.

We should be consistently enforcing strict mode everywhere that we eval code.

[1]: https://github.com/pouchdb/pouchdb/blob/c93fe274492fb660ea5228d68577cc9e77fe4337/packages/node_modules/pouchdb-mapreduce/src/evalFunctionWithEval.js
[2]: https://github.com/pouchdb/pouchdb/blob/c93fe274492fb660ea5228d68577cc9e77fe4337/packages/node_modules/pouchdb-core/src/evalView-browser.js
[3]: https://github.com/pouchdb/pouchdb/blob/c93fe274492fb660ea5228d68577cc9e77fe4337/packages/node_modules/pouchdb-core/src/evalFilter-browser.js

nolanlawson:
The patch here would be to 1) write a new mapreduce test in tests/mapreduce that tests what happens if the user supplies a map function that should throw in strict mode. The test should fail if the code is evaluated successfuly. Then 2) fix the eval wrapper to enforce strict mode, then 3) ensure that the test now passes. This may need to be modified in both the Node.js version and the browser version since there are two for that particular codepath (see the file tree).

nolanlawson:
Note this would technically be a breaking change, so we would need to release PouchDB 7.0.0.

PR #6247, opened 2017-02-18
(#6234) throw when map function violates strict mode

jk47:
I am not sure my test case is testing what it should be. Some guidance on how to catch the error thrown by map would be appreciated. 

Addresses #6234


tests/mapreduce/test.mapreduce.js
```diff
@@ -3697,5 +3697,15 @@ function tests(suiteName, dbName, dbType, viewType) {
...
     });
+
+    it('should throw when user supplies map func unallowed by strict mode (#6234)', function () {
+      var db = new PouchDB(dbName);
+      return db.query("(function () { globalVar = 'hello'; })()")
+      .then(function (res) {
+        should.not.exist(res);
+      }).catch(function (err) {
+        err.message.should.be.a('string');
+      });
```
nolanlawson: Technically this would need to be a second-arg error handler (i.e. `then(success, failure)`) rather than a `.catch()`, because if `should.not.exist()` throws then you end up in the `.catch()` function, which is not what we want. There are some details in [this blog post](https://pouchdb.com/2015/05/18/we-have-a-problem-with-promises.html); it's subtle.

tests/mapreduce/test.mapreduce.js
```diff
@@ -3697,5 +3697,15 @@ function tests(suiteName, dbName, dbType, viewType) {
         });
       });
     });
+
+    it('should throw when user supplies map func unallowed by strict mode (#6234)', function () {
+      var db = new PouchDB(dbName);
+      return db.query("(function () { globalVar = 'hello'; })()")
```
nolanlawson: I believe this should be a persisted query rather than a temp view (because that's the code path that actually does an `eval()`), also passing in a string here doesn't work IIRC. You may need to add it to `test.persisted.js` instead, since that's where we keep the persisted tests.

nolanlawson:
@daleharvey since this is a breaking change, once it's ready let's merge it to a branch called `BREAKING` or something, and then when we're ready to release 7.0.0 we can merge it back in. I have more breaking changes I'd like to fit in as well, and there will be less pressure if it's on a separate branch.

jk47:
I took out the 'use strict' and changed up my test case. Is this what you meant by a persisted query?

tests/mapreduce/test.persisted.js
```diff
@@ -670,6 +670,23 @@ function tests(suiteName, dbName, dbType) {
       });
     });
 
+    it('should throw when user supplies map func unallowed by strict mode (#6234)', function () {
+      var db = new PouchDB(dbName);
+      var doc = {
+        _id: '_design/barbar',
+        views: {
+          scores: {
+            map: "function() { globalVar = 'hello'; })()"
```
nolanlawson: the extra `}()` is not necessary here; you just define a function string

tests/mapreduce/test.persisted.js
```diff
@@ -670,6 +670,23 @@ function tests(suiteName, dbName, dbType) {
...
+          }
+        }
+      };
+      return db.post(doc).then(function (success, failure) {
+        should.not.exist(success);
+        should.exist(failure);
+      });
```
nolanlawson: I should have been clearer, sorry about that! What I meant is:

```js
db.query('barbar/scores').then(function () {
  throw new Error('expected an error');
}, function (err) {
  should.exist(err);
});
```

This ensures that if the promise _succeeds_, the test will fail. But if the promise _fails_, the test will succeed.

Also I believe you will have to add a non-design doc as well, because otherwise the view just won't run because there are no non-design documents.

nolanlawson:
Yup that is exactly what I meant by "persisted query" (i.e. a query defined in a design doc), but there are a few more things to make the test work correctly. :)

jk47:
Thanks for all the feedback and guidance on this! Thought I understood promises but this has been eye opening.

nolanlawson:
Great, looks like you've added a new test but the code is unchanged. :) Also I'm not sure how this test will run on CouchDB; Couch may allow strict mode, so you might have to wrap the test in an `if` to ensure it only runs against local DBs.

jk47:
I've included the code change now and moved the test to run against localdb only. I still see the test fail however when I run locally.v

jk47:
Code change added and test moved to execute only against local DB.

nolanlawson:
I'm not sure why this got closed, but we'd be happy to mark it as "work in progress" and help update it to pass the tests. :)

daleharvey:
Assuming jk47 didnt want to work on it any longer, closing due to inactivity

PR #6313, opened 2017-02-28
DO NOT MERGE - Fixes #6277

### LINKED ISSUE #6277: TypeError: Cannot use 'in' operator to search for 'startkey' in undefined(…)
_From @agozie on January 29, 2016 12:18_

I upgraded to 0.9.0 from 0.7.0. I started getting this: `TypeError: Cannot use 'in' operator to search for 'startkey' in undefined(…)` for this query :
 `db.find({
        selector: {
          tID: this.test.tID,
          due: {
            $exists: true
          },
          terms: {
            $exists: true
          }
        }
      })`
What am I doing wrong please?

_Copied from original issue: nolanlawson/pouchdb-find#153_

botverse:
This is a mirror of nolanlawson/pouchdb-find#256 and fixes #6277.

It will break when you pass an empty object to the selector. You can reproduce it like this:

```javascript
db.createIndex({
  index: {
    fields: ['field1', 'field2'],
  },
})
  .then(() => {
    console.log('this will work');
    return db.find({
      selector: { field1: 'xxx', field2: null },
      sort: [ { field1: 'asc' }, { field2: 'desc' }],
    });
  })
  .then(() => {
    console.log('this won\'t');
    return db.find({
      selector: { field1: 'xxx', field2: {} },
      sort: [ { field1: 'asc' }, { field2: 'desc' }],
    });
  });
```

[The `null` value assigned in this line](https://github.com/nolanlawson/pouchdb-find/blob/master/lib/adapters/local/find/query-planner.js#L353) is never converted to an object, so doing `'x' in null` later throws an error.

Other option is to say `var combinedOpts = {};` but we can save some runtime exiting earlier as in the PR.

### DIFF
--- a/packages/node_modules/pouchdb-find/src/adapters/local/find/query-planner.js
+++ b/packages/node_modules/pouchdb-find/src/adapters/local/find/query-planner.js
@@ -336,7 +336,7 @@ function getMultiFieldQueryOpts(selector, index) {
 
     var matcher = selector[indexField];
 
-    if (!matcher) { // fewer fields in user query than in index
+    if (!matcher || !Object.keys(matcher).length) { // fewer fields in user query than in index
       finish(i);
       break;
     } else if (i > 0) {


garrensmith:
@botverse could you write a test to prove this is working now.

botverse:
Just to clari

garrensmith:
@botverse If you want. I can help you make this fix. Start with a test case. We can then iterate on fixing it together.

nolanlawson:
closing in favor of https://github.com/pouchdb/pouchdb/pull/6358

PR #6338, opened 2017-03-10
pouchdb-find: don't emit missing fields

jlami:
This will make the local index smaller.
(But might break `'$exists': false` queries)

In our case it shaves of 100MB with 8 indices to produce a 40MB local indexedDb (30k docs).

### DIFF
--- a/packages/node_modules/pouchdb-find/src/adapters/local/abstract-mapper.js
+++ b/packages/node_modules/pouchdb-find/src/adapters/local/abstract-mapper.js
@@ -23,7 +23,7 @@ function createDeepMultiMapper(fields, emit) {
         var key = parsedField[j];
         value = value[key];
         if (!value) {
-          break;
+          return; // don't emit
         }
       }
       toEmit.push(value);


garrensmith:
@jlami this looks great. I'll try and review over the next day or two

packages/node_modules/pouchdb-find/src/adapters/local/abstract-mapper.js
```diff
@@ -23,7 +23,7 @@ function createDeepMultiMapper(fields, emit) {
         var key = parsedField[j];
         value = value[key];
         if (!value) {
```
garrensmith: Instead of `if (!value)`, could you do a check if the value is undefined or null. I can't remember if we have an exists:false test. If we don't could you add one in so we can prove it doesn't break exists: false.

jlami:
I'll see what I can do, but this mirrors the other case of `createDeepSingleMapper`, maybe they should both test for undefined/null then? Or maybe even better use something like `key in value`?

As far as the testing goes. I think the current tests only work on non-indexed queries, there is only one test for $exists that uses an index. And I believe $exists: false already doesn't work when indexed. The inmemory tests that do use $exists: false are not effected by this change. But I will see if I can add some tests for the indexed versions.

garrensmith:
@jlami you are correct. I forgot that `$exists` is done in memory. I think this PR is good to go then.
+1.

PR #6339, opened 2017-03-11
Selector support for `changes()`

willholley:
As a precursor to supporting selectors in replication, add support for filtering _changes using a selector. Note that this doesn't work in CouchDB 1.x which predates the find/selector feature.

The bulk of this involves splitting out the core selector logic from the pouchdb-find plugin into a new module, pouchdb-selector-core, that can be used in the pouchdb core without pulling in all of pouchdb-find. There are no unit tests for the new module - we rely on the integration tests
for changes() and pouchdb-find.

### DIFF
--- a/packages/node_modules/pouchdb-core/src/changes.js
+++ b/packages/node_modules/pouchdb-core/src/changes.js
@@ -12,6 +12,7 @@ import {
   collectLeaves,
   collectConflicts
 } from 'pouchdb-merge';
+import { matchesSelector } from 'pouchdb-selector-core';
 import inherits from 'inherits';
 import { EventEmitter as EE } from 'events';
 import evalFilter from './evalFilter';
@@ -98,11 +99,11 @@ function Changes(db, opts, callback) {
       } else if (self.isCancelled) {
         self.emit('cancel');
       } else {
-        self.doChanges(opts);
+        self.validateChanges(opts);
       }
     });
   } else {
-    self.doChanges(opts);
+    self.validateChanges(opts);
   }
 }
 Changes.prototype.cancel = function () {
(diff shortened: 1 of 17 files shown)


packages/node_modules/pouchdb-core/src/changes.js
```diff
@@ -169,6 +170,11 @@ Changes.prototype.doChanges = function (opts) {
     opts.filter = '_view';
   }
 
+  /* istanbul ignore if */
+  if (opts.selector && !opts.filter) {
```
garrensmith: Does a filter function take precedence over a mango selector? Is this the same in CouchDB 2.0?

garrensmith: Does this mean that a _view filter will take precedence over a mango selector? Is this how CouchDB treats it?

willholley: Good thought. In CouchDB, a selector requires passing `filter=_selector` as a query parameter as well as a `selector` field in the request body. I need to check what CouchDB does if the `filter` value is inconsistent, but we should probably have some validation on the Pouch side to prevent people specifying both.

willholley: On investigation, the CouchDB will only use the `selector` filter in the body if `filter=_selector` is specified in the query string. The implementation in this PR deviates from this (similar to our `doc_ids` filter), in that we do not require the `_selector` filter to be specified - we inject it automatically if a `selector` is provided. My inclination is to add validation that throws in the case that both `filter` and `selector` are specified, and `filter` is not `_selector`.

garrensmith: That makes sense. I agree with that.

prakashsys:
Can I use a selector approach for the pouch to couch sync and vice versa, like using live filter document?

willholley:
@prakashsys yes - specify `options.selector` when configuring the replication as documented [here](https://pouchdb.com/api.html#replication).

Issue #5717, opened 2016-09-26
support _selector filter in the replicator


willholley:
### Issue

CouchDB 2.0 adds a [new way](https://docs.couchdb.org/en/2.0.0/api/database/changes.html#selector) of filtering `_changes` based on the Mango query language. This is significantly more efficient than JavaScript based filtering (up to 10x in my tests), so it would be useful for Pouch's replicator to support it. The CouchDB replicator supports it (though it isn't yet documented) by adding a `"selector"` field to the replication doc/option. This instructs the replicator to POST to `_changes?filter=_selector` with the selector in the body.
### Info
- Environment: all
- Platform: all
- Adapter: HTTP
- Server: CouchDB/Cloudant
### Questions
1. Is it acceptable to support this for the HTTP adapter only, at least as a first pass?
2. ~In time it would be good to support this in PouchDB local adapters as well. Any thoughts as to how we might manage the dependency on PouchDB-find?~ PouchDB-find is now integrated into the monorepo so this should be more straightforward

Also a reminder that we'll need to update the checkpointing logic to account for the selector.

ronag:
Also the '_design' filter would be useful!

jefbarn:
If anyone wants to experiment with this...

```
PouchDB.plugin((PouchDB) => {
  let pouchReplicate = PouchDB.replicate

  PouchDB.replicate = (source, target, repOptions) => {

    let sourceAjax = source._ajax
    source._ajax = (ajaxOps, callback) => {
      if (ajaxOps.url.includes('_selector')) {
        ajaxOps.url = ajaxOps.url.replace('filter=_selector%2F_selector', 'filter=_selector')
        ajaxOps.method = 'POST'
        ajaxOps.body = {
          selector: repOptions.selector
        }
      }
      return sourceAjax(ajaxOps, callback)
    }
    return pouchReplicate(source, target, repOptions)
  }
})
```

then you can...

```
     replication = PouchDB.replicate(remote, local, {
        filter: '_selector',
        selector: {
          collection: {
            $in: [
              'sites',
              'structures',
              'people',
            ]
          }
        }
      })
```

burggraf:
@jefbarn can you explain exactly how to implement this as a plugin?  I'd like to try this out.

jefbarn:
[...]

PR #6340, opened 2017-03-13
(#5717) - Support selector filter for replication

willholley:
Building on the _changes support for selector, this allows a selector filter to be specified during replication.

Checkpoint behaviour has been updated to account for selectors, accepting that depending on JSON.stringify may be unreliable due to instability in the implementation.


jo:
\o/

nolanlawson:
Nice, just checked the impact on bundle size, and it's not bad at all: 46947 to 46999 min+gz. Awesome new feature too.

nolanlawson:
Ohhh I see, we already added pouchdb-selector-core in e971ff896eb69c57a190686bac4b0d192109a61c.  Looks like before that, the build size was 45518. Still not a huge change! :smiley:

PR #6344, opened 2017-03-16
Couch master test fixes

willholley:


nolanlawson:
just a quick note that I've confirmed that the CORS empty header error still exists in CouchDB 2.0.0. So this would be a good one to fix. :) https://github.com/pouchdb/pouchdb/issues/6349#issuecomment-287576023

willholley:
@nolanlawson I believe the CORS empty-header issue is fixed in CouchDB master. I deliberately pinned the CouchDB CI build against the latest stable release that has a DockerFile (2.0.0) so as to avoid chasing a moving target - perhaps this is worth revisiting (-master and -stable targets)?

nolanlawson:
Yeah I probably just had an out-of-date CouchDB. Thanks for the update!

nolanlawson:
And yes, if we can get the tests passing, let's update our Couch 2 version.

daleharvey:
Will is there any chance you could wrap your valid fixes here up into a PR, doesnt matter if it not 100% yet, will take a look at getting it green this week too

willholley:
closing this in favour of #6489 to work through the Travis failures - this is no longer about having the tests pass against Couch 2.0

PR #6346, opened 2017-03-17
(#6106) deprecate .type() in favor of ._remote

nolanlawson:
My proposed solution to #6106:

- Deprecate `.type()`. It was undocumented and always a bit strange. This PR itself only adds a warning for it, though.
- Add `._remote`, a boolean that each adapter can define which says whether they want to opt-in to certain optimizations for "remote" adapters
- Update our code as much as possible to remove references to `.type()`.

In PouchDB 7.0.0 we can just remove `.type()` entirely. Also, plugins like `socket-pouch` and `worker-pouch` can just add `this._remote = true` and get the fixes they need.

### DIFF
--- a/packages/node_modules/pouchdb-abstract-mapreduce/src/index.js
+++ b/packages/node_modules/pouchdb-abstract-mapreduce/src/index.js
@@ -1,7 +1,8 @@
 import {
   flatten,
   guardedConsole,
-  nextTick
+  nextTick,
+  isRemote
 } from 'pouchdb-utils';
 
 import {
@@ -811,13 +812,12 @@ function createAbstractMapReduce(localDocName, mapper, reducer, ddocValidator) {
   }
 
   function queryPromised(db, fun, opts) {
-    if (db.type() === 'http') {
-      return httpQuery(db, fun, opts);
-    }
-
     /* istanbul ignore next */
     if (typeof db._query === 'function') {
       return customQuery(db, fun, opts);
+    }
+    if (isRemote(db)) {
+      return httpQuery(db, fun, opts);
     }
 
     if (typeof fun !== 'string') {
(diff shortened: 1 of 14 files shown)


bin/build-modules.js
```diff
@@ -31,7 +31,13 @@ function buildPackage(pkg) {
...
 
 readDir('packages/node_modules').then(function (packages) {
-  return Promise.all(packages.map(buildPackage)).catch(function (err) {
+  var promise = Promise.resolve();
+  packages.forEach(function (pkg) {
+    promise = promise.then(function () {
+      return buildPackage(pkg);
+    });
+  });
+  return promise.catch(function (err) {
```
nolanlawson: For the life of me I can't tell why, but this parallel build process stopped working with my commit. Maybe there's a dependency issue. Maybe I was trying to be too clever. Who knows.

The build is a bit slower now, but it works! shrugmoji

packages/node_modules/pouchdb-abstract-mapreduce/src/index.js
```diff
@@ -811,14 +812,13 @@ function createAbstractMapReduce(localDocName, mapper, reducer, ddocValidator) {
...
-      return httpQuery(db, fun, opts);
-    }
-
     /* istanbul ignore next */
     if (typeof db._query === 'function') {
       return customQuery(db, fun, opts);
     }
+    if (isRemote(db)) {
+      return httpQuery(db, fun, opts);
+    }
```
nolanlawson: this ought to come after custom queries/viewCleanups because adapters can override this regardless of whether or not they're remote.

packages/node_modules/pouchdb-checkpointer/src/index.js
```diff
@@ -16,7 +16,7 @@ var LOWEST_SEQ = 0;
 function updateCheckpoint(db, id, checkpoint, session, returnValue) {
   return db.get(id).catch(function (err) {
     if (err.status === 404) {
-      if (db.type() === 'http') {
+      if (db.adapter === 'http' || db.adapter === 'https') {
```
nolanlawson: this is not really about remote vs not-remote; it's about warning the user when there's an http error in the console

nolanlawson:
Weird, somehow this refuses to build. I'll look into it:

    Error: Cannot find module 'pouchdb-core' from '/home/travis/build/pouchdb/pouchdb/packages/node_modules/pouchdb/src'

nolanlawson:
blocked by #6347

nolanlawson:
I wanted Dale to have a look at this one, but also didn't want to spam him if he's busy. Let's go ahead and merge.

Issue #6251, opened 2017-02-19
Migrate pouchdb-find documentation into core


nolanlawson:
Now that we have `pouchdb-find` itself in core (including code and tests), we also need to move over the documentation. As for http://nolanlawson.github.io/pouchdb-find/ it is probably still very useful as a demo app and worth keeping around, but the API docs themselves should be migrated.

PR #6351, opened 2017-03-19
(#6251) - import pouchdb-find API docs, update guide

nolanlawson:
This imports the `pouchdb-find` docs (mostly from the `README`) into our own API docs. I also wrote a new section in the Guide to describe how to use `pouchdb-find`.

The trickiest part I found is that we are adding a whole new querying API, with lots of confusing names like "Mango," "Cloudant Query Language," "pouchdb-find", etc. So I tried to strike a nice balance by making it clear which parts of the API correspond to pouchdb-find and which correspond to map/reduce, while also promoting pouchdb-find over map/reduce. (For instance, it comes up first in both the API docs and the guide now.)

pouchdb-find is still an external plugin, but the UMD builds (`pouchdb.find.js`, `pouchdb.find.min.js`) will be in the npm/bower package and hosted in GitHub (assuming #6347 gets merged), so I was also careful to include instructions for installing it.

### DIFF
--- a/docs/_guides/mango-queries.md
+++ b/docs/_guides/mango-queries.md
@@ -1,0 +1,330 @@
+---
+index: 13
+layout: guide
+title: Mango queries
+sidebar: guides_nav.html
+---
+
+Mango queries, also known as `pouchdb-find` or the `find()` API, are a structured query API that allows you to build _secondary indexes_ beyond the built-in `allDocs()` and `changes()` indexes. 
+
+This API is useful for answering questions like:
+
+- find all documents where the `type` is `'user'`
+- find all users whose `age` is greater than `21`
+- find all Pokémon whose `name` starts with `'pika'`
+- etc.
+
+{% include anchor.html title="Installation" hash="installation" %}
+
+The `find()` API is currently offered as a separate plugin, meaning that you must install it on top of `pouchdb.js`. Here's how to do so:
+
+### Script tags
+
+```html
+<script src="pouchdb.js"></script>
+<script src="pouchdb.find.js"></script>
+```
+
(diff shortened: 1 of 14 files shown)


nolanlawson:
BTW in the next PouchDB major version (7.0.0), I'd recommend we just remove map/reduce entirely and have both pouchdb-find and mapreduce as separately installable plugins. Now that we have a reasonably well-defined system for building plugins within the monorepo and documenting them in one place, while still allowing them to be externally installable, I think we can continue to do this for more modules that we don't strictly need in core (e.g. `pouchdb-debug`, map/reduce).

nolanlawson:
I added `[skip ci]` so this will always be yellow.

docs/_guides/mango-queries.md
```diff
@@ -0,0 +1,330 @@
...
+
+* `$eq`: equals
+* `$gt`: greater than
+* `$gte`: greater than or equal to
+* `$lt`: less than
+* `$lte`: less than or equal to
+
+There are many more options besides these, although note that not all of them can take advantage of indexes. For instance, `$regex`, `$ne`, and `$not` cannot use on-disk indexes, and must use in-memory filtering instead.
+
+The most complete documentation for selector options can be found at [Cloudant Query Language](https://docs.cloudant.com/cloudant_query.html) (which is the same thing as Mango). Any differences between this specification and PouchDB's implementation should be considered a bug; they are designed to be identical!
```
willholley: I think the [CouchDB documentation](http://docs.couchdb.org/en/2.0.0/api/database/find.html) is cleaner then Cloudant's; the Cloudant documentation also covers the `text` index type which isn't supported by CouchDB and, in my opinion, confuses things considerably.

nolanlawson: cool, I wasn't aware of this page. I'll update the docs, thanks!

Issue #6352, opened 2017-03-19
Add support for "module" as well as "jsnext:main"


nolanlawson:
I can't see any reason not to support both of these. This should just be a change to our build/publish process to add a `"module"` entry as well as a `"jsnext:main"` one.

daleharvey:
Merged in https://github.com/pouchdb/pouchdb/commit/f5bfdfa37e972d66e1b34fcf3315d518536dc1f8

PR #6353, opened 2017-03-20
(#6352) - add "module" to pkg.json when publishing

nolanlawson:
Based on https://github.com/rollup/rollup/wiki/pkg.module it seems we might as well do this. Webpack 2 and Rollup essentially consider these interchangeable.


nolanlawson:
I added `[skip ci]` so this will always be yellow.

PR #6358, opened 2017-03-21
(#6277) - fix pouchdb-find null error

nolanlawson:
Rebased/tested version of https://github.com/pouchdb/pouchdb/pull/6313

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ src_browser/
 lerna-debug.log
 tests/integration/utils-bundle.js
 *.heapsnapshot
+/pouchdb-server-install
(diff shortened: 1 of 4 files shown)


nolanlawson:
I've written a test, which I can confirm fails before the fix but succeeds after. Also it's consistent with CouchDB 2.0's output.

nolanlawson:
Tests are failing because we need to link pouchdb-server's deps back to the packages in the monorepo. Tests should pass after that.

garrensmith:
The pouchdb-find test passes. Not sure why the other fail. I'm +1 on this it look great.

garrensmith:
Wait, it seems the pouchdb-find is failing.

nolanlawson:
Yeah the "select is null" test is failing. Apparently my `npm link` fix didn't work. I need to debug.

nolanlawson:
OK that should do it. express-pouchdb needed to be linked back to the monorepo too.

garrensmith:
Nice work

All misspelling thanks to my iPhone.




On Tue, Mar 21, 2017 at 5:09 PM +0200, "Nolan Lawson" <notifications@github.com> wrote:












OK that should do it. express-pouchdb needed to be linked back to the monorepo too.

PR #6370, opened 2017-03-24
(#6370) - fix building of plugins in dev mode

nolanlawson:
Plugins like pouchdb-find and pouchdb.memory.js weren't being correctly rebuilt in dev mode, which is a bit annoying. This fixes that.

### DIFF
--- a/bin/build-pouchdb.js
+++ b/bin/build-pouchdb.js
@@ -192,12 +192,6 @@ function doBuildNode() {
     .then(buildForNode);
 }
 
-function doBuildDev() {
-  return doAll(buildForNode, buildForBrowserify)()
-    .then(doAll(buildForBrowser, buildPluginsForBrowserify, buildPouchDBNext))
-    .then(buildPluginsForBrowser);
-}
-
 function doBuildAll() {
   return rimrafMkdirp('lib', 'dist', 'lib/plugins')
     .then(doAll(buildForNode, buildForBrowserify))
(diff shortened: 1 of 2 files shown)


bin/build-pouchdb.js
```diff
@@ -192,12 +192,6 @@ function doBuildNode() {
     .then(buildForNode);
 }
 
-function doBuildDev() {
-  return doAll(buildForNode, buildForBrowserify)()
```
nolanlawson: ^ classic promise mistake; this is a function that returns a promise rather than a promise itself

bin/build-pouchdb.js
```diff
@@ -208,8 +202,6 @@ function doBuildAll() {
 function doBuild() {
   if (process.env.BUILD_NODE) { // rebuild before "npm test"
     return doBuildNode();
-  } else if (DEV_MODE) { // rebuild during "npm run dev"
-    return doBuildDev();
```
nolanlawson: I can't see any reason why we shouldn't just do the full build; during debugging we still need to debug files like `dist/pouchdb.find.js` and `dist/pouchdb.memory.js`.

bin/dev-server.js
```diff
@@ -75,7 +75,7 @@ function rebuildPerf() {
 }
 
 function watchAll() {
-  watch(['packages/node_modules/**/src/**/*.js'],
```
nolanlawson: `src` is only one level deep

PR #6372, opened 2017-03-24
Update to leveldown@1.5.1

blahah:
leveldown v1.5.1 [includes a change](https://github.com/Level/leveldown/issues/330) that allows prebuild to work when yarn is used as the package manager. This greatly speeds up installation of leveldown and therefore pouchDB.

This should also make using yarn for pouchdb's own CI tests much faster (see https://github.com/pouchdb/pouchdb/pull/5931).

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "inherits": "2.0.3",
     "level-codec": "7.0.0",
     "level-write-stream": "1.0.0",
-    "leveldown": "1.5.0",
+    "leveldown": "1.5.1",
     "levelup": "1.3.5",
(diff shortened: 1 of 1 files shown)


jure:
Tests failing, looks like it's a known issue with both 1.5.1 and 1.6.0: https://github.com/Level/leveldown/issues/194#issuecomment-269923063

nolanlawson:
Yup it's a known issue unfortunately. If someone would like to hop in and help leveldown fix that bug, then we could upgrade and get the prebuild benefits. 🙂

PR #6378, opened 2017-03-27
(#6378) - remove scope-eval dep, use our own

nolanlawson:
I'm tired of seeing warnings in our build about this dependency (something about using a non-standard CommonJS export), and we can do it smaller if we just bring it into `pouchdb-utils`. That's what this PR does.

I've checked and it seems that this technique prevents Uglify from not mangling variables, so it should still result in a small `pouchdb.min.js` but it's worth double-checking once this is green.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,6 @@
     "memdown": "1.2.4",
     "readable-stream": "1.0.33",
     "request": "2.80.0",
-    "scope-eval": "0.0.3",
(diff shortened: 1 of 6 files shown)


nolanlawson:
Double-checked to make sure there was no uglify weirdness; `pouchdb.min.js` went from 47211 gzipped to 47141, so build size is smaller 😎

PR #6398, opened 2017-04-01
Fix #6364

### LINKED ISSUE #6364: Whitespace around built-in reduce functions
PouchDB seems to only recognize the built-in reduce functions if they are exactly the strings `'_sum'`, ...

When using a tool to write a design doc, e.g. erica, it can easily happen that there's a final newline in the file which then ends up in the reduce function string.

According to the internet CouchDB used to be strict about this, but now strips whitespace.

Should PouchDB do the same? That is, recognize `' _sum\n'` as the built-in _sum function?

As for a fix, would it be enough to change this line to strip whitespace? I could probably open a PR for that. https://github.com/pouchdb/pouchdb/blob/4e0c46894e588ed414e7fbac9c822633ce2628ca/packages/node_modules/pouchdb-mapreduce/src/index.js#L49

PS: Thanks for PouchDB. It's awesome!

fehrenbach:
I'll shortly push the fix to this same PR and we can squash it then. I just want to see it actually fail.

### DIFF
--- a/packages/node_modules/pouchdb-mapreduce/src/index.js
+++ b/packages/node_modules/pouchdb-mapreduce/src/index.js
@@ -33,6 +33,18 @@ var builtInReduce = {
   }
 };
 
+function getBuiltIn(reduceFunString) {
+  if (reduceFunString.startsWith('_sum')) {
+    return builtInReduce._sum;
+  } else if (reduceFunString.startsWith('_count')) {
+    return builtInReduce._count;
+  } else if (reduceFunString.startsWith('_stats')) {
+    return builtInReduce._stats;
+  } else if (reduceFunString.startsWith('_')) {
+    throw new Error(reduceFunString + ' is not a supported reduce function.');
+  }
+}
+
 function mapper(mapFun, emit) {
   // for temp_views one can use emit(doc, emit), see #38
   if (typeof mapFun === "function" && mapFun.length === 2) {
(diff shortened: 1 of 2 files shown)


fehrenbach:
Ugh, so the test I was expecting to fail did not, because I pushed the fix too early, before Travis ran, and it checked out the HEAD of the branch instead of the commit it was meant to build.

fehrenbach:
Oh come on, startsWith is not a thing in phantomjs?!

Is there a preferred solution? Shim? Library? Locally defined helper?

packages/node_modules/pouchdb-mapreduce/src/index.js
```diff
@@ -33,6 +33,22 @@ var builtInReduce = {
...
+  }
+  if (startsWith(reduceFunString, '_sum')) {
+    return builtInReduce._sum;
+  } else if (startsWith(reduceFunString, '_count')) {
+    return builtInReduce._count;
+  } else if (startsWith(reduceFunString, '_stats')) {
+    return builtInReduce._stats;
+  } else if (startsWith(reduceFunString, '_')) {
+    throw new Error(reduceFunString + ' is not a supported reduce function.');
+  }
```
nolanlawson: Instead of `startsWith()` it's a bit more concise to do:

```js
if (/^_sum/.test(reduceFunString)) {
```

etc.

PR #6450, opened 2017-04-25
6073 fix

daleharvey:


daleharvey:
https://github.com/pouchdb/pouchdb/pull/6073 had a small mistake added in the commit, fixed it and will just give it a test run and land

Issue #6456, opened 2017-04-26
Documentation - high Unicode character does not work with CouchDB/Chrome


bmihelac:
### Issue

This is not bug in PouchDB, but it updating docs could save same headaches.

PouchDB docs recommends using '\uffff' character for prefix search. This is problematic with CouchDB in Chrome. CouchDB docs [uses](http://docs.couchdb.org/en/2.0.0/couchapp/views/collation.html) '\ufff0' character instead.

Related: #5361

### Info
- Environment: browser
- Platform: Chrome
- Adapter: http
- Server: CouchDB

### Reproduce

Create pouchdb using URL. Use query with '\uffff':

```
  db.query('myview', {
    startkey: 'items-',
    endkey: 'items-\uffff',
  }).then(console.log);
```
This would yield 0 results. Using '\ufff0' would yield correct results.

ermouth:
>  with CouchDB in Chrome
Both \ufffe and \uffff are problematic with CouchDBs having mochiweb with https://github.com/mochi/mochiweb/issues/168 issue not fixed. The issue is in fact pretty dangerous since allows DoS using a serie of innocent requests, so pre-2.0 versions of CouchDB should be patched if used in prod.

willholley:
thanks for the PR @bmihelac. @ermouth - the DoS problem sounds like something worth raising as a CouchDB issue.

richard-delorenzi:
\uffff seems very low, unicode goes up to \u10ffff. Does couchdb support these higher code points?

acailly:
Somewhat related, in case it is useful for someone, I had some docs with `_id` starting with 📢 (`\u1F4E2`) and they were not found when I put `endkey: "xxx\uffff"`

I tried with `endkey: "xxx\u10ffff"` as @richard-delorenzi  suggested, but it didn't work

It was because node does not handle automatically unicode codes with more than 4 characters, I had to put `endkey: "xxx\u{10ffff}"` and it worked

PR #6457, opened 2017-04-26
update docs with Chrome safe high Unicode character (#6456)

bmihelac:
Use '\ufff0' instead of '\uffff' which does not work in Chrome with http
adapter.


willholley:
looks great - thanks for the  PR!

willholley:
@bmihelac minor nit - can you reformat the commit message using the standard pattern for this repo: `(#6456) - update docs with Chrome safe high Unicode character`?

bmihelac:
@willholley thanks for review and advice. I have updated commit message.

willholley:
thanks @bmihelac - merged!

Issue #6458, opened 2017-04-26
Map/reduce error behaviour changes in CouchDB master


willholley:
### Issue
In CouchDB 1.x and PouchDB, if an invalid value was fed into a built-in reduce function in a view, any view requests would fail with a 500 error. This has been fixed in CouchDB master so that now view requests will return successful values where possible. For example, given the database:

```
{"id":"1","key":1,"value":1},
{"id":"2","key":2,"value":2},
{"id":"3","key":"baz","value":"baz"}
```

and a view:
```
"map": "function(doc){emit(doc.val, doc.val);}",
"reduce": "_sum"
```

A query to the view with `reduce=true` will now return:

```
HTTP/1.1 200 OK

{"rows":[
{"key":null,"value":{"error":"builtin_reduce_error","reason":"The _sum function requires that map values be numbers, arrays of numbers, or objects. Objects cannot be mixed with other data structures. Objects can be arbitrarily nested, provided that the values for all fields are themselves numbers, arrays of numbers, or objects.","caused_by":"baz"}}
]}
```

and with `reduce=true&group=true` it returns:

```
HTTP/1.1 200 OK

{"rows":[
{"key":1,"value":1},
{"key":2,"value":2},
{"key":"baz","value":{"error":"builtin_reduce_error","reason":"The _sum function requires that map values be numbers, arrays of numbers, or objects. Objects cannot be mixed with other data structures. Objects can be arbitrarily nested, provided that the values for all fields are themselves numbers, arrays of numbers, or objects.","caused_by":"baz"}}
]}
```

The PouchDB mapreduce module currently expects invalid reduces to result in a request/promise failure, and so the tests fail against CouchDB master.

It feels like PouchDB should either copy the CouchDB 2.x behaviour or add a shim to detect the error in the response and fail the function call on the client to maintain the current client behaviour. Given that mapreduce is less useful on the client now we have PouchDB-find, I'm leaning towards the latter as it's less work and does not create a breaking change.

garrensmith:
I'm ok with either.

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/e648d1fe9bea28ab8aaa740cebc4027ab9b21c34

PR #6459, opened 2017-04-26
(#6458) - consistent error handling for mapreduce

willholley:
CouchDB 2 improves mapreduce error reporting in the face of invalid
values being passed to a built-in reduce function.

Previously, a query to a view that couldn't be reduced would result
in a 500 error. In 2.x, CouchDB returns a 200 response and contains
details of the error in the row.

This patch detects the presence of this error message and converts it
into a client-side error, emulating the CouchDB 1.x behaviour.

### DIFF
--- a/packages/node_modules/pouchdb-abstract-mapreduce/src/index.js
+++ b/packages/node_modules/pouchdb-abstract-mapreduce/src/index.js
@@ -285,7 +285,19 @@ function createAbstractMapReduce(localDocName, mapper, reducer, ddocValidator) {
         method: method,
         url: '_design/' + parts[0] + '/_view/' + parts[1] + params,
         body: body
-      }).then(postprocessAttachments(opts));
+      }).then(
+      /* istanbul ignore next */
+      function (result) {
+        // fail the entire request if the result contains an error
+        var reduceError = result.rows.find(function (row) {
+          return row.value && row.value.error && row.value.error === "builtin_reduce_error";
+        });
+        if (reduceError) {
+          throw new Error(reduceError.reason);
+        }
+        return result;
+      })
+      .then(postprocessAttachments(opts));
     }
 
     // We are using a temporary view, terrible for performance, good for testing


packages/node_modules/pouchdb-abstract-mapreduce/src/index.js
```diff
@@ -285,7 +285,17 @@ function createAbstractMapReduce(localDocName, mapper, reducer, ddocValidator) {
         method: method,
         url: '_design/' + parts[0] + '/_view/' + parts[1] + params,
         body: body
-      }).then(postprocessAttachments(opts));
+      }).then(function (result) {
+        // fail the entire request if the result contains an error
+        var reduceError = result.rows.find(function (row) {
```
garrensmith: Its pretty much the same functionality here, but you could use `result.row.some` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

packages/node_modules/pouchdb-abstract-mapreduce/src/index.js
```diff
@@ -285,7 +285,17 @@ function createAbstractMapReduce(localDocName, mapper, reducer, ddocValidator) {
...
+          return row.value && row.value.error && row.value.error === "builtin_reduce_error";
+        });
+        if (reduceError) {
+          throw new Error(reduceError);
```
garrensmith: We have pouchdb-errors module now. When do we use those over a custom string?

packages/node_modules/pouchdb-abstract-mapreduce/src/index.js
willholley: I started with `some` but wanted the long-form reason to be included in the Error. Have pushed an update to make this clearer.

packages/node_modules/pouchdb-abstract-mapreduce/src/index.js
willholley: The string is an implementation detail of CouchDB (master). I think it would be appropriate to move into `pouchdb-errors` if we expected pouchdb to surface it directly or used it for a string comparison in a test but in this case we're suppressing it.

garrensmith: +1 I agree

packages/node_modules/pouchdb-abstract-mapreduce/src/index.js
garrensmith: Yes good thinking

daleharvey:
This looks great cheers, it needs a test for though (and likely coverage wont be happy without it)

daleharvey:
Although saying that coverage wont be happy anyway since we dont currently run coverage on 2.0, so will need some istanbul ignores

willholley:
Coverage should be ok now. For context, this work was triggered by https://github.com/pouchdb/pouchdb/pull/6344 where the mapreduce tests run against master (and fail).

daleharvey:
This was merged with broken tests? should never happen, even if the tests are unstable need to restart until green to merge

willholley:
@daleharvey apologies that was my bad, taking a shortcut to get a different PR passing. I'll submit a new implementation that doesn't require `Array.prototype.find` which should keep all the browsers happy.

PR #6463, opened 2017-04-27
Undefined xhr fix

### LINKED ISSUE #6126: Replication state error
### Issue
error
stacktrace

### Info
- Environment: Browser, pouchdb v6.1.0
- Platform: Chrome
- Adapter: Whatever is default on Chrome
- Server: Replicates with CouchDB server
- Setup: Using [app-pouchdb](https://github.com/PolymerElements/app-pouchdb)

### Reproduce
No idea, seems to happen randomly. It seems to be a live replication issue. It attempts to retry indefinitely with the same error. Any ideas on why it might be happening at all?

mkazlauskas:
Fixes #6126 

After a decent amount of testing with this fix I haven't encountered any errors


daleharvey:
When you say it happens randomly, can you get this to happen often? This is the only report I have seen about this so I would like to make sure about the root cause before landing defensive fixes against it

Looking at the code, it seems like any nulling out of the event handlers should prevent them from ever being called again but it may be possible that there are already events on the queue?

I think instead of setting `xhr` in both these cases we would probably want to early return like:

    if (!xhr || xhr.readyState !== 4) {
        return;
    }

because if `xhr` is null after `cleanup` is called its because we want to stop processing things, if `cleanup` hasnt been called then it sounds like something outside pouchdb is breaking it

mkazlauskas:
Nope, I can't get this to happen often, just occasionally. Repeating the workflow in my app after the error doesn't reproduce it, although there are background things which I can't really repeat, like replication.

I realize these defensive fixes are more of a workaround than actual fixes, but it's all I could come up with. Your theory about events already being in the queue sounds reasonable, but I don't know JavaScript well enough to say whether it's possible. If you're sure cleanup is the only way for it being null I suppose I'm good with your proposed early return. Just let me know if you're good to merge the early return and I'll make the change.

daleharvey:
I think doing an early return here is safe, but the `xhr=this` pattern seems worrying to me and would prefer to have good justification for it, cheers

Issue #6473, opened 2017-05-03
Build broken


daleharvey:
Phantom, ie and android are consistently failing in PR's, no changes to master which was green. its both saucelabs and phantom so seeming like a sub dep update is breaking stuff

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/9911627ba98b2de6fefd84632c1c339b19116195

PR #6474, opened 2017-05-03
(#6473) - Revert "(#6458) - consistent error handling for mapreduce"

daleharvey:
This reverts commit ecf47e256fdfa16e3146814c5de370c4c4e9541b.


daleharvey:
This is passing phantom so enough tests to confirm fix

PR #6489, opened 2017-05-10
Default to couch master in Travis

willholley:


daleharvey:
Whats the most official + usable docker images for us to be using? I want to have us running against both latest stable release and bleeding edge latest nightly commit, the bleeding edge being in allowed failures so couchdb cant break our build but we get early warnings

Should we use klaemo, couchdbdev or official couchdb docker images for those?

My plan is to switch to those images, then get them green (hopefully node already is, working on browser now) then we can get them out of allowed failures, once they are out we can look at making couch master (stable) the default

willholley:
@daleharvey my understanding is that the `klaemo` images are no longer maintained and we're waiting on ASF to have them [moved into the official CouchDB repository](https://github.com/apache/couchdb-docker/issues/20). The CouchDB CI uses [`couchdbdev/debian-8-dev`](https://hub.docker.com/r/couchdbdev/debian-8-dev/) which builds the master branch from source and uses the `dev/run` script as an entry point (so must be configured with n=1 to ensure single node / read-your-writes consistency).

The 2.0 image is [`klaemo/couchdb:2.0.0`](https://hub.docker.com/r/klaemo/couchdb/2.0.0). This performs a CouchDB install and starts in single node mode by default.

Hopefully we can get some consistency when the repo comes across to Apache. It would also be helpful to have a pre-built nightly rather than building from source each time, I think.

garrensmith:
Once we get CouchDB 2.1 out the door we should have some nice standard docker images to test against.

daleharvey:
Woiuld love to get back to this but closing out inactive PR's (easier to focus)

Issue #4176, opened 2015-08-19
EventEmitter possible memory leak with 4.0.0


natcohen:
I didn't make more tests but I just upgraded from 3.6.0 and got this message at the initialisation:

(node) warning: possible EventEmitter memory leak detected. 11 listeners added. Use emitter.setMaxListeners() to increase limit.

I don't use 11 listeners (afaik) so I'm wondering if it's a bug or a change that make my code creating 11 listeners...

FYI never had this message with 3.6.0

nolanlawson:
Yup, this is a known issue and is fixed by https://github.com/pouchdb/pouchdb/commit/691df0d6b7fac08d8a281f169c2381f6424510d6. We will release a 4.0.1 soon; in the meantime you can do `PouchDB.setMaxListeners(20)` and it will fix the warning. Sorry about that!

Issue #4141, opened 2015-08-10
getting eventEmitter memory leak when updating from 3.6.0 to 4.0.0


barbalex:
this is the error message:

    (node) warning: possible EventEmitter memory leak detected. 11 listeners added. Use emitter.setMaxListeners() to increase limit.
      EventEmitter.addListener @ events.js:142
      EventEmitter.once @ events.js:174
      prepareForDestruction @ constructor.js:44
      (anonymous function) @ constructor.js:156
      tryCode @ utils.js:14
      (anonymous function) @ utils.js:34
      (anonymous function) @ index.js:803
      Item.run @ browser.js:62
      drainQueue @ browser.js:33

here is my app.js containing the code that uses pouchdb on startup: https://github.com/barbalex/ae/blob/master/src/app.js

nolanlawson:
Yes, we are adding slightly more listeners to the `PouchDB` constructor in 4.0.0 than we used to in 3.6.0, in order to properly handle `'destroy'` events. If you are creating lots of `PouchDB` objects (as I see you are), then you will need to add:

```js
PouchDB.setMaxListeners(20);
```

I would happily accept a pull request to the [common errors page](https://github.com/pouchdb/pouchdb/edit/master/docs/errors.md) to explain this issue. :)

barbalex:
thanks!

Issue #4093, opened 2015-07-31
Eventemitter leak in changes() listener in replication (yes, again)


nolanlawson:
Apparently we are still able to get memory leaks during replication if the server throws a 503 (https://github.com/nolanlawson/local-npm/issues/39). I have no idea how that's possible, though, since I thought we fixed and tested that in d8902a18. Maybe it doesn't work if you neglect to `cancel()` after an error?

daleharvey:
So this should hopefully be fixed in https://github.com/pouchdb/pouchdb/commit/691df0d6b7fac08d8a281f169c2381f6424510d6

Issue #3848, opened 2015-05-15
max listener limit hit with live replication


prust:
I've been following PouchDB for some time but I've only recently started using it so there may be a boneheaded mistake in my code. I'm using one memdown PouchDB on the server and one websql PouchDB in Chrome, doing live 2-way replication.

I'm running into two problems that are probably related. The first is that I get unexpected "pause" events in the browser, which makes me nervous (I'm running on localhost, so there shouldn't be a connection problem), and another is that I hit the `possible EventEmitter memory leak detected. 11 listeners added` error on the server if I hit refresh 11x with an empty DB.

Here is `server.js`:

```javascript
var express = require('express');
var app = express();
var PouchDB = require('pouchdb');
var expressPouch = require('express-pouchdb');
var memdown = require('memdown');

var InMemPouchDB = PouchDB.defaults({db: memdown});
 
app.use('/db', expressPouch(InMemPouchDB, {'mode': 'minimumForPouchDB'}));

app.use(express.static('./'));

var server = app.listen(8083, function() {
  var host = server.address().address;
  var port = server.address().port;
  console.log('Example app listening at http://%s:%s', host, port);
});

new InMemPouchDB('db_csnw');
```

Here is `index.html`:

```html
<!DOCTYPE html>
<script src="pouchdb.js"></script>
<script>
document.addEventListener("DOMContentLoaded", function() {
  var local_db = new PouchDB("db_csnw", {adapter: "websql"});
  var remote_db = new PouchDB("http://localhost:8083/db/db_csnw");
  var sync = local_db.sync(remote_db, {
    live: true
  }).on("change", function (change) {
    console.log("something changed");
  }).on("paused", function (info) {
    console.log("replication was paused, usually because of a lost connection");
  }).on("active", function (info) {
    console.log("replication was resumed");
  }).on("error", function (err) {
    console.error(err);
  });
  window.addEventListener("beforeunload", function() {
    sync.cancel();
  });
});
</script>
```

If I run this with node w/ the latest versions of express (`4.12.3`), express-pouch (`0.14.0`), memdown (`1.0.0`) and pouch (`master`).

I thought that maybe #3832 or #3825 would fix this, so I upgraded client and server from `3.5.0` to `master`, but I'm still getting the same behavior (though I think I'm running into slightly fewer "pause" events).

Any idea what I can do to debug this further? (if it helps, I'm running Win7, Chrome 42 and Node 0.10.32)

nolanlawson:
[...]

PR #6504, opened 2017-05-16
pouchdb-adapter-idb: fix "possible EventEmitter memory leak detected"

adrienverge:
### Problem

With the idb adapter (default in Chrome), calling `.changes()` or `.replicate()` with `{live: true}` multiple times results in this error, even when the handler is nicely terminated with `.cancel()`.

    (node) warning: possible EventEmitter memory leak detected. 11 listeners added. Use emitter.setMaxListeners() to increase limit.

Related to #4176, #4141, #4093, #3848, #2667, #2311, #2205, and maybe others.

### How to reproduce the bug?

Run this plunkr: http://plnkr.co/edit/JKCbQ6vwlYClAp6ZSAf4?p=preview

Alternatively, you can create + cancel a `.changes()` feed a few times:

```js
changesHandler = db.changes({live: true});
changesHandler.cancel();
repeat
```

### About this fix

The bug comes from [the `_change()` API call from the idb adapter](https://github.com/pouchdb/pouchdb/blob/49e4945/packages/node_modules/pouchdb-adapter-idb/src/index.js#L420-L422) returning `undefined` instead of the `{cancel: function() {...}}` object [it should return](https://github.com/pouchdb/pouchdb/blob/49e4945/packages/node_modules/pouchdb-adapter-idb/src/changes.js#L29-L33). Because of that, the cleanup function in the core module [doesn't call the adapter's `cancel()` function](https://github.com/pouchdb/pouchdb/blob/49e4945/packages/node_modules/pouchdb-core/src/changes.js#L201-L209), letting a listener into the wild.

This commit fixes that.

### DIFF
--- a/packages/node_modules/pouchdb-adapter-idb/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-idb/src/index.js
@@ -418,7 +418,7 @@ function init(api, opts, callback) {
   };
 
   api._changes = function idbChanges(opts) {
-    changes(opts, api, dbName, idb);
+    return changes(opts, api, dbName, idb);
   };
 
   api._close = function (callback) {
--- a/tests/integration/test.changes.js
+++ b/tests/integration/test.changes.js
@@ -88,6 +88,60 @@ adapters.forEach(function (adapter) {
       });
     });
 
+    it('Live changes should clean listener when cancelled', function () {
+      var db = new PouchDB(dbs.name);
+
+      // TODO: The bug was fixed for the 'idb' adapter in
+      // https://github.com/pouchdb/pouchdb/pull/6504, but still happens with
+      // the 'websql' adapter. It needs to be fixed!
+      if (db.adapter === 'websql') {
+        return;
+      }
+
+      return new testUtils.Promise(function (resolve, reject) {
+        // Capture logs
+        var logs = [];
+        var oldLog = console.error;
+        console.error = function () {
+          var args = Array.prototype.slice.call(arguments);
+          logs.push(args);
+          oldLog.apply(console, arguments);
+        };
+
+        // Try to trigger the problem
+        var changes;
+        var i = 0;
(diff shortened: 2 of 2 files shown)


daleharvey:
Sorry for the delay

So this looks great and harmless, however definitely needs a test to verify the bug and ensure we do not regress it. Do you think given your STR you can produce an automated test (hopefully there is one you can base it off in https://github.com/pouchdb/pouchdb/master/tests/integration/test.changes.js)

adrienverge:
Hi @daleharvey,

I added a test to check for the error when suscribing to + cancelling changes multiples times. This new test fails without my fix (see picture), and passes with it.

error


PS: The build "errored" on Travis, although I restarted it a few times. It does not seem related to my PR.

daleharvey:
That is an extremely comprehensive test, thanks, I am going to do a full run @ https://github.com/pouchdb/pouchdb/pull/6532 (some our tests (saucelabs) dont run for contributors sadly, trying to get that fixed)

I am generally extremely wary of setTimeout, do you think replacing the settimeout by using `changes.on("complete"` to trigger the renewChangesListener would work?

(can read https://pouchdb.com/2014/12/11/the-pains-of-being-async-at-heart.html for why I hate setTimeout, its generally a large cause of intermittent failures in our tests)

adrienverge:
> do you think replacing the settimeout by using `changes.on("complete"` to trigger the renewChangesListener would work?

Thanks for this good idea, I just tested it and it works! Moreover, with the change you propose, the other `setTimeout(…, 50)` can be replaced with a `setTimeout(…, 0)`. (Again, I've tested that the test fails without my fix and passes with it.)

I've just updated the pull request accordingly.

daleharvey:
Awesome thanks, will do a full run again @ https://github.com/pouchdb/pouchdb/pull/6542 but code looks great

PR #6508, opened 2017-05-18
Check if replication is defined

### LINKED ISSUE #6512: Cannot use pouchdb-replication with React SSR
### Issue
Currently when using with https://github.com/zeit/next.js/ (server rendered react) 

I get this error:
```sh
TypeError: Cannot redefine property: replicate
    at Function.defineProperty (<anonymous>)
    at replication (../node_modules/pouchdb-replication/lib/index.js:979:10)
```

I've made a fix in this PR: #6508

### Info
- Environment: Node.js & Browser
- Platform: Chrome/V8
- Adapter: WebSQL
- Server: PouchDB Server

### Reproduce
use https://github.com/zeit/next.js/
with https://pouchdb.com/guides/replication.html

albinekb:
Currently when using with https://github.com/zeit/next.js/ (server rendered react) 

I get this error:
```sh
TypeError: Cannot redefine property: replicate
    at Function.defineProperty (<anonymous>)
    at replication (../node_modules/pouchdb-replication/lib/index.js:979:10)
```
I think this is the correct way to fix that? 🤔 

Fixes #6512

### DIFF
--- a/packages/node_modules/pouchdb-replication/src/index.js
+++ b/packages/node_modules/pouchdb-replication/src/index.js
@@ -5,19 +5,21 @@ function replication(PouchDB) {
   PouchDB.replicate = replicate;
   PouchDB.sync = sync;
 
-  Object.defineProperty(PouchDB.prototype, 'replicate', {
-    get: function () {
-      var self = this;
-      return {
-        from: function (other, opts, callback) {
-          return self.constructor.replicate(other, self, opts, callback);
-        },
-        to: function (other, opts, callback) {
-          return self.constructor.replicate(self, other, opts, callback);
-        }
-      };
-    }
-  });
+  if (!PouchDB.prototype.replicate) {
+    Object.defineProperty(PouchDB.prototype, 'replicate', {
+      get: function () {
+        var self = this;
+        return {
+          from: function (other, opts, callback) {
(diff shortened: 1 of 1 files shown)


albinekb:
@daleharvey maybe? 🤔

packages/node_modules/pouchdb-replication/src/index.js
```diff
@@ -5,23 +5,25 @@ function replication(PouchDB) {
...
-        from: function (other, opts, callback) {
-          return self.constructor.replicate(other, self, opts, callback);
-        },
-        to: function (other, opts, callback) {
-          return self.constructor.replicate(self, other, opts, callback);
-        }
-      };
-    }
-  });
+  if (!PouchDB.prototype.replicate) {
```
albinekb: Or do i need to check with  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyDescriptor ?

daleharvey:
Hrm, is this because you are trying to include both `pouchdb` and then include `pouchdb-replication` as a plugin? any reason for `.replicate` getting redefined that cant be avoided by not redefining it?

albinekb:
Don't know, didn't have time to look into it more.

PR #6517, opened 2017-05-19
Angular Branding Guidelines

evansdianga:
The Angular team released naming conventions regarding the various versions of Angular.
As this part of the docs are targeted to 1.x versions, the convention is to use AngularJS and not refer to it as Angular.
This fix helps avoid confusion between the two libraries.

http://angularjs.blogspot.co.ke/2016/12/ok-let-me-explain-its-going-to-be.html#Naming_guidelines_48
http://angularjs.blogspot.co.ke/2017/01/branding-guidelines-for-angular-and.html


karlwestin:
Cheers!

Issue #6518, opened 2017-05-20
Slow createIndex when index already exists


natew:
### Issue
I noticed that creating indexes was running slowly. I added console.time and saw 50-80ms per database I was adding, even when it was already created!

I "fixed it" by just checking myself using getIndexes like this:

```js
      const { indexes } = await this.collection.pouch.getIndexes()
      const alreadyIndexedFields = flatten(indexes.map(i => i.def.fields)).map(
        field => Object.keys(field)[0]
      )
      // if we have not indexed every field
      if (
        intersection(this.settings.index, alreadyIndexedFields).length !==
        this.settings.index.length
      ) {
        await this.collection.pouch.createIndex({ fields: this.settings.index })
      }
```

After this change it went from 50 => 2-3ms 🎉 

### Info
- Environment: Chrome 58
- Adapter: IndexedDB

natew:
@willholley looks like you may have added some testing... probably worth keeping open.

daleharvey:
Code landed so closing out, cheer

PR #6519, opened 2017-05-22
(#6518) - improve createIndex() performance for named indexes

willholley:
createIndex() uses the md5 hash of the index definition to test for duplcates. There's no straightforward way to avoid this, but users can specify their own index/ddoc name. In this case, we should skip the md5 calculation to improve performance.


willholley:
cc @garrensmith

PR #6520, opened 2017-05-22
Update 2015-05-18-we-have-a-problem-with-promises.md

theartofme:
Added another option for passing promise results through promise chains.

### DIFF
--- a/docs/_posts/2015-05-18-we-have-a-problem-with-promises.md
+++ b/docs/_posts/2015-05-18-we-have-a-problem-with-promises.md
@@ -502,19 +502,29 @@ putYourRightFootIn()
 
 That's what promises are all about.
 
-Another option is to return nested data from the promises, either as an object or an array:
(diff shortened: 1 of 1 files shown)


docs/_posts/2015-05-18-we-have-a-problem-with-promises.md
```diff
@@ -502,6 +502,22 @@ putYourRightFootIn()
...
+getUserByName('nolan').then(function (result) {
+  var user = result;
+  return {
+    user: user,
+    userAccount: getUserAccountById(user.id)
+  };
+}).then(function (userData) {
+  var user = userData.user;
+  var userAccount = userData.userAccount;
+  // okay, I have both the "user" and the "userAccount"
```
daleharvey: This doesnt seem like it would be true, userData.userAccount will be a promise (and not the value) at this point right?

daleharvey:
So there are various ways to do this but the point of the blog post isnt to be exhaustive (Promise.all is mentioned in the post but not this section). Thanks for sending us a PR but particularly for the blog posts I think we are more looking for corrections than additions. Cheers

Issue #6510, opened 2017-05-18
The _changes feed being timed out due to lack of data causes the backoff function to be called


mjosephd:
### Issue
The issue is that the "_changes" feed being timed out due to lack of any new changes in the longpoll request the backoff function is getting called. As far as I know I only believed that hte backoff function gets called when an error happens to the "_changes" feed call. However timeout seems to also call the backoff function. Additionally a custom backoff function was used for testing and it was also called when the "_changes" call was timed out then as well.

The use case that this affects is the following
If a user leaves a device on for a couple of hours and for each "_changes" call there are no new changes the backoff function will get repeatedly called. And after some hours the when there is an actual change the user will need to wait upto 10 mins (I believe the defualt backoff function has a max limit of 10 minutes) to get the change, whereas the change should come instantly as soon as possible when it is available.

### Info
- Environment: Browser, but we are developing for a hybrid application
- Platform: Chrome for testing this issue, but again since hybrid will use the Chrome/Safari browser webview
- Adapter: default so far, Indexeddb
- Server: Couchdb for testing if the issue exists (however we noticed it on couchbase sync gateway originally which is what the app we are developing is using).

### Reproduce
1) Checkout todo sample app
2) Change CDN version to 6.2.0
3) Test change feed calling backoff function

![testing](https://cloud.githubusercontent.com/assets/422345/26193392/d668100a-3bd2-11e7-81f7-17edd96a4f6f.gif)

mjosephd:
@willholley just to clarify, this means I should consider that this functionality as an issue correct?
And the immediate remedy would be to put a custom backoff function?

mjosephd:
@willholley the current backoff function uses a random value. I remember reading somewhere it was said it was to avoid two requests being sent at the same time or something like that (couldn't find source). Would there be any implications of using a custom backoff function which returned 10 seconds as a hardcoded value (without any randomness selection logic and such)?

willholley:
@mjosephd only that PouchDB would continue to retry in the case of errors - I think that's an acceptable workaround. I'll try and look at why this is happening in this week.

mjosephd:
Also I forgot to test offline scenarios, what is the expected outcome of pouchdb being unable to do the "_changes" call, is it to call the backoff function or treat it as normal and not call the backoff function then.

willholley:
@mjosephd I believe it would call the backoff function if `retry: true` is specified.

willholley:
[...]

PR #6521, opened 2017-05-22
(#6510) - respect ajax timeout when fetching _changes

willholley:
PouchDB already has logic to create a buffer between the server side
timeout and the client timeout, providing a 5 second grace period
for the server to respond before the client will timeout. However, there
are some situations where this causes _changes to fail with a client
timeout. Specifically:

1. If the timeout is less than 5 seconds. Given the logic to add a 5
second grace period, we need the client request timeout to be at least
this.
2. If the heartbeat is specified that is greater than the client timeout.
If a heartbeat parameter is specified, Couch ignores any timeout parameter.

To handle (1), this commit adjusts the client side timeout to be at least 5
seconds. To handle (2), we make the client side timeout at least the heartbeat
value.


daleharvey:
These changes mostly look like good ideas to me but need the tests looked at, cheers

willholley:
@daleharvey apologies it took a while - coverage / tests now good.

PR #6527, opened 2017-05-25
feat: detect safari 10.1+ in IDB adapter

janl:
PR for https://github.com/pouchdb/pouchdb/pull/6173

### DIFF
--- a/packages/node_modules/pouchdb-adapter-idb/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-idb/src/index.js
@@ -807,6 +807,15 @@ IdbPouch.valid = function () {
   var hasFetch = typeof fetch === 'function' &&
     fetch.toString().indexOf('[native code') !== -1;
 
+  // Safari > 10.1 has a good IDB implementation
+  // this check is valid until Safari version 19.99
+  // at which point I hope this is all refactoreed anyway
+  var isSafariWithGoodIDB = isSafari && (
+    /Version\/10.[1-9]{1,2}/.test(navigator.userAgent)
+    ||
+    /Version\/1[1-9].[0-9]{1,2}/.test(navigator.userAgent)
+  )
+
   // Safari <10.1 does not meet our requirements for IDB support (#5572)
   // since Safari 10.1 shipped with fetch, we can use that to detect it
   return (!isSafari || hasFetch) &&


packages/node_modules/pouchdb-adapter-idb/src/index.js
```diff
@@ -807,6 +807,15 @@ IdbPouch.valid = function () {
   var hasFetch = typeof fetch === 'function' &&
     fetch.toString().indexOf('[native code') !== -1;
 
+  // Safari > 10.1 has a good IDB implementation
+  // this check is valid until Safari version 19.99
```
mgol: 19.99? It only checks for some 10.x & 11.x versions.

packages/node_modules/pouchdb-adapter-idb/src/index.js
```diff
@@ -807,6 +807,15 @@ IdbPouch.valid = function () {
...
+  // at which point I hope this is all refactoreed anyway
+  var isSafariWithGoodIDB = isSafari && (
+    /Version\/10.[1-9]{1,2}/.test(navigator.userAgent)
+    ||
+    /Version\/11.[0-9]{1,2}/.test(navigator.userAgent)
```
mgol: Shouldn't it be `/Version\/1[1-9].[0-9]{1,2}/`?

janl: you’re right, good catch :)

fixing in next push

nolanlawson:
I feel my original PR (https://github.com/pouchdb/pouchdb/pull/6173) may be doing a better job of feature detection… Instead of UA-sniffing, I'm sniffing for `fetch()` which happens to coincide with the release of a decent IDB engine in Safari. This will also work for browsers based on WebKit that for whatever reason might not match the UA string. Feels safer to me.

janl:
Oh, then I misunderstood the issue you were raising.

janl:
@nolanlawson I thought you were asking for an addition to your PR about the version number (just because the build status is still red on the PR). If that’s wrong, what are the next steps here?

nolanlawson:
@janl I just looked at the build output for the most recent job, and the Safari UA is `10.0.1`. So unfortunately it seems SauceLabs has not updated Safari yet.

nolanlawson:
I'd prefer to wait for https://github.com/pouchdb/pouchdb/pull/6173 to pass although at this point we've waited so long for SauceLabs to add Safari 10.3+ support that maybe we should just test manually and then cross our fingers.

mgol:
@nolanlawson BrowserStack has Safari 10.1 & iOS 10.3 if that changes anything for you.

Issue #5490, opened 2016-07-21
Feature request: add _id to 404/409 response


ssured:
### Environment (Node.js/browser/hybrid app/etc.)

All
### Browser/platform (Chrome/FF/Safari/Edge/iOS/Android/etc.)

All
### Adapter (IndexedDB/WebSQL/LevelDB/etc.)

All
### Server (CouchDB/Cloudant/Couchbase/PouchDB Server/etc.)

All
### Please describe your issue. Test cases (e.g. JSBin) are very helpful!

For debugging & logging purposes it would be nice to reference the `_id` of documents in various `???PouchError` objects. Errors like `CustomPouchError {status: 404, name: "not_found", message: "missing", error: true, reason: "missing"}` are sometimes helpful, but I often have the need to know which document it was. Same holds for a 409 conflict.

I know I could fix this by adding more granular error checking, but when quickly coding I am often lazy and only add a `.catch(=>)` at the end of a promise chain.

ssured:
BTW this holds less for the HTTP adapter, where you can always use the network tab of your browser. For the local adapters this comes in quite handy, as it is hard to traceback what happens in your indexeddb/websql/localstorage.

daleharvey:
There are some situations where we add information like this to error response, I think easing debugging would be a good thing here

danielo515:
In an async world this is definitively a help.

ZeroNine00:
Have someone a workaroubd for this ?  I need the info for 404 to see which request was failed

daleharvey:
For a workaround, you can pass the information around when you call the `get`

```js
db.get(id).then(gotDoc).catch(notFound.bind(id));
```

ZeroNine00:
Sorry can you  explain your workaround a little bit more ? 
At the moment I have only Callbacks ... 

for (var i = 0; i < aEntries.length; i++) {
// Here I need a check against pouchdb
// Somethine like this ? 	thatDocument._oUiComp._opdbmat.get(aEntries[i].key.matnr).then().catch(notFound.bind(id));
}

daleharvey:
You need to add context to the error callback that tells the function what id you were originally looking for, my `.bind` version wont actually work, but its a common requirement in javascript especially when dealing with loops, http://stackoverflow.com/questions/750486/javascript-closure-inside-loops-simple-practical-example may help

ZeroNine00:
thx,

I tried with 

var funcs = [];
			
```
[...]

PR #6529, opened 2017-05-26
Include doc ID in 404 (not_found) errors for IDB adapter

alxndrsn:
Adds a property `docId` to not_found error when trying to get an individual doc in the IDB adapter.

As posted on the issue:

> It doesn't really feel like the correct place to be fixing this, so if anyone's got any guidance it would be appreciated. This patch took me less time to write than I spent debugging an app to find the ID of a missing doc, so I'm very interested in getting this issue closed.

Issue: #5490

### DIFF
--- a/packages/node_modules/pouchdb-adapter-http/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-http/src/index.js
@@ -498,7 +498,10 @@ function HttpPouch(opts, callback) {
       }).then(function () {
         callback(null, res);
       });
-    }).catch(callback);
+    }).catch(function(e) {
+      e.requestedDocId = id;
+      callback(e)
+    });
   });
 
   // Delete the document given by doc from the database given by host.
--- a/packages/node_modules/pouchdb-adapter-http/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-http/src/index.js
@@ -498,9 +498,9 @@ function HttpPouch(opts, callback) {
       }).then(function () {
         callback(null, res);
       });
-    }).catch(function(e) {
+    }).catch(function (e) {
       e.requestedDocId = id;
-      callback(e)
+      callback(e);
     });
   });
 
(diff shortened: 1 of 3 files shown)


daleharvey:
Sorry for the delay

So this looks great however as you said it probably isnt the exact place, we aim for consistent behavior across all of our adapters so we wouldnt want a fix that only worked for idb.

The next thing this change needs is a test, this clarifies behaviour, tests it across adapters and makes sure it doesnt get regressed, copying something like https://github.com/pouchdb/pouchdb/blob/master/tests/integration/test.basics.js#L98, but change it to use a valid but non existent id should be good.

With a test added you will probably see it starting to fail now, I think at a guess the correct place to fix this would be @ https://github.com/pouchdb/pouchdb/blob/master/packages/node_modules/pouchdb-core/src/adapter.js#L607, but you will want to confirm that with testing

Thanks, any questions then please ask

alxndrsn:
@daleharvey thanks for all the pointers.  I've updated the implementation to work for remote and local adapters, adding property `requestedDocId` to the `Error` objects generated in `db.get()` calls.

alxndrsn:
N.B. this does not handle IDs in 409 errors, so does not close #5490.

packages/node_modules/pouchdb-adapter-http/src/index.js
```diff
@@ -498,7 +498,10 @@ function HttpPouch(opts, callback) {
       }).then(function () {
         callback(null, res);
       });
-    }).catch(callback);
+    }).catch(function(e) {
+      e.requestedDocId = id;
```
daleharvey: Our linting likes a space here

    }).catch(function (e) {

packages/node_modules/pouchdb-adapter-http/src/index.js
```diff
@@ -498,7 +498,10 @@ function HttpPouch(opts, callback) {
...
+      callback(e)
```
daleharvey: and a semi colon here

alxndrsn: 👍 sorry for not checking eslint output previously

daleharvey:
It seems seriously unlikely that that this commit breaks webpack

daleharvey:
Awesome all green, thanks :)

PR #6531, opened 2017-05-27
Prevent GitHub from suppressing diffs

henri-hulski:
GitHub by default suppresses diffs in node_modules directories.
This PR excludes `/packages/node_modules` from that behaviour.
Works for [Cerebral](https://github.com/cerebral/cerebral).


daleharvey:
This has been bugging me ever since they started this behaviour, thanks so much :D

PR #6536, opened 2017-05-30
Update leveldown to 1.6.0 to resolve installation fail behind proxy.

rsfernandes:
When using npm i leveldown behind corporate proxy occurs a timeout because leveldown tries to get prebuild binaries from github. The flag --build-from-source avoid this issue by always compiling the source code, but only works with leveldown 1.6.0.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "inherits": "2.0.3",
     "level-codec": "7.0.0",
     "level-write-stream": "1.0.0",
-    "leveldown": "1.5.0",
+    "leveldown": "1.6.0",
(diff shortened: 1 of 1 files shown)


nolanlawson:
Unfortunately we can't support leveldown 1.6.0 because there is a bug that breaks our tests:

- https://github.com/pouchdb/pouchdb/pull/6202
- https://github.com/pouchdb/pouchdb/pull/6372
- https://github.com/Level/leveldown/issues/194#issuecomment-269923063

If somebody could look into https://github.com/Level/leveldown/issues/194 then we'd be happy to update. :)

PR #6538, opened 2017-06-01
Improving UUID generation with better random numbers

SCdF:
Replaces Math.random() with the crypto library, if available.

I think it's mostly there. I have a couple of problems with it:
 - According to [node documentation](https://nodejs.org/api/crypto.html#crypto_determining_if_crypto_support_is_unavailable) node can be compiled without crypto support. It's not clear to me how to deal with this. `import … from` gets converted into a `require` for node, and I don't see how I can wrap that generated code in the a `try … catch` block to correctly detect if it's not available.
 - It is getting tested by virtue of [test-uuids.js](https://github.com/pouchdb/pouchdb/blob/master/tests/integration/test.uuids.js) existing, but it isn't specifically testing the `randomUInt8` function 
 - Not sure I'm supposed to [append anything to this copyright notice](https://github.com/pouchdb/pouchdb/compare/master...SCdF:better-random-uuids?expand=1#diff-b31fb7bfa63105b279cfcf023feb9263R10)

### DIFF
--- a/packages/node_modules/pouchdb-utils/src/uuid.js
+++ b/packages/node_modules/pouchdb-utils/src/uuid.js
@@ -1,3 +1,5 @@
+import randomUInt8 from './random';
+
 // BEGIN Math.uuid.js
 
 /*!
@@ -39,7 +41,17 @@ var chars = (
   'abcdefghijklmnopqrstuvwxyz'
 ).split('');
 function getValue(radix) {
-  return 0 | Math.random() * radix;
+  // Using maxRandom of 255 will produce an un-even spread of random numbers.
+  // Instead, find the biggest number (within 8-bit range) that is a multiple of
+  // radix.
+  var maxRandom = 255 - (256 % radix);
+
+  var randomValue;
+  do {
+    randomValue = randomUInt8();
+  } while (randomValue > maxRandom);
+
+  return randomValue % radix;
 }
 function uuid(len, radix) {
   radix = radix || chars.length;
--- a/packages/node_modules/pouchdb-utils/src/uuid.js
+++ b/packages/node_modules/pouchdb-utils/src/uuid.js
@@ -41,17 +41,7 @@ var chars = (
   'abcdefghijklmnopqrstuvwxyz'
 ).split('');
 function getValue(radix) {
-  // Using maxRandom of 255 will produce an un-even spread of random numbers.
-  // Instead, find the biggest number (within 8-bit range) that is a multiple of
-  // radix.
-  var maxRandom = 255 - (256 % radix);
-
-  var randomValue;
-  do {
-    randomValue = randomUInt8();
-  } while (randomValue > maxRandom);
-
-  return randomValue % radix;
+  return Math.floor(randomUInt8() / 256 * radix);
 }
 function uuid(len, radix) {
   radix = radix || chars.length;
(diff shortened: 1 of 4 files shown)


SCdF:
Some added context: we found that in a production CouchDB DB of a few million documents we had ~50 conflicts which, to the best of our understanding, came from two PouchDB instances generating the same UUID, and then replicating upward to CouchDB. This PR is an attempt to improve randomness of UUIDs generated from PouchDB by avoiding using `Math.random()` wherever possible.

packages/node_modules/pouchdb-utils/src/uuid.js
```diff
@@ -39,7 +41,17 @@ var chars = (
...
+  // Instead, find the biggest number (within 8-bit range) that is a multiple of
+  // radix.
+  var maxRandom = 255 - (256 % radix);
+
+  var randomValue;
+  do {
+    randomValue = randomUInt8();
+  } while (randomValue > maxRandom);
+
+  return randomValue % radix;
```
nolanlawson: This makes me nervous performance-wise, since we're spinning in a while loop waiting for a random value to exceed some threshold. Is there a simpler way we can use WebCrypto to generate a UUID? Note that UUIDs in PouchDB don't need to be valid uuid.v4() IDs or whatever; just as long as they're strings and not unreasonably long.

SCdF:
@nolanlawson I am (hopefully correctly) falling back on `Math.random()` for browsers that don't have it: https://github.com/pouchdb/pouchdb/pull/6538/files#diff-90fe6da3653718331d8603cde18234efR6

packages/node_modules/pouchdb-utils/src/uuid.js
SCdF: I get the nervousness! While loops based on random values are definitely cause for another look. tl;dr; though, in normal uuids it never loops, and worst case it almost never loops.

`randomUInt8` generates numbers between 0 and 255, and so the theoretical worst the threshold % chance can be is ~50% (a radix of 129). In legit use of `uuid()` with radixes under 62 the worst case (radix 52) is ~81%. In the standard UUID case where the radix is 16 it's actually *100%*: i.e. it will never loop.

The intention with this approach was to change as little as possible in regard with how you generated uuids. To match what you can do with web crypto (generate random bytes) with what you can do with Math.random (generate a float 0 < x < 1) I hid it behind generating the index into char based on the radix. The reason it can loop (on some radixes) is that if the random number is larger than the maximum multiple of the radix then we'll generate uneven random numbers (because the entire collection of possibilities now don't map evenly against every radix possibility).

A faster approach (in terms of UUID generation)--or at least one that doesn't have a while loop in it-- would be to just use the `node-uuid` library. I'm not sure how it's MIT licence gels with PouchDB's Apache one, but if it's acceptable we could pretty much just copy https://github.com/kelektiv/node-uuid/blob/master/v4.js whole cloth.

daleharvey:
So yeh also worried about the while loop, however in node we do not really care about the package size of the uuid generation, its only in the browser that it matters, so if you want to use lightweight browser code and the standard package for node, think that would be ok?

SCdF:
(actually don't merge this change. Doing some more testing and I'm not happy with the distribution. I suspect it has something to do with float precision)

SCdF:
I'm closing this in favour of trying a different approach: using the node-uuid library

Issue #6552, opened 2017-06-10
Feature request: add "added" property to changes feed


gr2m:
For apps that listen to the `.changes()` feed in order to update the UI there is currently no way to differentiate changes that created a new document to changes that updated a document, which is a big challenge. Previous attempts to differentiate `create` and `update` have been removed (https://github.com/pouchdb/pouchdb/releases/tag/5.0.0), and rightfully so, they only differentiated based on the `_rev` number.

The only way for an app to know when a change is an update to a previously existing document in the database or was added to the database (e.g. via a replication coming from a remote database) is to keep track of what exists in the database and check against that for each `change` event. 

I wonder if it would be possible to add an `"added"` property to the changes feed the same way there is a `"deleted"` property today. I know that this means to diverge from what CouchDB does, but I think the `"added"` event does not make sense in a CouchDB cluster while it makes a lot of sense in a client. It’s also just an added property and is not changing existing behavior beyond that.

Let me know what you all think :)

nolanlawson:
The only way to determine if a doc was added is to check that the revision number starts with `1-`. Even this isn't foolproof, though, since you can have generation-1 conflicts and thus multiple documents that start with `1-`.

This is something that it would be technically possible for PouchDB to expose, but it's also something that the developer themselves can reasonably infer from the changes feed – i.e. use `changes()` with `{style: 'all_docs'}`, check the `seq` of any `1-` documents, and see if they're lower than the `seq` of the `1-` document you just saw. Is there any reason that this would be an infeasible solution for you?

gr2m:
sorry I don’t mean to differentiate between create and update from a document’s perspective.

I want to differentiate between when a document revision added to a document to my database and when it was updating a document that existed in my database before.

Say I replicate from a remote database, a document with revision `3-abcdef...` should be marked as "added" in the changes feed, because this revision is when it was added to this database. 

I wonder if we could implement that on PouchDB’s changes feed somehow.

gr2m:
To add some context, this is a major problem we have with Hoodie. We want to emit an `add` event when a document gets to a database, no matter its revision. From a developer perspective I want to know when a new document was added to my store so I can update my view correctly

gr2m:
[...]

PR #6553, opened 2017-06-10
"added" flag in "change" events

gr2m:
# THIS IS WORK IN PROGRESS – DO NOT MERGE

see #6552


daleharvey:
Would you expect `added` when you add, delete then create another document with the same id? in this case we process it as an update, would probably be possible to avoid if you look at `wasDeleted`

gr2m:
I didn’t know, thanks for letting me know. I’ll add a test case for this if we decide to move forward with this PR

daleharvey:
Closing out for now

Issue #6308, opened 2017-02-27
Option to disable remote checkpoints


nolanlawson:
I can't find where this was originally discussed, but in a discussion about the performance of the replication protocol I suggested we add an option to disable remote checkpointing. We already do this in cases where the remote DB refuses writes.

Off the top of my head, API could look like this:

```js
db.replicate.to('http://remote', {
  remote_checkpoints: false
})
```

Or maybe:

```js
var remoteDB = new PouchDB('http://remote', {
  checkpoints: false
})
db.replicate.to(remoteDB)
```

I'm leaning towards the second one, because it doesn't encode a concept of "remote" or "local," since PouchDB is potentially p2p and also allows replication between locals and between remotes. `checkpoints` would default to `true`, though, which is awkward. We could also make it `disable_checkpoints: true` instead.

gr2m:
I’m looking into the code for this. I could use some help for documenting that feature and giving some insights on why we do the checkpointing in the first place, and why we do it by default

gr2m:
@nolanlawson could you give me a few more pointers? 

I’m looking into https://github.com/pouchdb/pouchdb/blob/master/packages/node_modules/pouchdb-checkpointer/src/index.js. I’ve set `this.readOnlySource = true;` in the constructor just to see what happens when I run tests, but all tests pass?

My guess is that if we set the options to disable remote checkpoints, I’ll avoid running `this.writeCheckpoint()`, so we won’t write checkpoints in neither source nor target. We also want `getCheckpoint` to behave the same it does if `readOnlySource` is true.

And I assume that we can’t just set `readOnlySource` with the new option, it should be a separate state, as it would be confusing with my source is not read-only but I still don’t want checkpoints?

daleharvey:
We checkpoint because we do not want to run entire replications from scratch every time you load your local application, a checkpoint says 'this is where we left off, start from here'. 

We write checkpoints to both the client and the server to handle the case where either gets deleted, if you know the server will never be deleted then its safe to skip writing the remote ones which is what we are talking about here, with the flag enabled we would still be writing local ones.

I dont think there is any need to distinguish this feature and `readOnlySource`, they do the same thing, I would say its best to rename it but reuse the same code letting it be set as an option.

I really think this needs to be an option on the replication as opposed to the PouchDB instance though, its changing the behaviour of the replicator works instead of how pouchdb functions

gr2m:
agree with all, will do, thanks Dale!

[...]

PR #6555, opened 2017-06-12
(#6308) - Add option to disable checkpoints

chadrien:
Checkpoints can be disabled either on source, target, or both.


jareware:
Been looking forward to this for a long time, 🎉 🎉 !!

daleharvey:
This is great, thanks a lot, some people that will be very happy about this :)

kanongil:
The `checkpoint = 'source'` option is broken. It doesn't check for the `opts.writeSourceCheckpoint` in `getCheckpoint()`, which results in a get on the source that 404's, leading to _writing a new checkpoint on the source_ that starts from `LOWEST_SEQ`.

I suspect the fix is to check for `opts.writeSourceCheckpoint` along with `readOnlySource` in `getCheckpoint()`, thus avoiding the `get` & `put`.

Issue #6502, opened 2017-05-15
changes() leaks memory


cztomsik:
Any `changes()` call without `return_docs: false` leaks memory

It seems that `return_docs` is effective even for http connections unlike from what is mentioned in the docs:
> Is available for non-http databases and defaults to true

### Reproduce
```
const db = new PouchDB('http://127.0.0.1:5984/skimdb')

db.changes({
  include_docs: true,
  batch_size: 1000
}).on('change', (c) => {
  // doesnt matter
})
```

What happens is that it will "slow down" after first ~50 000 seqs - because of excessive memory usage (few seconds and it's hundreds MBs). I am not sure if it relates to any existing issue so I am creating new one.

This might be resolved by:
1. updating docs (along with examples)
2. making `return_docs: false` default for http databases

vvo:
Just stumbled on this too, we have been facing a memory leak on https://github.com/algolia/npm-search and have been searching for it a bit.

I will use this `return_docs: false` and see what happens. The docs says `[return_docs] Is available for non-http databases and defaults to true`.

We are using the http adapter, but strangely the code of the http adapter sets return_docs to true by default, see https://github.com/pouchdb/pouchdb/blob/d483bdf78a295b71cb43419d285b0eed2fcf360d/packages/node_modules/pouchdb-adapter-http/src/index.js#L777

Not sure this as anything to do with our leak but will try.

Haroenv:
This did fix our memory leak, so maybe `return_docs` should always be `false` by default.

vvo:
> maybe return_docs should always be false by default.

Yes in the http couch adapter, will wait for a more knowledgeable person to confirm this.

vvo:
@willholley If you need more insights let us know, for now this is mostly a documentation issue. The doc says that if you are only using a http layer or DB then you are fine, changes will not leak. But that's not the case because the actual http pouchdb layer sets returnDocs: true by default.

willholley:
thanks @vvo - will do

daleharvey:
So yup the documentation is wrong here, however we should not be setting different defaults for http or not, we should fix the documentation for whether this option is available for http adapters

Seperately we could change the default for all adapters to returnDocs = false, or more drastically remove the option for `returnDocs` at all and require any changes listeners to collect their own results

cztomsik:
just fixing docs is fine :)

vvo:
[...]

PR #6559, opened 2017-06-15
(#6502) http adapter - changes() return_docs defaults false

avickers:
* resolves (#6502) and aligns with documentation by making return_docs
default false for the http adapter


daleharvey:
Closing this out, as discussed in related issue we would fix the docs, and / or change the default for all adapters (or remove returnDocs), but not change for one adapter

PR #6565, opened 2017-06-21
Replacing uuid function with node-uuid

SCdF:
Instead of trying to write out own that doesn't use Math.random(), let's use an existing library.

I think we can do this for two reasons:
 -  In the codebase we only use `uuid` for either generating a standard uuid, or generating a `_rev` (a lower case uuid without the dashes). So then, we don't need to continue to support arbitrary lengths and radixes
 - From tests, explicitly including `uuid/v4` means that the other library implementations (`v1` etc) do not get included in the final builds, which means we are not bloating builds with unnecessary code.


SCdF:
Hmm. Or so I thought. Forgot to run the web tests. Looks like browserify is converting:

```
import { v4 as uuid } from 'uuid/v4';
...
uuid()
```
into:
```
var uuid_v4 = require('uuid/v4');
...
uuid_v4.v4();
```

Which doesn't work, obviously.

nickcolley:
What impact does this have to the bundle size, is it bigger/smaller? I imagine this will determine if we'd want to do this?

SCdF:
@nickcolley it's just the right size :-P

The reason for doing this is that the current implementation uses `Math.random()`, which generates UUIDs that are not as random as they should be. In a production CouchDB database with a few million records we have a 50-100 records that have conflicted because they were given the same UUID by different PouchDB instances. 

This is, as a far as I can tell, the simplest way of achieving this (note that it currently doesn't work, I have to find time to work out why it doesn't compile into the browser properly). I had a custom-implemented version in another PR but it had a while loop in it and people weren't confident.

The only other option I can think of is hand-re-implement node-uuid myself, which will basically get me where this PR should get me in terms of size and quality, but with the added detriment of not getting updates / etc.

SCdF:
@daleharvey / @nolanlawson I'd appreciate if you could take a look at this second stab at generating UUIDs better. This time I'm just using node-uuid, but it's only using the bits of the library that it needs (that is, the v4 implementation and a couple of util functions).

I have confirmed that it's only bundling the bits that it needs, and not other implementations by grepping the built bundles in `./pouchdb/packages/node_modules/pouchdb/dist/`

Here are the byte differences, compiled on my local machine (`npm run build`), between master and this branch rebased over it:
image

The build fails on coverage, but I can't work out why that would be me, so I'm wondering if that's just a flakey test.

garrensmith:
This looks really nice. I'm happy with it. But I would prefer a +1 from @daleharvey and @willholley before merging.

garrensmith:
The other thing, is this a breaking change? Do we release this in the next 6.* release or put it in 7?

SCdF:
> The other thing, is this a breaking change? Do we release this in the next 6.* release or put it in 7?

As far as I can tell it isn't. However, I'm not too familiar with how plugins work. If they can access the old `uuid(len, radix)` function then there may be code out there asking for all sorts of variations on radix and length, which we are no longer supporting.

daleharvey:
Yeh its definitely not a breaking change, we only support the public API for semver. I think this is a good change, glad to get rid of some custom code for a well tested lib, thanks

tests/integration/test.uuids.js
```diff
@@ -37,32 +37,9 @@ describe('test.uuid.js', function () {
...
-    var length = 32;
-    var count = 10;
-    var uuids = makeUuids(count, length);
-    // Test single UUID wrapper
-    uuids.push(testUtils.uuid(length));
-    uuids.map(function (uuid) {
-      uuid.should.have.length(length, 'UUID length is correct.');
-    });
-  });
+  it('_rev generation', function () {
```
daleharvey: As a future note as this was broken before you touched it, but integration tests are for public api stuff in general, this could just be a unit test if you think its even needed at all

willholley:
This seems to be causing some issues with the bundling. I'm not sure how broad the impact is, but I see build warnings along the lines of:
```
'uuid/v4' is imported by packages/node_modules/pouchdb-utils/src/rev.js, but could not be resolved – treating it as an external dependency
```

and then the `update-package-json-for-publish.js` script fails to resolve the `uuid` package. @daleharvey @nolanlawson could use your advice.

PR #6599, opened 2017-07-13
(#6596) - Fix assertions for chai upgrade

daleharvey:


daleharvey:
This has weirdly broken the map reduce tests, lets just see if there are teething problems in chai that get fixed in next release

Issue #6607, opened 2017-07-14
'uuid/v4' is imported by ..., but could not be resolved – treating it as an external dependency


daleharvey:

daleharvey:
This error gets printed while building pouch and been reported that it effects builds

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/28fba0870b1438626ca57f2278fc94d102c4c5fa

PR #6609, opened 2017-07-14
(#6607) - Use plain imports for uuid

daleharvey:
With the fancy name we miss dont match https://github.com/pouchdb/pouchdb/blob/master/bin/build-module.js#L45 and treat the import as a local one, with this the only warning now is 

     'default' is imported from external module 'uuid' but never used

Which was there previously and should only mean extra code bundled, although would still like to get rid of it


daleharvey:
Sweet cheers

PR #6617, opened 2017-07-15
(#4575) - Remove strict test for status

daleharvey:


daleharvey:
Yeh thats what the reference to the bug is about, will be keeping the bug open and carry on working on it, cheers

Issue #6623, opened 2017-07-16
local build failure


daleharvey:
```
  ✓ wrote packages/node_modules/pouchdb/dist/pouchdb.js
events.js:160
      throw er; // Unhandled 'error' event
      ^

Error: write EPIPE
    at exports._errnoException (util.js:1018:11)
    at WriteWrap.afterWrite (net.js:800:14)
```

Isnt failing in CI, would prefer not to back anything out as I have seen this a lot and never properly debugged

daleharvey:
`TRAVIS=1 npm run build` works fine, it looks like an issue with https://github.com/pouchdb/pouchdb/blob/master/bin/build-utils.js#L36

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/b337b80bfc228ba04fba5f628d282e9dde338a10

PR #6624, opened 2017-07-16
(#6623) - Update code to uglify package

daleharvey:


daleharvey:
Something around uglifyjs changed, switching to the normal api fixes it. Not worried about losing any optimisations as I would much rather we optimise local dev by only building what we need and not the world. Plus now if there is an error around here travis will pick it up now

daleharvey:
Cheers :)

PR #6626, opened 2017-07-17
(#4575) - Refactor tests to cancel operations in sequence

daleharvey:


daleharvey:
This test fails regularly (but not consistently) locally running against master, the connection limit does seem like a push here although that wouldnt run against travis however this looks to be less faily

I would like to determine the exact failure but am also planning to clean up the ajax code to be far cleaner so dont want to spend too much time debugging old code

Issue #6551, opened 2017-06-08
Update and test with the latest leveldown


peakji:
### Issue
We've fixed the bug in `leveldown` that may cause a crash while closing db with open iterators: https://github.com/Level/leveldown/issues/194#issuecomment-306686747

However, the `Assertion 'dummy_versions_.next_ == &dummy_versions_' failed` error can be from anywhere in the C++ land, so I'm not sure if PR https://github.com/Level/leveldown/pull/368 solved the PouchDB case.

It'll be great if you could try with the latest master of leveldown, and ping me if the test still fails.

### Info
- Environment: Node.js
- Adapter: LevelDB

hubgit:
I can confirm that the `Closing db does not cause a crash if changes cancelled` test no longer crashes, which is great progress.

Unfortunately the "local" version of the test still doesn't pass (unlike the "http" version of the test, which is fine), as the new "iterator has ended" error that's thrown isn't being caught.

ralphtheninja:
There's an additional advantage to updating to `leveldown@1.7.2` since it uses `preubild-install` instead of `prebuild --install` which reduced the dependency tree substantially.

@hubgit Is there anything I can do to help with this? It seems we broke something by fixing something else.

hubgit:
@ralphtheninja [this](https://github.com/pouchdb/pouchdb/blob/master/tests/integration/test.changes.js#L2366) is the test that's still failing - I didn't manage to get to the bottom of how to catch the thrown error appropriately.

ralphtheninja:
@hubgit What is the command to run all the appropriate tests? Do I need some extra setup with couchdb etc?

daleharvey:
`npm test` will run all the tests, `GREP=test.changes.js npm test` will run only that file, you do need couchdb running on localhost:5984, if that is too much of a hassle 

```
npm install -g pouchdb-server
pouchdb-server
```

is an easier version

ralphtheninja:
@daleharvey Thanks! I'm in the middle of the bush right now and a huge party, but will hop on this when I'm back home :)

hubgit:
Here's what I've been able to surmise so far:

[...]

PR #6639, opened 2017-07-23
upgrade leveldown and fix iterator problems

ralphtheninja:
Closes https://github.com/pouchdb/pouchdb/issues/6551

I was a bit unsure on where to check for the error. We have different options here.

At first I wanted to put an error handler on `changeStream` [here](https://github.com/pouchdb/pouchdb/blob/master/packages/node_modules/pouchdb-adapter-leveldb-core/src/index.js#L1042), which could check for that particular error. But unclear to me how a different error should be handled. `changeStream` is also completely internal to the `api._changes` function so a bit clunky to handle the errors.

So I tried to go deeper instead and try to stop this error as early as possible and as close to `iterator.next(cb)` as possible and basically don't even emit it on the stream to begin with.

Let me know if you aren't happy with this solution and/or if it should be tweaked.

Cheers

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "inherits": "2.0.3",
     "level-codec": "7.0.0",
     "level-write-stream": "1.0.0",
-    "leveldown": "1.5.0",
+    "leveldown": "1.7.2",
     "levelup": "1.3.8",
     "lie": "3.1.1",
     "localstorage-down": "0.6.7",
--- a/packages/node_modules/sublevel-pouchdb/src/readStream.js
+++ b/packages/node_modules/sublevel-pouchdb/src/readStream.js
@@ -76,7 +76,7 @@ ReadStream.prototype._cleanup = function (err) {
 
   var self = this;
   /* istanbul ignore if */
-  if (err) {
+  if (err && err.message !== 'iterator has ended') {
     self.emit('error', err);
   }
 


ralphtheninja:
I have no idea about some of the errors. Might some of them be related to other things?

daleharvey:
And yup, the failing tests werent level related and can be a bit flaky, trying to get them more stable but occasionally need a restart

ralphtheninja:
> Ignoring this based on the string of an error message is quite fragile, it would be nice to be a little more robust and only ignore errors when we know for example the iterator died after we closed the db, or preferably to close the iterator ourselves.

Aye, I was thinking exactly the same. I was thinking we could clean that up later by e.g. named errors in e.g. `level/errors` that abstracts it better. But currently only `levelup` uses the error module and not leveldown so that had to be changed etc.

ralphtheninja:
@daleharvey What do you reckon the new version is going to be? A patch?

ralphtheninja:
Nevermind, I'm an idiot :) https://github.com/pouchdb/pouchdb/blob/master/package.json#L3

ralphtheninja:
No wait, that version is not correct either haha. Ok, I think I got it now. Next version will be `6.3.5`.

andrewholsted:
Any update on when this might land in the next version @daleharvey?

PR #6649, opened 2017-07-27
6643 - support seq_interval for changes and use during replication

willholley:
CouchDB 2.x adds a new parameter to _changes, `seq_interval`, that allows seq information to be skipped for intermediate changes within a batch. This can improve changes response time (and therefore replication throughput) considerably. It is safely ignored by CouchDB 1.x so I think it makes sense for PouchDB to use it during replication.

The potential breaking change here is if a user listens to the `change` event of `sync` or `replication` - seq values for those changes may now be null. If this impacts users, specifying `{seq_interval: false}` as an option to replication will disable the new behaviour.


daleharvey:
Awesome this looks great thanks

Issue #5547, opened 2016-08-09
Couchbase Sync Gateway + Pouchdb: Initial sync creates 1000s of requests


hermitdemschoenenleben:
Hello,

I'm using Couchbase Sync Gateway 1.3 and Pouchdb 5.4.5. If I have many documents on the server and start an initial sync, I noticed that PouchDB creates one request for every single document, which of course is very inefficient.
Here are screenshots of the first few requests (I tried creating an HAR file, but it got too big...):
https://www.dropbox.com/s/fg08d87sp4lvty2/screen1.png?dl=0
https://www.dropbox.com/s/5t57ynsoh3n1lrb/screen2.png?dl=0
I noticed that one of the first requests is one to _bulk_get?revs=true, which fails with 406 Not Acceptable:
{"error":"Not Acceptable","reason":"Response is multipart"}
Maybe this causes the problem?

hermitdemschoenenleben:
do you think that this is a bug or are multipart responses just not implemented? In other words (as the current behavior is a deal-breaker for me), is this something easy to fix?

nolanlawson:
Couchbase Sync Gateway does not have full support for PouchDB and might never have it, since we have not had many PRs from Couchbase folks, and the PouchDB maintainers are primarily interested in maintaining compatibility with CouchDB. YMMV

hermitdemschoenenleben:
In the last weeks I investigated the problem further and found out that the problem seems not to be linked to CSG, but apparently is caused by PouchDB's general replication protocol:

1) When performing a two-way replication with a server, pouch basically starts two independent replications that do not interact. If the pull replication fetches documents from the remote DB, the push replication afterwards makes a _revs_diff call to the remote server in order to find out whether to upload these documents again. If both replication processes communicated with each other, this call could be avoided, as PouchDB already knows that the document with [id, rev] was just fetched and doesn't have to be uploaded.
What makes things much worse is the fact, that when the pull replication fetches many documents at the same time, this doesn't result just in one _revs_diff call but in many small ones. Thus, when you have a fresh instance PouchDB that syncs for the first time with a server that has a lot of documents (let's say 500), this easily causes 50 unnecessary _revs_diff calls, although (with a high enough batch size) just a single _all_docs call was performed.
[...]

PR #6658, opened 2017-08-01
(#5547) - add multipart support for bulkGet api

xiaoyvr:
The bulk_get api from CSG is returning as multipart/mixed. But pouchdb
expecting application/json. This commit is to add the multipart/mixed
support for bulkGet api to support CSG.


cyberjar09:
why did you close the PR and delete the branch? @xiaoyvr

Issue #5820, opened 2016-10-29
Couchbase Sync Gateway is not supported at 100%


daleharvey:
So opening one issue that we can dupe any CSG bugs to

The state of CSG support is, we have it in our test suite @ https://travis-ci.org/pouchdb/pouchdb/jobs/171449096, it has been green forever despite fairly severe problems we have found with csg previously so I am not certain I trust it. Over the years we have had the occasional patch to fix something in CSG, it certainly does not get the use or the testing that CouchDB / Cloudant does.

The current active maintainers have all shown a preference towards focusing on CouchDB compatibility so I do not see a need to track CSG issues, we may want to remove mention of its support from the website, if you are using CSG and are seeing a bug, please feel free to submit a fix, I think we will be happy merging stuff that doesnt interfere with the code base too much but providing tests will still be necessary. 

I do see this as being a bit of a last call though, unless someone wants to take up maintenance of the tests to ensure full compatibility, it seems likely that we may want to drop support for csg explicitly in the future

nolanlawson:
We've always been open to Couchbase folks providing pull requests and have interacted with them frequently, but yeah, at the end of the day compatibility is impossible unless you consider _something_ a standard, and for us the de-facto standard has been CouchDB because there's never been a formal spec/standardization process (closest is http://replication.io).

A couple years ago we were in a similar position with Cloudant – most of the tests were failing, the API surface wasn't guaranteed to work 100%, and even replication was a little wonky. However @willholley came in and basically fixed all the Cloudant bugs as well as CouchDB 2.0 bugs and so that problem is largely fixed. We never saw the same from Couchbase (despite a few great PRs from @jchris), and so I think even mentioning CSG support on the website as "in progress" is just not an accurate depiction of reality at this point.

There are plenty of CouchDB-likes out there (rcouch, AvanceDB, pouchdb-server, pouchdb-express-router, Drupal Relaxed Web Services), but ultimately it's up to those maintainers to provide compatibility because we don't have time or resources to test them all, even if PouchDB has what amounts to the only "Couch compatibility test suite" in the Couch ecosystem. We've even been careful not to give pouchdb-server any special privileges (it's an allowed failure in CI), so we've been pretty even-handed about it.

[...]

PR #6660, opened 2017-08-02
(#5547) - add multipart support for bulkGet api

xiaoyvr:
The bulk_get api from CSG is returning as multipart/mixed. But pouchdb
expecting application/json. This commit is to add the multipart/mixed
support for bulkGet api to support CSG.

### DIFF
--- a/packages/node_modules/pouchdb-adapter-http/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-http/src/index.js
@@ -296,6 +296,35 @@ function HttpPouch(opts, callback) {
   api.bulkGet = coreAdapterFun('bulkGet', function (opts, callback) {
     var self = this;
 
+    function wrapCallbackForMultipart(cb) {
+      return function (err, obj, resp) {
+        if (err || !resp || !resp.contentType) {
+          cb(err, obj, resp);
+          return;
+        }
+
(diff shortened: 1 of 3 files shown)


daleharvey:
So if we say we accept multipart/mixed we dont just need to handle multipart json but also xml, it doesnt look like this patch covers that?

For us to merge anything really, we need to have the fixes tested to ensure they work + arent regressed, for a patch like this it would mean restoring the tests running against couchbase server and showing the change in behavious here in the tests

I apologise for this patch having a large barrier to entry re: reviving all the tests, but it takes a lot of work to support an extra platform, supporting it a little bit makes a lot more work for everyone involved and nobody as yet has put a lot of effort into CSG support

If that is something you want to do then can discuss it further @ https://github.com/pouchdb/pouchdb/issues/5820, cheers

xiaoyvr:
Thanks for the response. I do understand the workload to fully support CSG is quite huge. I failed to find an easy way to add tests around the commit as well. 

So I guess my question is, is there a way to partially support CSG instead of taking an all or nothing approach? So that we don't need to wait for all things to be ready to use/contribute pouchdb. At least for my scenario, there's no need to support attachment related features. 

In order to do this, the test suite might need to be changed. I'm not sure whether it's something worth to do.

PR #6695, opened 2017-08-14
(#4575) - Ensure databases are cleaned up

daleharvey:
Previously we used to only cleanup databases once a since suite was finished to help reduce the amount of requests made, we also used to do a pre cleanup before a test run as it made the suite behave better when encoutering errors.

When we switched to randomly generated db names for couchdb master tests we forgot to handle that and now running the test suite left thousands of databases on the server possibly causing bugs with open file handles etc, our test suite should clean up after itself


willholley:
nice find! LGTM when green.

daleharvey:
Green on Couch 2.0 in the browser!

PR #6699, opened 2017-08-16
update mac couchdb install method

solancer:
Installing couchdb through brew on mac will end up in connection refused error. The best and easy way is to use the official binary from couchdb website.


daleharvey:
I suspect the homebrew error was likely temporary, however the official packages are easy to install and dont depend on having installed homebrew so I think its best recommending them, thanks

solancer:
yes the official packages are easy to install indeed. BTW just a heads up, the setup guide is not updated with the changes from the pull request yet https://pouchdb.com/guides/setup-couchdb.html

PR #6722, opened 2017-08-30
(#4575) - Remove couch master from expected failures

daleharvey:


daleharvey:
timebombing, fairly uncontroversial

Issue #6730, opened 2017-09-04
Resuming a "checkpoint: target" replication will seek from the beginning


afinne:
### Issue
Situation: Replicating from a CouchDB over http to a PouchDB, and having the replication settings to only write checkpoints in the target PouchDB. If the replication is interrupted, and later resumed, it will not find the correct checkpoint, so it will seek the _changes feed from the beginning. 

### Info
- PouchDB 6.3.4 in Cordova app
- CouchDB 2.0 (macOS app)

The relevant parts should be around here: https://github.com/pouchdb/pouchdb/blob/master/packages/node_modules/pouchdb-checkpointer/src/index.js#L137

It seems like `getCheckpoint` is not aware of the checkpointer options.

BrunoDeBarros:
I'm having problems with this as well; not doing remote checkpoints saves a huge amount of time on the first load but PouchDB ends up just re-checking every change ever anyway, which defeats the point of it.

daleharvey:
and fixed

PR #6734, opened 2017-09-05
(#6730) - fix single-sided checkpointing

alxndrsn:


packages/node_modules/pouchdb-replication/src/replicate.js
```diff
@@ -392,6 +392,9 @@ function replicate(src, target, opts, returnValue, result) {
         return;
       }
       return checkpointer.getCheckpoint().then(function (checkpoint) {
+        if (typeof result.initial_checkpoint === 'undefined') {
```
willholley: I'm not keen on changing the result to support the tests here. I think you can test by intercepting the _changes call as in https://github.com/pouchdb/pouchdb/blob/master/tests/integration/test.replication.js#L2400

packages/node_modules/pouchdb-checkpointer/src/index.js
```diff
@@ -133,9 +133,21 @@ var comparisons = {
 
 Checkpointer.prototype.getCheckpoint = function () {
   var self = this;
+
+  if (self.opts && self.opts.writeSourceCheckpoint && !self.opts.writeTargetCheckpoint) {
```
willholley: I wonder whether this should be combined with `self.readOnlySource` - it seems like we have two properties to express the same thing

tests/integration/test.replication.js
```diff
@@ -729,6 +729,136 @@ adapters.forEach(function (adapters) {
...
+      var remote = new PouchDB(dbs.remote);
+
+      db.bulkDocs({ docs: docs.slice(0, 1) })
+        .then(function () {
+          PouchDB.replicate(db, remote)
+            .on('error', done)
+            .on('complete', function (result) {
+              result.docs_read.should.equal(1);
+              result.docs_written.should.equal(1);
+              result.initial_checkpoint.should.equal(0);
```
willholley: as above, suggest capturing the `since` parameter sent to `changes()` rather than changing the schema of `result`

tests/integration/test.replication.js
```diff
@@ -729,6 +729,136 @@ adapters.forEach(function (adapters) {
...
+                      done();
+                    });
+
+                });
+            });
+        })
+        .catch(done);
+    });
+
+    it('Test replicate on source only resumes OK', function (done) {
```
willholley: perhaps 'Test replication resumes when checkpointing on source only' is more descriptive

tests/integration/test.replication.js
```diff
@@ -729,6 +729,136 @@ adapters.forEach(function (adapters) {
...
+
+                  PouchDB.replicate(db, remote, replicateOpts)
+                    .on('error', done)
+                    .on('complete', function (result) {
+                      result.docs_read.should.equal(1);
+                      result.docs_written.should.equal(1);
+                      result.initial_checkpoint.should.equal(0);
+                      done();
+                    });
+
```
willholley: remove whitespace

tests/integration/test.replication.js
```diff
@@ -729,6 +729,136 @@ adapters.forEach(function (adapters) {
...
+                  PouchDB.replicate(db, remote, replicateOpts)
+                    .on('error', done)
+                    .on('complete', function (result) {
+                      result.docs_read.should.equal(1);
+                      result.docs_written.should.equal(1);
+                      result.initial_checkpoint.should.equal(1);
+                      done();
+                    });
+
```
willholley: remove whitespace

tests/integration/test.replication.js
```diff
@@ -729,6 +729,136 @@ adapters.forEach(function (adapters) {
...
+                });
+            });
+        })
+        .catch(done);
+    });
+
+    it('Test replicate on target only resumes OK', function (done) {
```
willholley: perhaps 'Test replication resumes when checkpointing on target only' is more descriptive

tests/integration/test.replication.js
```diff
@@ -729,6 +729,136 @@ adapters.forEach(function (adapters) {
...
+      var db = new PouchDB(dbs.name);
+      var remote = new PouchDB(dbs.remote);
+
+      var replicateOpts = { checkpoint: 'target' };
+
+      db.bulkDocs({ docs: docs.slice(0, 1) })
```
willholley: lots of duplication here with the previous tests. Consider refactoring so that the test body can be reused, whilst injecting different `replicateOpts`

packages/node_modules/pouchdb-checkpointer/src/index.js
alxndrsn: The way I read it, there are four different options:

`source` \ `target` | on | off
---|---|---
**on** | default | previously broken
**off** | previously broken | no checkpointing

How would you recommend combining?

willholley: `self.readOnlyCheckpoint` is set when we detect that we don't have write permissions on the source. It seems like, once set, we'd want the behaviour to be the same as if `opts.writeSourceCheckpoint` is `false`. Looking at it again, this line probably doesn't change, but we have 2 similar-looking flags adjusting checkpoint behaviour that look like they should be combined in general

packages/node_modules/pouchdb-replication/src/replicate.js
alxndrsn: I agree.  I'll look into the suggested approach.

alxndrsn:
@willholley thanks for the feedback :slightly_smiling_face:

tests/integration/test.replication.js
```diff
@@ -2643,15 +2759,13 @@ adapters.forEach(function (adapters) {
       var targetPut =  target.put;
       target.put = putter;
 
-      var changes = source.changes;
-      source.changes = function (opts) {
+      interceptChanges(source, function(opts) {
```
afinne: Travis complains about this one, `function(opts)` vs. `function (opts)`

alxndrsn: Thanks :¬)

packages/node_modules/pouchdb-replication/src/replicate.js
alxndrsn: Change and it works, although a little weird as `changes()` is called twice for each call to `replicate()`, first with the initial checkpoint value, and then with the final checkpoint value.

packages/node_modules/pouchdb-checkpointer/src/index.js
alxndrsn: I'm not sure it's that simple, though.  One flag is a request for behaviour, and the other is a reaction to discovered properties of a DB.  E.g.
```
Checkpointer.prototype.updateSource = function (checkpoint, session) {
  if (this.opts.writeSourceCheckpoint) {
    var self = this;
    if (this.readOnlySource) {
      return Promise.resolve(true);
    }
```
Perhaps the auto-detection could be removed and an error thrown if the source is discovered to be read only when the user has requested checkpointing on both sides?  OTOH the current behaviour is quite helpful...

alxndrsn:
Ready for review again :¬)

packages/node_modules/pouchdb-checkpointer/src/index.js
alxndrsn: @willholley PR for combining the two at https://github.com/pouchdb/pouchdb/pull/6738

alxndrsn:
Test failures for https://travis-ci.org/pouchdb/pouchdb/builds/272428063:

## 1. `CLIENT=selenium:firefox:47.0.2 COMMAND=test`
https://travis-ci.org/pouchdb/pouchdb/jobs/272428078:
```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

## 2. `CLIENT=node SERVER=couchdb-master COMMAND=test`
https://travis-ci.org/pouchdb/pouchdb/jobs/272428087
```
  1) suite2 test.replication.js-http-http Test replication resumes when checkpointing is enabled:
     expected '1-g1AAAAIZeJyV0FEKgjAYwPGvDKqnjlAniM3NOZ_yJrXtm5jYeuq5blI3qZvUTWzqg0gEyuAbbPx_g5UAsMgDhJU7o6UpDeMt8YuW_mKqQK-rqiryQMHJH8wNCxNBCcLy4tBmR2fxb6w3fupdrxdExMTQYX1a9_teHybWSM2H9Ye6v_bfV4ohk4N6N_MTbn7zxL0zGJdWxmqE8WiNZ2cozamyYoTxao13ZyCjaFQ2wvi0RvMfk8aILNNRyH-r4gtSH43J' to equal 1
```
Looks like seq numbers on the remote DB are strings rather than integers.  Can we just do something like `parseInt(string.substring(0, string.indexOf('-')))` in the test?

## 3. `CLIENT=selenium:firefox:47.0.2 SERVER=couchdb-master COMMAND=test`
https://travis-ci.org/pouchdb/pouchdb/jobs/272428088

I don't really understand what's going wrong with this build.

## 4. `COMMAND=report-coverage COVERAGE=1`
https://travis-ci.org/pouchdb/pouchdb/jobs/272428095
```
ERROR: Coverage for lines (99.98%) does not meet global threshold (100%)
```

-----

Any advice on getting these issues fixed gratefully received :slightly_smiling_face:

willholley:
> Looks like seq numbers on the remote DB are strings rather than integers. 
This is correct - sequence values are not guaranteed to be incremental and so cannot be compared. Is the issue that the tests assume specific values for the checkpoints? If so, you'd need to just capture the expected checkpoint value and ensure it's passed to `changes` verbatim.

For the firefox tests, I'll restart them - it's probably just saucelabs being flakey.

The code coverage errors indicate that some of the new code paths aren't being exercised by the tests. This is worth looking at and, ideally, adding covering tests.

afinne:
I'm not sure if this is 100% relevant, but it seems like sequence numbers can't be trusted to be ordered on CouchDB: https://stackoverflow.com/questions/44447505/sequence-number-bug-in-couchdb-2-or-is-there-another-way-to-compare-sequence-num

Can we think of another way to check whether we get the desired `since`?

tests/integration/test.replication.js
```diff
@@ -729,6 +729,155 @@ adapters.forEach(function (adapters) {
...
+          opts.since.should.equal(expectedSince);
+          expectedSince = false;
+        }
+      });
+
+      db.bulkDocs({ docs: docs.slice(0, 1) })
+        .then(function () {
+          PouchDB.replicate(db, remote)
+            .on('error', done)
+            .on('complete', function (result) {
```
afinne: Does the result contain anything that could be useful, like `seq_written` or something?

alxndrsn:
I think I might have the tests passing.  Just waiting for a full batch to complete on Travis…

alxndrsn:
The build is passing and this is now ready for review.  Happy to rebase and squash beforehand if that's helpful.

alxndrsn:
Thanks!  And extra thanks for including my unedited commit message log ;¬)

afinne:
<3

PR #6754, opened 2017-09-13
Ignore package-lock.json

alxndrsn:
If this file isn't going to be committed, it would be really handy to ignore it - it gets regenerated every time I run `npm install`.  Ignoring it in my global gitignore is not suitable, as some projects _do_ commit it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ lerna-debug.log
 tests/integration/utils-bundle.js
 *.heapsnapshot
 /pouchdb-server-install
+/package-lock.json


alxndrsn:
Tests are failing, but I don't understand how this change could cause that to happen.

daleharvey:
Probably intermittent, however been seeing that package-lock.json is designed to be committed so seems like maybe we should do so as well? is it something greenkeeper will keep up to date @boennemann / @janl ?

daleharvey:
Yeh we are supposed to commit this file and reproducible builds are a huge +1 for us, so closing this out (https://docs.npmjs.com/files/package-lock.json)

PR #6793, opened 2017-10-15
Fixes regression that causes incompatibility with pouchdb-wrappers.

loic:
`installWrappers()` from pouchdb-wrappers patches original methods with
the `_handlers` attribute, but because `Pouchdb.replicate()` is a
property which returns a new object every time, the patched `from` and
`to` methods were actually lost.

Refs ceaeeb72ae6c2ff653cac1ad09f5f242ab21accf.

PR #6799, opened 2017-10-19
(#6793) Fix incompatibility with pouchdb-wrappers

marten-de-vries:
Trying to get PR #6793 by @loic past Travis.

### DIFF
--- a/packages/node_modules/pouchdb-replication/src/index.js
+++ b/packages/node_modules/pouchdb-replication/src/index.js
@@ -8,14 +8,17 @@ function replication(PouchDB) {
   Object.defineProperty(PouchDB.prototype, 'replicate', {
     get: function () {
       var self = this;
-      return {
-        from: function (other, opts, callback) {
-          return self.constructor.replicate(other, self, opts, callback);
-        },
-        to: function (other, opts, callback) {
-          return self.constructor.replicate(self, other, opts, callback);
-        }
-      };
+      if (typeof this.replicateMethods === 'undefined') {
+        this.replicateMethods = {
+          from: function (other, opts, callback) {
+            return self.constructor.replicate(other, self, opts, callback);
+          },
+          to: function (other, opts, callback) {
+            return self.constructor.replicate(self, other, opts, callback);
+          }
+        };
+      }
+      return this.replicateMethods;
     }
   });
 


marten-de-vries:
Ok, this has taken long enough. Current master also fails the same job as this is stuck on, and I created an issue to look into it further: https://github.com/pouchdb/pouchdb/issues/6803

I'm going to merge this. If it turns out it specifically broke Safari later, which I consider to be extremely unlikely based on the code, I'll fix it. Promise.

Merged: 2f4614a0a17db40627cd40b190a3a24b41066c46

loic:
Awesome, thanks @marten-de-vries. If it turns out it breaks anything I'll be there as well to do the after-sales service ;)

loic:
Do you think it's possible to cut a new release so we can proceed with modernizing the `pouchdb-server` monorepo?

If it's not possible, a stop-gap solution could be to host a test/dev build on the `pouchdb` github account (like I did with https://github.com/newlogic42/dist-pouchdb) that we could use it on `pouchdb-server` master until we near a release of both packages. See https://github.com/newlogic42/pouchdb-server/pull/2/files#r146157616 as an example.

marten-de-vries:
@loic The solution we've historically always used is to (temporarily) set the version of pouchdb in pouchdb-server to point to the latest master at Github: https://docs.npmjs.com/files/package.json#github-urls

We'd then turn it back before the next pouchdb-server (and after the next pouchdb) release. Does that work for you?

loic:
@marten-de-vries unless I misunderstood the way it works, I think this stopped working since pouchdb itself became a monorepo. Didn't investigate in detail but I think it has to do with the repo not containing the "built" version of the project. That's why I created a `dist-pouchdb` for my personal usage.

marten-de-vries:
You're right, of course. The downside of that would be that a ``dist-pouchdb`` git repo is effectively manually generated, and as such hard to maintain.

I see that ``release.sh`` supports a ``BETA`` option, which might be a feasible solution here, as you'd be able to use npm packages, while we would at the same time not inconvenience other users. That said, I have never myself done a PouchDB release so I'd like someone else to weigh in before running that script. Any downsides I'm missing?

loic:
Indeed, cutting a pre-release hosted on NPM would be perfect.

loic:
@marten-de-vries it just occurred to me that since this is a merged PR it may not receive much attention anymore, are you reaching out through another channel?

marten-de-vries:
@loic Yes, I did put a message on IRC/Slack. I could temporarily push your repo to pouchdb/dist-pouchdb, but I think Travis variables are repo-specific, not organisation-specific, so I'm not sure if that would help in the meantime. And as I said, it's not really a long term solution anyway.

loic:
I see your message on IRC now! forgot to check sorry about that.

We don't need to change Travis variables, we just need a publicly available target for the `package.json` of `pouchdb-server` (e.g. L45 of https://github.com/newlogic42/pouchdb-server/pull/2/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R45). But I agree it's not a long term solution, the beta release idea is more promising.

marten-de-vries:
Oh, I thought you wanted it under /pouchdb/ so it could connect to Saucelabs etc. Never mind then.

PR #6804, opened 2017-10-22
fix Checkpointer.getCheckpoint if checkpoints are only written one db (#6730)

hermitdemschoenenleben:


daleharvey:
So we have an open PR in progress, @alxndrsn looks to have done a bunch of work on this but having trouble getting the tests passing - https://github.com/pouchdb/pouchdb/pull/6734

We will definitely need tests for this, will try to take a deeper look at both of these PR's shortly (doing a quick triage right now) and comment further, but if yall agreed on a best solution and worked together on getting the tests passing it would likely go a lot faster :) cheers all

daleharvey:
Fixed by https://github.com/pouchdb/pouchdb/pull/6734#pullrequestreview-74104961

PR #6812, opened 2017-10-27
(#5710) Add pending property support

### LINKED ISSUE #5710: Feature Request: Support CouchDB 2.0 'pending' property in _changes feed response
### Issue

CouchDB 2.0 has introduced a 'pending' property in the `_changes` feed response, which should aid in reporting **replication progress** to users during large data syncs. 

Previously, I used a combination of `last_seq` and `update_seq` to calculate a 'percentage complete' value, but I have found cases when after a completed replication, `last_seq/update_seq` is not always 1 (i.e. close to, but less than 1). 

Here is a screenshot of XHR JSON returned from a two-way, live replication between a PouchDB 6.0.5 local db and a remote CouchDB 2.0-rc4 db:

pendingpouchdb

So the data is already given, we just need to provide it to users in a PouchDB way.
## Expectation

I would love for this `pending` value to appear in the `db.replicate.from/to().on('change', (info) => {...})` handler.

 i.e. `info.pending` or similar ?
### Info
- Environment: All
- Platform: All
- Adapter: All
- Server: CouchDB 2.0

dharders:
This PR simply passes the 'pending' property to the replication 'change' event when provided by the server (CouchDB 2.0+).

Note: This PR only provides the pending property FROM the server, so it will only appear in the FROM change events. i.e. pending docs to DOWNLOAD. If you know of an easy way to calculate pending docs to UPLOAD, then I'm all ears and will add it to this PR. I tried with some `update_seq`, `last_seq` trickery but failed to get it working in all scenarios.

Tests included.
Closes #5710

### DIFF
--- a/packages/node_modules/pouchdb-adapter-http/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-http/src/index.js
@@ -954,7 +954,11 @@ function HttpPouch(opts, callback) {
             if (returnDocs) {
               results.results.push(c);
             }
-            opts.onChange(c, pending);
+            /* istanbul ignore if */
+            if (typeof pending === 'number') {
+              c.pending = pending;
+            }
+            opts.onChange(c);
           }
           return ret;
         });
--- a/packages/node_modules/pouchdb-adapter-http/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-http/src/index.js
@@ -935,9 +935,10 @@ function HttpPouch(opts, callback) {
       if (res && res.results) {
         raw_results_length = res.results.length;
         results.last_seq = res.last_seq;
+        var pending;
         // Attach 'pending' property if server supports it (CouchDB 2.0+)
         if (typeof res.pending === 'number') {
-          results.pending = res.pending;
+          pending = res.pending;
         }
         // For each change
         var req = {};
@@ -952,7 +953,7 @@ function HttpPouch(opts, callback) {
             if (returnDocs) {
               results.results.push(c);
             }
-            opts.onChange(c, results.pending);
+            opts.onChange(c, pending);
           }
           return ret;
         });
--- a/packages/node_modules/pouchdb-adapter-http/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-http/src/index.js
@@ -935,7 +935,7 @@ function HttpPouch(opts, callback) {
       if (res && res.results) {
         raw_results_length = res.results.length;
         results.last_seq = res.last_seq;
-        var pending;
+        var pending = null;
         // Attach 'pending' property if server supports it (CouchDB 2.0+)
         /* istanbul ignore if */
         if (typeof res.pending === 'number') {
--- a/packages/node_modules/pouchdb-adapter-http/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-http/src/index.js
@@ -935,6 +935,10 @@ function HttpPouch(opts, callback) {
       if (res && res.results) {
         raw_results_length = res.results.length;
         results.last_seq = res.last_seq;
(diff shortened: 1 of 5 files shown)


dharders:
Thanks @daleharvey 

Any idea on emulating a 'pending' value for upload?

daleharvey:
I think it would be great to support, not sure how couchdb calculates it but opening an issue and investigating their approach would probably be a good start for that

dharders:
Will do. 

I noticed that the xhr responses after an upload also contain a 'pending' value but it never makes the changes event feed as there are no changes (ie differences) to what we have in our local db as we just upload those docs.

A similar approach to couchdb will probably be required to get accurate  values, as currently last_seq and update_seq don't fully cut it in all scenarios.

Once upload 'pending' values land too we will finally have easy and accurate replication progress reporting and percentages !!!

dharders:
Created new issue #6818 for pending UPLOAD feature

dharders:
Hmmm, not sure if travis is being funky or there is a legitimate issue with the code here.

jareware:
Just watching from the sidelines (great work!), but I think you meant

> Closes #5710

instead of

> Closes #5170

:)

daleharvey:
The coverage tests do not currently run against 2.0, which means the pending part of that code will not run, put a ```/* istanbul ignore if */``` above the if statements you add and coverage should be good

dharders:
Done. Now we wait for old mate travis...

@jareware Oops and thanks! Fixed.

dharders:
I've got no idea how to fix these `selenium:firefox` failures. 😕 

The logs are saying `#3961 Many attachments on same doc` test is failing (not my test), but this test passes fine when I run it on my local dev machine, which makes it hard to troubleshoot.

Also, the error log information appears to be from a different test `'#3539 - Exception in changes is fine'`, which is what we'd expect as we intentially `throw new Error('an error')` within the client change event.

I think I need your help to continue @daleharvey

daleharvey:
I think these are intermittent, restarting

dharders:
Finally all green! 🎉

daleharvey:
Yup thanks so much :)

darshantejani007:
I am not yet getting `pending` property in `.on('change', info=>console.log(info))` on one-shot replication from remote `couchdb (v2.1.0)`. `pending` flag appears in fetched response from couchdb (consoled the response from pouchdb.js code). But it is not being transmitted to the `info` object on change listeners. Is that a bug? or am I missing something here?

Configurations: 
adapter: 'websql' (also tried on `cordova-sqlite` and `idb`)
pouchdb: `6.4.3`
remote couch: `2.1.0`
cordova-android : `7.1.0` (for ionic V1)

aneeshd16:
Hi, @dharders We managed to get pending docs to upload by this code:
```
import genReplicationId from 'pouchdb-generate-replication-id';

const localDB = new Pouch('dbname')
const remoteDB = new Pouch('http://host/dbname')
const selector = {...} // we do filtered replication from pouch -> couch based on a selector, we need to use the same selector here
return genReplicationId(localDB, remoteDB, {
	live: true,
	retry: true,
	include_docs: true,
	conflicts: true,
	selector
})
  .then(function (replicationId) {
    return localDB.get(replicationId)
  })

  .then(function (localReplicationDoc) {
    return localReplicationDoc.last_seq
  })

  .catch(function (err) {
    // if there was no replication yet, localDB.get will fail with 404. In that case, we want return all docs, as no doc has been synchronised yet
    return 0
  })

  .then(function (seqNum) {
    return localDB.changes({
      since: seqNum,
      include_docs: true
    })
  })

  .then(function (response) {
    return response.results.map(result => result.doc);
  })
```

A big chunk of this is from https://github.com/zoepage/pouchdb-hoodie-unsynced-local-docs

PR #6815, opened 2017-10-29
(#6814) - Add support to store cookies on Node.js

### LINKED ISSUE #6814: pouchdb-ajax: support storing cookies on Node.js
### Issue
On Node.js, PouchDB does not hold a cookie jar for cookies received from server during authentication. It uses the `request` Node.js module to make HTTP requests but it does not configure it correctly to hold a cookie jar.

For cookies to be stored on Node.js, `pouchdb-ajax`'s line https://github.com/pouchdb/pouchdb/blob/4bd9c087402102994f6cec6bfc54caaebafffb25/packages/node_modules/pouchdb-ajax/src/request.js#L3 should at least be
```js
export default require('request').defaults({
  jar: true
});
```

### Info
- Environment: Node.js server (not on same host as CouchDB)
- Platform: Node.js + Express
- Adapter: Http
- Server: Private remote CouchDB with SSL proxy, cookie authentication and properly configured CORS

### Reproduce

1. Connect from a Node.js server to a remote CouchDB with cookie authentication.
2. Make a simple get query to see that the AuthSession cookie is not transmitted.

I don't know how I could setup a jsbin or such to demonstrate it, because I have no idea how I should mock a CouchDB. Please tell me if and how it is possible to do that.

ptitjes:
On Node.js, PouchDB does not hold a cookie jar for cookies received
from server during authentication. It uses the request Node.js module
to make HTTP requests but it does not configure it correctly to hold
a cookie jar.

This PR adds the minimal required configuration for the request module.
Fixes #6814.


ptitjes:
BTW, I think this should fix many complains about cookie authentication with PouchDB on Node.js like https://github.com/pouchdb-community/pouchdb-authentication/issues/18 and https://github.com/pouchdb-community/pouchdb-authentication/issues/76.

daleharvey:
Will take a look into this, but wanted to ping @mikeal to point out wether this was obviously wrong or right for any reason?

ptitjes:
Thanks @daleharvey !

ptitjes:
@daleharvey, I took some time to have a second read  of the `README.md` file from request.

So, in [All Options](https://github.com/request/request#requestoptions-callback):
> ## request(options, callback)
>
> The first argument can be either a `url` or an `options` object. The only required option is `uri`; all others are optional.
>
> [...]
>
> - `jar` - if `true`, remember cookies for future use (or define your custom cookie jar; see examples section)

And at the bottom of [Examples](https://github.com/request/request#examples), there is a short discussion on cookies and cookie jars:

> Cookies are disabled by default (else, they would be used in subsequent requests). To enable cookies, set `jar` to `true` (either in `defaults` or `options`).
> 
> ```js
> var request = request.defaults({jar: true})
> request('http://www.google.com', function () {
>   request('http://images.google.com')
> })
> ```
>
> To use a custom cookie jar (instead of `request`’s global cookie jar), set `jar` to an instance of `request.jar()` (either in `defaults` or `options`)
> 
> ```js
> var j = request.jar()
> var request = request.defaults({jar:j})
> request('http://www.google.com', function () {
>   request('http://images.google.com')
> })
> ```

The latter means we could make the cookie jar available as a global variable in case a pouchdb preset build is used. Something along the lines (in `pouchdb-ajax`):

```js
var request = require('request');
var jar;
if (!!GLOBAL) {
  if (!GLOBAL.POUCHDB_AJAX_COOKIE_JAR) {
    GLOBAL.POUCHDB_AJAX_COOKIE_JAR = request.jar();
  }
  var jar = GLOBAL.POUCHDB_AJAX_COOKIE_JAR;
} else {
  jar = request.jar();
}
request = request.defaults({ jar });
```

Could we move forward about this ? Sorry to be pushy on this but, as PouchDB is a monorepo, it is impossible for us to depend on a custom branch in our `package.json` dependencies...

ptitjes:
I added a test that shows whether pouchdb-ajax uses cookies or not. This tests runs only on Node.js as it requires to setup an http server.

It fails without the `.defaults({jar: true})` and succeeds with it.

tests/integration/test.ajax.node.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
+
+// This test only runs on Node.js
+var http = require('http');
+
+var adapters = ['http', 'local'];
+
+
+adapters.forEach(function (adapter) {
```
daleharvey: So this isnt an intergration test, could you remove any of the `adapters` code here and put it in `/tests/components/`, the integration tests are (mostly) for core API stuff that works on all platforms, there is some node only stuff in there but that was before we have different test folders

ptitjes:
OK, moved and fixed the test as you requested. And rebased on master.

daleharvey:
Awesome thank you, I think its possible people are going to want this option to be configurable at some point, but this looks like a good start for that

ptitjes:
Thanks @daleharvey ! I'd be happy to work on some follow-up if we identify what is needed.

Issue #6819, opened 2017-11-02
Package versions are not up-to-date


ptitjes:
Package versions have not been updated during last releases:
Current PouchDB version is 6.3.4 but package versions are set to 6.2.1-prerelease.

daleharvey:
Fix landed

PR #6820, opened 2017-11-02
(#6819) Update package version numbers

ptitjes:
This PR updates the package version numbers to the last release version: 6.3.4.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pouchdb-monorepo",
-  "version": "6.2.1-prerelease",
+  "version": "6.3.5-prerelease",
   "homepage": "http://pouchdb.com/",
   "license": "Apache-2.0",
   "private": true,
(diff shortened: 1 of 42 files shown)


ptitjes:
Not sure to fully understand what you meant, so let me rephrase and you confirm/correct me if needed.

You tell me redo the set-versions fix with "6.3.5-prerelease"  and the change of the release workflow will be discussed elsewhere. That's it ?

Just to be complete, let me just take some time to give you the scenario under which one might need recent (at-least more than those in npm) version numbers. When you use PouchDB plugins that themselves use modules from PouchDB monorepo (say pouchdb-authentication that uses pouchdb-ajax) and that you have a fix pending acceptation for PouchDB, then you have to not have npm pull dependencies for you plugin from the npm repository but those that you manually installed.

ptitjes:
I made the change (provided I did correctly understand what you requested) and rebased on master.

daleharvey:
Thanks loads, yup thats exactly what I meant we put the next version -prerelease just to make sure that if someone builds a release from their checkout and uses it, it is clear that it is not a release and where in the release line it lives, if the version numbers were say the current release, 6.3.4 then it would be easy to get your custom master release and the real 6.3.4 mixed up, hopefully this works for your use case but if not then feel free to open an issue and we can discuss, cheers

Issue #6823, opened 2017-11-04
In pouchdb-utils guardedConsole, should `console === 'undefined'` be `typeof console === 'undefined`?


browndav:
### Issue

In https://github.com/pouchdb/pouchdb/blob/master/packages/node_modules/pouchdb-utils/src/guardedConsole.js#L3, it looks like there's an attempt to test for the existence of `console`, but – as far as I can tell – a `typeof` operator should be present.

The code in git tests `console === 'undefined'`, but looks like it should test `typeof console === 'undefined'`.

### Info
- Environment: Node.js, Rollup, PhantomJS, Mocha/Chai
- Platform: Mac OS X 10.12.6
- Adapter: WebSQL (PhantomJS)
- Server: Local only

### Reproduce
REPL> delete console;
REPL> console === 'undefined'
ReferenceError: console is not defined
REPL> typeof console === 'undefined'
true

alxndrsn:
Or `console !== undefined` 🙂

daleharvey:
If you wanted to PR this to typeof console I would be happy to merge, havent seen this cause any problems but I am more comfortable with typeof

alxndrsn:
@browndav PR submitted 🙂

daleharvey:
And fixed, cheers

PR #6847, opened 2017-11-16
(#6823) - make guardedConsole checks safer

alxndrsn:
Fix for #6823

### DIFF
--- a/packages/node_modules/pouchdb-utils/src/guardedConsole.js
+++ b/packages/node_modules/pouchdb-utils/src/guardedConsole.js
@@ -1,6 +1,6 @@
 function guardedConsole(method) {
   /* istanbul ignore else */
-  if (console !== 'undefined' && method in console) {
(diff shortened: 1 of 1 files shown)


packages/node_modules/pouchdb-utils/src/guardedConsole.js
```diff
@@ -1,6 +1,6 @@
 function guardedConsole(method) {
   /* istanbul ignore else */
-  if (console !== 'undefined' && method in console) {
+  if (typeof console !== 'undefined' && typeof console[method] === 'function') {
```
daleharvey: tiny nit, once conditions start getting longer in length I usually find it cleaner to split them, so

    if (typeof console !== 'undefined' && 
        typeof console[method] === 'function') {

Looks like this is exactly on the limit (80) and is fine, just a heads up for future

alxndrsn: 👍 great to know the house style 🙂

Issue #6854, opened 2017-11-19
allDocs with keys and IndexedDB should create only one transaction


nename0:
### Issue
Currently when calling allDocs with the `keys` option `pouchdb-core`([source](https://github.com/pouchdb/pouchdb/blob/master/packages/node_modules/pouchdb-core/src/adapter.js#L751)) splits them into allDocs calls with one `key`. On the indexDB adapters this will create as many transactions as keys. From a bit of pure IndexedDB testing I found out that this can be made more perfomant when using just one transaction.
Test Results with 10000 entries in IndexDB and fetching 500 keys:
Chrome: 8x faster
Firefox: 2x faster

### Info
- Platform: Chrome 62 / FF 57

### Reproduce
I wrote the test with the pouchdb performance tests. If you want to reproduce them, I put them [here ](https://pastebin.com/2stvqXfi). (The code is not perfect; I should use `transaction.oncomplete` to know when its finished)

daleharvey:
Fixed

alxndrsn:
From brief profiling:

Fetching all docs from a production db containing 20,000 docs, `6.4.2` is about 100-400 times faster in Chrome than `6.3.4` and `6.1.2`.

I suspect this fix is the cause of the speedup.  Thanks to all involved!

nename0:
Yeah, It really depends on the amount of keys you query. I even optimized my project (using pouchdb) to batch up key queries

Issue #6543, opened 2017-06-04
Implement pouchdb-find for idb-next


daleharvey:
I wanted to open a discussion thread about this as its long been the main point and blocker for releasing `idb-next`

daleharvey:
Ok, so this is one of the main reasons I wanted to do `idb-next` as it would be possible to implement `pouchdb-find` but not based on top of mapreduce. This could potentially fix a huge amount of performance and size issues in PouchDB.

Now onto implementation details, for one this will likely mean a change in syntax for `pouchdb-find`, in `next` we have a single objectstore with a 1:1 mapping of documents, so we can create an indexeddb index on that objectstore and use that for queries, the main issue is that currently pouchdb-find does `createIndex` and indexeddb requires indexes to be defined as part of the schema

I suspect we want this to look something like 

```
var db = new PouchDB('games', { 
  indexes: [{fields: ['name']}]
});

return db.find({
  selector: {name: {$gt: null}},
  sort: ['name']
});
```

(hopefully `indexes` could look a little nicer, just guessing from the example @ http://nolanlawson.github.io/pouchdb-find/)

garrensmith:
Using the built in indexing would be great. Is it possible to create indexes later at all?

daleharvey:
It would be super hard, even maintaining the indexes is going to be tricky, indexeddb requires them to be defined as part of the schema so each time an index changes it will also be a schema change (that we need to make sure we can also handle the typical schema upgrades). Its possible to do that at runtime, but it sounds extremely messy and likely to breakages.

garrensmith:
Ah ok, I remember we had a discussion around this before. Is there no way for us to do some kind of bitwise incrementing or something every time we need to add/remove an index? This isn't going to be easy but we do need a decent way to add/remove indexes

daleharvey:
oh yup sorry, we will definitely be able to add/remove them, just only on construction, so we do 

```
var db = new PouchDB('games', { 
  indexes: [{fields: ['name']}]
});
```
and on startup we can check if there is a name index (or any unused indexes) and create / delete them, I only think its too complicated to do them at runtime (ie `db.createIndex`) but we defintely need to be able to add / remove them.

daleharvey:
Trying to find the original discussion about this but not finding it yet

nolanlawson:
Maybe one of these?

- https://github.com/pouchdb/pouchdb/issues/3775
- https://github.com/pouchdb/pouchdb/issues/5216
[...]

PR #6858, opened 2017-11-21
idb and webSQL adapters should only use one transaction for allDocs with keys (6854)

nename0:
PR for #6854

### DIFF
--- a/packages/node_modules/pouchdb-adapter-idb/src/allDocs.js
+++ b/packages/node_modules/pouchdb-adapter-idb/src/allDocs.js
@@ -17,6 +17,25 @@ import {
 import runBatchedCursor from './runBatchedCursor';
 import getAll from './getAll';
 
+function allDocsKeys(keys, docStore, onBatch) {
+  // It's not guaranted to be returned in right order  
+  var valuesBatch = [];
+  var count = 0;
+  keys.forEach(function (key, index) {
+    docStore.get(key).onsuccess = function (event) {
+      if (event.target.result) {
+        valuesBatch[index] = event.target.result;
+      } else {
+        valuesBatch[index] = {key: key, error: 'not_found'};
+      }
+      count++;
+      if (count === keys.length) {
+        onBatch(keys, valuesBatch, {});
+      }
+    };
+  });
+}
+
 function createKeyRange(start, end, inclusiveEnd, key, descending) {
   try {
     if (start && end) {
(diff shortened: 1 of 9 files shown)


nename0:
I also rewrote the WebSql adapter to use the keys in a SQL "... WHERE id in (keys)" statement. This gave a moderate performance boost aswell.

Note: I will remove the deleted: 'ok' option from the allDocs method in WebSql and IDB. It's not documented and only used for the keys option. So I think it's ok to just remove it (?)

nename0:
Results of new [all-docs-keys](https://github.com/pouchdb/pouchdb/pull/6858/commits/b6977fcf2a737dbea8bef5cf3673267d4cb0f602#diff-43963c040e192bd7d42b1bec99c2fc5bR138) performance test:

| Median  | Before | After |
| ------------- |:-------------:| -----:|
| Chrome 62 idb  | 32.3 ms | 4.5 ms |
| Firefox 57 idb | 12.3 ms | 3.5 ms |
| Chrome 62 websql| 33.4 ms | 1.7 ms |

nename0:
#### Issuses: 

- [Build Job 32](https://travis-ci.org/pouchdb/pouchdb/jobs/305733391):
I traced down the reason for [this](https://travis-ci.org/pouchdb/pouchdb/jobs/305733391#L2870) error. It's because pouchdb-server does not use a updated/npm linked version of `pouchdb-adapter-leveldb-core`. Using a updated version of `bin/run-tests.sh` ([gist](https://gist.github.com/nename0/e78a8297aadef97b5808eb2665d37f83#file-run-tests-sh-L9)) it works. For some reason I also had to change the npm link commands to get it working.
So in order to fix this build you'd need to update the `bin/run-tests.sh` file, but it does not really fit in this PR

packages/node_modules/pouchdb-adapter-idb/src/allDocs.js
```diff
@@ -50,18 +69,23 @@ function idbAllDocs(opts, idb, callback) {
...
-    // can just do an empty query. Else need to throw
-    return callback(createError(IDB_ERROR,
-      keyRangeError.name, keyRangeError.message));
+  var keyRange ; 
+  var keyRangeError;
+  if (!keys) {
+    keyRange = createKeyRange(start, end, inclusiveEnd, key, opts.descending);
+    keyRangeError = keyRange && keyRange.error;
+    if (keyRangeError && !(keyRangeError.name === "DataError" &&
+        keyRangeError.code === 0)) {
```
daleharvey: nit: 

    if (keyRangeError && 
        !(keyRangeError.name === "DataError" && keyRangeError.code === 0)) {

packages/node_modules/pouchdb-adapter-indexeddb/src/allDocs.js
```diff
@@ -101,57 +137,56 @@ export default function (idb, metadata, opts, callback) {
...
+  var cursor = descending ?
+    docStore.openCursor(keyRange, descending) :
+    docStore.openCursor(keyRange);
+
+  cursor.onsuccess = function (e) {
+
+    var doc = e.target.result && e.target.result.value;
+
+    // TODO: I have seen this before, but want to make sure I
+    // know exactly why its needed
```
daleharvey: Honestly cant remember, it could be copied in from the oldest idb tutorials, if you can remove it and it doesnt fail tests it would be nice to get rid of

daleharvey:
This looks great, thanks so much, I have restarted #15 and #25, its likely they are intermittent failures unrelated to your PR, as for #32 its great you found that problem as another PR had the exact same issue, I will open a new issue for that if you could submit your updated run-tests for that, then when that lands rebase this and should be all good to go? thanks

daleharvey:
@nename0 https://github.com/pouchdb/pouchdb/issues/6859 for the pouchdb-server issue

packages/node_modules/pouchdb-adapter-indexeddb/src/allDocs.js
nename0: Ok a quick test shows that removing that creates errors. Investigating...

nename0:
Ok, the checking of the cursor [here](https://github.com/pouchdb/pouchdb/pull/6858#discussion-diff-152769256R177) is needed, because when allDocs is called without limit option the cursor just finishes normally with all docs. We don't have this in the `idb` adapter, because we have #6059 there.
As I already said/asked [here](https://github.com/pouchdb/pouchdb/issues/6543#issuecomment-346197236) the  optimizations in #6059 and #6033 are also suited for `indexeddb`

nename0:
After rebase on #6865 the pouchdb-server build passes now. The [error](https://travis-ci.org/pouchdb/pouchdb/jobs/306757153) with AUTO_COMPACTION and phantomjs seems to be another random one.

nename0:
All green. Should be ready to merge

Issue #6870, opened 2017-11-24
NEXT=1 integration test uses messed up query paramters


nename0:
### Issue
Another reason why #6864 did not error integration tests:
When running the NEXT=1 integration test, things are not quite right. [Here](https://github.com/pouchdb/pouchdb/blob/f4ccbcc1b22484baf418602f1b8096ab544e5309/bin/test-browser.js#L92) `NEXT=1` will be but in query paramters of the url which is passed to selenium. However on the [dev-server](https://github.com/pouchdb/pouchdb/blob/f4ccbcc1b22484baf418602f1b8096ab544e5309/bin/dev-server.js#L37) `src` will be set as query paramters, which are then logged to console, but not `NEXT=1`. The [webrunner.js](https://github.com/pouchdb/pouchdb/blob/f4ccbcc1b22484baf418602f1b8096ab544e5309/tests/integration/webrunner.js#L8) actually requires the later one to really use the pouchdb-next. However in [this test](https://github.com/pouchdb/pouchdb/blob/f4ccbcc1b22484baf418602f1b8096ab544e5309/tests/integration/test.compaction.js#L516) the NEXT=1 query paramter is also needed
 
### Info
- Environment: pouchdb-next

### Reproduce
Test the different query parameter and debug the first lines of webrunner.js to see whats actually loaded

PR #6872, opened 2017-11-24
(#6870) fix usage of query parameters with NEXT=1 integration test + fix bulkDocs of indexeddb

nename0:
PR for #6870 
I chose to just use just the `NEXT=1` query parameter in the test scripts and let webrunner.js load the `pouchdb-next.js`. If you specify both `NEXT=1` and `src` query parameters now, src will be used.

### DIFF
--- a/bin/dev-server.js
+++ b/bin/dev-server.js
@@ -34,7 +34,7 @@ if (process.env.ITERATIONS) {
   queryParams.iterations = process.env.ITERATIONS;
 }
 if (process.env.NEXT) {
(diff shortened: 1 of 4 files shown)


nename0:
The error [here](https://travis-ci.org/pouchdb/pouchdb/jobs/307054087#L2225) seems to be an actually error in the `indexeddb `adapter, that bulkDocs returns the winning rev instead of the inserted rev

nename0:
I added the fix for the error with `indexeddb `adapter to this PR.

nename0:
All green. Should be ready to merge

PR #6865, opened 2017-11-23
(#6859) Update run-tests.sh to use updated packages for the pouchdb-server test

nename0:
Some builds fail because on travis pouchdb-server does not use updated packages.
[Example 1 Build 32](https://github.com/pouchdb/pouchdb/pull/6858#issuecomment-346413720)
[Example 2](https://github.com/pouchdb/pouchdb/pull/6833#issuecomment-346588271)

I now updated the run-tests.sh to update ALL the packages it can find in the pouchdb-server-install folder. Maybe thats to much

PR #6873, opened 2017-11-25
(#6864) fix missing return statements on allDocs indexeddb adapter

### LINKED ISSUE #6864: Missing return on allDocs of indexeddb adapter
In @dharders PR there is a return missing in the [allDocs implementation of indexeddb adapter](https://github.com/pouchdb/pouchdb/pull/6833/files#diff-dbb317186915d268eca5f85f5456363dR49). When running the integration tests in chrome I noticed this. It will throw this error `once called more than once`, because callback is called multiple times on allDocs

dharders:
Closes #6864 

/cc @nename0

### DIFF
--- a/packages/node_modules/pouchdb-adapter-indexeddb/src/allDocs.js
+++ b/packages/node_modules/pouchdb-adapter-indexeddb/src/allDocs.js
@@ -46,7 +46,7 @@ function handleKeyRangeError(opts, metadata, err, callback) {
     if (opts.update_seq) {
       returnVal.update_seq = metadata.seq;
     }
-    callback(null, returnVal);
+    return callback(null, returnVal);
   }
   callback(createError(IDB_ERROR, err.name, err.message));
 }
@@ -65,7 +65,7 @@ export default function (idb, metadata, opts, callback) {
     if (opts.update_seq) {
       returnVal.update_seq = metadata.seq;
     }
-    callback(null, returnVal);
+    return callback(null, returnVal);
   }
 
   var results = [];


dharders:
Hmmm, the test `CLIENT=selenium:firefox:47.0.2 COMMAND=test-webpack` failed 1st run at a random test, which usually solves itself with the normal travis retry attempt logic, but weirdly the subsequent builds failed on retry 2 and 3 before even getting to run the tests. 

Unsure if this is a result of the latest `run-tests.sh` script update merge (see [pull/6865#issuecomment-346929550](https://github.com/pouchdb/pouchdb/pull/6865#issuecomment-346929550)) or another travis issue altogether.

nename0:
I would not trust travis to much, it has been failing at random lately ([What's this?](https://travis-ci.org/pouchdb/pouchdb/jobs/306746236)) . You only changed pouchdb-next code so i would onlx look in the 28th build test, which passes for you. Also the test result are more or less meaningless until #6896 and #6870 are fixed.

dharders:
@nename0 thanks for the explanation and thanks for addressing the travis issues. It will be great to get stability and confidence back from CI test results.

nename0:
@dharders dale restarted the build job now it passes!

daleharvey:
So basically, if a test you are touching turns red, or you make a change and there are 2+ of the same type of test (nodejs or browser) failing then its a good idea to take a look at the failure, otherwise yup its common for these to fail intermittently, they always will hoever they havent had much people look at them for a llong time so could do with some love

dharders:
> if a test you are touching turns red ... and there are 2+ of the same type of test (nodejs or browser) failing ...  look at the failure

That's the conclusion that I had basically come to @daleharvey .

The confusing part was the failed **build** on retry 2 and 3, even though 1st build succeeded.

PR #6877, opened 2017-11-26
Don't use hardcoded firefox version on travis

nename0:
What @daleharvey already tried in #6832. I was able to fix the build by updating the selenium version in `test-browser.js`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,9 +12,6 @@ git:
 sudo:
   false
 
-addons:
-  firefox: "47.0.2"
-
 before_install:
   - |
     if [[ $TRAVIS_SECURE_ENV_VARS == "false"  ]] && [[ $CLIENT == saucelabs:* ]]; then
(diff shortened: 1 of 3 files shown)


nename0:
I also added a fix for [this test](https://github.com/pouchdb/pouchdb/blob/63a336fa29fc44bc5350bcd4a41f3889187b2802/tests/integration/test.get.js#L357) which randomly errored lately because of socket timeouts. 
[Example 1](https://travis-ci.org/pouchdb/pouchdb/jobs/307536423#L2998)
[Example 2 before job was restarted](https://github.com/pouchdb/pouchdb/pull/6865#issuecomment-346930888)

daleharvey:
So the failing test here is a FETCH run, this I have noticed oto be one of our most faily tests, I am in the process of writing a patch where we switch our http adapter to use fetch directly then it can be polyfilled cleaner than our current setup, but one main issue we could be seeing here is that we have fetch support from before they could be cancelled, to be honest I was surprised they ever passed, but if you take a look at this and FETCH is the only thing blocking upgrading firefox then I think its fine to add to expected failures and I will get it in my patch

Also @nename0 thanks so much for your recent contributions, I have invited you to become a contributor so you can merge other people code as you see fit (it also lets you restart your own travis jobs which helps)

nename0:
Ok, Thank you @daleharvey.

The interesting thing is the errors in #6832 I was able to reproduce. However the FETCH=1 not.

daleharvey:
Yup, while not surprised FETCH is buggy, it should at least run and I have tests passing against it on 59, will try firefox 55 which is what is being run on saucelabs

nename0:
I also tried updating selenium-standalone and wd however this lead to not only FETCH but also phantomjs not working: [travis of my fork](https://travis-ci.org/nename0/pouchdb/builds/309212583)

nename0:
Ok I fixed the issue. Selenium version 3.6.0 seems to be buggy and phantomjs would not start. So I hardcoded 3.7.1. I also tried 3.8.0 but this started chrome instead of phantomjs:question:

daleharvey:
So its super weird that FETCH=1 wont start, its possible that with FETCH already being super unstable, some speedups cause it to fail consistently before reporting anything

Either way I am currently working on fetch support by default now that cancellation has been implemented and I would like to see this patch land (it may help out with moving safari on to indexeddb as well) so I think add FETCH=1 to allowed failures for now and we can merge this

nename0:
Ok seems like I found the issue:
I noticed this before, when the the browser is opened [here](https://github.com/pouchdb/pouchdb/blob/6ae4364246b954b09e11c7b84ec1fc7f6ff235c0/bin/test-browser.js#L207) then it always takes a bit of time until the callback is called and the interval is set. On firefox this callback seems to be called when the "page is loaded", meaning that the spinner in the tab bar has stopped and the reload button changes. However when running the integration test this already takes a bit of time (~30s) proably because the integration test does many http requests and firefox think they are page loading. Now my idea was that with FETCH firefox never gets out of the loading state. I truns out to be correct. I don't know if my fix is appropriate, but the FETCH=1 seems to work now ([travis on my fork](https://travis-ci.org/nename0/pouchdb/jobs/310027440))

nename0:
Ok, now all is green, but it's a bit strange that the FETCH=1 test now runs 10mins longer

PR #6880, opened 2017-11-27
update travis to skip saucelabs faster

nename0:
This should reduce build times on PRs by terminating saucelabs build jobs in `before_install` of travis, so `npm install` does not need to be run.

I also renamed an error which confused me a bit when I saw this in the travis log.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,11 @@ addons:
   firefox: "47.0.2"
 
 before_install:
+  - |
+    if [[ $TRAVIS_SECURE_ENV_VARS == "false"  ]] && [[ $CLIENT == saucelabs:* ]]; then
+       echo "Not running test, cannot connect to saucelabs in Pull Request"
+       travis_terminate 0
+    fi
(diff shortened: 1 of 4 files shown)


nename0:
Time results before: about [3 min 30 sec](https://travis-ci.org/pouchdb/pouchdb/jobs/307294843), after [1 min](https://travis-ci.org/pouchdb/pouchdb/jobs/308129113)
So 2.5 min improvment times 6 saucelabs tests makes about 15 min. Its not huge but at least something.

I also added a fix for [this test](https://github.com/pouchdb/pouchdb/blob/63a336fa29fc44bc5350bcd4a41f3889187b2802/tests/integration/test.get.js#L357) which randomly errored lately because of socket timeouts. 
[Example](https://travis-ci.org/pouchdb/pouchdb/jobs/307536423#L2998)

nename0:
Yeah removing stress tests is of course wrong. However with the http adapter the locking of the doc ( to return 409 / 404 in correct manner) should happen on the server test and are therefore not subject of the tests. Still why we get timeouts here is indeed strange

Issue #6898, opened 2017-12-02
COUCH_HOST=http://127.0.0.1:3000/ npm run dev causes test failures


daleharvey:
The trailing slash in the url causes requests to be to `http://127.0.0.1:3000//` when calling `.info()` which can cause failures

dharders:
Merged #6901

PR #6901, opened 2017-12-02
(#6898) - Remove possible trailing slash from couchHost

daleharvey:


daleharvey:
Ah good catch fixed, also remember that we had a copied version of this function for other tests and updated that too

Issue #6900, opened 2017-12-02
Get rid of is*server() functions in tests


daleharvey:
Mostly introduced for Couch Master support, we should switch to master by default and remove and of these tests, sync gateway and pouchdb-express should straight up go

daleharvey:
Not that fussed about tracking this at the moment

PR #6902, opened 2017-12-02
(#6900) - Remove isSyncGateway from tests

daleharvey:


daleharvey:
Yup, we dont officially support CSG with discussion @ https://github.com/pouchdb/pouchdb/issues/5820, happy for other people to maintain support but in their absence its not something I want to spend time on, would like to remove the is*servers for all types (even couch master) but for couch master I will do it in a way that supports older couch as well

PR #6922, opened 2017-12-07
(#5905) - Run saucelabs on all PRs

daleharvey:


daleharvey:
Hopefully works, the other secure env vars were cloudant credentials we no longer use

daleharvey:
@nename0 wanna give a 👍 here so we can not worry about your tests not running on saucelabs? cheers

PR #6924, opened 2017-12-07
FIX update_seq - change not affecting other tabs

pubkey:
When pouchdb is used in multiple tabs with the localstorage-adapter, the seq-number is not reflected between them.

To reproduce do the following:
- Open this [fiddle](https://jsfiddle.net/cuyhxefm/41/) on 2 browser-windows
- Open the developer-console on both windows.
- One the first window, click "insert" and "getAll"
- As you can see, the seq-number is now increased and displayed correctly
- In the second window, the seq-number is not updated. You have to reload the page to get the new seq-number

This PR will:
- Remove the caching of `db._updateSeq` whenever `pouch.info()` is called.
- Assign the new seq-number to `db._updateSeq` so the cache updates.

### DIFF
--- a/packages/node_modules/pouchdb-adapter-leveldb-core/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-leveldb-core/src/index.js
@@ -247,7 +247,7 @@ function LevelPouch(opts, callback) {
   };
 
   api._info = function (callback) {
-    return stores.metaStore.get(UPDATE_SEQ_KEY, function (err, value) {
+    stores.metaStore.get(UPDATE_SEQ_KEY, function (err, value) {
       if (err) {
           callback(err, null);
       } else {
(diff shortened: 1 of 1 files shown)


nename0:
You should probably do the same with `allDocs `and the `opts.update_seq` option. [Here ](https://github.com/pouchdb/pouchdb/blob/90f9dcb979ecf7a7e4cc4d39e1d79136aed0612e/packages/node_modules/pouchdb-adapter-leveldb-core/src/index.js#L893)and [Here](https://github.com/pouchdb/pouchdb/blob/90f9dcb979ecf7a7e4cc4d39e1d79136aed0612e/packages/node_modules/pouchdb-adapter-leveldb-core/src/index.js#L980)

daleharvey:
So the level adapter is most definitely assumed to be single process, even with these fixes we still have db._updateSeq being used which will cause bugs, especially the one in bulkDocs. 

If we want to go down the route of hardening the level adapter for multiprocess use, we are going to need to start with tests that expose the above bug, we use to have a class of tests like that @ https://github.com/pouchdb/pouchdb/blob/16a27720d58544d40a6b4308079b94e763ae9054/tests/test.changes.js#L101 that can be used as a base

A genuine alternative (it would probably be easier) is to write a localstorage adapter directly that does this correctly, any reason you are using localstorage over indexeddb?

pubkey:
I'm closing this. As @daleharvey said, leveldown-adapters are assumed to run in a single process.
As worksarround I am now using an own seq-number as local-document.

PR #6925, opened 2017-12-07
Retry of #6877 "Don't use hardcoded firefox version on travis" (#6905)

### LINKED ISSUE #6905: Revert "(#6877) - Don't use hardcoded firefox version on travis"
Sorry @nename0 this caused test failures, the eval(window.results) was running to early and may need a try catch around it or similiar, alternatively we might want to make sure the browser has completed loading before we start to run the tests, I remember we had a problem with infinite loading when we started xhr to soon

However main issue is here is that your test run didnt pick it up, you are part of the pouchdb org and those saucelabs tests should be running for your PRs

nename0:
As it caused saucelabs failiures, I now changed the code a bit to keep the old behavior for saucelabs and skip waiting for `.get(testUrl)` on Selenium

Fixes #6905 
Should probably wait on #6922

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,6 @@ sudo:
   false
 
 addons:
-  firefox: "47.0.2"
   jwt:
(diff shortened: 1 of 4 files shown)


daleharvey:
it may possibly be an idea to put https://github.com/pouchdb/pouchdb/blob/master/tests/integration/webrunner.js#L150 behind a domcontentloaded or similiar so the browser can be sure to have finished loading, if we can avoid different code paths it would be best

nename0:
Wouldn't it be better to wait after loading the scripts: https://github.com/pouchdb/pouchdb/blob/master/tests/integration/webrunner.js#L84 ; also I use the `load ` event

PR #6929, opened 2017-12-08
Fix Slack link

rfigueroa:


daleharvey:
So the slack link was correct, we ran slack.pouchdb.com to run open invites for people who wanted to join the slack, but aside from it going down need to now fix the hosting and setup certificates.

Does anyone know if slack has fixed this and has public signup? if not I think its inappropriate for us to use as the recommended communication on an OSS product and we should just get rid of the link here

daleharvey:
Cheers for the PR, but gonna close out and we can have a discussion about what to do with slack for the long term @ https://github.com/pouchdb/pouchdb/issues/6931

PR #6936, opened 2017-12-12
(#6936) - Upgrade to official apache docker images

daleharvey:


daleharvey:
@willholley / @garrensmith for some reason when I switch the 2.0 to the official apache images pouchdb-find breaks completely, will try to look into it but figured in case you know anything?

(we can still get the 1.X onto apache for now)

bin/run-couchdb-on-travis.sh
```diff
@@ -1,12 +1,12 @@
 #!/usr/bin/env bash
 
 if [ "$SERVER" = "couchdb-master" ]; then
-  # Install CouchDB Master
+  # Install CouchDB 2.X (clustered)
   docker run -d -p 3001:5984 couchdbdev/debian-8-dev --with-haproxy --with-admin-party-please -n 1
```
ptitjes: I believe that [couchdbdev/debian-8-dev](https://hub.docker.com/r/couchdbdev/debian-8-dev/) is older than the official [couchdb:latest](https://hub.docker.com/_/couchdb/) image.

bin/run-couchdb-on-travis.sh
```diff
@@ -1,12 +1,12 @@
...
   COUCH_PORT=3001
 else
-  # Install CouchDB Stable
-  docker run -d -p 3000:5984 klaemo/couchdb:1.6.1
+  # Install CouchDB 1.X
+  docker run -d -p 3000:5984 apache/couchdb:1
```
ptitjes: Shouldn't we use the official [couchdb:1](https://hub.docker.com/_/couchdb/) instead of the 'semi-official' [apache/couchdb:1](https://hub.docker.com/r/apache/couchdb/) ?

willholley: +1

bin/run-couchdb-on-travis.sh
willholley: The dev images will build from the couchdb master source. We may also want to test against couchdb:latest as an independent, stable test against 2.x.

daleharvey:
I am confused, https://groups.google.com/forum/#!topic/couchdb-user-archive/dh5LGyGZiRY suggests that its `apache/couchdb` that should be used, and that a `latest` could not possibly exist. Anyone able to clarify whats the relationship between `couchdb` and `apache/couchdb` on docker?

ptitjes:
@daleharvey It is indeed not clear from reading the thread you linked. However, [apache/couchdb](https://hub.docker.com/r/apache/couchdb/) says:
> Unofficial convenience binaries for CouchDB, [...]

and when you click the "more information" GitHub [link](https://github.com/apache/couchdb-docker) the header says:
> Semi-official Apache CouchDB Docker images

Whereas [couchdb](https://hub.docker.com/_/couchdb/) is said to be an 'official repository'.

That said, I agree this is confusing. Maybe one of @wohali or @janl can disambiguate that, and I'm interested by the answer too.

wohali:
apache/couchdb are the "semi-official" binaries from the Apache CouchDB project. We build and upload these directly. They are "semi-official" because the only thing we are allowed to call an official release of the Apache CouchDB project is the tarball, not any so-called "convenience binary" builds. This is an Apache Software Foundation policy decision that I am keen to get altered in 2018, but that discussion would best be had not in a GitHub PR.

The docker official couchdb image is built using the same definitions as the apache/couchdb docker image. You can see how it's built here: https://github.com/docker-library/official-images/blob/master/library/couchdb It is "docker official" because, for the Docker project, it is their official CouchDB image.

Feel free to use either.

daleharvey:
Thats great thanks @wohali, taking that as a +1

PR #6938, opened 2017-12-12
(#5572) - Use IndexedDB in Safari

daleharvey:


nolanlawson:
This seems passing, compaction tests are a bummer to remove but make sense, LGTM

thgreasi:
In the respective [LocalForage issue](https://github.com/localForage/localForage/issues/766), we are considering to start using user agent string parsing and version checking (I know it sounds bad), since `zone.js` monkey patches `fetch`. As a result, the above detection will not work for all angular 2 apps.
Any feedback from your side would be much appreciated.

belafarinrod91:
Any tipps for a temporary workaround? Having the same issue - indexeddb can not be loaded as zone.js is monkeypatching the fetch function.

PR #6939, opened 2017-12-15
Emit an error event if an adapter fails to initialise

ZeeCoder:
I'm using PouchDB for a queued logging solution.

To make sure logs are not lost, I have to make sure PouchDB initialised correctly, and if not, try another adapter.

Since the instance we get from the constructor is an event emitter, we can listen to the "create" event, but not for an "error" event.

This currently forces me to look for errors in the taskqueue.fail property, which gets updated with potential errors.

Now, I don't know it the modification in this PR would solve my issue, but after browsing the source it seemed to be an obvious place to start. 🤔

### DIFF
--- a/packages/node_modules/pouchdb-core/src/constructor.js
+++ b/packages/node_modules/pouchdb-core/src/constructor.js
@@ -81,6 +81,7 @@ function PouchDB(name, opts) {
 
   PouchDB.adapters[opts.adapter].call(self, opts, function (err) {
     if (err) {
+      self.emit('error', err);
       return self.taskqueue.fail(err);
     }
     prepareForDestruction(self);


daleharvey:
Have you tried putting the contructor in a try catch? thats how errors in the sync construction should be caught (https://github.com/pouchdb/pouchdb/blob/master/tests/integration/test.basics.js#L37)

ZeeCoder:
Tbh I haven't even considered that to be sync, since when I'm asking for persistent storage, the user is prompted to allow for that.
Which could happen any time in the future. 🤔 

I'll double check next monday. 😅

daleharvey:
Yeh the error handling has a confusing history here, we used to have the constructor be a promise but that isnt valid and caused problems, so now if the constructor has issues it will throw but we have tried to push most of the work out of the constructor so it will never through and the first data call will throw instead

ZeeCoder:
I'll have to check it when I get back to work, it's all on my work laptop.
😅

ZeeCoder:
@daleharvey I've double checked, and a try-catch doesn't catch my error.
I don't think it's realistic to expect a sync setup either imho.

So here's my test code:
```js
try {
    const db = new PouchDB('DBName', { storage: 'persistent' });
    console.log('Executing after pouchdb setup:');
    console.log('db:', db);
} catch (e) {
    console.error('==POUCHDB ERR: ', e);
}
```

When I run this, I get the following prompt:
allowingpersistentstorage

Both console logs execute immediately, before I grant storage permissions, and the db instance has `{isReady: false, failed: false}` in the `taskqueue`.

Once I deny permissions, failed becomes an object with details of the error, and my try-catch is not executed, as expected.

So currently the **only reliable way to detect errors** is to look for this taskqueue in a loop.

I think providing an error event would be consistent with the way the DB is set up currently, since it already emits a created event on success.
(Meaning I already wait for db creation asynchronously, so I might as well handle errors the same way.)

daleharvey:
So with a stateless constructor most of the errors are defected to the next call, what happens if you do: 

    var db = new PouchDB(); 
    db.info().then( .... )

? If you dont see an error with that then I it would be good to fix that

I would like to avoid adding to the events here, the code that is used to maintain what peoples expectations around the events is messy

ZeeCoder:
I assume it would give me an error.
But that is handled differently from the case when the DB was not created the first place.
Also: I think when there's an error, "create" never gets called either, so that's also unreliable. 🤔

I'm not sure what you mean by people's expectations, can you clarify what people usually expect in a situation like this?
My first instinct when I learned that there's a "create" event, that there must therefore be an "error" event too, which was not the case. 🤔

I'm not even sure how people currently handle the case when the DB was not created successfully.

Anyway, since I have a "good enough" solution if you don't think it's against the your design, then I can close the issue.

daleharvey:
>  I'm not even sure how people currently handle the case when the DB was not created successfully.

They catch the error on the first api request made against it, calling `info()` if they want verification that it has been created which as you confirmed works.

As for the expectations, I mean that the events in PouchDB's API have been a source of confusion and code complexity, look at https://github.com/pouchdb/pouchdb/blob/master/packages/node_modules/pouchdb-core/src/setup.js most of which is just to make the created / destroyed events. If you add events people have expctations about when they are fired, ordering etc. In hindsight I would reconsider adding them in first place but would prefer to avoid adding more especially if there is another way to do the functionality

Issue #6910, opened 2017-12-04
Upgrade leveldown


daleharvey:
It looks like this is going to be a bit of a challenge so opening its own issue

daleharvey:
See https://github.com/pouchdb/pouchdb/pull/6906

ralphtheninja:
@daleharvey Remind me again, what's the simplest way of running the tests locally for PouchDB? I forget.

daleharvey:
Its no bother, I keep meaning to get it to the point to where `npm test` just works, but you need to install a server right now first, easiest way is `npm install pouchdb-server && ./node_modules/.bin/pouchdb-server` then `GREP=migration npm test` should just work (fail with the level upgrades)

ZeeCoder:
So this issue is about the leveldown 2.1.1 version, right?
But doesn't it work with 1.7.2, as that is merged in already?

Couldn't that be released before the bigger upgrade to v2?

ralphtheninja:
> So this issue is about the leveldown 2.1.1 version, right?

Yes.

ralphtheninja:
I could take a stab at this, but might need someone to hold my hand.

daleharvey:
I would be happy to, can't schedule much IRC time but will make sure to get back to any questions in here

PR #6941, opened 2017-12-16
Update/Fix levelup and leveldown to v2

ralphtheninja:
Resolves https://github.com/pouchdb/pouchdb/issues/6910

Output of `GREP=migration npm test` (while running `pouchdb-server`):

test-output


packages/node_modules/pouchdb-adapter-leveldb/src/migrate.js
```diff
@@ -36,7 +36,7 @@ var doMigrationOne = function (name, db, callback) {
       };
     }
     var sub = db.sublevel(store, opts);
-    var orig = levelup(storePath, opts);
+    var orig = level(storePath, opts);
```
ralphtheninja: I tried doing

```js
var orig = levelup(leveldown(storePath), opts);
```

But then I got the following error:

levelup-leveldown-error-migrate

This is a bit trial and error and would really like to know _why_ that doesn't work. I assume it has to do with encodings.

cc @daleharvey @vweevers

ralphtheninja:
As a side note, I get a shit tonne of

```
The following options have been renamed — please update your config: dest -> file
The following options have been renamed — please update your config: dest -> file
```

While running the tests. Not sure if it's normal or not.

packages/node_modules/pouchdb-adapter-leveldb/src/migrate.js
vweevers: I *think* the code here uses two db instances. One is wrapped with sublevel (which handles encodings), one is plain levelup. Both are expected to support `binary` and `json` encoding.

The levelup instance doesn't, and that's probably why the value `metadata` here isn't an object:

https://github.com/pouchdb/pouchdb/blob/e8d8f79a776cfb7196344f80310d09935278834d/packages/node_modules/pouchdb-merge/src/winningRev.js#L10

Hence `Cannot read property 'slice' of undefined`.

So I think it also needs `encoding-down` @ralphtheninja.

vweevers:
Reasons for failing Travis jobs:

- [`CLIENT="saucelabs:iphone:8.4:OS X 10.11" ADAPTERS=fruitdown COMMAND=test`](https://travis-ci.org/pouchdb/pouchdb/jobs/317213478): `fruitdown` needs upgrade to latest `abstract-leveldown`
- [`CLIENT=selenium:firefox ADAPTERS=localstorage COMMAND=test`](https://travis-ci.org/pouchdb/pouchdb/jobs/317213480): `localstorage-down` need upgrade to latest `abstract-leveldown`
- [`COUCH_HOST=http://127.0.0.1:3001 CLIENT=node SERVER=couchdb-master COMMAND=test`](https://travis-ci.org/pouchdb/pouchdb/jobs/317213483): no idea. @daleharvey?

vweevers:
Hm, `localstorage-down` cannot be fully `abstract-leveldown` compliant as it cannot offer snapshot guarantees. Add to that the generally terrible performance of LocalStorage (blocking reads) and the inability to store Buffers (without serializing them to base64 or other)..

If it's advisable to use other adapters, and if those are available in all environments nowadays (I don't know), perhaps the time has come to deprecate `localstorage-down`. It's still seeing some use, though on the decline:

image

vweevers:
Same question for `fruitdown`. Maybe our time is better spent on updating `level-js`, which also uses IndexedDB, without the compatibility hacks for Apple's (old?) IndexedDB.

daleharvey:
This is awesome, thanks so much. As for fruitdown and localstorage down I have added them to allowed_failures in order to unblock this. Fruitdown was made because Apples Indexeddb support was bad, however that situation has improved so not sure if its needed any more, localstorages support was always of questionable value, we can have them in allowed failures and if it looks like nobody wants to pick up support for them then just drop them

packages/node_modules/pouchdb-adapter-leveldb/src/migrate.js
vweevers: @daleharvey probably best if you do use `levelup` rather than `level` here, to avoid version mismatches.

vweevers:
> localstorages support was always of questionable value, we can have them in allowed failures and if it looks like nobody wants to pick up support for them then just drop them

Alright, sounds good. While it was a certain kind of *fun* diving down this rabbit hole, I thank you for pulling me out ;)

nolanlawson:
fruitdown is legacy, yes. It's not needed now that Apple has largely fixed their IDB implementation and old iOS/Safari versions are falling out of use.

localstorage was largely a fallback for people trying to support IE<10 which is not a whole lot of people anymore.

We can probably just remove those from the test suite because they are of limited value now.

The `following options have been renamed` warnings are fixed in #6991

PR #6950, opened 2017-12-18
Fix minified build

quentinvernot:
Hello!

I noticed the minified build were broken for the 6.4.0 release, no idea when it happened, but here is a fix for the build.

### DIFF
--- a/bin/build-utils.js
+++ b/bin/build-utils.js
@@ -33,7 +33,7 @@ function writeFile(filename, contents) {
 }
 
 function doUglify(pkgName, code, prepend, fileOut) {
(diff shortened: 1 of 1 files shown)


daleharvey:
Ugh that is a very bad bug, obviously an uglify api  change that did not get tested by us correctly, have opened an issue so we can catch this stuff next time and will do a new release with this fix, thanks

quentinvernot:
No problem! There are several ways to test this (test the build code like normal code, min-size test on the builds, import test on builds, run normal tests from the builds), but I'm not familiar with the toolchain and wouldn't know where to start. I think I saw a size-test already, maybe this part of the build can be used to run more tests?

PR #6958, opened 2017-12-21
fix find index with zero as values

nerumo:
I found an issue with find indexes if the value of the indexed fields is zero. The if condition is only if (!value) which is true with 0 and therefore no value is emitted. 
I also added a test for testing null values because they should also be allowed, but since I haven't figured out how I run tests on Windows I have to wait for the CI to tell me if it works 100%

### DIFF
--- a/tests/find/test-suite-1/test.default-index.js
+++ b/tests/find/test-suite-1/test.default-index.js
@@ -357,4 +357,60 @@ testCases.push(function (dbType, context) {
       });
     });
   });
+  it('handles zero as a valid index value', function () {
+    var db = context.db;
+    db.createIndex({
+        index: {
+            fields: ['foo']
+        }
+    }).then(function () {
+        return db.bulkDocs([
+            {_id: '1', foo: 0},
+            {_id: '2', foo: 1},
+            {_id: '3', foo: 2},
+            {_id: '4', foo: 3}
+        ]).then(function () {
+            return db.find({
+                selector: {
+                    _id: {$eq: 0}
+                },
+                fields: ['_id']
+            }).then(function (resp) {
+                resp.should.deep.equal({
+                    docs: [
+                        {_id: "1"}
+                    ]
+                });
+            });
+        });
+    });
+  });
+  it('handles null as a valid index value', function () {
+      var db = context.db;
+      db.createIndex({
+          index: {
+              fields: ['foo']
+          }
+      }).then(function () {
+          return db.bulkDocs([
+              {_id: '1', foo: null},
+              {_id: '2', foo: 1},
+              {_id: '3', foo: 2},
+              {_id: '4', foo: 3}
+          ]).then(function () {
+              return db.find({
+                  selector: {
+                      _id: {$eq: null}
+                  },
+                  fields: ['_id']
+              }).then(function (resp) {
+                  resp.should.deep.equal({
+                      docs: [
+                          {_id: "1"}
+                      ]
+                  });
+              });
+          });
+      });
+  });
 });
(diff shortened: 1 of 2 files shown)


daleharvey:
Good job looks like its working great, the only test failing is coverage, its possible that its an intermittent failure so restarted it, your test looks like it should hit the new code thats added however if you want to check that that test will ensure that every line of code that you added will get executed reliably then once we have the coverage issue fixed this looks good to merge

nerumo:
the test that failes is `test.and.js` -> `does and for _id`
There's a correlation to the code change because in the test uses a selector `$gt: null`, but the reported error is:
```
"after each" hook for "does and for _id":
     Error: database is destroyed
      at PouchAlt.<anonymous> (packages/node_modules/pouchdb-for-coverage/lib/index.js:9:9918)
      at PouchAlt.<anonymous> (node_modules/argsarray/index.js:14:18)
      at packages/node_modules/pouchdb-for-coverage/lib/index.js:9:7385
      at PouchAlt.<anonymous> (packages/node_modules/pouchdb-for-coverage/lib/index.js:9:6746)
      at PouchAlt.bulkDocs (node_modules/argsarray/index.js:14:18)
      at tests/find/test-suite-1/test.default-index.js:395:21
      at process._tickCallback (internal/process/next_tick.js:109:7)
```
Do these kind of errors happen regularly? Any ideas? Since I only have a Windows machine I'm not able to run the tests on my machine (or do I miss something?)

nolanlawson:
If you're on the Windows Fall Creators Update, you should be able to run the PouchDB test suite in Bash on Ubuntu on Windows... I believe (LevelDB was giving me problems last time I tried this). Otherwise the best bet may be Ubuntu running inside of a VM.

nolanlawson:
Coverage appears to be a true failure but I tried restarting it

nename0:
Theres no need for Ubuntu Bash on Windows. I'm doing fine with git bash. Also an easy way to run the browser tests is to use `Web Server for Chrome`: Set it to your pouchdb folder and then visit `localhost:<someport>/tests/integration/index.html`

tests/find/test-suite-1/test.default-index.js
```diff
@@ -357,4 +357,60 @@ testCases.push(function (dbType, context) {
       });
     });
   });
+  it('handles zero as a valid index value', function () {
```
nolanlawson: @nerumo It appears that the reason this is failing is that this test doesn't _return_ a promise; it merely calls `db.createIndex()` as a side effect. Thus is creates race conditions, and that's why we're seeing errors.

nolanlawson: I went ahead and made the fixup change; let's see if the tests pass...

nerumo: @nolanlawson ah true, I missed that. Tnx

nolanlawson:
Looks like there are true failures in the HTTP adapter. Does this new behavior match what CouchDB does?

```
pouchdb-find: http: test.zzz-suite-1.js handles zero as a valid index value:

       AssertionError: expected { docs: [] } to deeply equal { docs: [ { _id: '1' } ] } 
       + expected   - actual 

       {
       -  "docs": [] 
       +  "docs": [ 
       +    { 
       +      "_id": "1" 
       +    } 
       +  ] 
       }
        
      at assertEqual (node_modules/chai/lib/chai/core/assertions.js:485:19)
      at ctx.(anonymous function) (node_modules/chai/lib/chai/utils/addMethod.js:41:25)
      at doAsserterAsyncAndAddThen (node_modules/chai-as-promised/lib/chai-as-promised.js:296:33)
      at .<anonymous> (node_modules/chai-as-promised/lib/chai-as-promised.js:255:21)
      at ctx.(anonymous function) [as equal] (node_modules/chai/lib/chai/utils/overwriteMethod.js:49:33)
      at tests/find/test-suite-1/test.default-index.js:379:34
      at process._tickCallback (internal/process/next_tick.js:109:7)
```

Issue #6945, opened 2017-12-17
Remove lie


daleharvey:
Promises are now natively supported on all of our target platforms

daleharvey:
We would need https://github.com/pouchdb/pouchdb/issues/7047 to remove the lie dep completely

PR #6987, opened 2018-01-07
(#6945) - remove Promise polyfill (BREAKING)

nolanlawson:
Presumably we could also remove the polyfills for `Object.assign`, `Map`, and `Set`, but I believe this may bring minimal benefit (i.e. not reducing the bundle size very much) with some serious downsides (i.e. consumers need to bring their own Map/Set/Object.assign polyfills).

BTW this is a breaking change, intended for 7.0.

### DIFF
--- a/bin/rollupPlugins.js
+++ b/bin/rollupPlugins.js
@@ -16,13 +16,11 @@ function rollupPlugins(nodeResolveConfig) {
     inject({
       exclude: [
         '**/pouchdb-utils/src/assign.js',
-        '**/pouchdb-promise/src/index.js',
         '**/pouchdb-collections/src/**'
       ],
       Map: ['pouchdb-collections', 'Map'],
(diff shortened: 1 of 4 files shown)


nolanlawson:
Also our Map/Set polyfills are pretty awesome in that they don't bring in the _entire_ API, just what we need. Also they are smart enough to detect incomplete Map/Set implementations in old Edge/Safari.

broerse:
@nolanlawson is this related to the problems Angular 4 users have? https://github.com/pouchdb-community/relational-pouch/issues/86

daleharvey:
Yup Map and Set etc arent wide enough supported yet and such a small amount of code I dont think we should screw with them yet. I would like to remove the lie dependency completely, moving all the tests to platforms that have promises but this is a great first step 👍 thanks

Issue #6944, opened 2017-12-17
Switch HTTP Adapters to use fetch


daleharvey:
Requirement for https://github.com/pouchdb/pouchdb/issues/5322

Now that browsers have implemented the fetch api with cancellation, switch to using that in the http adapters, this removes all of pouchdb-ajax dependency. 

We will try to avoid polyfilling it, ie 11 is the only major holdout and people can polyfill themselves for that.

loic:
What's the plan on Node / `pouchdb-server`? I'm especially concerned about the cookies story.

Magellol:
@daleharvey 
>browsers have implemented the fetch api with cancellation

Didn't know fetch was cancellable natively. Are we sure about that, I wouldn't think native promises are cancellable in any way.

daleharvey:
Yup https://developers.google.com/web/updates/2017/09/abortable-fetch, I have a working implementation @ https://github.com/pouchdb/pouchdb/pull/7059

gr2m:
what do you plan to use for the fetch implementation in Node?

ptitjes:
@daleharvey I did not follow your latest developments so I ask. Does dropping cookie support on Node.js will make [tests/component/test.ajax.js](https://github.com/pouchdb/pouchdb/blob/master/tests/component/test.ajax.js) fail ? (I don't see it failing yet in CI on you WIP.)

Do you have any ideas yet about what actions we will have to take for authentication plugins, and other plugins that requires to share cookies with PouchDB itself, to still work with PouchDB 7.0 ?

daleharvey:
Yeh I don't think we can drop that test / cookies working out of the box, was assuming that there would be a working fetch + cookies module in mode but we may need to fix / write one ourselves, will probably leave it last. But plugin authors should mostly only need to service from Ajax to fetch

daleharvey:
https://github.com/pouchdb/pouchdb/commit/e35f949d1948913e4bb6a7f226de8b1af1c5d5c5

PR #6988, opened 2018-01-07
(#6944) - switch HTTP adapter to only use fetch (BREAKING)

nolanlawson:
Breaking change, intended for PouchDB 7.0.


daleharvey:
Closing out as https://github.com/pouchdb/pouchdb/pull/7059 is getting pretty close

Issue #6943, opened 2017-12-17
Deprecating WebSQL


daleharvey:
So happy days, IndexedDB is supported in all major browsers (https://caniuse.com/#search=Indexeddb) and once https://github.com/pouchdb/pouchdb/pull/6938 lands we will be using it everywhere, meaning webSQL will become legacy code

I suggest we do a 6.4.1 release that purely 6.4.0 + https://github.com/pouchdb/pouchdb/pull/6938, users who are worried about migrations can then use this release, detect if they have existing webSQL installations and migrate those over to indexedDB

Then at that point, 7.0 can be webSQL free, we can split it out into a new repo as an optional adapter for anyone to use, just not maintained + tested as part of core

nolanlawson:
https://github.com/pouchdb/pouchdb/pull/6938 is merged

nolanlawson:
There's probably no need to split it out into a new repo. Even if we remove it as a dep, running the publish scripts will continue publishing it as an optional dependency, so anyone who wants to use it can. Plus removing all the tangled `websql-core` / `node-websql` deps is probably not worth the trouble.

This plan sounds solid to me!

daleharvey:
Yeh so while ultimately I would prefer them in a seperate repo, I think thats a bit of a big change to make suddenly, can keep them here for now and close this out, after a couple of release / months can take a look at moving them. Thanks :)

moses:
Hi folks: Since the SQLite Plugin depends on the WebSQL adapter, will native SQLite support be deprecated as well within PouchDB?

Thanks for all that you do!

PR #6990, opened 2018-01-07
(#6943) - WIP - remove WebSQL adapter (BREAKING)

nolanlawson:
Opening a PR to see if any tests break. I think PhantomJS uses IndexedDB now, but I'm not sure.

Open question is whether we should ship browser builds WebSQL as we do fruitdown, memory, and localstorage. ~~I think we should probably just stop doing that altogether – it's too much effort to build and maintain browser versions for each of these three; anyone who wants to use them can just build them with browserify/webpack/etc.~~ _(on second thought: this is too much work)_

Then again, such a change would make our tests a little harder to write, since they rely on being able to include basic `<script>` tags. There's some cleanup work here to do.

### DIFF
--- a/docs/_includes/api/create_database.html
+++ b/docs/_includes/api/create_database.html
@@ -13,7 +13,7 @@ This method creates a database or opens an existing one. If you use a URL like `
 **Options for local databases:**
 
 * `auto_compaction`: This turns on auto compaction, which means `compact()` is called after every change to the database. Defaults to `false`.
-* `adapter`: One of `'idb'`, `'leveldb'`, `'websql'`, or `'http'`. If unspecified, PouchDB will infer this automatically, preferring IndexedDB to WebSQL in browsers that support both (i.e. Chrome, Opera and Android 4.4+).
+* `adapter`: One of `'idb'`, `'leveldb'`, or `'http'`.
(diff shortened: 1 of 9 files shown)


nolanlawson:
I'm going to try to make websql a separate plugin, ala fruitdown/memory/localstorage. It seems way easier to just hook into our existing build/test pipeline.

docs/_includes/api/create_database.html
```diff
@@ -31,7 +31,9 @@
 
 **WebSQL-only options:**
 
-* `size`: Amount in MB to request for storage, which you will need if you are storing >5MB in order to [avoid storage limit errors on iOS/Safari](/errors.html#not_enough_space). Persistent views use a separate database per view which will also request storage space so keep in mind the upper storage limits when using persistent views on iOS.
+_Note:_ as of PouchDB 7.0.0, this requires installing WebSQL as an add-on adapter.
```
daleharvey: I think this and the other WebSQL note just below should probably move this to the WebSQL section

daleharvey:
woot merged in https://github.com/pouchdb/pouchdb/commit/179660996893cd3e13897aa9c3a7f10880a69393

PR #6998, opened 2018-01-10
(#2936) - Update to semi official couch master docker image

daleharvey:


marten-de-vries:
I think you wanted to reference #6936 instead of #2936 in the commit message. Otherwise, this looks good to me although I personally can't keep track of the differences between ``apache/couchdb``, ``couchdb`` & ``klaemo/couchdb`` anymore. The Travis failure seems genuine.

daleharvey:
So afaik apache/couchdb and couchdb are essentially the same, both based off the klaemo build thats being replaced

The failure is genuine though

daleharvey:
ok so it seems like some `bookmark` value has started being added to mango results which the tests do not expect

daleharvey:
Taking the previous 👍 for this now that its green

Issue #6964, opened 2017-12-25
Update dependencies to close warnings


seantanly:
### Issue
Did a check using `snyk test`, there are two dependencies that are highlighted, and perhaps should be updated as soon as possible.

- debug@3.0.1 have `ReDoS`
- request@2.80.0 relying on tunnel-agent@0.4.3 that have `Uninitialized Memory Exposure`

daleharvey:
Fixed

PR #7009, opened 2018-01-15
(#6964) - Update request

daleharvey:


daleharvey:
Self 👍 oning since its just a dep update

PR #7015, opened 2018-01-16
WIP - Only treat express-pouchdb running in fullCouchDB mode as a couchdb instance.

loic:
Refs https://github.com/pouchdb/pouchdb-server/pull/276.

cc @marten-de-vries

### DIFF
--- a/tests/integration/utils.js
+++ b/tests/integration/utils.js
@@ -185,7 +185,7 @@ testUtils.putTree = function (db, tree, callback) {
 testUtils.isCouchDB = function (cb) {
   testUtils.ajax({url: testUtils.couchHost() + '/' }, function (err, res) {
     // either CouchDB or pouchdb-server qualify here
-    cb('couchdb' in res || 'express-pouchdb' in res);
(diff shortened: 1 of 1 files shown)


marten-de-vries:
:+1: when green, although this will probably require a pouchdb-server release first, right?

loic:
We'd lose the ability to test against `express-pouchdb` until a new version is released. I think we don't need to rush merging this one, `pouchdb-server` has an interim fix in its test own suite.

Issuing a new release of `pouchdb-server` with all plugins published from the monorepo is going to take some time IMO (i.e. we still need to come up with a strategy re versioning, etc.).

marten-de-vries:
Ok, that sounds fine. I updated the title to prevent accidental merges in the meantime.

marten-de-vries:
I think this is no longer blocked now?

loic:
I think so, we just need to change the express-pouchdb server version in package.json.

daleharvey:
Closing, happy to merge this when its ready + updated pouchdb-server

Issue #7025, opened 2018-01-18
PouchDB fails silently when quota is exceeded 


bogdanlupuro:
### Issue
When you try and open a local PouchDB and your Chrome temporary storage quota is exceeded, PocuhDB fails silently and does not resolve nor reject any promise, leaving it hung. 

### Info
- Environment: Browser/hybrid app
- Platform: Chrome
- Adapter: IndexedDB

### Reproduce
1. Fill your browser to have exceeded all the quota. 
```javascript
navigator.webkitTemporaryStorage.queryUsageAndQuota(function (usage, quota) {
    //usage === quota
});
```
2. Initialize PouchDB and get info()
```js
localDb = new PouchDB(dbname, configs);
localDb.info().then(function (info) {
    console.log(info)
}).catch(function (error) {
    console.error(error)
});
```
Expec

- ted behavior: Catch QuotaException: CustomPouchError {status: 500, name: "indexed_db_went_bad", message: "unknown", error: true, reason: "QuotaExceededError"}
- **Actual behavior:** Nothing happens, none of the logs are run.

daleharvey:
Thanks, dupe of https://github.com/pouchdb/pouchdb/issues/6954 (been reported a few times)

bogdanlupuro:
The fix is pretty easy, if I am not missing anything. Only one line. And it is a really blocker for us. I can prepare a PR.

daleharvey:
That would be great, susprised it still exists as we specifically fixed and tested this previously, however I think it was a hassle to test in CI so maybe regressed

PR #7026, opened 2018-01-18
#7025 - fix silent hang when starting a db with exceeded quota

bogdanlupuro:
Add quota transaction abort handing when it is exceeded. #7025


bogdanlupuro:
Yeah, I was thinking of to do an effective CI test here, but nothing came to mind. BTW: do you have any idea when a new version with the fix is planned for release?

Issue #7019, opened 2018-01-17
Crash after have empty response in some xhr request in pouchDB


artemmorozov7:
## Issue
I have many crashes in app after some actions with pouchDB. I have problems to reproduce, but I have stack trace: 
errorStack":"TypeError: Cannot create property 'status' on string ''\n    at XMLHttpRequest.n.onreadystatechange (file:..../libs/pouchdb-6.1.2.release.js:9:20973"

I looked to stack trace, and can found place where I have exception:
https://github.com/pouchdb/pouchdb/blob/07dc369113d495d43f3ed415ba31f50cb734e283/packages/node_modules/pouchdb-ajax/src/request-browser.js#L221
I figured out what the problem was and realized that sometimes the response in this request(xhr) comes in the form of an empty string (""). In this case, JSON.parse does not work correctly and an error occurs.

Please, add check to this situation in line - 
https://github.com/pouchdb/pouchdb/blob/07dc369113d495d43f3ed415ba31f50cb734e283/packages/node_modules/pouchdb-ajax/src/request-browser.js#L216

**`} else if (typeof xhr.response === 'string' && xhr.response !== '') {`**
of some another check for empty string

## Info
- Environment: hybrid app
- Platform: Android
- Adapter: WebSQL

## Reproduce
 I have problems to reproduce, but I have stack trace: 
errorStack":"TypeError: Cannot create property 'status' on string ''\n    at XMLHttpRequest.n.onreadystatechange (file:..../libs/pouchdb-6.1.2.release.js:9:20973"

daleharvey:
Looks like you have done the investigation here and happy to accept a PR, there is a release coming soon and if the PR doesnt make it then this will be fixed by the switch to fetch anyway. If you have the time a test for this would be very helpful. Happy to help with it

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/42ca86e17fea182aaf3695484a83595fb68a718d

PR #7033, opened 2018-01-22
7019 - crash empty xhr response

artemmorozov7:
#7019 - Crash after have empty response in some xhr request in pouchDB

### DIFF
--- a/packages/node_modules/pouchdb-ajax/src/request-browser.js
+++ b/packages/node_modules/pouchdb-ajax/src/request-browser.js
@@ -214,15 +214,15 @@ function xhRequest(options, callback) {
         err = new Error('ETIMEDOUT');
         err.code = 'ETIMEDOUT';
       } else if (typeof xhr.response === 'string' && xhr.response !== '') {
-          try {
-              err = JSON.parse(xhr.response);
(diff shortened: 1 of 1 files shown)


packages/node_modules/pouchdb-ajax/src/request-browser.js
```diff
@@ -213,12 +213,14 @@ function xhRequest(options, callback) {
...
         err = new Error('ETIMEDOUT');
         err.code = 'ETIMEDOUT';
-      } else if (typeof xhr.response === 'string') {
+      } else if (typeof xhr.response === 'string' && xhr.response !== '') {
         try {
           err = JSON.parse(xhr.response);
         } catch (e) {}
       }
-      err.status = xhr.status;
+      if (typeof err === 'object') {
```
daleharvey: Since this isnt executed as part of the test suite its now failing coverage, usually we would require a test that covers this however since this code is going to be removed on the next release and the code change is fairly safe I think its ok to add an istanbul ignore in there

artemmorozov7: Ok, thanks, I will delete these changes and recreate pr.

Issue #6999, opened 2018-01-10
6.4.2 Release


daleharvey:
So for 7.0 we will be making a few breaking changes with the biggest being dropping webSQL, now we have merged the cod eto use indexeddb in safari (https://github.com/pouchdb/pouchdb/commit/49bdffd1bc2a6ec41d7d7c8c90782bf22bc73adc) we need to do a release that lets people who have safari users to migrate towards indexeddb, once they have done that they can carry on using the 7.X release

So need to write a little migration example code and and ensure it works correctly for the blog post

daleharvey:
So drafted up some example code for how to do a migration, anyone want to chime in on changes I should make to this

```
function openDB(name, opts) { 
  return new Promise(function(resolve, reject) { 
    var localdb = new PouchDB(name, opts);
    localdb.info().then(function(info) { 
      
      if (info.adapter !== 'websql') { 
        return resolve(localdb);
      }
      
      var newopts = opts || {};
      newopts.adapter = 'idb';
      
      var newdb = new PouchDB(name, opts);
      var replicate = localdb.replicate.to(newdb);
      replicate.then(function() { 
        resolve(newdb);
      }).catch(reject);
    }).catch(reject);
  });
}
```

daleharvey:
Released, cheers all -  https://github.com/pouchdb/pouchdb/commit/2d6577ce0dde4c695ea79b2ba62b1228934feaf6

PR #7037, opened 2018-01-23
(#6999) - 6.4.2 Release post and add deprecation warning

daleharvey:
It would be great if I could get eyes on this, would like to make sure its right


marten-de-vries:
Looks good to me. Nice work!

PR #7040, opened 2018-01-25
Use fixed-length array in idb/indexeddb allDocs

alxndrsn:
I've been benchmarking `db.allDocs({ keys: someIds })`, and this PR may have a significant effect for some users.

Initialising arrays with a known length seems to be much faster in Chrome than using `[]`.  In Firefox, there seems to be little difference.

https://jsperf.com/random-array-inserts

Running the same benchmark directly in the Chrome console* seems to have even more marked differences, but presumably there's some other factor at play:

screen shot 2018-01-25 at 14 27 55

N.B. `COUNT = 20000`

*application which includes PouchDB was also loaded in this browser window.

### DIFF
--- a/packages/node_modules/pouchdb-adapter-idb/src/allDocs.js
+++ b/packages/node_modules/pouchdb-adapter-idb/src/allDocs.js
@@ -19,7 +19,7 @@ import getAll from './getAll';
 
 function allDocsKeys(keys, docStore, onBatch) {
   // It's not guaranted to be returned in right order  
-  var valuesBatch = [];
+  var valuesBatch = new Array(keys.length);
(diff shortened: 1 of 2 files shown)


daleharvey:
So there has been efforts to try and get our performance tests ready for production, fail CI if we get performance regressions and a dashboard of what our performance looks like. If you are interested in doing that would be happy to discuss in a new issue.

As for this improvment its a simple change that doesnt complicate the code so happy to merge, thank you :)

alxndrsn:
Thanks for merging!

Ref perf testing, I'm very interested.  Please keep @SCdF in the loop too.

Issue #7020, opened 2018-01-17
Global DB failure PouchDB/IndexedDB caused by different ID


bogdanlupuro:
### Issue
We encountered on a user database a global database failure making any read/write of data impossible.  No syncing is possible or allDocs or anything, It goes to the idb error handler and closes the connection. No way of recovery from what I've seen.

It tries to get a property '_attachments' from a doc to sync it to the server. But when it tries to get it, the document is undefined and results in a error and consequently in a global DB error and close. When trying to do an allDocs also results in a error. 

Affected code area is:
```javascript
function fetchDocAsynchronously(metadata, row, winningRev$$1) {
    var key = metadata.id + "::" + winningRev$$1;
    docIdRevIndex.get(key).onsuccess = function onGetDoc(e) {
        row.doc = decodeDoc(e.target.result);
        if (opts.conflicts) {
            var conflicts = collectConflicts(metadata);
            if (conflicts.length) {
                row.doc._conflicts = conflicts;
            }
        }
        fetchAttachmentsIfNecessary(row.doc, opts, txn);
    };
}
function fetchAttachmentsIfNecessary(doc, opts, txn, cb) {
    var attachments = Object.keys(doc._attachments || {});
    if (!attachments.length) {
        return cb && cb();
    }
....
```

doc is undefined, so doc._attachments gives the above error. On close inspection this happens because a document stored in the DB: lm_note_3b548873-bc4e-4b4b-240b-7042c059a4bd::1-83e5ef9a59e34693a2fb8faf2f27f75b When it tries and get the document id does not find also the containing document so it only returns:

```json
{
  "id": "lm_note_3b548873-bc4e-4b4b-240b-7042c059a4bd",
  "key": "lm_note_3b548873-bc4e-4b4b-240b-7042c059a4bd",
  "value": {
    "rev": "1-83e5ef9a59e34693a2fb8faf2f27f75b"
  },
  "doc": undefined
}
```

On a even closer inspection I noticed that in the 3 different places in which the document is stored in IndexededDB by PouchDB it has a slighty different ID. Check pictures:
lm_note_3b548873-**bc4e**-4b4b-240b-7042c059a4bd - deletedOrLocal
lm_note_3b548873-**bc4**-4b4b-240b-7042c059a4bd::1-83e5ef9a59e34693a2fb8faf2f27f75b - _doc_id_rev
[...]

PR #7051, opened 2018-01-27
allDocs safely in case docs get corrupted, not resulting in complete failure

bogdanlupuro:
This is the fix for #7020 and solves the problem in case for some reason there is a mismatch between document store ID's and getting the document from the docIdRevIndex results in a empty result which causes a global error, failing the DB
This fix should at least get all the other documents, not resulting in a complete loss and can keep the conflicts which could be used to get older version in case they can be used.
This would be impossible to test as the only scenario (I have) this would happen is when IDB does not create correctly the index.


daleharvey:
So I would prefer to avoid this type of code, errors should be handled appropriately but there isnt any way to appropriately handle the underlying storage breaking in this way so happy to merge so that at least its effect isnt fatal, thanks

Issue #6655, opened 2017-07-30
Replication events are inconsistent with their handling of `start_time` (and likely `end_time`) fields


flimzy:
### Issue
Replication inconsistently uses `Date` objects and `strings` for `start_time` field.

### Info
- Environment: Chrome, node
- Adapter: any, apparently
- Server: CouchDB 1.6.1

### Reproduce

Run this simple script via node (replacing the remote url with any readable, remote database):

    var pouchdb = require('pouchdb');
    pouchdb.replicate("http://admin:abc123@localhost:5984/user-1fbf880e-e057-41a4-b56d-6dd974c1c30d", "foo")
        .on('change', function(i) {
            console.log("change: " + i.start_time + " : " + typeof(i.start_time));
        })
        .on('complete', function(i) {
            console.log("complete: " + i.start_time + " : " + typeof(i.start_time));
        })

Observe the following output:

    change: 2017-07-30T14:31:27.867Z : string
    complete: Sun Jul 30 2017 16:31:27 GMT+0200 (CEST) : object

I would expect the `start_time` field to consistently be either a string, or a date object (ideally, a date object).

flimzy:
I would love to hear from the project maintainers/regular contributors if this is intended behavior, or if it's worth trying to fix.  I imagine it's a low-priority issue, but it would probably also be relatively easy for someone familiar with the codebase.

daleharvey:
Yup this isnt intended behaviour, these should be consistent. If we get any of these fields from couchdb as a string then we should match couch, if we always generated these they should probably be Dates

daleharvey:
Weird, we only use `new Date()` in replication code so they should all be objects  however https://ginger-purpose.glitch.me/ shows this bug still happens

daleharvey:
ok this is because the change objects get cloned and during clone they lose their types, easiest just to switch to strings here

daleharvey:
@flimzy want to 👍 the PR ? then can get this merged, cheers

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/6b79dbbcb887decdd1c49ecfe6c42eb468b10faa

PR #7052, opened 2018-01-28
(#6655) - Use consistent types for replication progress values

daleharvey:


flimzy:
Looks reasonable to me. :+1:

PR #7060, opened 2018-01-31
(#6944) - Remove unneeded response modification

daleharvey:


daleharvey:
Not 100% sure this isn't needed, see what the tests say. Would like to do a bunch of cleanup on the http adapter while I am working on it and we already have a breaking release

daleharvey:
Timebombing this simple PR, but anyone following along even if you arent a comitter if you could take a quick look at these it would be great, cheers

marten-de-vries:
I just looked through git blame - it seems to have been added years ago for sync gateway. It makes sense to clean this up while doing the ``http`` adapter rewrite. If it's still a problem, we'll probably hear through a bug report in due time.

Issue #6946, opened 2017-12-17
PouchDB 7.0


daleharvey:
Opening a new issues as https://github.com/pouchdb/pouchdb/issues/6356 has become pretty stale, we have a few upcoming breaking changes I would like to make, so lets get them into a single release to make things easier for all.

If there are things you would like to copy over from the old issue or other breaking changes then this is a good time to mention them

### TODOs

- [x] Deprecate WebSQL - https://github.com/pouchdb/pouchdb/issues/6943
- [x] Remove Promise polyfill - https://github.com/pouchdb/pouchdb/issues/6945
- [x] Switch to fetch api - https://github.com/pouchdb/pouchdb/issues/6944
- [x] allDocs broken on iOS - https://github.com/pouchdb/pouchdb/issues/7210
- [x] iOS broken - https://github.com/pouchdb/pouchdb/issues/7057
- [x] Chrome broken  - https://github.com/pouchdb/pouchdb/issues/7098
- [x] iPhone tests broken - https://github.com/pouchdb/pouchdb/issues/7187
- [x] Edge / ie isnt passing in travis - https://github.com/pouchdb/pouchdb/issues/7047

marcoturi:
Hi, and thank you for this library @nolanlawson and @daleharvey. I'm using PouchDB in ionic applications. I have a question about the deprecation of WebSQL. At the moment indexDB is not supported in iOS envirorments: if you set the adapter to idb and use chrome to simulate an iOS device, pouchDB will not work (no adapter found). There are already issue opened:
[#6741](https://github.com/pouchdb/pouchdb/issues/6741)
[#7057](https://github.com/pouchdb/pouchdb/issues/7057)
[#6701](https://github.com/pouchdb/pouchdb/issues/6701)

And stackoverflow questions without solutions, for example [here](https://stackoverflow.com/questions/41937630/pouchdb-use-indexeddb-instead-of-websql-in-chrome-not-working-anymore).

What is the correct upgrade path that you suggest for iOS?

daleharvey:
Indexeddb is supported in ios environments (https://caniuse.com/#search=indexeddb) and is passing our test suite in full on every commit to pouchdb (https://travis-ci.org/pouchdb/pouchdb/builds/336891564) 

I havent had time to look into what exactly the bug @ #7057 is, but I would suggest helping debug / fix that issue

marcoturi:
[...]

PR #7086, opened 2018-02-15
 (#7085) - Accept all IDB versions as valid 

### LINKED ISSUE #7085: fetch polyfills break `valid` function for Safari after 10.1
### Issue

Following [this comment](https://github.com/pouchdb/pouchdb/issues/6946#issuecomment-362921474), I looked a bit deeper why pouchdb was telling that I could not use idb in Safari 11 (while it should be available).

The culprit seems to be [this line](https://github.com/pouchdb/pouchdb/blob/master/packages/node_modules/pouchdb-adapter-idb/src/index.js#L814). Safari on a page with a polyfill that replaces `fetch` is detected as older than 10.1. 

So it seems that we need to find a better way to detect version of Safari after 10.1...

sgenoud:
As described in #7085, the fact that people tend to polyfill `fetch` a lot makes some trouble with our Safari version detection.

As the Gamepad interface appeared [in the same release](https://webkit.org/blog/7477/new-web-features-in-safari-10-1), and it is a less commonly polyfilled API (I do not have hard numbers, but I think it is a safe bet), let's use it.

Unfortunately I do not have access to a machine with an older version of Safari and could not test that the test fails there...

Closes #7085

### DIFF
--- a/packages/node_modules/pouchdb-adapter-idb/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-idb/src/index.js
@@ -800,26 +800,15 @@ function init(api, opts, callback) {
 }
 
 IdbPouch.valid = function () {
-  // Issue #2533, we finally gave up on doing bug
-  // detection instead of browser sniffing. Safari brought us
-  // to our knees.
-  var isSafari = typeof openDatabase !== 'undefined' &&
-    /(Safari|iPhone|iPad|iPod)/.test(navigator.userAgent) &&
-    !/Chrome/.test(navigator.userAgent) &&
-    !/BlackBerry/.test(navigator.platform);
-
-  // Safari <10.1 does not meet our requirements for IDB support (#5572)
-  // since Safari 10.1 shipped with fetch, we can use that to detect it
-  var hasFetch = typeof fetch === 'function' &&
-    fetch.toString().indexOf('[native code') !== -1;
+  // Following #7085 buggy idb versions (typically Safari < 10.1) are
+  // considered valid.
 
   // On Firefox SecurityError is thrown while referencing indexedDB if cookies
   // are not allowed. `typeof indexedDB` also triggers the error.
   try {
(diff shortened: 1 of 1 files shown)


daleharvey:
I was a little wary of using one feature to detect another to start with, now that it's hit problems I think we should probably not try it again and either use an explicit version check, a test for the actual bug in the implementation or drop old support

sgenoud:
> a test for the actual bug in the implementation 

It seems that the current method was a way to [admit defeat ](https://github.com/pouchdb/pouchdb/blob/master/packages/node_modules/pouchdb-adapter-idb/src/index.js#L803) against this type of approach

> use an explicit version check

How would you do this (I always thought that this was nearly impossible to do correctly)?

> drop old support

This is probably the way forward to me. But you probably would like to only drop support in the 7.0.0. I would propose to keep my fix for 6.x.x and drop support for 7.0.0.
Normally the project has always aimed at being as backward compatible as possible. Dropping support would go against this philosophy. But on the other hand it would make the support load lighter...

daleharvey:
our next release is 7.0 and Safari <10.1 is quickly dropping down to 0 usage so I think we can drop the check here, we have always been backwards compatible with pouch versions (and still are) but pretty agressive with killing support for outdated browsers

sgenoud:
I have rebased my commit – removing the test all together.

jpike88:
This is a royal PITA, took me down a rabbit hole to isolate this one. Can we have a release at some point?

daleharvey:
https://github.com/pouchdb/pouchdb/issues/6946 is the release issue, any help welcome

jpike88:
Is it possible to back-fix this one and release a minor fix for now? It's a pretty important bug fix imo

jpike88:
https://github.com/rollbar/rollbar.js/issues/616

I think this one is pretty serious, who knows what other libraries out there are polyfilling fetch. Would recommend a minor update with this.

PR #7041, opened 2018-01-25
WIP - Use edge in CI

daleharvey:

PR #7104, opened 2018-02-21
(#6944) - Switch http adapter to use fetch

daleharvey:


daleharvey:
So this is fully passing now, I added IE to allowed failures but also need to work on ie for https://github.com/pouchdb/pouchdb/pull/7041 so it will be fixed before release but didnt want to block this large patch

packages/node_modules/pouchdb-core/src/setup.js
```diff
@@ -122,4 +123,8 @@ PouchDB.defaults = function (defaultOpts) {
   return PouchAlt;
 };
 
+PouchDB.fetch = function (url, opts) {
```
marten-de-vries: Could this just be ``PouchDB.fetch = fetch``? Both the browser and ``fetch-cookie`` at least currently don't export more than a plain function.

tests/component/test.ajax.js
```diff
@@ -16,23 +16,14 @@ describe('test.ajax.js', function () {
...
-          method: 'GET',
-          url: 'http://127.0.0.1:6000/check-cookie',
-          timeout: 10
-        }, function (err, res) {
-          server.close();
-
-          should.equal(res.ok, true, "Cookie not set");
-          done();
-        });
+      PouchDB.fetch('http://127.0.0.1:6000/install-cookie').then(function () {
```
marten-de-vries: ``should.equal(res.ok, true, "Server not responding");`` is missing here. Not sure if important.

daleharvey: Those status checks are a bit more useful for callbacks to ensure you error out early, but for promises we would error out here if the server doesnt respond so happy to skip it

packages/node_modules/pouchdb-core/src/setup.js
daleharvey: Yeh changed that cheers

packages/node_modules/pouchdb-fetch/src/fetch.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+import nodeFetch from 'node-fetch';
+import fetchCookie from 'fetch-cookie';
```
ptitjes: Shouldn't that be `'fetch-cookie/node-fetch'` to support redirects ?
cf. [fetch-cookie: cookies and redirection](https://github.com/valeriangalliat/fetch-cookie#cookies-and-redirection)

tests/component/test.ajax.js
marten-de-vries: Ok, just checking. :)

packages/node_modules/pouchdb-find/src/adapters/http/index.js
```diff
@@ -1,35 +1,50 @@
...
+    return response.json();
+  }).then(function (json) {
+    if (!ok) {
+      json.status = status;
+      var err = generateErrorFromResponse(json);
+      callback(err);
+    } else {
+      callback(null, json);
+    }
+  });
```
sgenoud: Same as before, if there is a network error or a json parsing error the callback will not be called.

packages/node_modules/pouchdb-adapter-http/src/index.js
```diff
@@ -201,6 +184,54 @@ function HttpPouch(opts, callback) {
...
+          }
+        });
+      }
+
+      if (callback) {
+        callback(null, result.data);
+      } else {
+        return result;
+      }
+    });
```
sgenoud: Not sure it is relevant, but if there is a network issue or a json parsing error the callback will not be called – a `.catch(...callback...)` seems a good idea

packages/node_modules/pouchdb-fetch/src/fetch.js
daleharvey: I think we can take a look at that in a follow up, would be curious as to why that isnt the default

packages/node_modules/pouchdb-adapter-http/src/index.js
daleharvey: The callers of this are responsible for calling if the whole method throws (they all have `.catch(callback)`)

packages/node_modules/pouchdb-find/src/adapters/http/index.js
daleharvey: This one however the .catch is needed I think, thanks

package.json
```diff
@@ -34,11 +34,13 @@
     "test-webpack": "bash bin/test-webpack.sh"
   },
   "dependencies": {
+    "add-cors-to-couchdb": "0.0.6",
```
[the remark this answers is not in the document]
daleharvey: I kinda left it in but yeh I should take it out, cheers

packages/node_modules/pouchdb-fetch/src/fetch.js
ptitjes: @daleharvey Well, they explain, on the link that I referenced, that they have to rely on APIs that only `node-fetch` provides. More precisely:
> the generic fetch API does not allow any way to hook into redirects

daleharvey: Ah ok, since this is green gonna marge as is and if you wanna file and PR that as a follow up it would be awesome, thanks

daleharvey:
Ok cheers for taking a look, this has had a bunch of eyes and is a big patch so gonna get it merged and we can start letting it settle, thanks all

PR #7110, opened 2018-02-23
(#6946) - Remove debug dep

daleharvey:
Would like to move pouchdb-debug to pouchdb-community, but this is a first step


daleharvey:
timebombing uncontreversial PR (was already proposed by nolan)

Issue #6809, opened 2017-10-24
doc_ids and batch_size bug during replication


mikeymckay:
### Issue
If the number of docs in doc_ids is larger than batch size, then not all docs will be replicated

### Info
- Environment: browser
- Platform: Chrome
- Adapter: default, replicating with http
- Server: CouchDB

### Reproduce
cloudDB.replicate.to database,
  doc_ids: [
    "user.mmckay"
    "user.hmkali"
    "user.mtash"
  ]
  batch_size: 1
.on "complete", (r) ->
  console.log r

This results in:
Object {ok: true, start_time: Tue Oct 24 2017 11:26:58 GMT+0300 (EAT), docs_read: 1, docs_written: 1, doc_write_failures: 0…}

Change the batch_size to 3 and you get:
Object {ok: true, start_time: Tue Oct 24 2017 11:27:29 GMT+0300 (EAT), docs_read: 3, docs_written: 3, doc_write_failures: 0…}

jjal:
This is painful for us at the moment, don't know if anyone has time for a fix

PR #7114, opened 2018-02-26
(#6809) - Add test to ensure doc_ids dont prevent one-shot replication

tlvince:
Seems like when both `batch_size` and `doc_ids` are specified and the batch size is smaller than the doc IDs length, replication stops after the batch size.

### DIFF
--- a/tests/integration/test.replication.js
+++ b/tests/integration/test.replication.js
@@ -4263,6 +4263,25 @@ adapters.forEach(function (adapters) {
       });
     });
 
+    it('#6809 doc_ids dont prevent one-shot replication', function () {
+
+      var db = new PouchDB(dbs.name);
+      var remote = new PouchDB(dbs.remote);
+
+      var writes = [];
+      for (var i = 0; i < 20; i++) {
+        writes.push(remote.put({_id: i + ''}));
+      }
+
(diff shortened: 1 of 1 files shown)


tlvince:
Seems like tests are timing out. You can see if fail on my fork [here](https://travis-ci.org/tlvince/pouchdb/jobs/346335857#L4117).

tlvince:
I believe the source of the issue is around `changesCompleted`/`completeReplication`:

https://github.com/pouchdb/pouchdb/blob/014a34dd7c40379987bd88a30c9533ec932ec8db/packages/node_modules/pouchdb-replication/src/replicate.js#L207-L213

It looks like an empty batch is used to determine when replication has finished. Looks like additional handling is needed for `doc_ids`?

daleharvey:
I vaguely remember this being reported previously and not getting fixed, usually making the test is the hardest part and yeh looks like you have found where the fix is needed, thanks

daleharvey:
If you have any other questions about the fix please feel free to ping me

daleharvey:
Merged along with the fix, thanks so much :) https://github.com/pouchdb/pouchdb/commit/9cfad1d0afb6c136793c815390fd4475bd4b067a

PR #7115, opened 2018-02-26
remove catastrophic backtracking vulnerability

davisjam:
Problem:
A regex to parse function names was vulnerable to catastrophic
backtracking.

Solution:
Alter the regex to make it safe.
The new regex matches almost the same language.
The only differences should not be a problem in the input space.

This regex is not exploitable for REDOS as currently used.
This change is for future-proofing.

### DIFF
--- a/packages/node_modules/pouchdb-utils/src/functionName.js
+++ b/packages/node_modules/pouchdb-utils/src/functionName.js
@@ -15,7 +15,13 @@ if (hasName) {
   };
 } else {
   res = function (fun) {
-    return fun.toString().match(/^\s*function\s*(\S*)\s*\(/)[1];
+    var match = fun.toString().match(/^\s*function\s*(?:(\S+)\s*)?\(/);
+    if (match && match[1]) {
+      return match[1];
+    }
+    else {
+      return '';
+    }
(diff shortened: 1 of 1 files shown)


davisjam:
One point of confusion.

Railroad diagram: [old](https://regexper.com/#%2F%5E%5Cs*function%5Cs*%28%5CS*%29%5Cs*%5C%28%2F)
Railraod diagram: [new](https://regexper.com/#%2F%5E%5Cs*function%5Cs*%28%5C%28%5B%5E%5C%28%5C%29%5Cs%5D*%5C%29%29%5Cs*%5C%28%2F)

Should the final character in the regex be an opening brace '{' rather than an opening paren '('?

daleharvey:
It's pulling the function name out of function hello() { so ( seems correct, this is failing our lint rules about regex escapes however

davisjam:
@daleharvey 

1. Fixed the linting problem.
2. Thanks for clarifying the input format.

daleharvey:
We dont have unit tests for all functions and most of the time depend on the integration tests to ensureall is good which they are here so happy to merge, thanks so much :)

PR #7136, opened 2018-03-08
Deterministic revision numbers with md5

### LINKED ISSUE #4642: Make _rev generation deterministic (similar to CouchDB)
PouchDB currently generates _rev values using a uuid. This presents problems when, for instance, an application has conflict resolution logic which might run on different devices concurrently (because they would both resolve the conflict but, in doing so, generate a new one).

CouchDB has an optimisation here where the _rev is deterministic, based on the contents of the document, attachments, deleted flag and rev history (see http://stackoverflow.com/questions/5954864/how-does-couchdb-calculate-the-revision-number).

It would be useful for PouchDB to have a similarly deterministic rev generation, though it's not really necessary to match the CouchDB one.

garrensmith:
This uses md5 to create the the revision number for a document. This
means that each document will have a deterministic revision nubmer.
This makes client side conflict resolution easier and will reduce the
number of conflicts

By default this is turned on, but a user can fallback to generating
random revision numbers by passing `useMd5Rev: false' as an option to a
new PouchDB instance

Fixes #4642

### DIFF
--- a/packages/node_modules/pouchdb-adapter-utils/src/parseDoc.js
+++ b/packages/node_modules/pouchdb-adapter-utils/src/parseDoc.js
@@ -76,7 +76,12 @@ function makeRevTreeFromRevisions(revisions, opts) {
 
 // Preprocess documents, parse their revisions, assign an id and a
 // revision for new writes that are missing them, etc
-function parseDoc(doc, newEdits) {
+function parseDoc(doc, newEdits, dbOpts) {
+  if (!dbOpts) {
+    dbOpts = {
+      useMd5Rev: true
+    };
+  }
 
   var nRevNum;
   var newRevId;
@@ -90,7 +95,7 @@ function parseDoc(doc, newEdits) {
     if (!doc._id) {
       doc._id = uuid();
     }
-    newRevId = rev();
+    newRevId = rev(doc, dbOpts.useMd5Rev);
     if (doc._rev) {
       revInfo = parseRevisionInfo(doc._rev);
       if (revInfo.error) {
--- a/packages/node_modules/pouchdb-adapter-utils/src/parseDoc.js
+++ b/packages/node_modules/pouchdb-adapter-utils/src/parseDoc.js
@@ -79,7 +79,7 @@ function makeRevTreeFromRevisions(revisions, opts) {
 function parseDoc(doc, newEdits, dbOpts) {
   if (!dbOpts) {
     dbOpts = {
-      deterministic_rev: true
+      deterministic_revs: true
     };
   }
 
@@ -95,7 +95,7 @@ function parseDoc(doc, newEdits, dbOpts) {
     if (!doc._id) {
       doc._id = uuid();
     }
-    newRevId = rev(doc, dbOpts.deterministic_rev);
+    newRevId = rev(doc, dbOpts.deterministic_revs);
     if (doc._rev) {
       revInfo = parseRevisionInfo(doc._rev);
       if (revInfo.error) {
--- a/packages/node_modules/pouchdb-adapter-utils/src/parseDoc.js
+++ b/packages/node_modules/pouchdb-adapter-utils/src/parseDoc.js
@@ -79,7 +79,7 @@ function makeRevTreeFromRevisions(revisions, opts) {
 function parseDoc(doc, newEdits, dbOpts) {
   if (!dbOpts) {
     dbOpts = {
-      useMd5Rev: true
+      deterministic_rev: true
     };
   }
 
@@ -95,7 +95,7 @@ function parseDoc(doc, newEdits, dbOpts) {
     if (!doc._id) {
       doc._id = uuid();
     }
-    newRevId = rev(doc, dbOpts.useMd5Rev);
+    newRevId = rev(doc, dbOpts.deterministic_rev);
     if (doc._rev) {
       revInfo = parseRevisionInfo(doc._rev);
       if (revInfo.error) {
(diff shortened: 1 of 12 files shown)


karlwestin:
i have been bitten by this myself quite a few times, and i'm really happy to see this PR

This PR adds `pouchdb-md5` to `pouchdb-core`, which makes it around 2kb bigger uncompressed. PouchDB core is currently ~40kb uncompressed/unminified. It's always a tradeoff but i think this solves a really important and annoying gotcha in the differences between pouch and couch.

tests/integration/test.md5rev.js
```diff
@@ -0,0 +1,119 @@
...
+      dbs.name1 = testUtils.adapterUrl(adapter, 'testdb');
+      dbs.name2 = testUtils.adapterUrl(adapter, 'testdb2');
+    });
+
+    afterEach(function (done) {
+      testUtils.cleanup([dbs.name1, dbs.name2], done);
+    });
+
+    it('useMd5Rev=true so revision for two docs that are the same will be equal', function () {
+      const doc = {
```
daleharvey: We dont use es6 in pouch yet, havent looked at the test failures but imagine this could be a problem so think its best we stick to `var` for the moment

packages/node_modules/pouchdb-adapter-utils/src/parseDoc.js
```diff
@@ -90,7 +95,7 @@ function parseDoc(doc, newEdits) {
     if (!doc._id) {
       doc._id = uuid();
     }
-    newRevId = rev();
+    newRevId = rev(doc, dbOpts.useMd5Rev);
```
daleharvey: A while back we switched from camel case to snake case for our options so for consistency should switch to that, I think `deterministic_rev` is best here, saying we use md5 is a leaky abstraction in case we need to stop using md5, `deteministic` is a bit long / unweidly however I doubt this is going to be use much, more of a precaution in case it causes unexpected problems when released

daleharvey:
`pouchdb-md5` was always in our browser builds though right? its a dep I would like to see if we could reduce its size since I think most of the size comes from doing binary md5's on attachments but if its already in browser builds then think its fine to add to pouchdb-core

karlwestin:
you are right, md5 is already included in the browser builds, i was not looking things up properly

packages/node_modules/pouchdb-adapter-utils/src/updateDoc.js
```diff
@@ -10,7 +10,7 @@ import {
 function updateDoc(revLimit, prev, docInfo, results,
                    i, cb, writeDoc, newEdits) {
 
-  if (revExists(prev.rev_tree, docInfo.metadata.rev)) {
+  if (revExists(prev.rev_tree, docInfo.metadata.rev) && newEdits === false) {
```
daleharvey: if (revExists() && !newEdits)

packages/node_modules/pouchdb-adapter-utils/src/parseDoc.js
```diff
@@ -76,7 +76,12 @@ function makeRevTreeFromRevisions(revisions, opts) {
 
 // Preprocess documents, parse their revisions, assign an id and a
 // revision for new writes that are missing them, etc
-function parseDoc(doc, newEdits) {
+function parseDoc(doc, newEdits, dbOpts) {
+  if (!dbOpts) {
+    dbOpts = {
+      deterministic_rev: true
```
daleharvey: Sorry if I got it wrong in previous comment, I think it should be `determinstic_revs` here

packages/node_modules/pouchdb-utils/src/rev.js
```diff
@@ -1,7 +1,15 @@
 import uuid from 'uuid';
+import { stringMd5 } from 'pouchdb-md5';
+import { clone } from 'pouchdb-utils';
 
-function rev() {
-  return uuid.v4().replace(/-/g, '').toLowerCase();
+function rev(doc, deterministic_rev) {
+  var clonedDoc = clone(doc);
```
daleharvey: So I really want to reduce thte amount of cloning we do, plus this is quite fragile (there are other private fields that get set), we make the metadata vs data split futher down in parseDoc, it would be great to refactor so we could just md5(result.data) here instead of cloning

daleharvey: We may want to punt on that for its patch though if it ends up too complicated

garrensmith: Ok cool. Ideally we want to md5 the previous revision as well. I'll take a look into that.

garrensmith:
The idb-next tests are failing. So I will work on those tomorrow plus the changes you need.

rnewson:
I suggest sha-256 instead of md5. md5 is fine for the purpose but sha-256 is really the base algorithm for any hashing duty beyond the trivial, it's often a cpu instruction in its own right so might well be faster.

garrensmith:
sha-256 is supported natively in most [browsers](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest). So if we moved our attachment sha to that as well we could reduce the size of pouchdb be removing pouchdb-md5. We could also hopefully get CouchDb to move to sha-256 which means we might be able to get deterministic revs across couchdb and pouchdb

packages/node_modules/pouchdb-utils/src/rev.js
garrensmith: At this point, from what I can see. The doc normally looks something like this:

```js
{
    _id: '12342349b5904acaa7161f12a0fb31b59c'
   _rev: '1-456149b5904acaa7161f12a0fb31b59c'
   ...  other doc contents...
  __revtree: []
}
```
So for this pr I would like to do the clone and remove the _rev_tree. Then look at another PR that does the rev of the result.data

daleharvey:
Will open a new isue to switch to sha-256, no need to block so we can get this landed and see if there are other teething issues, but sha-256 is gonna need another refactor since its async so can do it seperately

daleharvey:
Thanks Garren :)

garrensmith:
Awesome. Thanks for the help.

PR #7141, opened 2018-03-10
#7140 ios blob support silent fail

### LINKED ISSUE #7140: iOS db.info() silent fail on indexeddb
### Issue
iOS db.info() silent fail on indexeddb 

### Info
- Environment: browser/hybrid app
- Platform: iOS 11.2.5 and 11.2.6
- Adapter: IndexedDB
- Server: -

### Reproduce
In our app at startup we have a DB migration process, which first checks an old DB exists and replicates it 
to the new DB. For this we fist do a open the old DB with default adapter and do a db.info and check doc_count. 
Until PouchDB 6.4.2 everything was fine (WebSQL adapter), but since the new version which has IDB as default adapter it hangs in about 50% of the time. Non deterministic. 

We tracked the problem to the `checkBlobSupport(txn)` where the put request fails sometimes (after refresh works, then again does not):
`var req = txn.objectStore(DETECT_BLOB_SUPPORT_STORE).put(blob$$1, 'key');`

The problem that that returns an error but does not go `txn.onabort`.

And because of that error `blobSupportPromise` is not finished and neither `completeSetup`.

I am preparing a fix to catch the error and not silent fail.

bogdanlupuro:
This PR fixes #7140 in which checkBlobSupport fails silently on iOS 11.2 idexeddb hanging db.info()


bogdanlupuro:
I do not think the failing tests are from my changes.

daleharvey:
Yup we have some intermittent failures in our test runs since the suite is pretty large, all green after a restart though this looks good, thanks

PR #7155, opened 2018-03-14
fix: #6809 use _changes results last seq for since param

kdoran:
Closes: https://github.com/pouchdb/pouchdb/issues/6809

Requests to CouchDB 1.6 and 2.X `/db/_changes` with a `_doc_ids` filter does not return the `last_seq` number of the filtered changes as expected. In PouchDB's `replicate.js`, that `last_seq` is used as the `since=` argument on the next call to changes, which means documents are ocassionally missed when replicating on IDs. 

Related: https://github.com/apache/couchdb/issues/1221
Related: https://github.com/apache/couchdb/issues/748

Instead of using `last_seq`, this PR proposes using the `seq` number of the last change in the results array, ensuring no changes are missed. The last result in the array should be the latest -- we don't supply a `descending` param and its default in CouchDB is false:

[from the CouchDB 2.1 _changes doc:](http://docs.couchdb.org/en/2.1.1/api/database/changes.html#post--db-_changes
)
> descending (boolean) – Return the change results in descending sequence order (most recent change first). Default is false. 

Unfortunately, this would mean having to remove `seq_interval` as an available parameter.


daleharvey:
I split out the commits to the original test PR and this one so you boh get listed as authors, thanks

kdoran:
@daleharvey 👍 thanks for the fast look at this!

willholley:
I don't fully understand this fix. `last_seq` should always contain a correct `since` token to continue `_changes` - is the suggestion that there's a bug in CouchDB 2.x when using the `doc_ids` filter that it does not?

wohali:
@willholley Yeah, see https://github.com/apache/couchdb/issues/1221 .

Issue #7135, opened 2018-03-08
Remove websql


garrensmith:
### Issue
Safari 10.1 onwards has a good implementation of IndexedDB, we no longer need to support the fruitdown adapter. Nolan has also stopped maintaining [fruitdown](https://github.com/nolanlawson/fruitdown)

I propose we either just delete the fruitdown adapter or move it to pouchdb-community. I think its nice to try and keep the pouchdb repo to just the packages that are maintained.

### Info
- Environment: Browser
- Platform: Browser
- Adapter: Fruitdown

daleharvey:
I agree although I would like to see https://github.com/pouchdb/pouchdb/issues/7057 fixed and the dust settle on the 7.0 release before we get into a position where we need to do a lot of work to reenable ios if idb is still not actually usable

daleharvey:
To make a start on this however, removing phantom to clean some things up and save precious travis cycles

daleharvey:
pingbot

PR #7156, opened 2018-03-14
(#7135) - Remove testing in phantom

daleharvey:


daleharvey:
Yup good call, cheers

Issue #7160, opened 2018-03-15
Couch Master should pass tests by default


daleharvey:
Right now we have a bunch of isCouchMaster tests which will fail if you dont remember to do `SERVER=couch-master` when testing, thats super fragile and confusing so lets make them pass by default

daleharvey:
So my plan for this, I am going to remove the tests that dont pass on master since that is our baseline and skip tests that pass on master but fail on 1.X (or pouchdb-server), get a baseline of tests that pass on all platforms and then additional tests that only work on particular platforms (master) can be enabled (not disabled) by passing additional flags

That means we dont limit ourselves to what we can test, but out of the box people can expect our test suite to pass whatever server they use

PR #7161, opened 2018-03-15
(#7160) - Remove tests not valid for 2.0

daleharvey:


daleharvey:
See https://github.com/pouchdb/pouchdb/issues/7160#issuecomment-373660716 for plan / motivation, this removes a bunch of tests that were failing in 2.0 (mostly testing assumptions that no longer hold) but also reenables mot of the compaction tests which is nice

Issue #7120, opened 2018-03-04
process.nextTick() in shipped browser code for pouchdb-find


mkubilayk:
### Issue
Both `pouchdb-find/lib/index-browser.es.js` and `pouchdb-find/lib/index-browser.js` seem to have `process.nextTick` which is not available in the browsers. 

It looks like other uses of `process.nextTick()` were removed in #5898. This probably precedes `pouchdb-find`. 

### Info
- Environment: browser
- Platform: Chrome
- Adapter: IndexedDB
- Server: Cloudant

### Reproduce
```
yarn add pouchdb-find@6.4.3
```

See `node_modules/pouchdb-find/lib/index-browser.es.js#115`.

daleharvey:
I thought we would expect browserify to polyfill that, we aren't seeing test failures but still likely best to replace vwith settimeout

mkubilayk:
Thanks!

PR #7175, opened 2018-03-19
(#7120) - Dont use nextTick in browser builds, and switch to setTimeout

daleharvey:


daleharvey:
Removes the immediate dep which we originally had for free since we poyfilled promise, but now this gets rid of 2KB from the build size

Issue #7180, opened 2018-03-20
Remove isChromeApp


daleharvey:
Chrome apps are deprecated

Issue #7179, opened 2018-03-20
Remove isCordova


daleharvey:
We dont actually use it anywhere

PR #7181, opened 2018-03-20
(#7180, #7179) - Remove isChromeApp + isCordova 

daleharvey:
(#7181) - Remove workaround for fixed firefox bug


janl:
@daleharvey: I trust you know which versions of FF we support and that the fix is in the baseline at least. Just mentioning it because it wasn’t clear in the commits/issues: +1

daleharvey:
Ah yup, the fix was firefox 46 which is way beyond what we would consider supporting, cheers for the review

Issue #4362, opened 2015-09-23
Option to include (or not) attachment data in replication


cquillen2003:
Image attachments are very large in comparison to their associated CouchDB/PouchDB documents.  A few hundred images in a database can easily exceed the storage capacity of IE and Mobile Safari, so it would be great to have the option to not include attachment data when replicating or syncing.

kxepal:
Option to not include attachments or modify in any way documents during replication breaks the meaning of the word "replication" since target will be not a replica of source, but a product of F(source). And second replication without such option will produce conflicts for all documents which may not be available since they'll have lesser revision.

You can keep attachments and their metadata in separate databases while both documents shares the same id. Metadata is light and replicates fast, while images are big and could be fetched by demand.

cquillen2003:
To clarify, I want the attachment stubs (metadata) to always replicate, but make replicating the attachments themselves (the binary data) optional.  It seems this would align perfectly with the CouchDB/PouchDB API as it relates to attachments because attachments are not included by default when fetching a document, only the attachment stubs are.

Also, not including attachments in a replication seems kinda like a filtered replication - a standardized filtered replication.

daleharvey:
Yup this is a long requested feature, I actually filed this myself almost 2 years ago, will dupe

mmpowers:
I had until now assumed we could do this.  Now I'm going to need it.  Could this be as simple as ignoring the binary during replication if an option is set?  Or perhaps storing a dummy binary value holding the URL to the attachment at the replication source, and fetching the attachment on demand if needed?

nolanlawson:
Yep, you could also do this trick yourself by not using attachments and simply storing the md5sum, then keeping the attachments in a separate server. Incidentally this is how npm  does it, because they ran into scaling issues with CouchDB attachments.

daleharvey:
So the actual option to not fetch the attachment would be fairly simple, it would be an option to control https://github.com/pouchdb/pouchdb/blob/master/lib/replicate/getDocs.js#L25 (may need a few other small changes)

I dont think we should store urls to the remote attachments etc, with `attachments: false` the user will at least see the stubs I believe so knows there are supposed to be attachments there, they can manually pick them up from the remote if they wish

daleharvey:
Another thing would be to ensure that this option became part of the replication id, so if the user did another replication without that option, it would fill in the attachments

mmpowers:
[...]

PR #7185, opened 2018-03-21
Basic WIP idea of attachment:false replication

garrensmith:
This is a very basic idea for implementing #4362 which is to be able to replicate documents that have attachments from one database but not replicate the attachments. Rather just add stubs.

At this stage, it can replicate from a CouchDB to PouchDB and from PouchDB to PouchDB. I don't think it would be possible to do a replication from PouchDB to CouchDB with this option without first modifying CouchDB. 

What I'm not sure how to do is if a second replication happens later with `{attachments: true}` how to get it to replicate the attachments this time. I'm not sure where I would do a check to see if we have the attachments. 

I would like some feedback before proceeding any further on this.

### DIFF
--- a/packages/node_modules/pouchdb-adapter-idb/src/bulkDocs.js
+++ b/packages/node_modules/pouchdb-adapter-idb/src/bulkDocs.js
@@ -172,15 +172,16 @@ function idbBulkDocs(dbOpts, req, opts, api, idb, callback) {
 
     var req = attachStore.get(digest);
     req.onsuccess = function (e) {
-      if (!e.target.result) {
-        var err = createError(MISSING_STUB,
-          'unknown stub attachment with digest ' +
-          digest);
-        err.status = 412;
-        callback(err);
-      } else {
-        callback();
-      }
+      callback();
+      // if (!e.target.result) {
+      //   var err = createError(MISSING_STUB,
+      //     'unknown stub attachment with digest ' +
+      //     digest);
+      //   err.status = 412;
+      //   callback(err);
+      // } else {
+      //   callback();
+      // }
     };
   }
 
--- a/packages/node_modules/pouchdb-adapter-idb/src/bulkDocs.js
+++ b/packages/node_modules/pouchdb-adapter-idb/src/bulkDocs.js
@@ -1,4 +1,3 @@
-import { createError, MISSING_STUB } from 'pouchdb-errors';
 import {
   preprocessAttachments,
   processDocs,
@@ -171,17 +170,10 @@ function idbBulkDocs(dbOpts, req, opts, api, idb, callback) {
   function verifyAttachment(digest, callback) {
 
     var req = attachStore.get(digest);
-    req.onsuccess = function (e) {
+    req.onsuccess = function () {
+      // Due to the option of having attachments: false for replications
+      // We now allow having a doc without the actual attachment
       callback();
-      // if (!e.target.result) {
-      //   var err = createError(MISSING_STUB,
-      //     'unknown stub attachment with digest ' +
-      //     digest);
-      //   err.status = 412;
-      //   callback(err);
-      // } else {
-      //   callback();
-      // }
     };
   }
 
--- a/packages/node_modules/pouchdb-adapter-idb/src/bulkDocs.js
+++ b/packages/node_modules/pouchdb-adapter-idb/src/bulkDocs.js
@@ -1,3 +1,4 @@
+import { createError, MISSING_STUB } from 'pouchdb-errors';
 import {
   preprocessAttachments,
   processDocs,
@@ -39,6 +40,7 @@ function idbBulkDocs(dbOpts, req, opts, api, idb, callback) {
   var docInfoError;
   var metaDoc;
 
+  console.log('dd', docInfos);
   for (var i = 0, len = docInfos.length; i < len; i++) {
     var doc = docInfos[i];
     if (doc._id && isLocalId(doc._id)) {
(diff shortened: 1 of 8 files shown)


broerse:
Nice work!

`stub: true` indicates data not only exist but also reachable. So we need something like `unreacheable: true` showing data exist but is unreacheable. This will ensure data integrity.

It will also solve what already replicated revs to replicate with attachments when you later do a replication with {attachments: true}.

sebastianrothbucher:
I take it you want to treat a stub like an attachment so you can skip the actual attachment from getting transmitted, right? So for all intents and purposes, the system believes there is an attachment and will eventually reply with "haha, not in THIS pouch" when the attachment actually gets queried. I don't see an issue with doing that - as one would have to turn attachments=false on explicitly - which means one can write code that can just handle that.
From a use case perspective: would it not be a ton easier to have a doc w/out attachment (having, say, id: <docid>) and a 2nd document with id: <docid>_attachments? One could add attachments to <docid>_attachments and download them on demand while the normal filter function blocks all ids ending '_attachments'... Would work without a PR or anything just out of the box. 
Anyways: I'd be +0 on this. Maybe I didn't get the urgency of it quite well - but I don't see a problem with it.

packages/node_modules/pouchdb-generate-replication-id/src/index.js
```diff
@@ -35,6 +35,11 @@ function generateReplicationId(src, target, opts) {
   return Promise.all([src.id(), target.id()]).then(function (res) {
     var queryData = res[0] + res[1] + filterFun + filterViewName +
       queryParams + docIds + selector;
+      
+    if (opts.attachments === false) {
+      queryData += 'attachments:false';
```
daleharvey: if ('attachments' in opts) {
      queryData += 'attachments:' + opts.attachments;

We probably arent going to be changing defaults or anything here, but best practise just in case we would

packages/node_modules/pouchdb-adapter-idb/src/index.js
```diff
@@ -379,6 +379,10 @@ function init(api, opts, callback) {
     var type = attachment.content_type;
 
     txn.objectStore(ATTACH_STORE).get(digest).onsuccess = function (e) {
+      if (!e.target.result) {
+        callback(null, null);
+        return;
+      }
```
daleharvey: Will expand more outside a line comment, but I think we need pouch to know that we have a stub without an attachment and to expect that + not try to fetch the attachment, otherwise I think this could hide errors

daleharvey:
Sorry for the delay, busy few weeks. 

So I think this could make for a nice feature, it doesnt seem like it should be too risky but we definitely need a lot of testing, particularly a test to make sure that after an attachments=false replication, the use can then do an attachments=true replication and the data will be filled in

I think we need to add something to the stub to tell pouchdb that we know there is a missing attachment, as said in the comments I think otherwise we run the risk of hiding errors in attachments, I think we should also expand a little on the ergonomics of users who do this, when they have a database without attachments how will they get them? do they need to rereplicate with attachments, manually construct a url for the original attachment, should we store the url we would have replicated from so they could just fetch() it?

broerse:
Perhaps we should call the `unreacheable: true` flag `"unfetched": true`

garrensmith:
@daleharvey and @broerse I've added the `unfetched: true` flag. This works really well. And cleans up the code nicely. The code can now replicate to and from a source repo. I'm a bit stuck now when it comes to being able to replicate again from the source repo and get the attachments. 

Currently when the replication starts it does a diff and sees that the doc's are already in the target and doesn't replicate them. Is there a way to force the replication to send all docs again otherwise we need to introduce a new way of forcing replication to re-send docs with attachments. Else we need a work around i.e a user has to replicate the attachments to a new db or fetch them manually.

broerse:
@garrensmith This is looking good. I think if `attachment: true` the diff will have to diff the `unfetched:` flag where `unfetched: false` shoud be considered the same as a not existing `unfetched:` flag on for instance the CouchDB side. This will ensure that if both sides have `unfetched: false` or both sides have `unfetched: true` flags no replication is done. Not sure where to do this in the code :smile: I hope @daleharvey can point you to it. It seems you are almost there.

packages/node_modules/pouchdb-adapter-idb/src/bulkDocs.js
```diff
@@ -168,17 +169,28 @@ function idbBulkDocs(dbOpts, req, opts, api, idb, callback) {
...
     var req = attachStore.get(digest);
     req.onsuccess = function (e) {
       if (!e.target.result) {
-        var err = createError(MISSING_STUB,
-          'unknown stub attachment with digest ' +
-          digest);
-        err.status = 412;
-        callback(err);
+        console.log('ATT HERE', att);
+        if (att.unfetched) {
```
daleharvey: I think this should be moved to before verify attachment is even called since we already know to not expect it with an `if (att.unfetched` check

packages/node_modules/pouchdb-adapter-idb/src/index.js
```diff
@@ -379,6 +379,11 @@ function init(api, opts, callback) {
     var type = attachment.content_type;
 
     txn.objectStore(ATTACH_STORE).get(digest).onsuccess = function (e) {
+      if (!e.target.result && attachment.unfetched) {
```
daleharvey: Same as above

daleharvey:
Yeh the rereplicating is a major problem, https://github.com/pouchdb/pouchdb/blob/master/packages/node_modules/pouchdb-replication/src/replicate.js#L149 here we send the list of source changes that we see into the target and if they already exist we dont do anything

We cant bypass this optimisation because the user 'might' have done an attachments false replication, thats harming the performance for everyone, even having the ability to force replicating all docs if attachments=true is harming the replication performance a lot

Could look into what effect it would have to have revsDiff return false if it has unfetched attachments?

daleharvey:
Aother idea that may be worth considering. Once idb-next is landed I would like to spend time working on replication optimisations, there are lots of different approaches that could work, not all of them are compatible and not all of them make sense inside core pouchdb, adding code to support each optional feature of the replicator soon gets unmaintainable

I have considered (briefly) whether it would be a good idea to simplify the replication code in a way that made it super simple for people to write a new replicator with completely different behaviour without needing to touch the core of pouch

ie what could we do inside pouch to make it easier for someone to have a replicator that supported attachments=false, without us actualling making changes to core pouch and without the effort to write a replicator supporting attachments=false being huge

broerse:
Except form the `attachments: false` option there is not much completely different behavior needed I think. The scenario where you want to refetch all unfetched attachments later is an unusual one. I don't think we should check all attachments for the unfetched flag in `revsDiff`.  Just document it will not refetch for performance reasons. 

We could make a plugin with a `fetchAttachment` function that can fetch one of the lets say four unfetched attachments. Because of the `unfetched:` flag we than still have data integrity. We don't have to stop replication and start it again if we have a `fetchAttachment` function.

So I am for implementing this without a refetch.

packages/node_modules/pouchdb-adapter-idb/src/bulkDocs.js
garrensmith: I originally had it there. But this for the case we replicate back to the original source db. Then we can remove the unfetched flag since the attachment exists.

packages/node_modules/pouchdb-adapter-idb/src/index.js
garrensmith: I originally had it there. But this for the case we replicate back to the original source db. Then we can remove the unfetched flag since the attachment exists.

garrensmith:
@daleharvey making the replication pluggable would be awesome. I think that would be the only way to get this to work in the case where we want to do a new replication and include the attachments the second time around.

At this stage without making replication pluggable, we can get the replicator to replicate to a source without attachments and replicate any new changes on the target back to the source. But we can't get it to replicate the missing attachments to the target via a new replication. I think this could still be useful in the state it is in. But we would have to put some warnings and workarounds in the documentation. Dale would this be ok or would you prefer we waited until we made the replication section pluggable?

daleharvey:
> Dale would this be ok or would you prefer we waited until 
> we made the replication section pluggable?

So the replication is already pluggable / just a plugin, it just that right now the code isnt in a hugely reusable + modifiable state, I like the idea of this as a feature but am definitely leaning more towards a modified replicator than an option. particularly since developers can work around this with a slightly different data model

I dont think we can release this without being able to backfill the attachments, the second we release it people will be asking for that feature / will be counted as a bug and I am hesitant to diverge pouchdb core apis from couchdb (ie the revsdiff change that would let us backfill)

There are lots of different ways to improve replication and its going to be unweildy to stack all of these option on top of each other, I think lets start to split them out now, see how good we can make them and get the things that work best merged back to core

garrensmith:
@daleharvey that makes sense. I'm going to close this PR but I will keep the branch around as a reference. And we can look at ways of adding extension points. Thanks for the help.

Issue #7057, opened 2018-01-30
Can't create databases in iOS 10 on Safari or Chrome


rdwoodring:
### Issue
Can't create PouchDB on iOS10 Safari or Chrome

### Info
- Environment: (Browser)
- Platform: (iOS - Chrome 63.0.3239.73 and Safari)
- Adapter: (IndexedDB (letting PouchDB choose))
- PouchDB Version: 6.4.2

### Reproduce
1. Have a page that includes PouchDB 6.4.2
2. Add some code to create a database
3. Hit the page from an iOS 10 device in Safari or Chrome

### Expected Result
The database should be created using the IndexedDB adapter.

### Actual Result
The database is not created and the following error is logged:

Database has a global failure
message: An unknown error occurred within Indexed Database
name: UnknownError

Failed to open indexedDB, are you in private browsing mode?

objectStore@[native code] i@pouchdb-6.4.2.min.js:7:83327 e@pouchdb-6.4.2.min.js:7:89900 onupgradeneeded@pouchdb-6.4.2.min.js:7:89906

### Test Case
https://attractive-brand.glitch.me/

Daniel-mumms:
I can confirm the same problem on an iOS 11.2.5 device. However if using private/incognito browsing, I do not get the error, and PouchDB seems to be created normally.

SteffenDE:
I can confirm this on iOS 11.2.2, PouchDB 6.4.2.
It doesn't happen on every page load, but if it doesn't happen the creation of the PouchDB database seems to hang.

htejwani:
Using PouchDB 6.4.2 in an iOS app, I get the same error. The database is created fine but the sync fails with this error after a download of around 50-100 MB.

InvalidStateError: Failed to execute 'objectStore' on 'IDBTransaction': The transaction finished.

daleharvey:
> The database is created fine but the sync fails with this error after a download of around 50-100 MB. 

mobile safari has had a 50MB limit with websql, would expect some similiar limit to be implemented for indexeddb. This isnt the same as the reported bug

The ios tests are currently fully passing using indexeddb on 8.4, would be good to figure out exactly where the error is here

rdwoodring:
[...]

Issue #7187, opened 2018-03-22
Iphone saucelabs failing repeatedly


daleharvey:
Noticed the last few days that this test is now super faily

WorldMaker:
May be related to #7057 ? Passes in #7194 at least.

daleharvey:
Not fixed yet

daleharvey:
So this is the failure that I see 90% of the time
```
  "completed": true,
  "failed": 1,
  "passed": 548,
  "failures": [
    {
      "message": "done() invoked with non-Error: [object PromiseRejectionEvent]",
      "stack": "test.setup_global_hooks.js:17:7",
      "title": "\"after each\" hook for \"Test basic events\""
    }
  ],
  "lastPassed": "Test basic events",
```

This fail started happening reasonably recently, but I cant reproduce locally and the way travis is setup is a nightmare to bisect so not sure if it was a change by us or a browser change

PR #7194, opened 2018-03-25
(#7057) - Remove Blob Support Check in IndexedDB

WorldMaker:
Versions of Chrome < 43 supported IndexedDB but not Blob attachments.

Everything on the current browser support horizon supports Blobs [1]
today and the Blob support check is initialization time that could go to
use elsewhere.

Additionally, this may solve a timing issue in database creation on
recent versions of iOS (#7057).

[1] https://caniuse.com/#search=blob


WorldMaker:
I find it interesting to point out that this branch does seem to fix the iPhone test failure plaguing all the other PRs, but leave it to Safari to be the odd duck out to fail a test here.

daleharvey:
It doesnt fix the iphone issue, thats just an intermittent issue, in your run it failed once but the restart fixed it.

Sadly the Safari failure is genuine though, its using a recent safari "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_1) AppleWebKit/604.3.5 (KHTML, like Gecko) Version/11.0.1 Safari/604.3.5" and fails with 

>  DataCloneError (DOM IDBDatabase Exception 25): Failed to store record in an IDBObjectStore: BlobURLs are not yet supported.

So it doesnt look like we can take this blob check out

WorldMaker:
Looking at the test it appears to me it may just be an issue in converting Base64 to Blob rather than an issue with Safari's Blob storage support.

Some possibly supporting evidence:

- https://stackoverflow.com/questions/17924501/javascript-convert-base64-to-blob-fails-in-safari-only
- https://stackoverflow.com/questions/14347534/blob-constructor-not-working-in-safari-opera

The conversion path here resembles the not-working examples in those threads and doesn't show the `.buffer` recommended in the accepted answers.

daleharvey:
Ah that would make sense and be nice, in that case the code for that is @ https://github.com/pouchdb/pouchdb/tree/master/packages/node_modules/pouchdb-binary-utils/src

packages/node_modules/pouchdb-adapter-idb/src/index.js
```diff
@@ -700,31 +696,29 @@ function init(api, opts, callback) {
...
-      if (typeof blobSupport === 'undefined' || !storedMetaDoc) {
+      if (!storedMetaDoc) {
+        guardedConsole('error', 'Meta document may not have been stored');
         return;
       }
       api._meta = {
         name: dbName,
         instanceId: instanceId,
-        blobSupport: blobSupport
+        blobSupport: true // Versions of Chrome < 43 supported IndexedDB, but not Blob
```
daleharvey: We dont need this if its hard coded, just take out the falsy paths for anywhere that compares against this

WorldMaker:
Strange. In the latest build Travis claims it built https://github.com/pouchdb/pouchdb/commit/a021f0e49612f45cf70be551cdee661786759d68, which is the same it built the last time (merge of 780c685 with `master`). Is there maybe a Travis config issue that it is using stale fork merges and not actually picking up new commits?

daleharvey:
Travis does wierd stuff with how it merges branches, but never had an issue with it not testing the code that is shown in the PR, so I would assume its accurate 

I did a quick test, https://slash-gray.glitch.me/ seemed to suggest I could store a blob in safari, http://robnyman.github.io/html5demos/indexeddb/ looks to be broken although not sure where its breaking, the test failure is consistently 

>    DataCloneError (DOM IDBDatabase Exception 25): Failed to store record in an IDBObjectStore: BlobURLs are not yet supported.

Are you able to reproduce the failure and fine the issue? trying to fix this blind and hoping CI passes eventually will take a really long time, I will be able to help out soon but want to get the false negatives @ https://github.com/pouchdb/pouchdb/issues/7187 fixed first

PR #7237, opened 2018-04-19
Request text change in FAQ doc

LarsVoith:
Since IndexedDB is available now can we change the text?
The FAQ says **iOS Web Application** .... IndexedDB is not available ---- change to --- >  IndexedDB is now available as of iOS.3

https://caniuse.com/#feat=indexeddb


daleharvey:
Awesome thanks 👍

PR #7297, opened 2018-05-26
 Add promise polyfill for ie tests

daleharvey:


nolanlawson:
If history is any indication, you may have the best luck with IE10+Win8.

daleharvey:
Ok managed to get these green, the skips are mostly needed for ie and not edge, however I wanted to make sure that ie was tested since its currently still more used than edge, we can only get edge 14 to run but can work on getting 17 up and running and we can work on removing the skips, although not too fussed about them since its fairly clear these are ie intermittent bugs that got fixed in edge

PR #7300, opened 2018-05-29
Add documentation for deterministic_revs

garrensmith:


docs/_includes/api/create_database.html
```diff
@@ -15,6 +15,7 @@
 * `auto_compaction`: This turns on auto compaction, which means `compact()` is called after every change to the database. Defaults to `false`.
 * `adapter`: One of `'idb'`, `'leveldb'`, or `'http'`.
 * `revs_limit`: Specify how many old revisions we keep track (not a copy) of. Specifying a low value means Pouch may not be able to figure out whether a new revision received via replication is related to any it currently has which could result in a conflict. Defaults to `1000`.
+* `deterministic_revs`: Use a md5 hash to create a deterministic revision number for documents. Setting it to false will mean that the revision number will be a random number. Defaults to true.
```
daleharvey: I think we should have `mean that the revision will be a random string` here. uuid is a number in theory but in practice a string and I dont think we want to confuse people by giving the impression that its a usable data instead of just a random id, what you think?

garrensmith: Good point.
I've changed it to random UUID

Issue #7259, opened 2018-05-04
this is not scoped in toPromise()


alxndrsn:
### Issue
`pouch-utils` complains if PouchDB functions are not bound to the pouchdb instance.

### Info
- PouchDB 6.4.3

#### `PouchDB.get()`
```
db = new PouchDB(`test-${Date.now()}`);
Promise.resolve('some-doc-id').then(db.get);
-> TypeError: this is undefined
```
#### `PouchDB.put()`
```
db = new PouchDB(`test-${Date.now()}`);
Promise.resolve({}).then(db.put);
-> TypeError: this is undefined
```

daleharvey:
File is @ https://github.com/pouchdb/pouchdb/blob/master/packages/node_modules/pouchdb-utils/src/toPromise.js, anyone cleverer than me know how to fix this off hand? trying to get a clean sweep of the issues before release

PR #7304, opened 2018-05-30
(#7259) Bind standard functions to PouchDB instances

alxndrsn:


daleharvey:
Awesome thank you, is there any chance you can add a quick test to this? a simple .get case as you did in the issue in test.basics.js would be perfect. It seems like we may want to change up how AbstractPouchDB is defined to avoid having to rebind these and would be good to make sure it doesnt get regressed

PR #7308, opened 2018-05-31
(#7307) - Remove unneeded tests

daleharvey:
So the chrome run was because the chrome 66 that travis was using had a weird fatal bug that didnt exist in any other chrome, and is now working

For IE, IE 11 is too big to drop support for however its pretty faily in the test suite, the issues are all intermittent issues unlikely to be seen by real code and are fixed in edge so not sure its worth it / possible to fix them but I want to make sure we dont completely break ie11, the other ie test run is 800 tests including basics and is much more reliable so happy testing that + edge

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,10 +70,8 @@ env:
 
   # Testing in saucelabs
   - CLIENT=saucelabs:chrome COMMAND=test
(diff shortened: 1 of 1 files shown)


daleharvey:
Gonna merge as test only

Issue #7360, opened 2018-06-28
IE 11 fetch requirement not clear


jpike88:
It is unclear in the documentation or on the website that a fetch polyfill is required for IE 11.

daleharvey:
Cheers for the report, added to the docs in ^

PR #7364, opened 2018-06-30
(#7360) - Add fetch / promise polyfill requirements to docs

daleharvey:
[skip ci]


daleharvey:
Docs only, so timebombing

PR #7365, opened 2018-06-30
(#7135) - Remove WebSQL from repo

daleharvey:


daleharvey:
As mentioned in https://github.com/pouchdb/pouchdb/issues/7135, if someone wants to use / maintain these modules happy to give you access to the pouchdb-community repo, but doesnt need to block removing them from this repo (the codes in git, it will stay around)

Issue #6266, opened 2017-02-19
Sorting with selector $exists uses the wrong index


nolanlawson:
_From @dolfje on September 8, 2016 10:27_

When trying to sort on a field that is only selected by the $exists selector, gives an error.

> testDB = new PouchDB('test');
> testDB.createIndex({index: {fields: ["test"]}}).then(function() {
>  return testDB.put({_id: ""+Math.random(), "test": 5});
> }).then(function() {
>   return testDB.find({selector:{"test": {$exists: true}}, sort: ["test"]});
> }).then(function(ret) {
>   // Expect result, only get 
>   //Error: Cannot sort on field(s) "test" when using the default index
> });`

_Copied from original issue: nolanlawson/pouchdb-find#212_

nolanlawson:
_From @dolfje on September 8, 2016 10:47_

I think the checkFieldsLogicallySound is wrong. There is checking if the index can be used. But that only works when the indexes are one of '$eq', '$gt', '$gte', '$lt', '$lte'. So $exists isn't added there. I've added $exists temporarely to logicalMatchers and my function works. Although I have to be honest, I don't have enough knowledge about the code base, to know the implications.

Also the code for isInvalidNe in that function does nothing. Because the hasLogicalOperator is false when $ne is used. So that part of the code cannot be triggered with $ne.

nolanlawson:
_From @dolfje on September 8, 2016 12:43_

And also $regex should be added to the logicalMatchers variable

nolanlawson:
_From @yoonwaiyan on December 14, 2016 14:54_

Any update on this issue? I'm having the same problem here, trying to sort based on a "date" field but getting the same error even after using indexes:

```
    expenses.createIndex({
      index: {fields: ['date']}
    }).then(function (result) {
      console.log("created index", result);
      return expenses.find({
        selector: {
          date: {'$exists': true}
        },
        sort: ['date']
      }).then(function (result) {
        console.log("result", result);
        dispatch({
          type: GET_ALL_EXPENSES,
          payload: result.rows
        });
      }).catch(function (err) {
        console.log("err", err);
        dispatch({
          type: GET_ALL_EXPENSES,
          payload: []
        });
      });
    }).catch(function (err) {
      console.log("index error", err);
    });
```

nolanlawson:
_From @htrampe on January 6, 2017 20:2_

Same problem here - any ideas?

flamusdiu:
[...]

PR #7368, opened 2018-07-02
(#6266) - Allow regex, type and exists checks to use index when sorting

daleharvey:


garrensmith:
This should fail against CouchDB. CouchDB doesn't allow that behaviour. But this brings up an interesting point. Some of the way we do sorting, querying in CouchDB has different constraints to pouchdb-find. For example sorting without an index could be possible in PouchDB but in CouchDB there isn't a way of doing that at the moment without causing an out of memory issue. I think it can be ok for PouchDB-find to deviate from the behaviour of CouchDB.

daleharvey:
so not sure about the effect on the other tests, but all the newly added tests passed against couchdb without modification

garrensmith:
Right. Ok. That is cool and a little suprising. My guess is then that uses the index for sorting and the filter in memory afterwards.

daleharvey:
Yeh its looking like this patch breaks elsewhere, probably for the same reason you didnt expect it to work may just be an exception in couch, not looking to deviate from couch at the moment, mostly looking at pouchdb-find patchs so I can get familiar enough to work on native indexes

Issue #7362, opened 2018-06-29
In memory PouchDB cannot be destroyed


kevinkleine:
### Issue
When calling destroy() on an in-memory PouchDB instance, a never resolving promise is returned.

### Info
- Environment: Nodejs v8.9.3
- Platform: Windows
- Adapter: pouchdb-adapter-memory => MemDown
- Server: local db

### Reproduce
```javascript
'use strict'

const   PouchDB         = require( 'pouchdb' )
const   InMemPouchDB    = PouchDB.defaults({ db: require( 'memdown' ) })

PouchDB.plugin( require( 'pouchdb-adapter-memory' ) )

const   normalDB        = new PouchDB( 'normal' )
const   inMemDB         = new InMemPouchDB( 'inmem' )

normalDB.destroy()
    .then( () => console.log( 'normal db destroyed' ) )
    .catch( console.error )

inMemDB.destroy()
    .then( () => console.log( 'in-mem db destroyed' ) )
    .catch( console.error )

setTimeout(
    () => console.log( 'timeout' )
,   60 * 1000
)

// outputs:
// normal db destroyed
// timeout
```

daleharvey:
Yeh the code for that is @ https://github.com/pouchdb/pouchdb/blob/master/packages/node_modules/pouchdb-adapter-leveldb-core/src/index.js#L1492, we just leave it hanging if there is no destroy method, not actually sure what the best thing to do is here, just calling back will be confusing since the data will still be there, clearing the data is a bunch of code that seems like it should be the adapters responsibility, or it should throw?

daleharvey:
So actually, while it wouldnt actually be enforced, we dont actually support using the `db` object after a destroy is called on it so we could just callback here pretty safely

daleharvey:
Fixed in ^

PR #7377, opened 2018-07-09
(#7362) - Callback when destroying level pouchdb without an underlyin…

daleharvey:
…g destroy


daleharvey:
Timebombing since its a trivial PR

PR #7384, opened 2018-07-12
getKeyFromDoc does not support dot notation

ultravail:
The [CouchDB API](http://docs.couchdb.org/en/2.1.2/api/database/find.html#subfields) supports filtering on sub fields by using hierarchical selectors or by providing the field names in dot notation. Currently fields in dot notation are silently ignored because their values are not extracted from the document. The root cause is the `getKeyFromDoc` function that expects all fields in the index to be "flat" (i.e. top-level fields).

### DIFF
--- a/packages/node_modules/pouchdb-find/src/adapters/local/utils.js
+++ b/packages/node_modules/pouchdb-find/src/adapters/local/utils.js
@@ -47,7 +47,9 @@ function getKeyFromDoc(doc, index) {
   var res = [];
   for (var i = 0; i < index.def.fields.length; i++) {
     var field = getKey(index.def.fields[i]);
-    res.push(doc[field]);
+    var parsedField = parseField(field);
+    var docFieldValue = getFieldFromDoc(doc, parsedField);
+    res.push(docFieldValue);
   }
   return res;
 }


daleharvey:
@ultravail Thanks for the PR :)

It looks like you missed out the definition of `parseField`, eslint is failing on that. Also for a patch like this it would be great if you could attach a test, something that fails before the fix and passes after just to make sure it stays fixed, should be a matter of copying a test from https://github.com/pouchdb/pouchdb/tree/master/tests/find/test-suite-1 that looks similiar to your use case and modifying it to use dot notation

HAppy to give any guidance if you get stuck, can ask here or irc etc

martinhbramwell:
**Does this issue affect only db.find(), or does it also affect filtered replication?**

I have a filter that I simply cannot get to do anything: 

    "by_new_inventory": "function(doc, req) { if(typeof doc == 'object' && typeof doc.data == 'object' ) { return true; } }",

If I strip it back to just...

    "by_new_inventory": "function(doc, req) { if(typeof doc == 'object' ) { return true; } }",

... it works.

The record written in both cases looks like this:

    {
      "_id": "Movement_1_000000180903123155012459",
      "_rev": "1-cbe9ea0c1a4d6d47fe1a0c11ce554f6a",
      "data": {
        "id": "000000180903123155012459",
        "type": "movement",
        "status": "new",
        "inOut": "out",
        "bottles": [
          100037
        ],
        "customer": 12459,
        "inventory": 12345
      }
    }

My plan was to constrain to only records of type 'movement' in status 'new', but it is beginning to seem impossible.

Am I making one of those "too obvious to see" mistakes?
Is there any possible way to see  in the browser log (or somewhere else) the output of `log()` statements embedded in the filter itself?
_____


    "@movilizame/relational-pouch": "^3.0.1",
    "pouchdb-browser": "^7.0.0",
    "pouchdb-find": "^7.0.0",

    chrome browser: 68.0.3440.106

PR #7386, opened 2018-07-15
docs: upsert had existed in SQL prior to MongoDB

reconbot:
By at least a year or two https://en.wikipedia.org/wiki/Merge_(SQL)


daleharvey:
I think most people came to hear of upsert via mongo, so could be argued either way, but with this being a little more correct happy to have that change, cheers

PR #7387, opened 2018-07-15
docs: remove a sexual reference from the replication guide

reconbot:


daleharvey:
Yeh Nolan is going for a friendly tone but probably overly familiar there, cheers

Issue #7391, opened 2018-07-18
withCredentials default true in 6.4.3, credentials = 'same-origin' in 7.0.0


jlami:
The move to `fetch` has changed the default behavior for cross-domain requests: where cookies were sent in 6.4.3, in 7.0.0 they are not.

I don't know if `credentials = 'include'` would be a safe default value. But this would fix any issues.

Otherwise it might be prudent to document this change.

default 6.4.3: https://github.com/pouchdb/pouchdb/blob/6.4.3/lib/index-browser.es.js#L8408

daleharvey:
Yeh we should match the fact we set withCredentials in previous releases, good call, cheers

PR #7395, opened 2018-07-21
(#7391) - Restore default of sending credentials with fetch requests

daleharvey:


steodor:
If you're using `pouchdb-authentication` and the upgrade to `pouchdb@7.0.0` broke your login, this PR fixes it!

daleharvey:
Will take that as a 👍 thanks :)

barbalex:
Has this made it into pouchdb-browser v7.0.0? I have this exact same problem using that. And I can't use full pouchdb because that causes errors in gatsby v2.

Looks likle pouchdb-browser has not been updated in 4 months. Please update. Otherwise I am completely blocked.

barbalex:
My problem is: I am not building the pouches myself. It is [rxdb](https://github.com/pubkey/rxdb) that does that. So I can not even apply https://github.com/pouchdb-community/pouchdb-authentication/issues/239#issuecomment-410489376 to circumvent the problem.

barbalex:
please publish a new version

PR #7398, opened 2018-07-22
(#XXXX) - Split up Edge test suite

daleharvey:


.travis.yml
```diff
@@ -68,14 +68,17 @@ env:
   # Testing in saucelabs
   - CLIENT=saucelabs:chrome COMMAND=test
   - SKIP_MIGRATION=true CLIENT=saucelabs:safari COMMAND=test
-  - SKIP_MIGRATION=true CLIENT="saucelabs:MicrosoftEdge:14:Windows 10" COMMAND=test
+
+  # Split up the ie test suites (they are faily)
+  - GREP=suite2 SKIP_MIGRATION=true CLIENT="saucelabs:MicrosoftEdge:14:Windows 10" COMMAND=test
+  - GREP=suite2 INVERT=true SKIP_MIGRATION=true CLIENT="saucelabs:MicrosoftEdge:14:Windows 10" COMMAND=test
```
garrensmith: What is the difference between these two test runs?

daleharvey: I put `suite2` on some of the longer test suites (replication etc) till it seemed like it was about half. half can run with `GREP=suite2` and the other half run with `GREP=suite2 INVERT=true` which does a negation on the grep

Issue #7403, opened 2018-07-25
Error when requesting all_docs and their attachments in a binary blob if one doc doesn't exist in db


Blafy:
### Log

```
allDocs [{…}]
    0: attachments: true
        binary: true
        include_docs: true
        keys: (6) ["a_r", "a_z", "b_r", "b_z", "c_r", "c_z"]
        __proto__: Object
        length: 1
        __proto__: Array(0) 
TypeError: Cannot read property '_attachments' of undefined
    at readAttachmentsAsBlobOrBuffer (index-browser.es.js:8703)
    at Array.forEach (<anonymous>)
    at eval (index-browser.es.js:9435)
    at ZoneDelegate.invoke (zone.js:388)
    at Object.onInvoke (core.js:4745)
    at ZoneDelegate.invoke (zone.js:387)
    at Zone.run (zone.js:138)
    at eval (zone.js:858)
    at ZoneDelegate.invokeTask (zone.js:421)
    at Object.onInvokeTask (core.js:4736)
```

### Issue
The above error occurs when requesting all_docs and their attachments in a binary blob, if one of the docs does not exist ( here ‘b_z’ is inexistant ).
Therefore the entire request fails, even if the other docs are present.
It looks like the `readAttachmentsAsBlobOrBuffer` function is assuming the document always exists.

### Reproduce
Send a request with the following parameters : 
```
attachments: true
binary: true
include_docs: true
keys: (2) ["thisDocExists", "thisOneDoesnt"]
```

### Possible fix

A possible fix might be to verify `doc` before `doc._attachments` in `pouchdb-adapter-http/src/index.js`, the way it was before this commit https://github.com/pouchdb/pouchdb/commit/f00b62e50abe0a0d51d778452e139b4465ef7767

```
-- var atts = doc._attachments;
++ var atts = doc && doc._attachments;
```

Blafy:
Ok, working on it ! :)

Blafy:
Ok, I'm working on it !

daleharvey:
Fixed in ^, many thanks

PR #7410, opened 2018-07-30
Check if doc exist before _attachment attributes #7403

Blafy:
Fixing the error when requesting all_docs and their attachments in a binary blob if at least one doc doesn't exist in the DB.
More details are available in the issue #7403.

### DIFF
--- a/packages/node_modules/pouchdb-adapter-http/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-http/src/index.js
@@ -40,7 +40,7 @@ var supportsBulkGetMap = {};
 
 function readAttachmentsAsBlobOrBuffer(row) {
   var doc = row.doc || row.ok;
-  var atts = doc._attachments;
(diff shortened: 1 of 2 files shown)


daleharvey:
Awesome thanks, so we would need a test for this to confirm behaviour and make sure it doesnt get regressed, copying an all_docs test with keys and modifying it to references one that doesnt exist would be good, like: https://github.com/pouchdb/pouchdb/blob/master/tests/integration/test.all_docs.js#L76

Give us a heads up if you need a hand, apologies for the late review will get back quicker next time

Blafy:
I pushed a test as requested, however I'm not sure why some other tests failed in 3 differents builds.

daleharvey:
Awesome will take a look, our test suite is a bit fragile at the moment so they are probably unrelated intermittent fails

daleharvey:
Thats awesome thanks so much, the fix looks great 👍

PR #7411, opened 2018-07-30
Add some find perf tests

SCdF:
There is probably a lot more interesting stuff we can do here. These
follow the level of complexity that existing perf tests look to have,
and try to keep their run times reasonably low while testing useful
things.

This is some pre-amble work before implementing find for idbnext (the
'indexeddb' adapter, as opposed to the existing 'idb' adapter), so we
can understand performance differences between the adapters.

### DIFF
--- a/tests/performance/perf.attachments.js
+++ b/tests/performance/perf.attachments.js
@@ -52,6 +52,6 @@ module.exports = function (PouchDB, opts, callback) {
     }
   ];
 
-  utils.runTests(PouchDB, 'views', testCases, opts, callback);
+  utils.runTests(PouchDB, 'attachments', testCases, opts, callback);
 
 };
--- a/tests/performance/perf.find.js
+++ b/tests/performance/perf.find.js
@@ -1,0 +1,123 @@
+'use strict';
+
+module.exports = function (PouchDB, opts, callback) {
+
+  var utils = require('./utils');
+
+  function makeTestDocs() {
+    var docs = [];
+    for (var i = 0; i < 1000; i++) {
+      docs.push({
+        key: i % 47,
+        even: i % 2 === 0,
+        str: ['foo', 'bar', 'smang', 'foobar'][i % 3]
+      });
+    }
+    return docs;
+  }
+
+  var testCases = [
+    {
+      name: 'create-index',
+      assertions: 1,
+      iterations: 1,
+      setup: function (db, callback) {
+        db.bulkDocs(makeTestDocs())
+          .then(function () {
+            callback();
+          }, callback);
+      },
+      test: function (db, itr, doc, done) {
+        db.createIndex({
+          index: {
+            fields: ['key']
+          }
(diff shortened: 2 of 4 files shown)


tests/performance/perf.attachments.js
```diff
@@ -52,6 +52,6 @@ module.exports = function (PouchDB, opts, callback) {
     }
   ];
 
-  utils.runTests(PouchDB, 'views', testCases, opts, callback);
+  utils.runTests(PouchDB, 'attachments', testCases, opts, callback);
```
SCdF: Guessing this was a bad copy / paste? Apologies if it's supposed to say this…

tests/performance/perf.find.js
```diff
@@ -0,0 +1,123 @@
...
+          index: {
+            fields: ['key']
+          }
+        }).then(function () {
+          done();
+        }, done);
+      }
+    },
+    {
+      name: 'simple-find-query',
```
garrensmith: Could you add another test like this `simple-find-query` but it doesn't use an index but rather just `_all_docs` - basically just don't do the createIndex part. It would be nice to have that as a comparison to see how it compares.

SCdF: Done :-)

daleharvey:
Sorry for the delay, so for doing work on perf tests, I wanted to give a quick mention of my 'plans' for them, doesnt need to block or affect your work just a heads up in case you fancied doing things along the same lines. 

I had 2 main things I want to do, 1. the browserify + tape setup is a little clunky here, I think switching to mocha or a dedicated performance test framework would be best, benchmark never used to have async setup but may be fixed by now

2. Make the results actionable, I wanted to run a service, like our coveralls that checked the last run of the performance tests and make a PR turn red if there was a regression

These have been my plans for many years and unlikely to get round to them any time soon, they wont block this being merged, as said just mentioning.

As for this, its awesome to have extra performance test in place for the work on .find, thanks

daleharvey:
Weird not sure why but its not finding `createIndex` here, the perf tests are failing

SCdF:
> Weird not sure why but its not finding createIndex here, the perf tests are failing

That is weird. It originally didn't until I changed index.html to include find. Maybe I need to change it elsewhere as well? I can have a look on Monday (just got back from a work trip, wading through jetlag today)

re: your plans @daleharvey, agree with all of it. Dependent on how much time I get to work on Pouchdb / how long I need to keep pushing to get the performance I need I could end up getting around to doing this. It would be after implementing idbnext's find and implementing an idbnext-specific mapreduce though IMO, and at the service of finding even more things to work on.

SCdF:
Hello I am back on this stuff! It has been awhile! I'll get this working and then re-request a review

SCdF:
@daleharvey I fixed the mystery of the lack of `createIndex`! The other two erroring builds are mysteries to me, I don't think they have anything to do with anything though

PR #7442, opened 2018-08-27
Fix broken link to CouchDB _find API in docs.

dsabanin:


daleharvey:
Hey thanks so much for this, however since we dont want to be doing this for every couchdb release, any chance you could update this to use http://docs.couchdb.org/en/latest/api/database/find.html ?

PR #7445, opened 2018-08-28
Fix breaking change on server info with no JSON

### LINKED ISSUE #7444: Breaking change in 7.0.0: cannot replicate from some CouchDB servers
Some CouchDB servers give private information on `GET /` (software version, a
welcome message...) but some other don't. For instance, they can return `200 OK`
with an empty body.

With version 6.4.3 and before, this wasn't a problem. Since 7.0.0, replication
fails because JSON-parsing the server response fails and the error is not
catched.

adrienverge:
This commit fixes a breaking change introduced in version 7.0.0 by
commit e35f949 "Switch http adapter to use fetch".

Since version 7.0.0, it is no longer possible to replicate a database
from a server that replies an empty response on `GET /` (even with a
`200 OK` code). This worked fine with version 6.4.3 and before.

See code from 6.4.3, that returned an ID even if the response wasn't
JSON-parsable, or if there was an error:
https://github.com/pouchdb/pouchdb/blob/6.4.3/lib/index-browser.js#L8927

Closes #7444.

### DIFF
--- a/packages/node_modules/pouchdb-adapter-http/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-http/src/index.js
@@ -301,12 +301,13 @@ function HttpPouch(opts, callback) {
   api.id = adapterFun('id', function (callback) {
     ourFetch(genUrl(host, '')).then(function (response) {
       return response.json();
+    }).catch(function () {
+      return {};
     }).then(function (result) {
+      // Bad response or missing `uuid` should not prevent ID generation.
       var uuid = (result && result.uuid) ?
           (result.uuid + host.db) : genDBUrl(host, '');
       callback(null, uuid);
-    }).catch(function (err) {
-      callback(err);
     });
   });
 
--- a/tests/integration/test.replication.js
+++ b/tests/integration/test.replication.js
@@ -3958,27 +3958,55 @@ adapters.forEach(function (adapters) {
       }).catch(done);
     });
 
-
-    it('4094 cant fetch server uuid', function (done) {
-
+    it("#4094 can't fetch server uuid (HTTP error)", function (done) {
       var db = new PouchDB(dbs.name);
-      var remote = new PouchDB(dbs.remote);
-
-      var ajax = remote._ajax;
-
-      remote._ajax = function (opts, cb) {
-        var uri = testUtils.parseUri(opts.url);
-        if (uri.path === '/') {
-          cb(new Error('flunking you'));
-        } else {
-          ajax.apply(this, arguments);
+      var remote = new PouchDB(dbs.remote, {
+        fetch: function (url, opts) {
+          var uri = testUtils.parseUri(url);
+          if (uri.path === '/') {
+            // Network error, as returned by the Fetch API:
+            return Promise.reject(new TypeError('Failed to fetch'));
+          }
+          return PouchDB.fetch(url, opts);
         }
-      };
+      });
 
       var _complete = 0;
       function complete() {
         if (++_complete === 2) {
-          remote._ajax = ajax;
+          done();
+        }
+      }
+
+      var rep = db.replicate.from(remote, {live: true, retry: true})
+        .on('complete', complete);
+
+      var changes = db.changes({live: true}).on('change', function () {
+        rep.cancel();
+        changes.cancel();
+      }).on('complete', complete);
+
+      remote.post({a: 'doc'});
+    });
+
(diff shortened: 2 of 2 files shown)


daleharvey:
Good catch, and the fix looks good (sorry about the failing tests, most of those arent genuine) however we need a test here (coverage is genuinely breaking for the same reason)

You might be able to test this by just overriding `new PouchDB('name', { fetch: function ... ` but if not https://github.com/pouchdb/pouchdb/tree/master/tests/component has examples where you could have a server that errors on / requests

adrienverge:
@daleharvey thanks for the review, yes I should have thought about tests in the first time!

I rewrote the test for #4094, which tries to solve the same problem, by removing AJAX and using the Fetch API (I assumed that now tests should use `fetch()`, please correct me if I'm wrong).
I also added a test for valid HTTP responses which are not JSON.

adrienverge:
Hey @daleharvey, have you had a chance to see the coverage issue you wanted to look at?
Do you need anything from me?

daleharvey:
@adrienverge Sorry for the delay, it seems that the test cant be hitting the newly added catch, maybe the error is being caught elsewhere?

adrienverge:
@daleharvey thanks for your answer.

I have to say I'm a bit confused. I just spent two hours on debugging this coverage (including one hour to have tests running on my Linux machine, I have to say) and *the newly added catch is really hit*!

To be 100% sure, I've tried your `console.log()` trick, and it confirms it:
```diff
--- a/packages/node_modules/pouchdb-adapter-http/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-http/src/index.js
@@ -302,6 +302,7 @@ function HttpPouch(opts, callback) {
     ourFetch(genUrl(host, '')).then(function (response) {
       return response.json();
     }).catch(function () {
+      console.warn('Hey, execution went here!');
       return {};
     }).then(function (result) {
       // Bad response or missing `uuid` should not prevent ID generation.
```
capture d ecran de 2018-11-14 12-14-19

To be 200% sure, I've removed my fix, and both tests fail.

A last argument that this fix really does its job: we've been using it in production for 3 months (by patching PouchDB once installed), and if we remove it, replication fails.

daleharvey:
Sorry for the delay here, we need to make coverage much easier to debug, I keep forgetting how to run it, 

    $ POUCHDB_SERVER_FLAGS=--in-memory TRAVIS_REPO_SLUG=pouchdb/pouchdb COVERAGE=1 PLUGINS=pouchdb-find SERVER=pouchdb-server npm run test-coverage

'should' work, but right now keeps failing at different points for me, will keep plugging away at this and this time for sure make it easier to setup coverage, I keep having to do this every few months

adrienverge:
Don't worry, I can imagine that maintaining such a large project and keep tests perfectly running is a heavy task.

About the fix itself, the summary was: **the newly added `catch` is really hit**. Tests show it, as well as manual debugging with `console.log()` (as you suggested), but also with real-life websites serving PouchDB since August 2018.

daleharvey:
So there are 3 options, 

  1. You hit them locally but not during the test run (we run different tests in saucelabs - https://github.com/pouchdb/pouchdb/blob/master/tests/integration/test.replication.js#L10)

  2. They are not hit deteministically

  3. Your new catch prevents a higher level catch from being hit

I expect its the last one, it should be fairly simple to verify by running the performance tests and seeing which lines arent hit

adrienverge:
Thanks for the clarification. I'm pretty sure this is not option 2.

I lack time to further debug this, if anyone subscribed to this issue want to help, feel free!

daleharvey:
Heh by chance it ran for me this time, yup its #3 this stops 

https://github.com/pouchdb/pouchdb/blob/master/packages/node_modules/pouchdb-adapter-http/src/index.js#L308

from running, can just remove those now and all should be good :)

adrienverge:
Ho, nice, well seen!

I'll adapt the commit.

adrienverge:
Updated with:
```diff
--- a/packages/node_modules/pouchdb-adapter-http/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-http/src/index.js
@@ -308,8 +308,6 @@ function HttpPouch(opts, callback) {
       var uuid = (result && result.uuid) ?
           (result.uuid + host.db) : genDBUrl(host, '');
       callback(null, uuid);
-    }).catch(function (err) {
-      callback(err);
     });
   });
```

Crossing fingers for tests & coverage!

adrienverge:
Thanks! :+1:

PR #7447, opened 2018-08-29
Update 2014-04-14-pagination-strategies-with-pouchdb.md

iugo:


iugo:
When options be set as {skip : 5}, returned data should start from 6. But prev post set the “First doc returned” arrow above 5. I set the arrow above 6.

Issue #7467, opened 2018-09-16
Implement partial_filter_selector.


garbados:
### Issue

The index attribute `partial_filter_selector`, which CouchDB uses to implement [partial indexes](http://docs.couchdb.org/en/stable/api/database/find.html#find-partial-indexes) is not present in PouchDB-Find. Indexes created with this attribute do not apply this selector to the index prior to querying.

### Info
- Environment: Node.js v10.7.0
- Platform: Ubuntu 18.04.1 LTS
- Adapter: LevelDB
- Server: N/A

### Reproduce
Consider this index:

```javascript
await this.createIndex({
  index: {
    partial_filter_selector: {
      _id: {
        $regex: '^bookmark:.+$'
      }
    },
    fields: ['tags']
  },
  ddoc: 'bookmark-tag-search'
})
```

Queries against this index, forced using the `ddoc` query attribute, will return results which do not pass the partial filter selector.

garbados:
Here is a script for reproducing the error:

```javascript
const PouchDB = require('pouchdb')
PouchDB.plugin(require('pouchdb-find'))

async function test () {
  const db = new PouchDB('.test')
  await db.createIndex({
    index: {
      partial_filter_selector: {
        _id: {
          $regex: '^bookmark:.+$'
        }
      },
      fields: ['_id']
    },
    ddoc: 'bookmark-tag-search'
  })
  await db.bulkDocs([{
    _id: 'bookmark:1',
  }, {
    _id: 'list:1',
  }])
  const docs = await db.find({
    selector: {
      _id: {
        $gt: null
      }
    },
    use_index: 'bookmark-tag-search'
  })
  // should only contain the one bookmark
  return docs.length === 1
}

test().then((ok) => {
  console.log(ok)
}).catch((error) => {
  console.error(error)
})
```

detroitenglish:
![image](https://user-images.githubusercontent.com/18492423/45612459-43fd4300-ba63-11e8-9a59-8ebadc61887f.png)

garbados:
I'm sorry, I'm not sure I understand. That image indicates, "all of the fields in a candidate index must exist in a query's selector." In the example script, only `_id` is specified in the selector, and it is the only field which is indexed or used by the partial_filter_selector.

Can you explain what you mean?

garrensmith:
@garbados it is not working purely because this hasn't been implemented in PouchDB yet.

ptitjes:
Nope stale bot !

[...]

PR #7472, opened 2018-09-25
(#7467) - Implement partial indexes

ptitjes:
This commit implements [partial indexes](http://docs.couchdb.org/en/stable/api/database/find.html#find-partial-indexes) for #7467.

It attempts to mitigate the work of matching selectors (and follows the style of having the "smallest possible map function") by doing all the massage work of the selector outside of the map functions and duplicating the create functions to avoid a test in the simpler cases.

The added test is derived from the test script provided by @garbados in https://github.com/pouchdb/pouchdb/issues/7467.

### DIFF
--- a/packages/node_modules/pouchdb-find/src/adapters/local/abstract-mapper.js
+++ b/packages/node_modules/pouchdb-find/src/adapters/local/abstract-mapper.js
@@ -1,5 +1,5 @@
 import abstractMapReduce from 'pouchdb-abstract-mapreduce';
-import { parseField } from 'pouchdb-selector-core';
+import { massageSelector, parseField, rowFilter } from 'pouchdb-selector-core';
 
 //
 // One thing about these mappers:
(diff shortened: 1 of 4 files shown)


ptitjes:
I should maybe add 3 more tests to test all the 4 new map functions, right ?
What else is to enhance in my PR ?

garrensmith:
I’m away this week but will take a look next week. Ideally this PR should just run the match selector on each doc before it’s passed to the map function. That is how I did it for CouchDB.

ptitjes:
Hi @garrensmith.

I don't understand what you mean because currently `createIndex()` has no access to the documents. `pouchdb-find` relies on `pouchdb-abstract-mapper` to actually create the index. Are you suggesting that I modify `pouchdb-abstract-mapper` to introduce a pre-filter function in addition to the mapper ?

So I'll wait for you to come back and take a look at the PR.

daleharvey:
I will defer to @garrensmith's review here especially as the one implementing it in CouchDB, however going over the code looks good to me

packages/node_modules/pouchdb-find/src/adapters/local/abstract-mapper.js
```diff
@@ -98,8 +192,9 @@ function mapper(mapFunDef, emit) {
   // mapFunDef is a list of fields
 
   var fields = Object.keys(mapFunDef.fields);
+  var selector = mapFunDef.partial_filter_selector;
 
-  return createMapper(fields, emit);
+  return createMapper(fields, selector, emit);
```
garrensmith: This is nicely done. I'm just wondering if we could remove the amount of code that is copied and pasted from the previous map functions. Would it be possible to do something like this here instead of having different map functions:
```
if (!selector) {
    return createMapper(fields, emit);
} else {
    var mapper = createMapper(fields, emit);
    return function (doc) {
         if (!rowFilter(doc, selector, inMemoryFields)) {
            return;
         }

          return mapper(doc);
    };
}
```

SCdF:
@garrensmith no, indexeddb does not natively support partial indexes AFAIK.

(side note: I have been slammed at "real work" for a couple of months now and haven't made any progress on idbnext. My fond hope is that starts back up next week!)

I imagine for the first cut then this would mean dropping support for them. However, in later cuts it is entirely possible that we just fall back on the same strategy I'd like to employ for mapreduce against real indexes. This would mean that we have to manually keep the index up to date (as you do with mapreduce) by having a separate objectstore for the thing you're indexing. In the case of map reduce that's where the view emits go, and in the case of this that's where a subset of the data goes, but only for documents that pass the defined filter query.

garrensmith:
Partial index selectors are really useful in a clustered database where an index can be really huge and a user wants to narrow down the size of the index and be very specific around what goes into an index. I'm not 100% if this is applicable for PouchDB. Along with the fact that this would be difficult to do in Indexeddb with native indexes I would prefer we wait a bit more before adding this into PouchDB.

ptitjes:
@garrensmith, my rationale was that it is better to have more targeted indexes when you run the same query a lot of times. I don't know if I'm right here because I still can't grasp what's best in terms of performance with PouchDB. So please tell me if I am wrong.

My use case is queries that filter depending on a `type` field. All my documents have a `type` field and this is what allows me to discriminate them. But I often do queries above one type of documents and so those queries run over an index of all the documents when they could be run on a much smaller partial index, hence more efficiently. Am I wrong ? If partial indexes were not to be implemented, should I manually store the documents in separate databases, despite the added administrative cost on the server side ?

As I understand it, this would be difficult to implement for indexeddb but no more than map/reduce. I believe however that this is an important thing to have for collections of documents that don't change often but are often queried. Again, tell me if I am wrong.

I'm willing to do the requested work if ever you guys decide it is a go.
- [ ] Remove the duplicate code as per @garrensmith's suggestion
- [ ] Ensure the partial indexes are not selected for queries that don't select it explicitly
- [ ] Implement tests along the ones of CouchDB

SCdF:
FWIW I think we should eventually implement this, because it's a compatibility point with CouchDB, which is important.

However, with idbnext (where we use native indexeddb indexes) the implementation of this would have to be generating normal full indexes and just merging the `partial_filter_selector` via `$and` with whatever the user is requesting each time they do it, and then passing that to the query planner and have it work out what to do.

I'm not sure and am not weighing in if it's worth implementing before idb next is done or after, just throwing in my thoughts on how idb next will effect it.

garrensmith:
You wouldn't need to do the $and of the partial_filter_selector. CouchDb requires you to specify the partitial_index_selector in the query along with the `use_index` for the partial filter selector index to be used. I would prefer to hold off on this work for awhile. Lets stabilise the pouchdb-find with idb-next and then we can come back to this. It doesn't offer a huge amount of gain for a in browser database where as a larger clustered database this is pretty useful.

daleharvey:
So happy to take @garrensmith's recommendation to hold off on having this merged until we have merged / stablised native pouchdb-find. We may want think about implementing these as some type of plugin as well (https://github.com/pouchdb/pouchdb/issues/7566#issuecomment-450229582)

garrensmith:
I'm +1 on having this as a plugin

PR #7477, opened 2018-09-27
Fix `allDocs` options formatting

craigshoemaker:
Adjusts the list item position for `conflicts`, `attachments`, and `binary` options.


daleharvey:
Yup I cant see a reason for their current indentation, looks like a mistake, thanks for the fix :)

Issue #7495, opened 2018-10-17
Fetch doesn't resolve absolute URIs


ptitjes:
### Issue
PouchDB exposes its `fetch` instance at `db.fetch(...)`. As a `fetch` implementation, it is supposed to follow the `fetch` specification. However, the code of the `http` adapter does not resolve absolute URIs as it should.

### Info
- Environment: All
- Platform: All
- Adapter: Http
- Server: None

### Reproduce
See test cases in PR https://github.com/pouchdb/pouchdb/pull/7496.

PR #7496, opened 2018-10-17
(#7495) - Fix how `fetch` resolves absolute URIs

ptitjes:
If the provided path is an absolute path, then resolve the path on the host directly.

### DIFF
--- a/packages/node_modules/pouchdb-adapter-http/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-http/src/index.js
@@ -435,7 +435,7 @@ function HttpPouch(opts, callback) {
 
   api.fetch = function (path, options) {
     return setup().then(function () {
-      let url = path.startsWith('/') ?
+      var url = path.startsWith('/') ?
         genUrl(host, path.substring(1)) :
         genDBUrl(host, path);
       return ourFetch(url, options);
--- a/packages/node_modules/pouchdb-adapter-http/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-http/src/index.js
@@ -435,7 +435,10 @@ function HttpPouch(opts, callback) {
 
   api.fetch = function (path, options) {
     return setup().then(function () {
-      return ourFetch(genDBUrl(host, path), options);
+      let url = path.startsWith('/') ?
+        genUrl(host, path.substring(1)) :
+        genDBUrl(host, path);
+      return ourFetch(url, options);
     });
   };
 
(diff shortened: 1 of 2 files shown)


ptitjes:
@daleharvey, @garrensmith can one of review my PR please ?

Resolution of #7495 is mandatory for any plugin that needs to communicate with CouchDB directly, as does PouchDB Authentication.

It would be great if it could be solved and that we have a patch release along with https://github.com/pouchdb/pouchdb/pull/7395.

garrensmith:
Sure. But I’ll only get to it on Monday.

daleharvey:
Hey sorry for the delay, you posted just as I went on holiday, will get to this by tomorrow latest, cheers

ptitjes:
No problem guys. I was just worried to miss the release, which I feel might come soon :)

daleharvey:
lol so I looked into this, silly that I missed it before, we use es5 aside from modules so `let` needs to be `var`, this straight up breaks the entire build during minification which is a silly bug on our end somewhere but an easy fix

ptitjes:
Oh sorry. That's too much switching between projects... I'll try to fix that tomorrow!

ptitjes:
@daleharvey So there is only a failure with message `Unable to get property 'rows' of undefined or null reference (http://127.0.0.1:8000/tests/integration/test.design_docs.js:90)"` on both `IE11:Windows 10` tests. Any hint on how it is related to my changes ?

daleharvey:
@ptitjes Its very unlikely this bug is related to your changes, ie (and iphone) are quite unstable browsers in when run in selenium, will restart to confirm and likely this is good to go when green

daleharvey:
Actually I should have looked more carefully, the failures are both ie11 and you are using `startsWith`, probably easiest to just switch that to a regex or can just use a replacement function (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith) but not polyfill, we dont want to touch the string prototype

ptitjes:
@daleharvey I fixed it by using `substr`. There is now only one test failing with a timeout :D

ptitjes:
Finally passing on IE :) <3

ptitjes:
@daleharvey Is the « don't use the big green button » still the policy ? Or we can use the "Squash and merge" or "Rebase and merge" ?

daleharvey:
Did we write dont use the big green button down anywhere? github fixed it a while ago so we have set so it will squash and merge without making a merge commit

daleharvey:
ie using the big green button is fine now

ptitjes:
Yep, still written here: https://github.com/pouchdb/pouchdb/blob/master/CONTRIBUTING.md#committers

PR #7503, opened 2018-10-29
typo

leoelz:
Before that line it says: "... Remember to replace ```user```, ```pass``` and **```myname.example.com```** with the credentials..."


daleharvey:
Awesome thanks for that

PR #7506, opened 2018-11-01
Create Tswana_README.md

KaboLetsogo:
Translate basic introduction version of repository to Tswana-Translation, a language spoken by 4 million in Southren Africa

label:first time only


daleharvey:
So this is interesting, we have had discussions about translating the documentation however that always looked like an unsustainable effort for a probject this size. However maintaining a basic translated README doesnt take much maintenance at all. I think this is a great idea thank you.

If we receive a lot of translations I would be worried about having them all in the root folder however we can cross that bridge when we come to it. Could you rename this to README.tn.md (or tsn, whichever is most used)? Thanks

daleharvey:
Closing for now, but if you want to do the rename would be happy to merge, cheers

PR #7514, opened 2018-11-06
Use esr to fix firefox selenium issue

daleharvey:


daleharvey:
Self merging to get tests fixed

PR #7518, opened 2018-11-11
DONT MERGE

daleharvey:
Just testing a rebase of https://github.com/pouchdb/pouchdb/pull/7496


ptitjes:
I'm surprised that all these failures could come from my patch ! :anguished:

daleharvey:
I am also very surprised, particularly with the `verify-build` as you are only adding 3 lines, but it seems so, will take a look into it but you probably want to look into those failures as well as they are the same as your PR

ptitjes:
I will rebase and run the tests again at home.

PR #7522, opened 2018-11-14
Update databases.md

hacktron95:
the db.info() method is chained with "then", so it needs a catch block, or it will produce an error of type " Unhandled promise rejection ".


daleharvey:
Thanks for the PR, however for code samples in documentation we try to use the minimal example code as possible, this means we ignore what would be standard error handling in cases so going to leave this as is, cheers

PR #7523, opened 2018-11-14
Update documents.md

hacktron95:
`db.put(doc)` will return a promise, so it needs 
`.catch(err => console.log(err))` or it will generate an error of " Unhandled promise rejection "


daleharvey:
Thanks for the PR, however for code samples in documentation we try to use the minimal example code as possible, this means we ignore what would be standard error handling in cases so going to leave this as is, cheers

PR #7525, opened 2018-11-14
**WIP** Native index find support in IDBNext [not ready to merge]

SCdF:
The general base case of this works, at least as far as tests are concerned.

I'm interested in opinions on the general strategy so far.

This is very much a draft, and is missing some key things before it is worth considering for a merge:
 - Much better performance tests so we can see what we're getting ourselves in for
 - IndexedDB does not natively support indexing the values `true`, `false`, and `undefined`, so I need to implement a remapping strategy to support this
 - IndexedDB does not support index key paths which are not valid JS variable names (eg you couldn't index the field `foo-bar` because of the `-`, or `0cool` because of the starting digit), so I need to implement a remapping strategy to support this
 - Generating multiple indexes at the same time causes race conditions which break the idb handle references. Need to either guard against it or queue index creation (the latter probably makes more sense in idb)
 - There is no migration path from the existing IndexedDB implementation to this one.


tests/find/test-suite-1/test.ltgt.js
```diff
@@ -62,41 +95,65 @@ testCases.push(function (dbType, context) {
...
+          { _id: '4', foo: 'abo'}
+        ]);
+      }).then(function () {
+        return db.find({
+          selector: {foo: {"$gte": "ebb"}},
+          fields: ["_id", "foo"],
+          sort: [{foo: "asc"}]
+        });
+      }).then(function (resp) {
+        console.log(resp);
```
garrensmith: You can remove

tests/find/test-suite-1/test.exists.js
```diff
@@ -41,7 +40,6 @@ testCases.push(function (dbType, context) {
       ]).then(function () {
         return db.find({
           selector: {
-            _id: { $gt: null},
```
garrensmith: Nice clean up. Could you add another test that runs the explain method of query.  An `exists: false` and `$exists: true` selector on its own should not use an index but always fallback to all_docs.

tests/find/test-suite-1/test.escaping.js
```diff
@@ -0,0 +1,134 @@
+'use strict';
+
+testCases.push(function (dbType, context) {
+
+  describe.skip(dbType + ': test.escaping.js', function () {
```
garrensmith: Will this be something we can support later?

tests/find/test-suite-1/test.default-index.js
```diff
@@ -390,4 +390,90 @@ testCases.push(function (dbType, context) {
...
+      }).then(function () {
+          return db.bulkDocs([
+              {_id: '1', foo: true},
+              {_id: '2', foo: false},
+              {_id: '3', foo: 2},
+              {_id: '4', foo: 3}
+          ]).then(function () {
+              return db.find({
+                  selector: {
+                      foo: {$ne: 2}
```
garrensmith: A `$ne` should not use an index. So you can remove the createIndex above or you need another selector on `foo` to force it to use the `foo` selector. This could be why PouchDB and CouchDB are different here. Its worth running an explain and seeing what PouchDB and CouchDB are doing here. But PouchDB could be using the index.

tests/find/test-suite-1/test.default-index.js
```diff
@@ -390,4 +390,90 @@ testCases.push(function (dbType, context) {
...
+      }).then(function () {
+          return db.bulkDocs([
+              {_id: '1', foo: null},
+              {_id: '2', foo: 1},
+              {_id: '3', foo: 2},
+              {_id: '4', foo: 3}
+          ]).then(function () {
+              return db.find({
+                  selector: {
+                      foo: {$ne: 1}
```
garrensmith: I'm pretty sure CouchDB won't use the foo index for a `$ne`. You need to use another selector on foo to use the index.

tests/find/test-suite-1/test.ddoc.js
```diff
@@ -235,5 +235,30 @@ testCases.push(function (dbType, context) {
...
+
+    it('supports creating many indexes at the same time', function () {
+      // Currently this will timeout. This happens because an index creation
+      // also queries the view to "warm" it. This triggers our code to decide
+      // that the idb handle needs to be dropped and re-created. This is fine if
+      // it's done one by one, but if you're doing multiple at the same time
+      // they aren't getting queued one after the other and their handles are
+      // getting invalidated.
+      //
+      // We could fix this by being better with how we hold idb handles, or by
```
garrensmith: I think running createIndexes in a queue are a good idea.

tests/find/test-suite-1/test.array.js
```diff
@@ -79,6 +76,10 @@ testCases.push(function (dbType, context) {
         });
       });
 
+      // { name: 'James', _id: 'james',  favorites: ['Mario', 'Pokemon'], age: 20 },
+      // { name: 'Mary', _id: 'mary',  favorites: ['Pokemon'], age: 21 },
+      // { name: 'Link', _id: 'link', favorites: ['Zelda', 'Pokemon'], age: 22},
+      // { name: 'William', _id: 'william', favorites: ['Mario'], age: 23}
       it('should return docs match single value in array with defined index', function () {
```
garrensmith: Since we checking which index this runs, can you convert this to use the explain method to check its actually using the correct index not just banking on the results. I think you might have to edit a few of these test for that.

packages/node_modules/pouchdb-find/src/adapters/local/abstract-mapper.js
```diff
@@ -126,4 +126,8 @@ var abstractMapper = abstractMapReduce(
   ddocValidator
 );
 
-export default abstractMapper;
+export default function (db) {
+  // REVIEWER: a better name than customFindAbstractMapper
+  // REVIEWER: is passing the DB around the right way to do this?
```
garrensmith: I think @daleharvey will have a good answer on this.

tests/find/test-suite-1/test.escaping.js
SCdF: Yeah, it's skipping currently because if it runs it crashes the whole rest of the test stack. When we merge tests early we'll re-enable this

tests/find/test-suite-1/test.array.js
SCdF: Hmm yeah, looking at lots of these tests they are specifying that they're using an index but we're not checking via explain. I suppose we should change them all? Did you want to hit that as part of your refactor, or should I do that here just for this file?

garrensmith:
I can do it as the refactor

SCdF:
I've split the test changes off into: https://github.com/pouchdb/pouchdb/pull/7529/files

I'll leave this one open for now for comment, but will ultimately probably close it and reopen a new one when I am ready for a merge

SCdF:
Reopening as a new PR

PR #7529, opened 2018-11-16
Test changes in preperation for IDBnext

SCdF:
Some test refactoring, as well as additional tests that test edge
cases specific to IDBNext


tests/find/test-suite-1/test.deep-fields.js
```diff
@@ -54,32 +54,6 @@ testCases.push(function (dbType, context) {
       });
     });
 
-    it('deep fields escaped', function () {
```
SCdF: Moved into an expanded upon in an entirely new test suite

PR #7533, opened 2018-11-19
Use index even for non-logical operators

garrensmith:
When selecting an index for a query don't reject the index if the field
found in the selector is non-logical. Rather use the index with the
default start key. Using the index will automatically
narrow down the list of documents so that is better than using
_all_docs. This fix helps pouchdb-find match CouchDB Mango on index selection.


schie:
@daleharvey will we soon be getting a release with this update in it?


I'm running into an issue that would be resolved by this fix.

PR #7554, opened 2018-12-04
(#7554) - Add convenience function to run coverage

daleharvey:


daleharvey:
Timebombing minor patches

PR #7556, opened 2018-12-04
(#7556) - Make a smoke test ie run and add full run to allowed failures

daleharvey:


daleharvey:
Tired of restarting internet explorer tests, I dont want to ad them all to allowed failures since then we wont notice if we break ie completely so I added a test.basics run which needs to stay green and added the rest to allowed failures, I would like at some time to stabalise these tests on ie but its hard

daleharvey:
Timebombing minor patches

PR #7612, opened 2019-01-22
node-fetch Headers not from default export

jlami:
See https://github.com/bitinn/node-fetch/issues/569 for further information


jlami:
No problem. I see you already rebased yourself. So this time I was the slow one ;)

jlami:
What is the roadmap for a new release?
For now I'm fixing this in my own node_modules directory, but that's not a great solution.

PR #7616, opened 2019-01-24
Some IDBNext fixes

SCdF:
So I started using IDBNext at work to prototype perf increases in our app, and I found a couple of bugs.

First, `rewrite`'s aggression turns out to matter much more than I thought it did, making view generation very slow (well, as slow as current master is, but idbnext is supposed to be faster than that!). Adding a check to the front of rewrite to only rewrite if you need to makes view generation 20x faster for me. I tried writing even more optimal algorithms than that, but I didn't really any more substantial improvement that what is in this PR, so I didn't think it was worth the complexity.

Finally, I made a boo boo when generating native indexes, and accidentally generated a base index if you had any normal, non-mango views.

[side note: using IDBNext in our app I was able to increase refresh time by 2x, and the time it takes it to initially start and generate indexes by 50x. Looking forward to mainlining this thing!]

### DIFF
--- a/packages/node_modules/pouchdb-adapter-indexeddb/src/setup.js
+++ b/packages/node_modules/pouchdb-adapter-indexeddb/src/setup.js
@@ -58,7 +58,7 @@ function maintainNativeIndexes(openReq, reject) {
       return Object.keys(ddoc.views).reduce(function (acc, viewName) {
         var fields = rawIndexFields(ddoc, viewName);
 
-        if (fields) {
+        if (fields && fields.length > 0) {
           acc[naturalIndexName(fields)] = correctIndexFields(fields);
         }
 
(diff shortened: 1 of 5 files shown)


packages/node_modules/pouchdb-adapter-indexeddb/src/setup.js
```diff
@@ -135,7 +135,7 @@ function openDatabase(openDatabases, api, opts, resolve, reject) {
     };
 
     idb.onversionchange = function () {
-      console.log('Database was made stale by being opened in another tab');
+      console.log('Database was made stale, closing handle');
```
SCdF: I also made this error a little more generic since it can happen for lots of reasons other than opening another tab

daleharvey:
That is some side note :) Sounds awesome, I am trying to get through some work on the performance tests now. This patch is looking good but hitting some genuine failures it looks like

daleharvey:
Just a small note, we try to keep the issue number in the commit message, its nice for consistency and also helps refer back to the bug when looking at the commit (https://github.com/pouchdb/pouchdb/commits/master) 

I would have merged it myself but was trying to get android green although that isnt particularly necessary, may move that to allowed_failures if I cant get it greener

SCdF:
@daleharvey whoops sorry! I'll keep that in mind in the future

PR #7618, opened 2019-01-24
Making the indexeddb adapter publicly consumable

SCdF:
It's time to take indexeddb out from under NEXT=1 and into the spotlight. Or at least the `npm install` light..

This PR is attempting to do such a thing. I am probably missing steps. For now I'm creating this PR so travis runs tests, as it seems profoundly disinterested in this branch otherwise.

Things done in this PR:
 - Trying to open an `idb`-created DB with the `indexeddb` adapter errors
 - Trying to open an `indexeddb`-created DB with the `idb` adapter errors
 - Removed the `private` flag from the adapter, which I presume is all I need to do to enable eventual npm publishing
 - Removed the NEXT=1 stuff and replaced it with referencing the adapter directly
 - Unfortunately massively increased the number of tests travis runs.

I'm not sure what to do about that last one. We really should be testing each supported adapter in each supported browser, but that's a matrix of doom. This will drop down once we drop / deprecate `idb` support, but there are still lots of combinations we're not testing here (eg most of Safari).


SCdF:
No we definitely want to do this, I just need to not be crushed at Normal Work™...

alexfedoseev:
@SCdF I'm ~1-2 weeks away from initial release and I'm not sure if I should switch to the new adapter or stick with the current for now. AFAIU those are not compatible and there's no way yet to easily upgrade dbs on switch. What would you recommend?

SCdF:
@alexfedoseev for now I wouldn't, though I would greatly appreciate any testing you want to perform! 

Unfortunately getting this merged is a secondary concern for me as there has been some Super Urgent Critical™ stuff at work for the past few months that has reduced my ability to contribute here. I'm hoping in a month or so I can get back on track and find the time to fix the build and get this work moving again.

On the migration, worst case the recommendation would be to replicate between the two to perform the migration, but I'm hopeful that it's not that much pain to just alter the internal structures. I haven't looked into it deeply yet.

alexfedoseev:
@SCdF Thanks for the feedback! I will stay on the current version then. I’m not familiar with the way pouchdb gets built. If you can give me some pointers how can I consume next adapter, I’ll check it out sometime. Api is the same, right?

SCdF:
@alexfedoseev unfortunately it's hard to use the new adapter right now (this is what this PR is meant to solve), you need to basically use the hidden NEXT=1 build, which I do not believe is possible via standard npm-based imports.

SCdF:
@daleharvey / @garrensmith so I think this is ready to look at. **I have no idea what I'm doing**, and I don't _actually_ know if this will expose this adapter on npm once a publish kicks off.

I also fixed the Chrome tests, and they now run on Saucelabs. This change contains a very large travis yml, which basically tests everything on both adapters (new and old) as well as on Chrome and FF. I think at a minimum we should keep that latter change (everything on Chrome and FF) since they are both important browsers.

tests/integration/test.compaction.js
```diff
@@ -515,7 +515,11 @@ adapters.forEach(function (adapter) {
     // of pouch-adapter-(idb|websql|leveldb), skip the tests if
     // HTTP or if using new adapters (NEXT=1).
 
-
+    // TODO BEFORE MERGING: Do we still need to block this off for the indexeddb
+    // adapter?
+    // Note that we're no longer invoking that adapter by passing NEXT=1, so
+    // this code isn't firing. If these tests work with indexeddb builds on travis
+    // we can drop this.
```
daleharvey: Yup lets drop

daleharvey:
Sorry for the long long delay here, in future feel free to 'timebomb' these commits, ie if you dont get a timely review (2 weeks) and have commit access, feel free to land

So I think this is very nice, we can drop the autoCompaction thing as you mention, the one issue is I would like to avoid adding a whole bunch of new test runs, especially if we need to move chrome to saucelabs for now

We dont need (and will never get) full coverage, while `indexeddb` is still optional, making sure it works in firefox + chromium + node is fine, then as we switch it to be the default it can get full browser coverage

We can take out the AUTO_COMPACTION runs completely at this point since nobody has touched that code in a long time

daleharvey:
Tell you want, lets merge this and can drop the uneeded test runs in a follow up

SCdF:
(Hey! Sorry I was so delayed following up with this, I've been travelling for work and hadn't found time to read your comments yet. Thanks for the merge!)

PR #7630, opened 2019-01-27
node-fetch Headers not from default export

daleharvey:


daleharvey:
Rebasing https://github.com/pouchdb/pouchdb/pull/7612

PR #7641, opened 2019-02-07
pouchdb-utils: fix circular dependency

Ti-webdev:
I got Circular dependency: ../node_modules/pouchdb-utils/lib/index.es.js -> ../node_modules/pouchdb-utils/lib/index.es.js
/
```js
node_modules/pouchdb-utils/lib/index.es.js:import { clone } from 'pouchdb-utils';
```


daleharvey:
Awesome thanks, I think there are a few of these left and they dont seem to break anything, but fixing one at a time is good

PR #7646, opened 2019-02-12
Fix _rev option in fetch document doc

sonnyp:
with `rev`instead of `_rev`:

```
{ error: 'bad_request',
  reason: 'Invalid rev format',
  status: 400,
  name: 'bad_request',
  message: 'Invalid rev format',
  docId: 'hello' }
```


sonnyp:
oops, my bad

Issue #7571, opened 2018-12-16
Links to couchdb wiki need updating, when available


anaran:
### Issue
See apache/couchdb#1813

Also, some still use http: instead of https:, but are upgraded to https.

k-j-kleist:
Please, stale bot, leave this open.

k-j-kleist:
I believe that this issue can be closed.

I have checked all `.md` files in `docs/`: no references to `wiki.apache.org` are left.

daleharvey:
Awesome job thank you

PR #7653, opened 2019-02-18
(#7571) - Remove rotten link to CouchDB wiki.

k-j-kleist:
Instead, link to CouchDB docs. Also, some rephrasing.


daleharvey:
This is great, thanks a lot for going through these

Issue #7647, opened 2019-02-12
Interest in having TypeScript annotations + TSC Compiler checkJS?


phil-lgr:
Hi, I'm a new user of PouchDB.

I wanted to ask if you'd be interested in having TypeScript annotations (see an example of jsdoc TypeScript types annotations here: https://github.com/webpack/webpack/blob/master/bin/webpack.js#L47) together with TypeScript's `checkJS` (info here https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html)

PR when they added it to webpack: https://github.com/webpack/webpack/pull/6862

The benefits I'd see:
- we could generate the *.d.ts in the build instead of relying on @types/pouchdb
- potentially identify code improvements (based on the experience the webpack folks had)
- we can add types to only a subset of the files, no need for huge changes, it would be incremental

Thank you,

Phil

phil-lgr:
I'd start with one of the core file, for example https://github.com/pouchdb/pouchdb/blob/master/packages/node_modules/pouchdb-core/src/adapter.js, add a few types annotations, check what comes out of TypeScript's compiler with checkJS enabled

karlwestin:
Thank you for referencing that Webpack thread, it was really interesting.
I'm looking at the issues that they found and it did not seem like they found a lot of stuff.

We've generally said (if you search for other issues) that we don't have a core effort to maintain typescript. This is not because we think typescript is bad, but because the pouchdb team is super small and there is too much to do already.

So, typescript in the pouchdb repo, sorry not at this time
If you're able to run this separately and file issues/PRs as a result from it, it's very welcome

phil-lgr:
Ok, I understand.

The one huge plus that it had for webpack is you no longer need to rely on hand typed @types/webpack, whenever you npm i webpack your editor will be able to auto complete like a boss regardless if you are writing .js or .ts

Feel free to ping me if you’d want to explore this option in the future!

garrensmith:
I know we typically say we don't support typescript. But I would like us to reconsider this. I've had to work on a few projects where we used typescript and PouchDB. Currently using PouchDB with typecsript is really horrible. I haven't read all the links yet, but I think we should consider this. It would be a huge help for the end users and if we can do some basic type checking on PouchDB it would be useful for us as well. I'm +1 on this.

garrensmith:
@phil-lgr if you would like to do a basic POC for this, that would be cool. We can then see how it goes. Start really simply and just do some of the basic api

[...]

PR #7662, opened 2019-02-23
(#7647) - Convert pouchdb-core/adapter.js to ES6 Syntax

phil-lgr:


phil-lgr:
@garrensmith what do you think of those changes?... after spending a couple of hours with the codebase this weekend..  I realize that this is going to be substantial work

just replacing this helper function:

```js
function adapterFun(name, callback) {
  return toPromise(getArguments(function (args) {
    if (this._closed) {
      return Promise.reject(new Error('database is closed'));
    }
    if (this._destroyed) {
      return Promise.reject(new Error('database is destroyed'));
    }
    var self = this;
    logApiCall(self, name, args);
    if (!this.taskqueue.isReady) {
      return new Promise(function (fulfill, reject) {
        self.taskqueue.addTask(function (failed) {
          if (failed) {
            reject(failed);
          } else {
            fulfill(self[name].apply(self, args));
          }
        });
      });
    }
    return callback.apply(this, args);
  }));
}
```

to a more ES6 style, is going to require a lot of changes 🤔..

garrensmith:
Hey @phil-lgr thanks for trying this. At this stage we can't reach a consensus on this so I think we going to have to push pause on it. Sorry about that.  I like the idea of the es6 code but PouchDb is a complex code base but I think we need to find an approach that is easy and doesn't hinder development on it.

PR #7681, opened 2019-03-08
enable stateful filtering by adding the options object as third argument to the filter function

svogl:
We are partially replicating databases based on a number of external conditions. For this to work, we need an object that is accessible over multiple calls to filter().

We're using typescipt natively so passing an reference to a class instance looks cleaner than a global object, which would be an alternative. 

A basic usage example looks like:

```
class DBGetter {
  // db declarations for local and remote db
  counter = 0
  myFilter(doc: any, req: any, opts: any): boolean {
    const self = opts.self
    self.counter++
    return (self.counter % 2 === 1)
  }

  getHalfOfIt() {
    this.db.replicate.from(this.remoteDB, {
      filter: this.myFilter,
      self: this
    })
  }
}
```

### DIFF
--- a/packages/node_modules/pouchdb-utils/src/filterChange.js
+++ b/packages/node_modules/pouchdb-utils/src/filterChange.js
@@ -1,8 +1,8 @@
 import { createError, BAD_REQUEST } from 'pouchdb-errors';
 
-function tryFilter(filter, doc, req) {
+function tryFilter(filter, doc, req, opts) {
   try {
-    return !filter(doc, req);
+    return !filter(doc, req, opts);
   } catch (err) {
     var msg = 'Filter function threw: ' + err.toString();
     return createError(BAD_REQUEST, msg);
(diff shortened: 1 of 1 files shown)


karlwestin:
Did you know you can achieve this effect completely without this change?
Please see an example here http://jsbin.com/rarosiqele/edit?js,output

I understand this is a simplified example you're providing, but here's the deal:
Any time you use a function for filtering, the filtering will be done in-memory, client side (whether node or browser). You have access to anything variable you create, whether global variables or in the closure you're running.

The suggested change would introduce incompatibilities with [CouchDB filtered replication here](http://docs.couchdb.org/en/stable/ddocs/ddocs.html#filterfun) and we try really hard to match that implementation.

svogl:
Well good point - I can work around in a similar way I guess but I had hoped for a more elegant solution.
Thanks anyway for the example!
Simon

Issue #7417, opened 2018-08-01
live replicating HTTP databases keeps something open and the process hangs


tyler-johnson:
After writing some replication tests for my application I began having some weird issues with the tests never completing. Jest kept giving me `Jest did not exit one second after the test run has completed.` After a lot of head scratching, I decided to work backwards and discovered this appears to be an issue with HTTP databases and live replication.

To demonstrate the issue, I have put to together a small project with the absolute minimum needed to reproduce: https://github.com/tyler-johnson/pouchdb-sync-issues

This comes with two js files `no-destroy.js` and `with-destroy.js`. Both these files are almost identical, they create a local in-memory database and a remote HTTP database and then sync them with each other. When the sync becomes "paused" (all caught up), I cancel the sync. 

`no-destroy.js` does not destroy the remote database and as a result this file will never exit. `with-destroy.js`, on the other hand, does destroy the remote database after canceling and this file exits immediately afterwards.

### Info
- Environment: Node.js
- Platform: Mac OS X
- Adapter: HTTP and LevelDB (memdown)
- Server: CouchDB

tyler-johnson:
I decided to see if this was just PouchDB 7 with the problem. Turns out that it is, testing with PouchDB 6, both files exited properly.

https://github.com/tyler-johnson/pouchdb-sync-issues/tree/pouch-6

tyler-johnson:
Alright so I've determined the issue. This is specific to the Changes API on Node.js. Since PouchDB now uses node-fetch to make requests, [the changes API relies on AbortController](https://github.com/pouchdb/pouchdb/blob/b666cfc51c375d8fbb49f8323c180bdbfbd9a5d5/packages/node_modules/pouchdb-adapter-http/src/index.js#L925) to cancel the longpoll request. BUT, node-fetch hasn't implemented the abort API (see https://github.com/bitinn/node-fetch/issues/95 and https://github.com/bitinn/node-fetch/pull/437) and right now PouchDB is using [a stub AbortController](https://github.com/pouchdb/pouchdb/blob/master/packages/node_modules/pouchdb-fetch/src/fetch.js#L10) that does nothing.

So I'm not sure what to do until the patch lands on node-fetch. This seems like kind of a big issue given that this is effectively a memory leak in Node.js applications.

Pinging @daleharvey @nolanlawson for guidance.

gkiko:
[...]

PR #7740, opened 2019-04-25
(#7417) - spec compliant AbortController to fetch

tyler-johnson:
So a little while ago, node-fetch finished up added proper support for abort signals (bitinn/node-fetch#437, bitinn/node-fetch#539), I figure it's over due for PouchDB to support them as well.

I thought I would start this by just adding the basic requirements for support: upgrading node-fetch to `^2.3.0` and replacing the AbortController mocks with [abort-controller](https://github.com/mysticatea/abort-controller).

I'm not sure how you guys want me to test this or what else is required, but I am happy to make this pull request a bit more complete.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "test-webpack": "bash bin/test-webpack.sh"
   },
   "dependencies": {
+    "abort-controller": "^3.0.0",
     "argsarray": "0.0.1",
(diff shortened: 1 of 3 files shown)


daleharvey:
It looks like the node-fetch update on its own is causing our build to fail

PR #7745, opened 2019-04-29
fix: handle corrupt rev tree

ronag:
If a document has a corrupt rev tree it can cause an unhandled exception.

This PR fixes it by fallback to minimal rev tree if the real one is corrupted.

Alternatively we should probably do a `cb(new Error('corrupt rev tree'))` instead of unhandled null reference exception.

### DIFF
--- a/packages/node_modules/pouchdb-core/src/adapter.js
+++ b/packages/node_modules/pouchdb-core/src/adapter.js
@@ -633,6 +633,13 @@ AbstractPouchDB.prototype.get = adapterFun('get', function (id, opts, cb) {
         }
       }
 
+      /* istanbul ignore if */
+      if (!path) {
+        err = new Error('invalid rev tree');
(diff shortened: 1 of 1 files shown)


daleharvey:
I would be far happier with us throwing an error here, that way if we hit this particularly in testing / new PR's we can debug what caused the invalid rev_tree instead of silently carrying on, want to make a PR for that? thanks

ronag:
@daleharvey updated PR

ronag:
@daleharvey I can't quite find what the failure here is?

daleharvey:
Sorry for missing this, the error here is in code coverage, since there are no tests that create an invalid tree, need either a test or more likely an instanbul ignore

ronag:
How do I do a "istanbul ignore"?

ronag:
checks passed! wooho

Issue #7767, opened 2019-05-19
Expose fetchJSON to API


leonid-shevtsov:
### Issue

The `pouchdb-adapter-http` has a private `fetchJSON` function that is useful to any plugins making requests to the CouchDB API. 

For example, I made a [crude implementation](https://github.com/pouchdb-community/pouchdb-authentication/blob/d01851a14ee51de9575b1b08e258239ed7a35daa/src/utils.js#L19) in `pouchdb-authentication`.

Besides being a gross repetition of code, I have to pull in a side dependency on `pouchdb-fetch` just to create a `Headers` object. If you use the `pouchdb-browser` bundle or any other bundle, you'll get an extra copy of the `pouchdb-fetch` library in your build.

### Solution

Expose `fetchJSON` along with the plain `fetch`.

PR #7768, opened 2019-05-19
[pouchdb-adapter-http] Expose api.fetchJSON and api.pathToUrl

leonid-shevtsov:
Closes (#7767)

Besides exposing `fetchJSON`, I also exposed `pathToUrl`, which was similarly hidden in the private implementation and is similarly useful to other plugins.

### DIFF
--- a/packages/node_modules/pouchdb-adapter-http/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-http/src/index.js
@@ -434,12 +434,41 @@ function HttpPouch(opts, callback) {
     }).catch(callback);
   };
 
+  // Converts `path` relative to DB to a fully resolved URL.
+  // If given an absolute path (starting with a slash):
+  //   will base the path off the CouchDB server URL
+  // If given a relative path (not starting with a slash):
+  //   will base the path off the DB URL
+  api.pathToUrl = function (path) {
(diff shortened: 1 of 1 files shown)


daleharvey:
I feel bad to -1 stuff when I am not active enough these days, but I think this will make maintenance increasingly more burdonsome, every api was exposed means support / bugs etc, also this breaks the consistency of the API across the various adapters

If the `pouchdb-fetch` dep is not getting dedupped when building, that needs fixed, otherwise its a coupld of lines of code duplicated which I think is fine to enforce consistency and reduce maintenance burden

garrensmith:
Fair point @daleharvey, specially around breaking api dependencies with the other adapters.

garrensmith:
Thanks for you effort. Sorry we can't merge it in.

broerse:
To get `pouchdb-authentication` to use PouchDb 7 Expose api.fetchJSON would be logical.

broerse:
For www.cloudstation.com we needed to overrule a signup request when logged in as a non-admin user to be able to prevent credentials from being included. I still hope Exposeing api.fetchJSON so pouch-authentication can be improved and work with PouchDB 7 will be included in 7.1

PR #7774, opened 2019-05-27
[pouchdb-replication] add replicating flag to pass through wrappered functions.

snowyu:
We can encrypt/decrypt data through wrappped functions transparently.  But we do not wanna decrypt data in the remote replication for e2e ecnryption.
So we need pass a `replicating` flag to the wrapped functions to specify under the replcation.


snowyu:
I have no idea about your thinking. The pouchdb-replication is the plugin of pouchdb.

daleharvey:
@snowyu That was @garrensmith's point, its a change to core pouchdb code but we dont know the reason for the change

Could you open an issue with the problem you are trying to solve, it may be that other people have already solved it with the current api

Thanks
Dale

PR #7779, opened 2019-06-04
Remove argsarray dependency from PouchDB

### LINKED ISSUE #7775: Remove or relicense argsarray dependency
### Issue

The `argsarray` package used by PouchDb is released under the WTFPL license.  Usages can probably be modernized to use  [Rest Parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters).

This WTFPL license has a large number of issues (lack of warranty disclaimer, very vague rights grant), and was also rejected as an open source license by OSI.

lindner:
- Replace with standard javascript rest parameters
- Fixes #7775

### DIFF
--- a/bin/build-pouchdb.js
+++ b/bin/build-pouchdb.js
@@ -14,7 +14,6 @@ var rollupPlugins = require('./rollupPlugins');
 var rimraf = denodeify(require('rimraf'));
 var mkdirp = denodeify(require('mkdirp'));
 var all = Promise.all.bind(Promise);
-var argsarray = require('argsarray');
 var buildUtils = require('./build-utils');
 var addPath = buildUtils.addPath;
 var doUglify = buildUtils.doUglify;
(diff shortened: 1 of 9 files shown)


garrensmith:
@lindner this all looks good. Traditionally we haven't started using es6 and above features for PouchDB. So the only thing holding me back from merging this is that. @SCdF and @daleharvey what do you think? Do you think there is enough support for rest arguments that we can support this?

One thing, can you look at the failing tests on travis. I'm not sure why its failing on firefox.

daleharvey:
So yeh, not sure if what happened here is by starting to use es6 it triggered more work on the webpack side and thats why the build is bigger?

I think its probably a good time to start looking at introducing es6, it might end up being a big change so I think we should hold of with this and look at merging it after the release

daleharvey:
Closing while this is still failing but feel free to check on the failures and reoppen if needed, cheers

PR #7798, opened 2019-06-13
chore: pin node-fetch to 2.4.1

garrensmith:
@greenkeeperio-bot says that anything above 2.5 is breaking tests and recommends we should pin it. Thanks Greenkeeper


daleharvey:
Yeh happy is this gets a green build

PR #7806, opened 2019-06-18
Tyler johnson fetch abort controller

fluidnotions:
This pull request was rejected due to a package version conflict... see https://github.com/pouchdb/pouchdb/pull/7740#issuecomment-488222013

However it still solves #7417 (Which was closed by a bot, but is actually still an issue, causing a slow memory leak in node.js where zlib related objects seem to accumulate over time)

Updating the conflicting package versions does not effect the original fix by tyler-johnson, which includes abort-controller in place of the current stub.


garrensmith:
@fluidnotions code looks good, but the tests are still failing. I've restarted the tests if they still fail, I need you or Tyler to try and fix those tests.

daleharvey:
Are safari / ie failing because they dont implement abort-controller? can we fallback to the stub in those cases?

fluidnotions:
@daleharvey umm this is a node issue, so why would the CI be running a browser test which could fail on this code @garrensmith I my own project just upgraded to 7.1.1 on node 12.7 to see if this was resolved. It wasn't. This is a major issue, my node apps memory usage increases over 5 days until it crashes. I am not really sure why the CI build is failing, can you have a look at this, it really is a rather trival fix which offers alot of value to pouchdb.

daleharvey:
Pretty much any change to PouchDB will effect both node and the browser, and we run all the tests on every change to ensure a patch doesnt break anything as this patch does.

Removing the changes to `packages/node_modules/pouchdb-fetch/src/fetch-browser.js` might fix this, as I explained in https://github.com/pouchdb/pouchdb/pull/7806#issuecomment-507026434 its failing because we are not falling back correctly in safari and ie

I dont have a lot of time to commit to PouchDB so the time I do have I spend on working on other peoples pull requests and not peoples requests to fix bugs for them, if you want to fix up this trivial fix for a major issue you are having I would be happy to

fluidnotions:
I monkey patched this code in my node_modules on a production instance to test if it does in fact solve the node rss memory leak I am seeing when using changes with live from browser a remote pouch on node... it doesn't solve the issue, so will try logging a new bug related to this memory leak

daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/commit/61e3e01909a56c9131ba5ea9a8888c7ae985f4c6

PR #7828, opened 2019-07-10
Update index.html

Rajchowdhury420:
needed


daleharvey:
Whatever your issue is, this is most definitely the wrong fix, please open an issue to ask a question if something isnt happening that you expect

PR #7830, opened 2019-07-10
Don't require an opaque string in the rev

### LINKED ISSUE #7829: PouchDB cannot sync a document with rev of "1-"
### Issue
A valid CouchDB rev consists of a number, followed by a dash, followed by an opaque string of 0 or more characters. Relevant to this issue, is that the opaque string can be 0 characters long. In particular, these revs all work on CouchDB: `1-`, `2-`, `9999-`, etc.

However, when syncing from CouchDB to PouchDB, `1-` does not work, while the others do.

### Info
CouchDB 2.3.1 in Docker, PouchDB 7.1.1 in Chrome 73.

### Reproduce
Create a document with `new_edits=false`, and rev "1-":

    curl -v -X PUT https://example.com/dbname/foo\?new_edits=false -H 'Content-Type: application/json' -d '{"_rev":"1-"}'

Then sync to PouchDB, and you'll get the error `bad_request: Invalid rev format`

Do the same with a higher rev:

    curl -v -X PUT https://example.com/dbname/foo\?new_edits=false -H 'Content-Type: application/json' -d '{"_rev":"2-"}'

and sync to PouchDB works fine.

If necessary, I can come up with a complete reproduction case, but I imagine this is a simple fix in the rev-validation code.

flimzy:
This aims to fix #7829.

PouchDB has a rev-validating regular expression in two places, and they don't match. This makes them match.

    packages/node_modules/pouchdb-adapter-utils/src/parseDoc.js-49-  if (!/^\d+-./.test(rev)) {
    packages/node_modules/pouchdb-adapter-utils/src/parseDoc.js:50:    return createError(INVALID_REV);

and

    packages/node_modules/pouchdb-core/src/adapter.js-584-          if (!(typeof (l) === "string" && /^\d+-/.test(l))) {
    packages/node_modules/pouchdb-core/src/adapter.js:585:            return cb(createError(INVALID_REV));

Notice that the first one requires at least one character after the `-`, the second does not. With this patch, they are consistent.

It might be wise to define this RE in a single location, to avoid this class of error in the future, but I'll leave that as an exercise for someone with more JS experience.

### DIFF
--- a/tests/integration/test.bulk_docs.js
+++ b/tests/integration/test.bulk_docs.js
@@ -162,6 +162,21 @@ adapters.forEach(function (adapter) {
       });
     });
 
+    it("7829 bare rev 1- with new_edits=false", function (done) {
+      var db = new PouchDB(dbs.name);
+      var docs = [
+        {
+          _id: "foo",
+          _rev: "1-",
+          integer: 1
+        }
+      ];
+      db.bulkDocs({ docs: docs }, { new_edits: false }, function (err) {
+        should.not.exist(err, 'no error');
+        done();
+      });
+    });
+
     it('Test empty bulkDocs', function () {
       var db = new PouchDB(dbs.name);
       return db.bulkDocs([]);
--- a/packages/node_modules/pouchdb-adapter-utils/src/parseDoc.js
+++ b/packages/node_modules/pouchdb-adapter-utils/src/parseDoc.js
@@ -46,7 +46,7 @@ var dataWords = toObject([
 ]);
 
 function parseRevisionInfo(rev) {
-  if (!/^\d+-./.test(rev)) {
+  if (!/^\d+-/.test(rev)) {
     return createError(INVALID_REV);
   }
   var idx = rev.indexOf('-');


flimzy:
@daleharvey That should be done now.

tests/integration/test.bulk_docs.js
```diff
@@ -162,6 +162,23 @@ adapters.forEach(function (adapter) {
       });
     });
 
+    if (adapter === 'local') {
```
daleharvey: Is `1-` a valid rev in CouchDB? We very much try to avoid needing these type of switches and having consistent behaviour between couch and pouch

flimzy: > Is `1-` a valid rev in CouchDB?

Yes, it is. This is actually how I discovered the PouchDB bug--I was trying to sync a `1-` rev from CouchDB to PouchDB, when Pouch failed.

flimzy: Maybe that switch isn't necessary. I now can't remember why I put it there. Maybe related to my local test environment. Let me investigate.

flimzy: Well, without that switch, the test passes, too, so I've removed it and re-pushed that commit.

daleharvey: Awesome, I think you need to push the commit here though, it still exists in this PR

flimzy:
Whoops. Pushed now.

tests/integration/test.bulk_docs.js
```diff
@@ -162,6 +162,21 @@ adapters.forEach(function (adapter) {
...
+      var docs = [
+        {
+          _id: "foo",
+          _rev: "1-",
+          integer: 1
+        }
+      ];
+      db.bulkDocs({ docs: docs }, { new_edits: false }, function (err) {
+        should.not.exist(err, 'no error');
+        done();
```
daleharvey: Actually just a note, we can use promises here, 

    return db.bulkDocs({ docs: docs }, { new_edits: false });

Will throw if there was an error and fail the tests, no need to call `done()` manually, will merge for now but if you want to clean in a follow up would be ideal, cheers

PR #7832, opened 2019-07-12
Change potentially offending term

sto3psl:
In the same vain of removing [sexual references](https://github.com/pouchdb/pouchdb/pull/7387) it makes sense to remove other potentially offending terms. CouchDB had a discussion about that too [COUCHDB-2248](https://issues.apache.org/jira/browse/COUCHDB-2248).

I'm not set on the term `follower`. Could be something else and I will happily change it.


daleharvey:
So CouchDB isnt 100% set on a replacement here, it used single/master in the linked issue but follower makes more sense in this situation so happy to merge, can always change it later, cheers

Issue #7756, opened 2019-05-08
pouchdb-server crashes due to request containing invalid rev


kwpeters:
### Issue
When POSTing to `_bulk_get`, an invalid rev causes PouchDB Server to crash.

### Info
- Environment: Node.js v8.11.1
- Platform: No browser involved. Both client and server running in Node.js v8.11.1.
- Adapter: LevelDB
- Server: PouchDB Server

### Reproduce

All of the code needed to reproduce this issue can be found in [this repo](https://github.com/kwpeters/pouchdbserver-invalid-rev).  I did not use glitch.me in order to demonstrate the server crash.

1.  Start PouchDB Server.
    ```
    pouchdb-server --port 3030 --dir ./dbs
    ```
    In the sample repo, this can be done with `npm run start-server`.

2.  Run the code found in [this file](https://github.com/kwpeters/pouchdbserver-invalid-rev/blob/master/src/invalid_rev_post.js)

    This code sends a POST request to the `_bulk_get` endpoint.  The code demonstrates that when `rev` is set to a valid value, the request succeeds. But, when `rev` is set to an invalid value, it causes the PouchDB Server to crash.

3.  Observe that the second POST request causes the PouchDB Server to crash with the following error and stack trace:
    ```
    Error: Unable to resolve latest revision for id person_1, rev 1-ecdf613faa5b4ba68c610c4d54dc0bca
        at Object.latest (/Users/kwpeters/dev/pouchdb_invalid_rev/node_modules/pouchdb-merge/lib/index.js:462:9)
        at /Users/kwpeters/dev/pouchdb_invalid_rev/node_modules/pouchdb-adapter-leveldb-core/lib/index.js:445:42
        at /Users/kwpeters/dev/pouchdb_invalid_rev/node_modules/sublevel-pouchdb/lib/index.js:245:9
        at /Users/kwpeters/dev/pouchdb_invalid_rev/node_modules/sublevel-pouchdb/lib/index.js:76:13
    ```

    Granted, the request is invalid, but I would expect the server to return an appropriate error rather than crash.

I have created [a duplicate issue in the pouchdb-server repo](https://github.com/pouchdb/pouchdb-server/issues/391), because that is probably the more appropriate place for it.

garrensmith:
I'm closing this based on the pouchdb-server issue.

mohlsen:
[...]

PR #7834, opened 2019-07-12
Bugfix #7756 bulk get exception

mohlsen:
**Issue**
When POSTing to _bulk_get, an invalid rev (not existing revision) with latest=true causes PouchDB to throw and unhandled exception.  This is a discrepancy from CouchDB, which returns `{error: 'not_found', reason: 'missing'}`

Also see comments in related issue in pouchdb-server, https://github.com/pouchdb/pouchdb-server/issues/391#issuecomment-507383847.


packages/node_modules/pouchdb-adapter-leveldb-core/src/index.js
```diff
@@ -383,7 +383,17 @@ function LevelPouch(opts, callback) {
...
+        if (opts.latest) {
+          // #7756: If the revision is a valid pattern but it does not exist, this will throw. In this case just
+          // continue using the non-existing revision, it will not be found and handled as any other (MISSING_DOC).
+          try {
+            rev = getLatest(opts.rev, metadata);
+          } catch (ex) {
+            rev = opts.rev;
+          }
+        } else {
+          rev = opts.rev;
```
daleharvey: This seems like it could be 

    rev = opts.rev;
    if (opts.latest) { 
      try { ....

And be a little cleaner?

mohlsen: sure, made this change

mohlsen:
> Codewise looks good, had a fairly minor comment, not certain I am not missing something there

change made for that, thanks for the comment

> Very much appreciate the tests, however failures are genuine so will need to look into them and get them green before we can merge, cheers

Regarding the failing tests ...  from what I see, the failing tests are all the new tests that were added to prove the broken functionality.  The fix is actually in all the various adapters where the exceptions can be thrown.  I made the fix in the leveldb-core adapter and the tests pass when running in node. 

To take this fix in the leveldb-core adapter, would we also need to make a fix in all the other adapter?  What's the strategy there?  Or can we run these new tests only when running in node with that adapter?

To be honest, you could argue that this fix is *only* appropriate in node.  Throwing an unhandled exception in node takes down the process, while running in browsers that is not necessarily the case.  Thoughts?

tests/integration/test.bulk_get.js
```diff
@@ -74,6 +74,98 @@ adapters.forEach(function (adapter) {
...
+      var db = new PouchDB(dbs.name);
+      db.put({_id: 'foo', val: 1}).then(function () {
+        db.bulkGet({
+          docs: [
+            {id: 'foo', rev: 'BAD_REV'}
+          ]
+        }).then(function (resp) {
+          var result = resp.results[0];
+          result.id.should.equal('foo');
+          if (adapter === 'local') {
```
daleharvey: Ideally we should have the same error API between local and remote, but if not then you can make this assertion a little less specific to avoid the switch here, can just check that it is an error doc

tests/integration/utils.js
```diff
@@ -337,6 +337,15 @@ testUtils.sortById = function (a, b) {
...
 
+testUtils.getInvalidRev  = function (rev) {
+  // Create an invalid revision by incrementing the last digit.
+  var lastDigit = parseInt(rev[rev.length - 1], 16);
+  var newLastDigit = lastDigit === 0xf ? 0 : lastDigit + 1;
+  var invalidRev = rev.slice(0, -1) + newLastDigit.toString(16);
+  return invalidRev;
+};
+
+
```
daleharvey: tiny nit: extra newline

daleharvey:
> To merge this PR with the leveldb-core adapter changes, would we also need to make a fix in all the other adapters? What's the strategy there? Since we improved one of them, and the rest remain as it today, would it be acceptable? If we can take the code, but dont want failing tests for the other adapters, can we run these new tests only when running in node with that adapter?

That is possible and we make exceptions for that sometime however by far we prefer to have the same behaviour across the adapters. We have tried pretty hard to abstract the code away so as much as possible is shared so it should mostly be a copy and paste for it it to work in other adapters. You can search or grep for what looks like the equivalent line (https://github.com/pouchdb/pouchdb/blob/master/packages/node_modules/pouchdb-adapter-idb/src/index.js#L346 in idb) then replace that with your node version.

Can also extract that code out into a shared function `getLatestRev` or something, but dont need to do that

Cheers

PR #7862, opened 2019-08-02
response.buffer() and blob fields doesn't exist in ReactNative

### LINKED ISSUE #7688: Sync attachments from CouchDB : Error in fetchData function response.buffer() not a function
### Issue
Sync with attachments fails-->
fetchData throws error at calling response.buffer()

### Info
- Environment: Ionic 4, PouchDB 7.0, CouchDB 2.3.1 on Ubuntu Server
- Platform: Android
- Adapter: IndexedDB
- Server: CouchDB 2.3.1

### Reproduce
To reproduce this issue you have to follow this steps:

1. Enable any DB Remote CouchdDB 
2. Save an attachment in Android PouchDB
3 Call PouchDB.replicate.to ionic app to replicate document to Remote CouchDB
4 Erase PouchDB on Android
5 Replicate to new PouchDB the documents with attachments from CouchDB

I found this on the code 

function fetchData(filename) {
        var att = atts[filename];
        var path = encodeDocId(doc._id) + '/' + encodeAttachmentId(filename) +
            '?rev=' + doc._rev;
        return ourFetch(genDBUrl(host, path)).then(function (response) {
          if (typeof process !== 'undefined' && !process.browser) {
            return response.buffer();
          } else {
            /* istanbul ignore next */
            return response.blob();
          }

So I solve this issue on my environment by forcing the response to be a blob(), I don't know if this could be a problem in other platforms, but for Ionic it works.

Also had to change (remove) some code that has a TODO comment, // TODO: Can we remove this?

For me this solve everything, my app is working fine but I think you should check that part so we can have a official working version of PouchDB for Ionic 4.

Hope I can help you in the future.

garfieldnate:
Add a check for ReactNative next to the checks for node/browsers so that we don't throw errors using unsupported API's.

Fixes #7688.

### DIFF
--- a/packages/node_modules/pouchdb-adapter-http/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-http/src/index.js
@@ -506,7 +506,7 @@ function HttpPouch(opts, callback) {
         var path = encodeDocId(doc._id) + '/' + encodeAttachmentId(filename) +
             '?rev=' + doc._rev;
         return ourFetch(genDBUrl(host, path)).then(function (response) {
-          if (typeof process !== 'undefined' && !process.browser) {
+          if ('buffer' in response) {
             return response.buffer();
           } else {
             /* istanbul ignore next */
@@ -514,8 +514,8 @@ function HttpPouch(opts, callback) {
           }
         }).then(function (blob) {
           if (opts.binary) {
-            // TODO: Can we remove this?
-            if (typeof process !== 'undefined' && !process.browser) {
+            var typeFieldDescriptor = Object.getOwnPropertyDescriptor(blob.__proto__, 'type');
+            if (!typeFieldDescriptor || typeFieldDescriptor.set) {
               blob.type = att.content_type;
             }
             return blob;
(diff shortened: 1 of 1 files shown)


garfieldnate:
If this needs to be cleaned up or moved into a method or something, please let me know and I'll fix the PR up a bit. For now I would like to know whether the logic change is acceptable or not.

There are probably other environments where this is applicable: Ionic, VueNative, etc. I really just want support for ReactNative :)

daleharvey:
If we could switch these to feature detecting I think that would be ideal, they probably should have been already but definitely if we are going to handle more environments then definitely better than checking for environment names, something like:

```
if ('buffer' in response) { 
```
?

garfieldnate:
@daleharvey Following your suggestion I revised the commit.

daleharvey:
Codewise I like the look of the changes, but need to investigate the test failures, they are genuine, cheers

garfieldnate:
@daleharvey All fixed.

daleharvey:
Awesome thanks, looks good, we usually require tests but we dont test in reactnative so I think thats fine for now, cheers

garfieldnate:
I've seen you comment before that you do not plan on supporting react native because you believe everything can be done in plugins. Is that still the case? Getting PouchDB to work on RN was such a nightmare for me, but it's such convenient tech that it's hard not to try.

If I opened another PR to add testing in react native, would you accept it?

Issue #7810, opened 2019-06-19
since 7.1.1 a query will not find documents when an index exists but is not used


pubkey:
### Issue
Since version 7.1.1 a query will return no documents, when there is an index used.

### Info
- Environment: Node.js
- Platform: Linux
- Adapter: memory

### Reproduce

I created this test which fails ~9 out of 10 times. I think it is a timing problem~. With the older pouchdb-version, it works.

```js
const PouchDB = require('pouchdb-core');
PouchDB.plugin(require('pouchdb-find'));
PouchDB.plugin(require('pouchdb-adapter-memory'));
const db = new PouchDB(
    'things',
    {
        adapter: 'memory'
    }
);
await db.createIndex({
    index: {
        fields: [
            'passportId'
        ]
    }
});
await db.put({
    _id: 'foobar',
    passportId: 'z3i7q29g4yr1',
    firstName: 'Edison',
    lastName: 'Keebler',
    age: 24
});

const docs = await db.find({
    selector: {
        _id: {}
    },
    limit: 1
});


console.dir(docs); // <- empty array
assert.equal(docs.docs.length, 1);
```

karlwestin:
there seems to have been some change in how things work, but there also seems you're not actually using the index you create in the example above

maybe that's happened between the versions?
see console output from 7.0.0 here: (**note the warning about no index**)
![Screen Shot 2019-06-20 at 9 36 05 AM](https://user-images.githubusercontent.com/359947/59830001-ffc70800-933e-11e9-9c94-7a95fe0bfa98.png)

and the console output from 7.1.1 here:
![Screen Shot 2019-06-20 at 9 36 11 AM](https://user-images.githubusercontent.com/359947/59830024-0e152400-933f-11e9-931b-26c81e7d6d94.png)

Here are links to the JSbins
http://jsbin.com/juhinawire/1/edit?html,js,output
http://jsbin.com/pewitofuma/1/edit?html,js,output

pubkey:
@karlwestin thanks for investigation.
Yes, my query does not use the index. But I think having an index should not affect the result of an unrelated query.

karlwestin:
ok that sounds fair but how about this –  
if you use the index that you create, do you still have the bug or does it only show when not using the index?

garrensmith:
This is kind of weird. The only change we have to pouchdb-find is https://github.com/pouchdb/pouchdb/commit/8aed67088b32f972149d52ba74ca5db044fc3174

[...]

PR #7895, opened 2019-08-30
Reproduce error for #7810

pubkey:
This PR adds a test which reproduces #7810

### DIFF
--- a/tests/integration/test.issue7810.js
+++ b/tests/integration/test.issue7810.js
@@ -5,7 +5,8 @@ describe('#7810 query will not find documents when an index is used', function (
         });
     });
     it('should find the documents even if a index is used', function () {
-        var db = new PouchDB('mydb');
+        var CustomPouch = PouchDB.defaults({db: require('memdown')});
+        var db = new CustomPouch('mydb');
         return db.createIndex({
             index: {
                 fields: [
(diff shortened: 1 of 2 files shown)


tests/integration/test.issue7810.js
```diff
@@ -0,0 +1,39 @@
...
+        }).then(function () {
+            return db.put({
+                _id: 'foobar',
+                passportId: 'z3i7q29g4yr1',
+                firstName: 'Edison',
+                lastName: 'Keebler',
+                age: 24
+            });
+        }).then(function () {
+            return db.find({
```
garrensmith: I don't quite follow this test. So why do you use `memdown`? 
I can't see how `_id: {}` would cause it to use the above defined index?

pubkey: I use the memory adapter because I copied that from another test in the suit. It does not matter which adapter is used for this bug.
The query does intentional NOT use the index. The bug is that if the index exists but is not used, this has a side-effekt on the query which then returns the wrong result. See #7810

garrensmith: Is it for any selector or only for `_id: {}` as the selector?

pubkey: It is for any selector which does not trigger the usage of the index

pubkey: Is there any information missing? @garrensmith

garrensmith:
No. I just don’t have time to work on this. My time is fully taken up working on CouchDB 4.0


daleharvey:
Closing out due to inactivity, its unlikely that failing tests will get merged as we dont have a lot of activity on Pouch these days and so fixes would likely need to be authored, I promise to get to any new PR's in a timely manner, apologies for being so slow recently

Issue #7739, opened 2019-04-25
Script not stop when cancel changes


La-source:
### Issue
When I cancel the `change` function, the script does not stop while there is nothing else waiting.

Curiously, if I make a change in the database via Fauxton the script stops immediately afterwards. (does not call the callback change)

### Info
- Environment: Node.js
- Platform: nodejs 10
- Adapter: http
- Server: CouchDB

### Reproduce
```ts
import * as PouchDB from "pouchdb";

const db = new PouchDB("http://db:5984/test", {
    auth: {
        username: "test",
        password: "test",
    },
});

console.log("start");

const change = db
    .changes({
        since: "now",
        live: true,
        include_docs: true,
    })
    .on("change", data => console.log("data", data))
    .on("complete", status => console.log("complete", status))
    .on("error", e => console.error("error", e))
;

setTimeout(() => {
    console.log("cancel");
    change.cancel();
}, 5000);
```

La-source:
the problem is still here

PR #7900, opened 2019-09-06
(#7739) - Use proper AbortController implementation in pouchdb-fetch

mariocasciaro:
This allows to cancel long polling changes requests. Using a fake abort controller will leave live requests hanging and cause serious memory leaks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "ltgt": "2.2.1",
     "memdown": "1.2.4",
     "node-fetch": "2.4.1",
+    "abort-controller": "3.0.0",
     "promise-polyfill": "8.1.2",
     "readable-stream": "1.0.33",
     "request": "2.87.0",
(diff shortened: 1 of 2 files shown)


daleharvey:
Cheers for the PR. Restarting in case it was an intermittent failure but it was iphone that was failing which would hint that this doesnt fallback correctly when there is no abortController implementation?

mariocasciaro:
This is happening on Node.js using the http adapter. I was investigating a pretty bad memory leak and I got here. The requests are not really closed, they stay open and data keeps flowing into buffers that are never emptied because nobody is reading from them.

daleharvey:
Yup, I know the issue affects node more as browsers cleanup the requests themselves, but the fix needs to work in both node and the browser since they are using the same code path

mariocasciaro:
Oh sorry I misunderstood, I didn't see the failure. I though that fetch.js was used only on Node while fetch-browser.js on the rest of the platforms. The package `abort-controller` doesn't have automatic fallback for browsers, but we can easily polyfill with the same package. Safari 11 doesn't have AbortController so yes, that may be the problem.

daleharvey:
You are right, I didnt check the implementation and forgot there was a -browser specific one, passing now so perfect cheers :)

mariocasciaro:
Thanks for merging @daleharvey . Do you have a rough estimate on when the PR will be included in a release?

PR #7913, opened 2019-09-23
Add Kivik to list of PouchDB-related tools

flimzy:


flimzy:
I don't know if this PR should include a link directly to https://github.com/go-kivik/pouchdb or not. If that seems better I can update it.

daleharvey:
Sorry for the delay, cheers

Issue #7914, opened 2019-09-25
Bug: put() return an empty array([]) when `options.new_edits = false`


snowyu:
### Issue

`put()` return an empty array([]) when `options.new_edits = false`

```ts
const result = await db.put(doc, {new_edits: false});
console.log(result);
expect(result.ok).to.be.true;
```
expect return `{ok: true}` if succesful created.

### Reproduce

```ts
const newDoc = {_id: 'newkey', _revisions: {start: 1, ids: ['first-rev']}};
const result = await db.put(newDoc, {new_edits: false});
console.log(result);
```

snowyu:
Keep alive bot.

PR #7915, opened 2019-09-25
fix(put): should not return an empty array with new_edits:false #7914

snowyu:


daleharvey:
https://github.com/pouchdb/pouchdb/issues/7916#issuecomment-560028271

Issue #7916, opened 2019-09-25
[Pouchdb-transform][Bug] afterIncoming can not get result.ok when new_edits is false


snowyu:
### Issue

I've used the `result.ok` to check the saving whether successful in the `transform`.
But put or bulkDocs with options.new_edit: false successfully returns result.ok is undefined.

```ts
aDB.transform({
  afterIncoming: async (result: ITransformIncomingResult, type: TransformPouchType) => {
    if (result.ok) {
       // do it if ok
    }
  }
})
```
I've find this is reason #5775 and #7572 .
So how to determine if the document has been successfully saved, through the strange couchdb bug?

snowyu:
Ok, I have fixed the couchdb bug by refilling the empty array. 

https://github.com/pouchdb/pouchdb/compare/master...snowyu:fix/bulkDocsWithNewEdits

karlwestin:
can you please expand the description of this bug?

- what adapter is this? http? indexeddb? If this is against couchdb, what couchdb version?
- can you add a short test case in the bug description?

thank you

snowyu:
1. CouchDB 2.3.1
  * Some guy decides the pouchdb should simulate this bug for local storage too.
  * This make nobody know what doc is successful when new_edits is false. So `pouchdb-transform` has always been failed if new_edits is false.
2. the shortest test case here

```js
const docs = [{_id: "doc1", _rev: "1-init"}, {_id: "doc2", _rev: "1-init"}];
const result = await db.bulkDocs({ docs: docs }, { new_edits: false });
expect(result).to.have.length(1);
expect(result[0]).have.property('ok', true);
```
**Expect** the pouchdb bulkDocs is always return document id array.

```js
[
  {
    "ok": true,
    "id": "doc1",
    "rev": "1-init"
  },
  {
    "ok": true,
    "id": "doc2",
    "rev": "1-init"
  },
]
```

https://pouchdb.com/api.html#batch_create :
> The response contains an array of the familiar `ok/rev/id` from the put()/post() API...

daleharvey:
> Some guy decides the pouchdb should simulate this bug for local storage too

We aim for API compatibility with CouchDB, if you want to fix this bug it would be best to do so in CouchDB then we can copy its behaviour

PR #7921, opened 2019-10-03
fix: #7916 the remote couchdb bulkDocs with new_edits:false always return empty array

snowyu:
Refilling the result of the remote couchdb returns to correct.


karlwestin:
thank you but we can't really review this until you provide more information here https://github.com/pouchdb/pouchdb/issues/7916

daleharvey:
https://github.com/pouchdb/pouchdb/issues/7916#issuecomment-560028271

PR #7923, opened 2019-10-04
docs(setup): fix typescript import

alexbonhomme:
Hi !

I was setting up a Nestjs project with Pouchdb and I'd to face an issue with TS imports. I've follow the documentation, but I've got an `pouchdb_1.default is not a constructor` error. I've manage to make it work by importing the whole package using an alias.

Hope this could help :-)


daleharvey:
If it works for you, happy to merge, cheers

Issue #7836, opened 2019-07-13
HTTP getAttachment broken in Electron ( nodeIntegration: true )


knubie:
### Issue
When calling `db.getAttachment` in the Electron renderer process, PouchDB incorrectly guesses that the function is being called in a node (or non-browser environment) and calls `response.buffer()` instead of `response.blob()`. `buffer` is undefined and raises an error. (Incidentally, this error message is not bubbled up to the promise, which made it difficult to debug).

The relevant code can be found here: https://github.com/pouchdb/pouchdb/blob/160f481e08d052919219e0115b6e65034303d7ec/packages/node_modules/pouchdb-adapter-http/src/index.js#L622-L627

Where `process` is defined, but `process.browser` is not.

**NOTE**: I believe that `nodeIntegration` must be set to true in the `BrowserWindow` options for `process` to be defined in the renderer process. In Electron <5.x.x  this value is set to `true` by default.

#### Workaround
At the beginning of your code you can add `process.browser = true` to work around the issue.

### Info
- Environment: browser
- Platform: Electron
- Adapter: IndexedDB
- Server: CouchDB

### Reproduce
- Create a new electron process.
- Create two a CouchDB database with a document with at least one attachment.
- Call `getAttachment` from the renderer process.

MichaelJCole:
Wild-ass guess.  Try the browser version of pouchdb: `npm install pouchdb-browser` instead of `npm install pouchdb`

DarrenCook:
Thanks so much, for that simple workaround! I've been collecting similar bug reports as I try to troubleshoot this since yesterday afternoon, and don't know how I missed your bug report until now.

BTW, I was getting this yesterday in chromium (Quasar, mode SPA) on first load, but then it always worked after a hot reload. I.e. not just electron (but with electron a reload never helped). However I haven't been able to reproduce it today.

https://stackoverflow.com/q/54595908/841830  Not Electron-specific, and his workaround was a nasty hack, and then his other workaround (https://stackoverflow.com/q/54595908/841830) looked angular specific. But I think this is the same bug, and I suspect `process.browser=true` would fix it there too.

https://stackoverflow.com/q/51672227/841830 looks to be the same (but connecting to Cloudant not CouchDB, and again using angular, not electron).

Same error message, but probably different cause: https://github.com/pouchdb-community/pouchdb-replication-stream/issues/58

[...]

PR #7933, opened 2019-10-20
Make sure `buffer` is callable before calling it

knubie:
See issue https://github.com/pouchdb/pouchdb/issues/7836 for full details on this bug.


daleharvey:
thanks for the investigation + fix

PR #8026, opened 2020-03-10
Fix inheritance of network error properties

H--o-l:
The goal of this commit is to copy error properties in the case the error is a
network [`TypeError` error] from the [`Fetch` API].

I don't know precisely why, but [`TypeError` error] properties are not
iterable with javascript `for in`.
This html example shows it:
```html
<html>
  <script>
    fetch('missing.png').catch(function(error) {
      for (var i in error)
        console.warn(i); // nothing
      for (var i of Object.getOwnPropertyNames(error))
        console.log(i); // stack, message
    });
  </script>
</html>
```

Thus, when a replication failed on such error, we get almost no details and
can't handle it:
```json
{
  "result": {
    "ok": false,
    "start_time": "2020-03-10T14:03:22.316Z",
    "docs_read": 0,
    "docs_written": 0,
    "doc_write_failures": 0,
    "errors": [],
    "status": "aborting",
    "end_time": "2020-03-10T14:03:22.333Z",
    "last_seq": 0
  }
}
```

To quickly test it (inside a browser that use the Fecth API):
```html
<html>
  <script src="packages/node_modules/pouchdb/dist/pouchdb.js"></script>
  <script>
    var localDB = new PouchDB('mylocaldb')
    var remoteDB = new PouchDB('http://localhost:5984/myremotedb')
    localDB.replicate.to(remoteDB).on('error', function (err) {
      console.warn('failure', err)  // Go look inside `err`
    });
  </script>
</html>
```

[`TypeError` error]: https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/TypeError
[`Fetch` API]: https://developer.mozilla.org/fr/docs/Web/API/Fetch_API

### DIFF
--- a/packages/node_modules/pouchdb-errors/src/index.js
+++ b/packages/node_modules/pouchdb-errors/src/index.js
@@ -48,7 +48,7 @@ function createError(error, reason) {
     // inherit error properties from our parent error manually
     // so as to allow proper JSON parsing.
     /* jshint ignore:start */
-    for (var p in error) {
+    for (var p of Object.getOwnPropertyNames(error)) {
       if (typeof error[p] !== 'function') {
         this[p] = error[p];
       }


H--o-l:
Hey!

Can someone help me with this issue?

Due to the actual lack of detail when a replication failed, like I explained, in my web application, I have to fetch my servers once first to know if they are reachable, and if it's not have the detail of why, before starting an actual replication.
I could remove this delaying fetch call if PouchDB replication was able to forward more details when it failed on HTTP error.

I also think this change could help a few anonymous users to have better error handling in the future.

H--o-l:
Hey @garethbowen !

Thanks for the feedback.

> I think we want to keep the if (typeof error[p] !== 'function') { check just in case.

Ok good idea, I put it back.

> Have you tested this out? Does it work as expected?

I should have done it before right, here it's a simple test and its results (which I find conclusive):
```html
<html>
  <script src="packages/node_modules/pouchdb/dist/pouchdb.js"></script>
  <script>
    var localDB = new PouchDB('mylocaldb')
    var remoteDB = new PouchDB('http://localhost:5984/myremotedb')
    localDB.replicate.to(remoteDB).on('error', function (err) {
      console.warn('failure', err)
      // Before:
      //   CustomPouchError {
      //     result: {
      //       doc_write_failures: 0,
      //       docs_read: 0,
      //       docs_written: 0,
      //       end_time: "2020-05-15T09:25:06.778Z",
      //       errors: [],
      //       last_seq: 0,
      //       ok: false,
      //       start_time: "2020-05-15T09:25:06.762Z",
      //       status: "aborting",
      //     }
      //   }
      // After:
      //   CustomPouchError {
      //     message: "Failed to fetch",
      //     result: {
      //       doc_write_failures: 0,
      //       docs_read: 0,
      //       docs_written: 0,
      //       end_time: "2020-05-15T09:21:19.426Z",
      //       errors: [],
      //       last_seq: 0,
      //       ok: false,
      //       start_time: "2020-05-15T09:21:19.409Z",
      //       status: "aborting",
      //     },
      //     stack: "TypeError: Failed to fetch"
      //   }
    });
  </script>
</html>
```
I add this test example inside the commit message (a shorter version)

PR #8027, opened 2020-03-10
Added null check for value in getFieldFromDoc in pouchdb selector core utils

t-cho1:
Screen Shot 2020-03-10 at 1 38 34 PM
Sometimes `value` would be undefined, so in the for loop, it would be looking for `value[key]`, which spits out `Cannot read property <property> of undefined`


gaodeng:
@daleharvey 
Can you help merge this PR ?

dheimoz (approved):

gaodeng (approved):

dheimoz:
hello @t-cho1 I am terribly sorry for the time that it has taken to get  your contribution reviewed. As you have seen, tests has ben migrated to GitHub Actions. I wonder if you can get the latest changes and apply your contribution and check should tests passed?

Thanks.

AlbaHerrerias:
Hey @t-cho1 , could you rebase from master? It will allow me to trigger Github actions. Thanks!

rasgo-cc:
Any chance to get this reviewed? 
`getFieldFromDoc` is still crashing whenever the doc is null / undefined

garethbowen:
@rasgo-cc It seems like @t-cho1 is not responding so nobody can update the PR to run the GH actions which is required to merge. Would you mind creating a new PR from `master` which this change? Then we can close this one.

packages/node_modules/pouchdb-selector-core/src/utils.js
```diff
@@ -4,6 +4,9 @@ import { clone } from 'pouchdb-utils';
 // can be "deep" due to dot notation
 function getFieldFromDoc(doc, parsedField) {
   var value = doc;
+  if (!value) {
+    return
```
SourceR85, line 8: It makes more sense to me to define the return type explicitly:
```suggestion
    return undefined;
```

SourceR85 (changes requested):

SourceR85:
Assumption is wrong; this would break pouchdb.
Since it needs more work and the author isn't active, I suggest to abandon this patch.

PR #8033, opened 2020-03-29
Fix react-native Fast Refresh error.

amazingmarvin:
When using pouchdb with react-native 0.61's Fast Refresh, an error is
thrown when this `Object.defineProperty` is called again during fast
refreshes.


amazingmarvin:
How do I re-open this?

PR #8062, opened 2020-05-08
If it is still true, let's make it obvious!

bean5:
If it isn't true, is there a more updated "not a blog post"?


daleharvey:
Thanks for the PR, but I think we only want notifications if there are things that are out of date, otherwise we kinda need to ekep adding this notice repeatedly forever

PR #8066, opened 2020-05-11
Improved reproduction for #7810

pubkey:
This PR adds an improved test ( previous #7895 ) which reproduces #7810

PR #8068, opened 2020-05-12
Fix pouchdb-find inconsistencies - issue #7810

mardii:
## Yet another PR for issue #7810
This PR (imho) replaces #8066 and #7895.

I've took the work @pubkey on the issue, and took the liberty to expand on it. Hopefully adding value.

I spend quite some time to get to understand the _pouchdb-find_ module (or couchdb/pouchdb for that matter).
I added some more test cases for the issue, hopefully covering the desired behavior when it come to the results certain selectors should give.

All of the tests written test the behaviour on both a db with indexes, and without.
Also the testsuite now runs on the remote as well, to make sure the local and remote behave similar as well.

## How to test

Make sure you have CouchDB installed and you ran `npm install` (see [TESTING.md](https://github.com/pouchdb/pouchdb/blob/master/TESTING.md)). 
Then run:
```bash
TYPE=find PLUGINS=pouchdb-find npm run dev
```

To run all the tests in the browser visit [go here](http://127.0.0.1:8000/tests/find/?adapter=idb&plugins=pouchdb-find)

Or only the new test suite from this PR [go here](http://127.0.0.1:8000/tests/find/?adapter=idb&plugins=pouchdb-find&grep=test%5C.issue7810%5C.js)

## Feedback

Hopefully this is a stab in the right direction. But by all means feel free to share light how to improve (Code-wise, problem-solving-wise, but process-wise as well).


vazra:
@mardii  are you working on this?

daleharvey:
Apologies for not looking earlier, but this PR is looking really good, if you are going to finish it up can help out with the failing tests, right now it looks like you might have just got hit with couchdb failing to start for no reason to do with your code, restarted them, there was one issue of `testUtils` not being defined, will see if that happens again

daleharvey:
That the error talks about `testUtils` is confusing, but fairly sure it will be down to the use of `Object.assign` and `Object.values` in the PR that arent supported in ie.

I am up for revisiting how we support older browsers but I think those functions are quite easily replaced for the time being I think there is an `extend` function in pouchdb-utils that can do what `Object.assign` does

daleharvey:
@mardii @pubkey either of you wanna push this one over the line? I can clean it up if not just dont wanna take the attribution from the work youve both put into it

mardii:
> @mardii @pubkey either of you wanna push this one over the line? I can clean it up if not just dont wanna take the attribution from the work youve both put into it

I'm looking into it. Sorry it took so long to respond.

mardii:
@daleharvey I think the usage of Object.assign is allowed (it's used in different places I saw). But indeed Object.values seems not to be supported. So I replaced it with `Object.keys(myObject).map(key => myObject[key])`.
Although, now I wonder if `[].some` will give me similar issues. How do I reproduce the test to fail locally? Run the test on an old version of node? (I'm on a Mac).

daleharvey:
The eaiest way to know for sure is to run the virtual machines microsoft provide, https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/

However that can be a hassle, another option is to edit .travis.yml and delete all the test runs aside from the one you care about, debug via travis and restore travis.yml when you get it working

mardii:
@daleharvey The problem was indeed `Object.values`, and I used arrow functions, which IE 11 did not like. So I fixed those, And now all tests seem to pass except for [.40](https://travis-ci.org/github/pouchdb/pouchdb/jobs/702893519). Is that just a fluke? Because both [.39](https://travis-ci.org/github/pouchdb/pouchdb/jobs/702893518) and [.41](https://travis-ci.org/github/pouchdb/pouchdb/jobs/702893518) did pass (same basic tests, just different node versions), and so did [.1](https://travis-ci.org/github/pouchdb/pouchdb/jobs/702893480) which is exactly the same as [.40](https://travis-ci.org/github/pouchdb/pouchdb/jobs/702893519), isn't it?
So what I am asking, am I missing something? I went through the logs of both Travis' subtasks, but can't find a clue as to what I did wrong.

daleharvey:
@mardii Awesome, the failure was just an intermittent, passed when I retriggered @ https://travis-ci.org/github/pouchdb/pouchdb/builds/702893479

daleharvey:
So I think we can drop the ES5 requirement but I would like to do it as part of https://github.com/pouchdb/pouchdb/issues/8097 and not rush it.

However thanks so much for sticking with this, I merged all of your commits into 2 seperate ones and landed in 
https://github.com/pouchdb/pouchdb/commit/ac03f8a6a6407e38de2b089a80755b15265a5b03
https://github.com/pouchdb/pouchdb/commit/2a064045f2a27d4d4fe697fee85c614ab260094e

Cheers

PR #8069, opened 2020-05-12
    Fix warning no index found message (altered in couchbase 3.1)

mardii:
In this [commit](https://github.com/apache/couchdb/commit/f86747f02a8b066f136a275fa28e92200404d0e7) CouchDB changed the formatting of the `no matching index found, create an index to optimize query time` warning message.

This PR applies the same change to the _pouchdb-find_ implementation, and updated the corresponding tests to support both old and new message. That way both CouchDB <3.1 and >3.1 are supported to test against.

Updated corresponding documentation as well.

There are a couple of tests in pouchdb-find that use the No matching index found warning. We need to support both variations (CouchDB pre 3.1 and post 3.1).
I did not change the tests themselves. Most of them are still skipped, or not even part of the describe scope, therefor not run.

The pouchdb-find test suite and code still needs work, imo. But that's for another PR.

### DIFF
--- a/docs/_guides/mango-queries.md
+++ b/docs/_guides/mango-queries.md
@@ -314,7 +314,7 @@ You may also want to pay attention to the `"warning"` value included in your res
 ```js
 {
   "docs": [ /* ... */ ],
-  "warning": "no matching index found, create an index to optimize query time"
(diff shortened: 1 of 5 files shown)


garethbowen:
Merged in #8073

PR #8071, opened 2020-05-13
fix(package): update leveldown to version 5.6.0

julianhille:
Closes #8017

Can we have this? So that 32bit windows builds are working again?


garethbowen:
Good call! There are still some builds failing but this fixes some of them.

PR #8073, opened 2020-05-14
Fix testrunners of couchdb v2 and add couchdb v3

julianhille:
Lets see if it builds.
If it does it fixes some build issues with COUCH_HOST

This would need another build to ensure tests against v3 (latest).
But for a couch v3 we need to wait for: https://github.com/pouchdb/pouchdb/pull/8069

Merged 8096 to see if this builds with v3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,6 +66,7 @@ env:
 
   # Test pouchdb-find
   - COUCH_HOST=http://127.0.0.1:3001 TYPE=find PLUGINS=pouchdb-find CLIENT=node SERVER=couchdb-master COMMAND=test
+  - COUCH_HOST=http://127.0.0.1:3002 TYPE=find PLUGINS=pouchdb-find CLIENT=node SERVER=couchdb-v2 COMMAND=test
   - TYPE=find PLUGINS=pouchdb-find CLIENT=selenium:firefox ADAPTER=idb SERVER=pouchdb-server COMMAND=test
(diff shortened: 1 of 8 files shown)


julianhille:
Last build was succesfull besides one test having a timeout.
This tests now couchdb v3 and v2 successful

julianhille:
@garethbowen this should fix all test failures, but it now failes with a `pouchdb.min.js` size failure and all the tests according to that file fail too. The size is "lower then 20" so this might be a caching issue.
My local builds do not have that issue, the personal biulds on travis had the same until "cache clear and rerun" maybe there is a similar issue here.

Please take a look if you can see something what might broke the `pouchdb.min.js` or clear cache and rerun those tests.

julianhille:
@daleharvey may you reset caches and rebuild? This should fix all test fails.

garethbowen:
Caches cleared and build restarted.

julianhille:
rebased on master

julianhille:
@garethbowen please restart again its running but one job was too slow on sauce labs ios.
this PR makes Travis green again.

garethbowen:
Restarted.

julianhille:
These sauce labs tests are really flaky. I'd say take a look and maybe merge cause it fixes Travis mostly.

garethbowen:
Agreed - it's better than `master`. Merging.

garethbowen:
Merged. Thanks!

PR #8074, opened 2020-05-16
move txn before .then()

dholth:
Seems to help Safari


daleharvey:
Fixed in https://github.com/pouchdb/pouchdb/pull/8095

PR #8077, opened 2020-05-19
Fix use regular `for` loop for IE

H--o-l:
Commit 7d058496 _"Fix inheritance of error properties"_ use a `for ... of` loop
which is not available on IE.

This new version have been tested manually on IE 11.


H--o-l:
Sorry I'm fixing my own mistake here, introduce in commit 7d05849

PR #8087, opened 2020-06-02
updated http-server to latest 0.12.3

Neilblaze:
devDependency 'http-server' was updated from 0.12.0 to 0.12.2.


daleharvey:
Sorry didnt realise this was a person, usually a bot does out updates

daleharvey:
And looking at the failured, it was only iphone failed so going to merge and maybe take the iphone run out of there

Issue #8084, opened 2020-05-27
Safari can't open transaction after .then() for pouchdb-adapter-indexeddb


dholth:
See https://github.com/pouchdb/pouchdb/pull/8074

daleharvey:
And merged fix, cheers

PR #8095, opened 2020-06-19
(#8084) - Move transaction before .then

daleharvey:


daleharvey:
Just cleaned up https://github.com/pouchdb/pouchdb/pull/8074 so can merge without waiting

PR #8109, opened 2020-06-21
issue#-8096

leroy-douglas:
Changed to test for abscence of the nodejs global object


daleharvey:
Awesome thanks, so the test suite seems to be happy with that change and it makes sense, however I think it would be cleaner if we started removing some of these build quirks instead of of adding tests for them, `global` came around because we were using browserify to build which mimicked a node environment so we could use the same code in node + browser. 

However this code is specifically only supposed to work in the browser, so do you want to try replacing `global.` with `window.` and see how the tests like that?

daleharvey:
Apologies that one was my mistake, while this is to be used in the browser it is run inside a webworker so there is no `window`, could you try replacing `window.` with `self.`

And sorry if my previous comment was unclear but we want to take the check out, it is not clear why it is needed and so when there is a doubt if some code is needed its best to try and leave it out

Issue #8089, opened 2020-06-07
Fix EventEmitter import 


lucidNTR:
The browser releases
https://unpkg.com/pouchdb@7.2.1/lib/index-browser.es.js
and 
https://unpkg.com/pouchdb-browser@7.2.1/lib/index.es.js

have the following defects:

- import statement for EventEmitter is wrong:
the events package exports EventEmitter as the default but it is  importet as a named import
fix: change `import { EventEmitter } from 'events';` to `import EventEmitter from 'events';`

- a line that was supposed to be removed in the browser release somehow got into the browser module:

search in the referenced files for 
```
// this is not used in the browser [sic!]

var setImmediateShim = global.setImmediate || global.setTimeout;
```

this will obviously break as global is undefined in a browser

fix: optimally fix the release system or for a quick fix, change the line to:
var setImmediateShim = global ? (global.setImmediate || global.setTimeout) : setTimeout;

daleharvey:
So that `global` issue should be fixed, as for the EventEmitter import, looks like a few uses across the code

```
~/src/pouchdb (master)$ grep -r "EventEmitter" ./packages/node_modules/*/src/**
./packages/node_modules/pouchdb-core/src/adapter.js:import EventEmitter from 'events';
./packages/node_modules/pouchdb-core/src/changes.js:import { EventEmitter as EE } from 'events';
./packages/node_modules/pouchdb-core/src/setup.js:import { EventEmitter as EE } from 'events';
./packages/node_modules/pouchdb-replication/src/replication.js:import { EventEmitter as EE } from 'events';
./packages/node_modules/pouchdb-replication/src/sync.js:import { EventEmitter as EE } from 'events';
./packages/node_modules/pouchdb-utils/src/changesHandler.js:import { EventEmitter as EventEmitter } from 'events';
./packages/node_modules/pouchdb-utils/src/listenerCount.js:import { EventEmitter as EE } from 'events';
```

lucidNTR:
thanks! any idea why it does not break on some of those? maybe a version mismatch / events changed the export format?

daleharvey:
I expect similiar to the global issue, the build system does some magic to try and handle both cases and that magic doesnt get applied to the ESM builds. I did a little test and removed the brackets around one of the imports and the tests were still passing.

Ultimately we need to be testing the ESM published module which I dont think we are but will keep that in mind as part of https://github.com/pouchdb/pouchdb/issues/8097

[...]

PR #8122, opened 2020-06-25
update EventEmitter imports to use the default export

jshemas:
issue: https://github.com/pouchdb/pouchdb/issues/8089


jshemas:
`force-pushed` was for fixing the commit message.

PR #8128, opened 2020-07-01
Emit checkpoints

chrisekelley:


daleharvey:
Hey @chrisekelley thanks for the PR :)

Can you give a bit of an explanation as to what you need these events for? While adding events is a simple change, they need to be tested, documented, people will see the events and have their own use cases which can get complicated very quickly so we want to be very sure when we add anything that its commonly required and conceptually simple.

chrisekelley:
Ooops - sorry about that - I was simply trying to do a compare during a conf call and didn't mean to do that here! 

Since the cat is out of the bag, here is the back story: Our project does data collection on low-memory, low-bandwidth Android tablets using an encrypted sqlite db. When we do a big sync (~5000 docs) to our master Couchdb from a tab, all is good, but subsequent syncs take A VERY LONG TIME. It seemed like a lot of mysterious stuff was happening during the diffing of the _rev's and checkpoint phase,so I thought it might be useful to emit some checkpoint-related events. The emits in my PR are my first try. 

Anyway, this may be completely unnecessary. The fact that the app is spending so much time w/ checkpointing may indicate that I've screwed up somewhere. But it might be useful to emit some of these events anyways. I'll close this PR - it is waaay too early, and if it makes sense, and might be useful to other folks, I'll re-open it. And add some decent docs. And tests. etc 

:-)

PR #8140, opened 2020-07-17
(#8134) - filter with regex for multifield indexes

### LINKED ISSUE #8134: [Find plugin] with multiple indices: TypeError: Cannot use 'in' operator to search for 'startkey' in undefined
### Issue

Error when trying to filter DB with regex and more than one index:
```
TypeError: Cannot use 'in' operator to search for 'startkey' in undefined
```
```js
  db.createIndex({
    index: {
      fields: ['name', 'type'],
    },
  });
...
  // thorws error!
  const result = await db.find({
    selector: {
      name: {
        $regex: new RegExp('bar', 'i'),
      },
    },
    sort: ['name'],
  });
```
### Info
- Environment: browser
- Platform: Chrome
- Adapter: IndexedDB - default
- Server: -

### Reproduce

Reproduced here https://pouchdb-find-issue.glitch.me/

darkowic:
Fixes #8134

### DIFF
--- a/packages/node_modules/pouchdb-find/src/adapters/local/find/query-planner.js
+++ b/packages/node_modules/pouchdb-find/src/adapters/local/find/query-planner.js
@@ -368,11 +368,10 @@ function getMultiFieldQueryOpts(selector, index) {
     if (!matcher || !Object.keys(matcher).length) { // fewer fields in user query than in index
       finish(i);
       break;
+    } else if (Object.keys(matcher).some(isNonLogicalMatcher)) { // non-logical are ignored
+      finish(i);
+      break;
     } else if (i > 0) {
-      if (Object.keys(matcher).some(isNonLogicalMatcher)) { // non-logical are ignored
-        finish(i);
-        break;
-      }
       var usingGtlt = (
         '$gt' in matcher || '$gte' in matcher ||
         '$lt' in matcher || '$lte' in matcher);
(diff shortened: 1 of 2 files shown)


daleharvey:
Hey, thanks for the PR! It looks great, thanks a lot for the new test. Codewise its not the area I am most familiar with but happy to depend on the tests here.

However this failed tests when running against CouchDB, I have restarted the tests in case they were intermittend failures however seems likely that the failures were genuine. Can you take a look at the failures / running this test against CouchDB. We very much like to maintain the same behaviour between the 2 implementations. If its also a bug that CouchDB needs fixed then great we can fix it in both places.

darkowic:
The error is `No index exists for this sort, try indexing by the sort fields.`. The error is gone when I either add filtering by the second index:
```js
        return db.find({
          selector: {
            name: {$regex: "^Luig"},
            debut: {$gt: 1980}
          },
          sort: ['name']
        }).then(function (resp) {
```
or remove the `sort` from find:
```js
        return db.find({
          selector: {
            name: {$regex: "^Luig"},
          },
        }).then(function (resp) {
```
for now, I will just add filtering by the second field. It is another, unrelated issue with compatibility with CouchDB implementation. Is it ok for you?

daleharvey:
@darkowic That looks great and all passing now, thank you :)

PR #8165, opened 2020-08-27
Initial dollar sign escaped with unit tests.

filionf:
I am not sure if I have done everything properly, please let me know.

The unit test does not work when using pouchdb-server, but I suspect this is a difference between pouchdb-server and the real couchdb.

### DIFF
--- a/packages/node_modules/pouchdb-selector-core/src/utils.js
+++ b/packages/node_modules/pouchdb-selector-core/src/utils.js
@@ -33,15 +33,12 @@ function parseField(fieldName) {
   var current = '';
   for (var i = 0, len = fieldName.length; i < len; i++) {
     var ch = fieldName[i];
-    if (ch === '.') {
-      if (i > 0 && fieldName[i - 1] === '\\') { // escaped delimiter
-        current = current.substring(0, current.length - 1) + '.';
-      } else { // not escaped, so delimiter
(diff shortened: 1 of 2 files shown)


daleharvey:
Hey cheers for this, and sorry for the delay

This isnt the area of code I am most familiar, however it looks like just above the clause you added there is the exact same clause but for . instead of $, I think we would probably want to have a single bit of code handle the escaping for both situations?

filionf:
@daleharvey Done, I find this version clearer. Any idea why all the builds are failing?

daleharvey:
Thanks, I think the build was failing due to saucelabs issues and nothing to do with your code, if you can rebase that would be great then I think this code would be good to go :) Apologies for the delay, working on making sure that doesnt happen again

daleharvey (changes requested):
Meant to flag as requesting changes to keep track

daleharvey:
Hrm, the test was failing on mapreduce tests but only in firefox + one adapter, not convinced the failure was down to this patch but couldnt restart, reopened is runnin gthe tests on .com so will see with that

daleharvey:
And cheers for the rebase

daleharvey:
ok and all passing now, dunno whats up with .org but new PR's shouldnt hit this issue.

Many thanks for the PR 👍

Issue #8167, opened 2020-09-07
Unable to build PouchDB documentation


waitem:
### Issue
I was unable to build the documentation following the instructions in the [Building PouchDB Documentation section in the  CONTRIBUTING guide](https://github.com/pouchdb/pouchdb/blob/master/CONTRIBUTING.md#building-pouchdb-documentation).

I will raise a PR to show how I resolved the issue (which included reverting to jekyll version 3.9)

### Info
- Environment: Ubuntu 20.04

### Reproduce
Follow instructions on the CONTRIBUTING.md page

shadow81627:
https://github.com/pouchdb/pouchdb/issues/8149

PR #8168, opened 2020-09-07
(#8167) - revert Jekyll to 3.9.0 to be able to build doumentation

waitem:
And also update domentation for building  on Ubuntu


daleharvey:
Thanks for this, hopefully can finish the switch away from Jekyll as well

PR #8172, opened 2020-09-09
(#8171) - Allow _access field in documents

### LINKED ISSUE #8171: Allow special document member _access
### Issue
With the release of [v3, CouchDB](https://docs.couchdb.org/en/latest/whatsnew/3.0.html#id2) allowed the the [special document member `_access`](https://github.com/apache/couchdb/pull/2388). For the moment it does nothing, but in the future it will be used for [per-document access](https://github.com/apache/couchdb/issues/1524).

PouchDB is one of the core use-cases for _per-document access_ (as described in the [RFC's Goals](https://github.com/apache/couchdb-documentation/blob/rfc/010-per-document-access/rfcs/010-per-document-access-control.md)). But it would require allowing `_access` in PouchDB, too.

Full access control would only be required within the [pouchdb-server](https://github.com/pouchdb/pouchdb-server) packages. For the browser and node.js it would be enough to allow `_access` just like CouchDB did with v3 (add it to [`reservedWords` list in pouchdb-adapter-utils/src/parseDoc.js](https://github.com/pouchdb/pouchdb/blob/e490d46624fc889365e07f7106ce0ea5d7f8053a/packages/node_modules/pouchdb-adapter-utils/src/parseDoc.js#L16))

### Info
- Environment: Node.js, browser, hybrid app
- Platform: All clients (Chrome/FF/Safari/Edge/iOS/Android/etc.)
- Adapter: All but pouchdb-adapter-http (IndexedDB/WebSQL/LevelDB/etc.)
- Server: CouchDB, Cloudant & etc.

### Reproduce
Create a document with an `_access` field in CouchDB v3 and in PouchDB.
- It will succeed in CouchDB >=v3.
- It will fail in PouchDB and CouchDB < v3.

```json
{
  "_id": "something",
  "_access": ["alice"]
}
```

Terreii:
Closes #8171.

This pull request adds the special document member field `"_access"` to `reservedWords` and `dataWords` in _pouchdb-adapter-utils_.

The objective is to store `_access` alongside all other fields, to be compatible to the _per document access_ RFC and future versions of CouchDB.
CouchDB v3.0 does store `_access`, too.


Terreii:
😕 this did work on my machine …

Terreii:
I did look into the failing tests. They all are one of two errors-types: Either the __http-adapter__ fails, or the test can't connect to a browser.

Given that the http-adapter depends on the used CouchDB/PouchDB-Server version (and it passing everything to it), it might be better to opt-out of testing for `_access` with the http-adapter. And add a warning in the release notes that replication of an access-enabled db between CouchDB v3 <-> PouchDB <-> CouchDB v2 will cause an error.

This pull request focus is to be compatible to future versions of CouchDB and ensure the the local adapters don't throw on `_access`.

Terreii:
🎉

daleharvey:
Awesome thanks for the rebase, one minor issue is that using promises instead of callback is preferred as its cleaner, at some point we will probably make the jump to async / await, however thats just a note for future it doesnt need to block this which looks good to me, cheers

daleharvey (approved):

PR #8187, opened 2020-09-23
Don't include leveldown in browsers

RangerMauve:
I'm getting errors trying to use this in browser environments due to the migrate.js script always requiring leveldown.

This will cue Webpack and Browserify to ignore that module.

### DIFF
--- a/bin/build-module.js
+++ b/bin/build-module.js
@@ -30,6 +30,9 @@ var AGGRESSIVELY_BUNDLED_PACKAGES =
 // packages that only have a browser version
 var BROWSER_ONLY_PACKAGES =
   ['pouchdb-browser'];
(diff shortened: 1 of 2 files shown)


daleharvey (changes requested):
This is part of the codebase I am least familiar with and will be looking to be doing major improvements on soon, but if it fixes it for you and doesnt break the tests I will be happy to merge.

The tests were broken at the time you submitted but if you rebase we should be able to see if its green. Sorry for the delay and thanks for the patch

RangerMauve:
@daleharvey Thank you for your patience, just did a rebase.

daleharvey:
No worries thanks for the update, all passing now 👍

daleharvey (approved):

PR #8191, opened 2020-09-24
8190 optimized filter by id

jeffsenn:
Copied from issue description.

# Issue
A replicate.from feed with a client side filter from a large/active remote when you just want to keep existing (already fetched) documents up-to-date is extremely inefficient.

# Info
When you specify filter:a_function on a replicate, PouchDB forces the changes feed (inside the replication) to specify include_docs:true. This will result in all remote changes coming over the wire. Normally this is what you want, however in the (extremely interesting) case of only wanting to keep documents you have already fetched up-to-date, it is very inefficient -- all you really need is the document _id.

With 2 small changes this can be made much more efficient.

One user visible API change: support a filter_id_only: true option to replicate. If specified this suppresses the additional include_docs:true and sends the change to the filter function as {_id: doc_id}

I will followup with a PR in case you want to include this feature

### DIFF
--- a/packages/node_modules/pouchdb-replication/src/replicate.js
+++ b/packages/node_modules/pouchdb-replication/src/replicate.js
@@ -421,8 +421,10 @@ function replicate(src, target, opts, returnValue, result) {
         };
         if (opts.filter) {
           if (typeof opts.filter !== 'string') {
-            // required for the client-side filter in onChange
-            changesOpts.include_docs = true;
+            if(!opts.filter_id_only || opts.include_docs) {
+              // required for the client-side filter in onChange
+              changesOpts.include_docs = true;
+            }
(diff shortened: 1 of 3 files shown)


jeffsenn:
CI Failures do not look like they are due to this PR.

janl:
why would you use a filter instead of the `doc_ids` option for replication?

lucidNTR:
@janl i think i understand the intend. with the doc_ids option it is only possible to filter for up front known doc ids. what @jeffsenn wants to do is check live in the filter function if an incoming doc id from the feed is an id that is in the local pouch without getting alls docs bodies. for dbs where there are many changes and large docs of which you just want a small subset locally, this could make quite a difference. is there any other way to accomplish this?

PR #8198, opened 2020-09-30
Handle non-buffer binary data in node

cortopy:
The difference between browser and node generally works well. However, there are some instances in which the boundaries are not that clear. For example:

- Jest runtime that loads jsdom in the background
- Server side rendering of single-page applications (that may also rely on some kind of DOM implementation in node to simulate a browser-like environment)

I bumped into this issue when testing attachments in my application. I believe these changes fix the issue.

Apologies for not adding any tests. This is my first contribution and for what I can see, node tests are run with mocha and browser tests with a headless browser.

If tests are required, I can add them but I'd need to know which would be the best way of replicating those two use cases above in current tests setup

### DIFF
--- a/packages/node_modules/pouchdb-adapter-leveldb-core/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-leveldb-core/src/index.js
@@ -781,13 +781,27 @@ function LevelPouch(opts, callback) {
           // small optimization - don't bother writing it again
           return callback(err);
         }
-        txn.batch([{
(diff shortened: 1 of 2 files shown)


janl:
we’d need tests, and no CI build failures :)

AlbaHerrerias:
Hey @cortopy, could you rebase from master? It will allow me to trigger Github actions. Thanks!

alxndrsn:
No action for 5 years.

PR #8204, opened 2020-10-04
pouchdb#8203

sennahLessert:
i added package-lock.json to .gitignore file


daleharvey:
As far as I can tell we are supposed to actually commit package-lock.json, I havent yet as I wasnt certain but everywhere I see comment on it seems to agree so closing this out sorry, if you wanted to make a PR to add it that would be great

Issue #8209, opened 2020-10-08
indexeddb adapter can break when opened in multiple tabs


srova:
### Issue
We have an application using pouchDB to make offline functionality possible, in live replication with a remote CouchDB cluster.
We recently moved from idb adapter to indexeddb adapter in order to perform better with native IdB indexes.
The app performance are now ok, but we have some cases where the end users experiment occasionally the following error:

Database has a global failure

launched by the new indexededb adapter in the idb.onabort callback.
When this happens, the download replication stream crashes, instead the upload one and the app local data browsing still work.
In all cases we had to start a fresh download of all local data from server to overcome this issue
Any ideas of possible causes or what to check?
Thanks

### Info
- Environment: (Node.js/browser/hybrid app.)
- Platform: (Electron app)
- Adapter: (IndexedDB)
- Server: (CouchDB)

srova:
An other addition info: the crash described above is preceeded by the following error, launched inside pouchdb.min.js:

Cannot read property filter of undefined

Unfortunatly since the js is minified and we are in production enviroment we are not able to retrieve any forther detail right now

elaynelemos:
@geidsonc

srova:
For anyone interested we were finaly able to figure out what was happening:
- our remote CouchDB database contains many attachments
- when running in the electron app, sometimes (rarely - but our application is used every day by dozens of end users), an attachment download fails, and a txn.abort() is called (we still have not understood why it fails writing locally the attachment)
- this causes that the onAbort callback inside **pouchdb-indexeddb** is called, but unfortunatly the error variable is null, so the callback seems to be invoked for a successful result, even if the results array is undefined 


```
preProcessAttachments().then(function () {
    api._openTransactionSafely([DOC_STORE, META_STORE], 'readwrite', function (err, _txn) {
....
      txn.onabort = function () {
        callback(error);
      };
....

```

- this causes an uncatched exeption in **pouchdb-core, at adapter.js line 887** (trying to call .filter on res, that is undefined, causing the downstream syncronization to freeze

```
  return this._bulkDocs(req, opts, function (err, res) {
    if (err) {
      return callback(err);
    }
    if (!opts.new_edits) {
      // this is what couch does when new_edits is false
      res = res.filter(function (x) {
        return x.error;
      });
    }
    // add ids for error/conflict responses (not required for CouchDB)
    if (!isRemote(adapter)) {
      for (var i = 0, l = res.length; i < l; i++) {
[...]

PR #8211, opened 2020-10-16
Typo in bulkDocs.js

msotnikov:
Error is not defined. I guess it was meant the `txn.error`

### DIFF
--- a/packages/node_modules/pouchdb-adapter-indexeddb/src/bulkDocs.js
+++ b/packages/node_modules/pouchdb-adapter-indexeddb/src/bulkDocs.js
@@ -372,7 +372,7 @@ export default function (api, req, opts, metadata, dbOpts, idbChanges, callback)
       txn = _txn;
 
       txn.onabort = function () {
-        callback(error);
+        callback(txn.error);
       };
       txn.ontimeout = idbError(callback);
 


msotnikov:
https://github.com/pouchdb/pouchdb/issues/8209

msotnikov:
Are there maintainers here?

daleharvey (changes requested):
Hey. Very sorry for the delay in reviewing this, had a bit of a break for a while but will be looking to keep up with upcoming changes.

Good catch, the change looks good to me I can see other idb code accessing txn.error. However the repo was broken when you submitted this which is why the tests arent passing. I think this is ok to merge but would like to make sure with a green build so if you could rebase this PR it would be a great help, thanks

daleharvey:
So error here actually is defined @ https://github.com/pouchdb/pouchdb/blob/master/packages/node_modules/pouchdb-adapter-indexeddb/src/bulkDocs.js#L28

Its assigned a value in some error conditions which you can see @
https://github.com/pouchdb/pouchdb/blob/master/packages/node_modules/pouchdb-adapter-indexeddb/src/bulkDocs.js#L244

This is to ensure the transaction has closed before we return, there are situations where the transaction can abort and the error is null, in that case we should probably fall back better, but this isnt correct as is

msotnikov:
May be solution is wrong, but error still occurs.

Issue #6260, opened 2017-02-19
and/or with regex on same field


nolanlawson:
_From @valepu on October 20, 2016 11:47_

Hello,
I have tried to use both an $and and an $or including a set of conditions ($regex) all on the same field, but it appears that only the last regex gets used (both using $and and $or)

Here's a plunker showing the issue: http://plnkr.co/edit/9N0xuC6k5iwPrY8QMca2?p=preview
This should show no results but it shows all of them instead, if you move the 2nd regex at the end it'll show only 1 result (the one starting with a b)

Checking a regex on different fields will work, but if the same field is checked multiple times then only the last regex will work

I'm not sure it's intended or not, i haven't found anything on the Cloudant API documentation but i admit i haven't read it thoroughly (although i've seen an example of an $or checking the same field multiple times, even though it was not a regex)

_Copied from original issue: nolanlawson/pouchdb-find#219_

LivingOnAPear:
I patched this locally on my version.  My fix is not very good but hopefully it will point someone who knows enough of the internals to do it right in the right direction.

The issue is that the mergeAndedSelectors function creates an js object with keys of the field names e.g.

query
```
{
  "$and":[
    {"field1":{"$regex":"a"}},
    {"field2":{"$regex":"b"}},
    {"field3":{"$regex":"c"}},
  ]
}
```

requestDef.selector becomes
```
{
  "field1":{
     "$regex":"a"
  },
  "field2":{
     "$regex":"b"
  },
  "field3":{
     "$regex":"c"
  }
}
```

Hence, if you have the same field name with the same operator in an $and you will only ever get the last value.  The right solution I am sure is to make an array under the operator level in requestDef.selector, but I don't know enough about the code to do that safely.  This is what I did:

Changed:
`fieldMatchers[operator] = value;`

to:
```
if(!fieldMatchers.hasOwnProperty(operator)){
  fieldMatchers[operator] = value;
  }else{
    var w = 1;
    while(fieldMatchers.hasOwnProperty(operator+"_"+w)){
      w+=1;
    }
  fieldMatchers[operator+"_"+w] = value;
}
```

suffixing each repeat operator with _(int)

then in function match() i just added this to the top:
`userOperator = userOperator.replace(/(_[0-9]+$)/,"");`

It is very dirty, but it seems to work well.  I was mostly interested in $and but it seems to work for $or as well.

mrzslr:
[...]

PR #8223, opened 2020-11-13
(#8222) - fix multiple $regex conditions on same field when using $and

### LINKED ISSUE #8222: pouchdb-find, $and and multiple $regex conditions
### Issue
using pouchdb with pouchdb-find, i can prepare a selector with two or more `$regex` conditions in an `$and` array and only the *last* of those conditions is ever respected.

### Info
- Environment: plain node (v12.18.4) script
- Platform: Windows
- Adapter: seems to affect both IndexedDB in Electron as well as LevelDB in node
- Server: pouchdb embedded code

### Reproduce
1. set up a db
2. insert some docs with an attribute that has minor overlap
3. create a selector similar to this in structure:
```javascript
{
  '$and': [
    { text: { $regex: '.*foo.*' } }, // condition1
    // [...] further conditions
    { text: { $regex: '.*bar.*' } } // conditionN
  ]
}
```
4. `condition1`...`conditionN-1` are ignored, only `conditionN` is used to return documents

i wrote a small example reproducing the behavior [here](https://gist.github.com/bogenpirat/63c5452cd43b7cbd10f6df7ffcf3158f)

is this a bug or limitation? is there a workaround?

AlansCodeLog:
Fixes #8222 / #6260

This takes the same approach as the `$ne` operator, placing the values to match in an array so they must all match if `$and` is used.

One thing I noticed but did not fix is that in the [couchdb docs](https://docs.couchdb.org/en/stable/api/database/find.html#condition-operators) it says `$regex` can only match against strings. But there is no check for this, regexes have always been able to match against whatever javascript can match. I looked quickly at the other operators, and for example, it seems `$size` too, does not check it's value is an array. Is a fix to conform more to how couchdb handles the conditions wanted? I can submit a separate PR.

### DIFF
--- a/packages/node_modules/pouchdb-selector-core/src/in-memory-filter.js
+++ b/packages/node_modules/pouchdb-selector-core/src/in-memory-filter.js
@@ -291,7 +291,9 @@ var matchers = {
   },
 
   '$regex': function (doc, userValue, parsedField, docFieldValue) {
-    return fieldExists(docFieldValue) && regexMatch(docFieldValue, userValue);
+    return fieldExists(docFieldValue) && userValue.every(function (regexValue) {
+      return regexMatch(docFieldValue, regexValue);
+    });
   },
 
   '$type': function (doc, userValue, parsedField, docFieldValue) {
--- a/packages/node_modules/pouchdb-selector-core/src/utils.js
+++ b/packages/node_modules/pouchdb-selector-core/src/utils.js
@@ -98,6 +98,8 @@ function mergeAndedSelectors(selectors) {
             return mergeNe(value, fieldMatchers);
           } else if (operator === '$eq') {
             return mergeEq(value, fieldMatchers);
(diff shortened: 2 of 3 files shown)


daleharvey:
Yeh if we can align more closely with CouchDB then it leaves less of a chance for users to be surprised when code works with a local backend but not a CouchDB one. 

This code looks good to me, apologies for the delay and the errors, we have removed saucelabs from the travis run so if you rebase then we will have a good idea if this is good to merge. Many thanks

daleharvey (changes requested):
Meant to flag as requesting changes to keep track

daleharvey (approved):

daleharvey:
Awesome thank you

AlansCodeLog:
Thanks and cool, I'll start working on the other stuff and open up an issue to discuss it.

PR #8230, opened 2020-11-20
(#8229) - indexeddb: ensure errors are propagated and handle database deletion

### LINKED ISSUE #8229: New indexedDB adapter: locks itself off if database is manually deleted, doesn't handle transactions exceptions
### Issue
If there's an open database, and user clears the website data & storage (or simply go to DevTools and delete the database directly), upon next transaction, the database gets silently recreated with version=1, which can give problems if there's cached data, and also will fail next time the user tries to reload the website, since the adapter will fail with `Incorrect adapter: you should specify the "idb" adapter to open this DB` due to the wrong version.

Also, on Chrome, the old database object will get closed, which will trigger an error on next time `$t` tries to create a transaction on which. These errors (both on `setup` and on transaction) were silently ignored, which made the respective callback/promise never resolve, not giving the user the opportunity to act over it.

### Info
- Environment: Browsers
- Platform: Chrome & Firefox
- Adapter: new IndexedDB
- Server: no

### Reproduce
- Open a database with `indexeddb` adapter
- Clear website data while db is open
- Try to make any request to it

andrevmatos:
Fixes #8229 

This will prevent the database from being reopened/recreated with `version=1` if it got deleted on Firefox or Chrome, which currently prevents it from ever being opened again.
Also, transactions errors are properly propagated.

### DIFF
--- a/packages/node_modules/pouchdb-adapter-indexeddb/src/setup.js
+++ b/packages/node_modules/pouchdb-adapter-indexeddb/src/setup.js
@@ -115,6 +115,11 @@ function openDatabase(openDatabases, api, opts, resolve, reject) {
       // migrate between the two. In the future, dependent on performance tests,
       // we might silently migrate
       throw new Error('Incorrect adapter: you should specify the "idb" adapter to open this DB');
+    } else if (e.oldVersion === 0 && e.newVersion < versionMultiplier) {
+      // Firefox still creates the database with version=1 even if we throw,
+      // so we need to be sure to destroy the empty database before throwing
+      indexedDB.deleteDatabase(opts.name);
+      throw new Error('Database was deleted while open');
     }
 
     var db = e.target.result;
@@ -146,6 +151,13 @@ function openDatabase(openDatabases, api, opts, resolve, reject) {
       console.log('Database was made stale, closing handle');
       openDatabases[opts.name].versionchanged = true;
       idb.close();
+    };
+
+    idb.onclose = function () {
+      console.log('Database was made stale, closing handle');
+      if (opts.name in openDatabases) {
+        openDatabases[opts.name].versionchanged = true;
+      }
     };
 
     var metadata = {id: META_STORE};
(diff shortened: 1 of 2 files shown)


daleharvey:
Hey so this looks great thanks, ideally we could have a test but startup failures are hard to test. If you could rebase on top of the current master then we should see this green hopefully

andrevmatos:
Hi, @daleharvey . Thanks for the headsup. Done and passing.

daleharvey (approved):

daleharvey:
Awesome thanks so much

packages/node_modules/pouchdb-adapter-indexeddb/src/setup.js
```diff
@@ -115,6 +115,11 @@ function openDatabase(openDatabases, api, opts, resolve, reject) {
       // migrate between the two. In the future, dependent on performance tests,
       // we might silently migrate
       throw new Error('Incorrect adapter: you should specify the "idb" adapter to open this DB');
+    } else if (e.oldVersion === 0 && e.newVersion < versionMultiplier) {
+      // Firefox still creates the database with version=1 even if we throw,
+      // so we need to be sure to destroy the empty database before throwing
+      indexedDB.deleteDatabase(opts.name);
+      throw new Error('Database was deleted while open');
```
karlwestin, line 122: if you don't mind, can i ask how do you catch this error?
the context is that i'm trying to make the frontend code choose between the idb or indexeddb adapter, in order to migrate users from the old to the new

i was hoping to do something like this:
```javascript
// example that did not work like i hoped
async function getDB () {
  var db = new PouchDB('myDb', { adapter: 'indexeddb' })
  try {
      await db.info()
  } catch (e) {
     if (e.message.startsWith('Incorrect adapter')) {
        return new PouchDB('myDb', { adapter: 'idb' })
     }
  }
}

but because the error is thrown, and not used to reject the promise, it seems like i'm not able to catch it, were you able to do this somehow?

andrevmatos: Hi , @karlwestin , sorry for the delay in replying
We don't catch specifically this error, but AFAIR, `onupgradeneeded` exceptions passed through some of these asynchronous calls (I think the first time we tried to access it, but not sure), so we could catch it more or less like you did inside an `async` function and it did work:
https://github.com/raiden-network/light-client/blob/master/raiden-ts/src/db/utils.ts#L44-L69

Issue #8011, opened 2020-02-18
Nested $or not working in pouchdb-find


DiamondYuan:
```
{
  "$and": [
    {
      "$or": [
        {
          "female": {
            "$size": 0
          }
        },
        {
          "female": {
            "$exists": false
          }
        }
      ]
    },
    {
      "$or": [
        {
          "male": {
            "$size": 0
          }
        },
        {
          "male": {
            "$exists": false
          }
        }
      ]
    }
  ]
}
```
But result is 

```
{
  id: '00577033-4ba8-49f2-9de8-e8c5a1f52c75',
  female: [ '030026c4-1284-4c70-8cc3-61f05027f77c' ],
  male: [],
}

**Comment by DiamondYuan** [comment]
use massageSelector 

```
console.log(
  JSON.stringify(
    massageSelector({
      $and: [
        {
          $or: [
            {
              female: {
                $size: 0,
              },
            },
            {
              female: {
                $exists: false,
              },
            },
          ],
        },
        {
          $or: [
            {
              male: {
                $size: 0,
              },
            },
            {
              male: {
                $exists: false,
              },
            },
          ],
        },
      ],
    })
  )
);

```


```
{"$or":[{"male":{"$size":0}},{"male":{"$exists":false}}]}
```

DiamondYuan:
same to https://github.com/pouchdb/pouchdb/issues/6879

DiamondYuan:
@nolanlawson

Issue #7446, opened 2018-08-28
How to match on nested attribute (subfield)?


martinhbramwell:
This is a generic question, independent of platform.
I'm using the latest version from NPM: `"pouchdb-find": "^7.0.0"`

I have some records like this:

    {
      "_id": "Movement_1_000000180828112434013019",
      "data": {
        "type": "movement",
        "inOut": "out",
        "widgets": [
          100043
        ],
        "customer": 13019,
        "inventory": 12345
      }
    }

This query succeeds with selector:

    this.$pouch.find({
      selector: { _id: 'Movement_1_000000180828110654013019' },
    }).then((res) => {
      console.log('== Result ==');
      console.log(res);
    });

I get ZERO results with a selector like this ...

    selector: { 'data.type': 'movement' },

... and, with a selector ...

    selector: { data: { type: 'movement' } },

... I get the error `Uncaught (in promise) Error: unknown operator "type" - should be one of $eq, $lte, $lt, $gt, $gte, $exists, $ne, $in, $nin, $size, $mod, $regex, $elemMatch, $type, $allMatch or $all`

Is matching nested attributes not possible?

Following [the note at the end of the ReadMe section, Examples](https://www.npmjs.com/package/pouchdb-find#examples), I tried to "refer to [Cloudant's _find documentation](https://docs.cloudant.com/api.html#finding-documents-using-an-index)", but **that link is dead.** [This page](https://console.bluemix.net/docs/services/Cloudant/api/cloudant_query.html#subfields) seems to be equivalent, and implies that both the methods I tried ought to work.

daleharvey:
It looks like this is something broken and was being worked on in https://github.com/pouchdb/pouchdb/pull/7384

martinhbramwell:
I am a bit mystified.  The code changes @ultravail submitted, unsuccessfully, are there in the master branch verbatim. So it appears someone else submitted a PR that includes his fix.

However, if it's part of the master branch, my problem must have a different cause, since I am using 7.0.0. and still cannot use subfield selectors. Or am I missing something here?

martinhbramwell:
@daleharvey -- I asked a further question in the PR 7384, but no one answers.  I'm beginning to worry that  PouchDB is slowly being abandoned.  Is that what's happening?

Issue #6879, opened 2017-11-27
$or in $and behaves unexpected


konradjurk:
Example code to be executed on your website http://nolanlawson.github.io/pouchdb-find/

```
db.find({
  selector: {
    $and: [
      {
        $or: [
          {
            name: 'Donkey Kong'
          },
          {
            name: 'Captain Falcon'
          }
          ]
      },
      {
        $or: [
          {
            debut: 1981
          },
          {
            debut: 1982
          }
          ]
      }
      ]
  },
  sort: ['_id']
});
```

This equals to the logic expression `((name == 'Donkey Kong' || name == 'Captain Falcon') && (deput == 1981 || deput == 1982))` which would resolve as `false` if `name` is Mario, hower the result of PouchDB as you can see on the website includes a record named Mario.

If this is a bug, how can we fix this? If not, how do I achieve something equal to the logic expression?

fenglianxu:
I have done many tests in pouchDB with a combination of $and and $or, as well with nested of both operators.  I have found a common pattern that the explicit $and does not work in PouchDB.
When you change the **$and** to be an implicit format such as  {$and[ {x > y}, {u <z}] }==>
{ x > y ,
  u < z
}
This implicit format works, however this needs each CouchDB client to change their code and it's very time consuming to make effort to make the more complex of mixture of the $and and $or operators.
It would be nice for PouchDB to support this explicit $and in the query so it behaves same as CouchDB.

tomjmul:
This does look like a bug. 

```
{
 "selector": {
   "$and": [
   {
     "$or": [
       {"a": 1},
       {"a": 2}
     ]
   },
   {
     "$or": [
       {"b": 1},
       {"b": 2}
     ]
   }]
 }
}
```

The above query gets rewritten to an implicit $and, and in so doing, the first $or key gets overwritten by the second:

{
 "selector": {
   "$or": [
       {"b": 1},
       {"b": 2}
     ]
 }
}

konradjurk:
Programmatically adding a "wontfix" label to a bug in one of the most basic query functionalities a database can have (and/or) is a big joke from my point of view. Thats probably why there are 3k closed issues...

konradjurk:
Oh it got removed. Intelligent bot

Issue #6664, opened 2017-08-02
Support nesting $or in selector properties for db.find


mattjennings:
### Issue
Using an $or selector nested in a property produces the following error (and same for $and)

> Error: unknown operator "$or" - should be one of $eq, $lte, $lt, $gt, $gte, $exists, $ne, $in, $nin, $size, $mod, $regex, $elemMatch, $type, $allMatch or $all

This kind of query works with Mango in CouchDB 2.0 (at least, in Project Fauxton if I test it in their "Run a Mango Query" section it works)

### Info
- Server: CouchDB 2.0, pouchdb 6.3.4

### Reproduce
Go to http://nolanlawson.github.io/pouchdb-find/ and put in:

```javascript
db.find({
  selector: {
      _id: {
        $or: [
          {$eq: "dk"},
          {$eq: "link"}
        ]
      }
  },
  sort: ['_id']
});
```

rasdaniil:
Same story with $and :(

mattjennings:
The reproduction method returns this error now:

`match@https://unpkg.com/pouchdb@6.2.0/dist/pouchdb.find.js:3065:20
every@[native code]
every@[native code]
filter@[native code]
filterInMemoryFields@https://unpkg.com/pouchdb@6.2.0/dist/pouchdb.find.js:2988:21
https://unpkg.com/pouchdb@6.2.0/dist/pouchdb.find.js:5507:40
promiseReactionJob@[native code]`

mjafarpour:
**Issue**
Is any way to use nested $and, $or operator?

> app.js:654 Error: unknown operator "0" - should be one of $eq, $lte, $lt, $gt, $gte, $exists, $ne, $in, $nin, $size, $mod, $regex, $elemMatch, $type, $allMatch or $all

**Info**

- server: Couchdb 2.2, pouchdb 7.1.1

**Reproduce**
`{ "$or": [
               {"$and": [
																	{"data.tags": { "$in": ["word1"] } }, 
																	{"data.tags": { "$in": ["word2"] } } 
														]},
														{"$and": [
																	{"data.tags": { "$in": ["word3"] } }, 
																	{"data.tags": { "$in": ["word4"] } } 
														]}														
													 ] 
											} `

PR #8237, opened 2020-11-26
Conform pouchdb find operators to couchdb's 

AlansCodeLog:
This pull request:

- Fixes pouchdb to conform to [couchdb's operator limitations](https://docs.couchdb.org/en/stable/api/database/find.html#condition-operators) for how it matches against a value, specifically `$size`, `$regex`, and `$mod`. I looked over the other operators and they seem to be fine. 
- Adds a function to validate the selector and check all operator value types (e.g. `$exists` can only be queried by a boolean, `$and` can only be an array, etc.) *before* running / sending the query. 

Notes:

I added a new file `validateSelector` to `pouchdb-find` since there did not seem to be a better place to put it and moved `$mod`'s checks there but did not move the check to see if a "matcher" exists (i.e. the operator is supported by pouch). I left that where it was in `in-memory-filter` because it should be possible to query couchdb with operators that pouchdb does not yet support.

`$regex` is of course also allowed to be an instance of `RegExp` ~~so the validation function won't save users from querying couchdb with one~~ (I fixed it), but I have an idea for how to allow cross-compatible queries. Will file a separate issue.

Other:

The .gitignore is not ignoring all temp db folders generated if they happen to still be around because I stopped a test, etc. Really annoying. I changed mine locally to `tmp*` instead of `tmp`. I can add the changes to this PR or would you prefer a separate one? Or was there some reason for it being just `tmp`?

Also being a double quotes person I tried as much as possible to use single quotes like most of the rest of the source code (though tests has quite the mix), but eslint is not set up with any quote style so some might have slipped through.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-tmp
+tmp*
 .DS_Store
 /node_modules
 docs/_site
--- a/packages/node_modules/pouchdb-find/src/adapters/http/index.js
+++ b/packages/node_modules/pouchdb-find/src/adapters/http/index.js
@@ -1,6 +1,7 @@
 import { generateErrorFromResponse } from 'pouchdb-errors';
 import { Headers } from 'pouchdb-fetch';
 import massageCreateIndexRequest from '../../massageCreateIndexRequest';
+import validateSelector from '../../validateSelector';
 
 function dbFetch(db, path, opts, callback) {
   var status, ok;
@@ -29,6 +30,7 @@ function createIndex(db, requestDef, callback) {
 }
 
 function find(db, requestDef, callback) {
+  validateSelector(requestDef.selector, true);
   dbFetch(db, '_find', {
     method: 'POST',
     body: JSON.stringify(requestDef)
--- a/packages/node_modules/pouchdb-find/src/adapters/local/find/index.js
+++ b/packages/node_modules/pouchdb-find/src/adapters/local/find/index.js
@@ -17,6 +17,7 @@ import {
   massageUseIndex
 } from '../utils';
 import {pick} from '../../../utils';
+import validateSelector from '../../../validateSelector';
 
 function indexToSignature(index) {
   // remove '_design/'
@@ -73,6 +74,8 @@ function doAllDocs(db, originalOpts) {
 
 function find(db, requestDef, explain) {
   if (requestDef.selector) {
+    // must be validated before massaging
+    validateSelector(requestDef.selector, false);
     requestDef.selector = massageSelector(requestDef.selector);
   }
 
@@ -91,16 +94,16 @@ function find(db, requestDef, explain) {
     db.constructor.emit('debug', ['find', 'planning query', requestDef]);
     var queryPlan = planQuery(requestDef, getIndexesRes.indexes);
     db.constructor.emit('debug', ['find', 'query plan', queryPlan]);
-    
+
     var indexToUse = queryPlan.index;
-    
+
     validateSort(requestDef, indexToUse);
 
     var opts = Object.assign({
       include_docs: true,
       reduce: false,
       // Add amount of index for doAllDocs to use (related to issue #7810)
-      indexes_count: getIndexesRes.total_rows, 
+      indexes_count: getIndexesRes.total_rows,
     }, queryPlan.queryOpts);
 
     if ('startkey' in opts && 'endkey' in opts &&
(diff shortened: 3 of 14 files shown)


daleharvey:
Just some quick comments, not much on the code itself, will look at that shortly

    The .gitignore is not ignoring all temp db folders generated if they happen to still be around because I stopped a test, etc. Really annoying. I changed mine locally to tmp* instead of tmp.

Yeh has annoyed me, no reason, feel free to include in this PR

    Also being a double quotes person I tried as much as possible to use single quotes like most of the rest of the source code

Single Quotes was a mistake, feel free to use double and can move towards eslint enforcing that at some point

AlansCodeLog:
Cool, I'll add the .gitignore commit then.  

One other thing, I just realized, the validation can be made to error if the user tries to send a regex over http. I can just add a parameter to tell it whether it's in "http" or "local" mode. Will edit the commit.

Regarding the single quotes I'll leave them as is for now if you don't mind (it's quite time consuming changing them manually while not touching other code). Also, can more modern stuff, like template strings, be used in the code? Eslint is setup for it, but erred on the side of not using them because I found none, but it would be nice to know if ES6 is/isn't supported in the code. If it is, a few eslint rules that have auto fix would go a long way towards modernizing everything a bit and keeping the style consisten. I can open a separate issue to discuss.

AlansCodeLog:
I added an extra check and tests for it so that the validation doesn't recurse into validated operators that are arrays. But in doing so I found other operators that have a problem. Any operator that is using any `arrayContains*` functions are working incorrectly because they are checking values like this ` docFieldValue.indexOf(val) > -1`, when if an object is passed it should be deep checked. 

http tests should pass in the build but local tests will error.

I think the values can be checked with `collate` instead? I'll take a look at this as soon as I can.

AlansCodeLog:
Okay I fixed the `arrayContains` issue. Also stumbled into two other issues I fixed as well.

The first is that implicit $eq against objects was not working (#7446, possibly others about deep nested field access, it's hard to search with everything closed). The way couchdb handles this is that it walks through the entire object *until* it meets an operator, whereas pouch was expecting the first key of an object to be an operator. I added test for both implicit and explicit $eq (which was already working) to make it clear they should behave differently with regards to operators nested inside.

The second is #6664 / #8011 / #6879 / possibly others. I fixed it to correctly massage `$or` instead of treating it just like any other `$and`. This allows pouchdb to correctly handle nested `$or`. The resulting selector looks quite different than couchdb when using `explain`, but I'm assuming the $and / $or selectors are massaged like this for a reason since they don't appear in the in memory filters.

Lastly there was a skipped test that I happened to notice didn't make much sense (probably from copying the `$lte` test above it). I fixed it.

daleharvey:
So this was very large and intimidating, apologies for taking so long to look at it, I would really like to get this merged, looks like there was only one failing test and have done a PR @ https://github.com/pouchdb/pouchdb/pull/8260 just to see whats up with that against master, it may just be a phantom failure

AlansCodeLog:
😅 No problem. It kind of got out of my hands, one thing led to another. My only real worry is the OR selector massaging. With normal sized queries should not be a problem, but am not sure that is the best way to handle it if it wasn't done like that for a reason (would like to know if it was), but it does fix the issue according to the original code's intentions and I would not know enough to fix it some other way (like sticking just that selector to be in memory).

Was there a failing test, like a specific test I mean? I only saw that one failed because it didn't meet the 100% coverage set in the params/command of the test. I assumed that was done on purpose given it was setup that way.

daleharvey:
! So I managed to get a little time to test the coverage error was from https://github.com/pouchdb/pouchdb/commit/fe7ffa9cf7306baf74856b74cf7734ebafac25d0#diff-79342eb93496efe970ba307fd40631f3c6c6c8aaa4db5abdd8593ff8ef1a3e16R135 and given how long this has been waiting I figured I would merge first and figure out whats up second since I didnt want this bitrotten, long overdue but thanks so much :)

AlansCodeLog:
Forgot about this tbh 😅 . I've actually been working on a boolean expression query lang / parser that can just be run against any object, and thinking back to this code, I would re-iterate that the way it's "massaging" the selectors is dangerous in a worse case scenario (I don't think developers will run into it if they're in control of the queries, but they might if they aren't).

I glanced at the code real quick again now, I'm still not sure why it's done this way. The queries end up mostly being done in-memory anyways. So unless massaging has some purpose I didn't catch, I don't think the small optimizations it does have much of an effect on performance, in fact, past a certain point it might be the opposite. Also although I've not gone to read CouchDB's code, I'm sure it doesn't do this because it doesn't change the query or choke on extremely complex ones.

When my library is done and I have some time, I'll do some perf tests.

Issue #8242, opened 2020-11-30
Possibly unhandled rejection upserting docs


garethbowen:
### Issue
On first load with an empty database we're getting this error in the browser console:

```
Error "Possibly unhandled rejection: {"status":404,"name":"not_found","message":"missing","error":true}"
```

It was very difficult to pin this down to exactly what's causing it but I believe I've traced the source to the PouchDB upsert function. It doesn't seem to cause any issues - I think there are retries built in that allow PouchDB to recover.

### Info
- Environment: Browser
- Platform: Chrome
- Adapter: IndexedDB
- Server: CouchDB

### Extra
It looks like the unit tests are silently failing on this too, for example [this build](https://travis-ci.com/github/pouchdb/pouchdb/jobs/447983992) outputs this to console...

```
[0m  test.mapreduce.js-upsert[0m
  [32m  âœ“[0m[90m should throw an error with no doc id[0m
(node:5729) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'get' of undefined
    at /home/travis/build/pouchdb/pouchdb/packages/node_modules/pouchdb-for-coverage/lib/index.js:872:8
    at new Promise (<anonymous>)
    at upsert (/home/travis/build/pouchdb/pouchdb/packages/node_modules/pouchdb-for-coverage/lib/index.js:871:10)
    at Context.<anonymous> (/home/travis/build/pouchdb/pouchdb/tests/unit/test.mapreduce.js:11:12)
    at callFn (/home/travis/build/pouchdb/pouchdb/node_modules/mocha/lib/runnable.js:348:21)
    at Test.Runnable.run (/home/travis/build/pouchdb/pouchdb/node_modules/mocha/lib/runnable.js:340:7)
    at Runner.runTest (/home/travis/build/pouchdb/pouchdb/node_modules/mocha/lib/runner.js:443:10)
    at /home/travis/build/pouchdb/pouchdb/node_modules/mocha/lib/runner.js:549:12
    at next (/home/travis/build/pouchdb/pouchdb/node_modules/mocha/lib/runner.js:361:14)
    at /home/travis/build/pouchdb/pouchdb/node_modules/mocha/lib/runner.js:371:7
    at next (/home/travis/build/pouchdb/pouchdb/node_modules/mocha/lib/runner.js:295:14)
    at Immediate.<anonymous> (/home/travis/build/pouchdb/pouchdb/node_modules/mocha/lib/runner.js:339:5)
[...]

PR #8243, opened 2020-11-30
(#8242) - unhandled promise rejection in upsert

garethbowen:


garethbowen:
@daleharvey Can I get a review for this? I _think_ it's safe enough...

daleharvey (approved):
Good catch, this looks good to me

Issue #8225, opened 2020-11-16
Remove setting blob type in http adapter


mduriancik:
### Issue

When synchronizing documents with attachements (replicateFrom) in our Electron application we get a TypeError on a Blob.
The code is running in the RenderProcess with NodeIntegration enabled (webPreferences: {nodeIntegration: true, enableRemoteModule: true, webSecurity: false}).

### Info

Windows 10
Angular 9.1.12 / Ionic 5.4.1
Electron 10.1.5 
Adapter: IndexedDB
CouchDB 2.3

### Details

The error occurs in [/pouchdb-adapter-http/src/index.js:632](https://github.com/pouchdb/pouchdb/blob/1ba879709fdf6306dc3f8e9d2d67f4060ade6e50/packages/node_modules/pouchdb-adapter-http/src/index.js#L632)

```javascript
   // TODO: also remove
   if (typeof process !== 'undefined' && !process.browser) {
     blob.type = contentType;
   }
```

In fact with Electron/RenderProcess blob is a standard Blob type, which is [read only](https://developer.mozilla.org/en-US/docs/Web/API/Blob).
(The condition `typeof process !== 'undefined' && !process.browser` in our Electron with NodeIntegration evaluates to true)

Possible solutions:

1. the code should be removed (I think this is what the comment meant)

2. if 1. is not possible please change the condition to something like:

`typeof process !== 'undefined' && window === undifined` or  `typeof process !== 'undefined' && process.versions.electron`

or make sure that "blob" type is not Blob (to differentiate it from Buffer, I suppose it is when using node-fetch) 

`! (blob instanceof Blob)`

### Info
- Environment: (Node.js/browser/hybrid app/etc.)
- Platform: (Chrome/FF/Safari/Edge/iOS/Android/etc.)
- Adapter: (IndexedDB/WebSQL/LevelDB/etc.)
- Server: (CouchDB/Cloudant/PouchDB Server/etc.)

Milan

pete-mcwilliams:
Have you seen these two issues?
https://github.com/pouchdb/pouchdb/issues/8154
https://github.com/pouchdb/pouchdb/issues/7836

daleharvey:
Happy to take a PR to remove that code, if I remember we were trying to have code that used blobs and buffers interchangably that we ended up removing anyway. 

I suspect there is a test that will check for this property, but if you make a PR with it removed we can see what the tests say

saschwarz:
[...]

PR #8255, opened 2020-12-30
8225 Remove assigning content type to blob

saschwarz:
Attempt at resolving issue in #8225

### DIFF
--- a/packages/node_modules/pouchdb-adapter-http/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-http/src/index.js
@@ -613,9 +613,7 @@ function HttpPouch(opts, callback) {
     var params = opts.rev ? ('?rev=' + opts.rev) : '';
     var url = genDBUrl(host, encodeDocId(docId)) + '/' +
         encodeAttachmentId(attachmentId) + params;
-    var contentType;
     ourFetch(url, {method: 'GET'}).then(function (response) {
-      contentType = response.headers.get('content-type');
       if (!response.ok) {
         throw response;
       } else {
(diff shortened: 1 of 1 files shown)


daleharvey (changes requested):
Hi @saschwarz 

Cheers for the patch, I am not sure why there are so many changes to the imports? In this bug we only need the last 3 lines of that diff that assign the `blob.type`, its very likely there will be test failures that come along with that but if you just push that change you can see where CI fails, right now the import changes are causing undefined errors

saschwarz:
Sorry, I'm using prettier in my editor and it reorders imports - I didn't notice because I'm not seeing any import errors when I run my app.

daleharvey (changes requested):
This is failing with @ https://travis-ci.com/github/pouchdb/pouchdb/jobs/470263668 with:

```
/home/travis/build/pouchdb/pouchdb/packages/node_modules/pouchdb-adapter-http/src/index.js

  616:9  error  'contentType' is assigned a value but never used  no-unused-vars
```

So that will need fixed

daleharvey (changes requested):
Thanks for the prompt update, however this is still failing tests, likely there are tests that check that the content-type was assigned. Please run the tests and ensure they are passing before requesting review. Feel free to ping me if you have questions though. Thanks

saschwarz:
@daleharvey I looked at the first few of the failing tests and don't understand why they are failing or how to fix them.

daleharvey:
ok will close out then

saschwarz:
@daleharvey I hoped you might be able to assist me with how to fix the failing tests. If you could give me some guidance I would try to get them to pass. I'm no longer using pouchdb but would like to help the project.

daleharvey:
@saschwarz If you have specific questions, "I cant get this to work, I tried doing X and Y", or "I have trouble figuring out what is supposed to happen here" then I can try to help out best I can

PR #8256, opened 2021-01-03
Fix #7331 - close leveldb file handlers

### LINKED ISSUE #7331: close does not close dependent dbs fully
### Issue
Closing a database does not seem to close the dependent dbs / views / indices.
So one can not repair / rebuild the indices without a restart

### Info
- Environment: node
- Platform: macos / Win
- Adapter: leveldb
- Server: Pouch db

### Reproduce
Flow:
- Create a db
- Create an index / view
- Close the database
- delete the view directories
- reopen the database
- try to use the index

```
const PouchDB = require('pouchdb')
PouchDB.plugin(require('pouchdb-find'))
const fs = require('fs')
const expect = require('chai').expect
const path = require('path')
const util = require('util')
const rimraf = require('rimraf')
const chai = require('chai')


describe('index directory creation issue', () => {
	const databaseDir = path.join(__dirname, 'database-view-recreate') + path.sep
	let database 
	createDatabase = async () => {
		let db = new PouchDB(databaseDir)
		await db.info()	
		return db
	}
	before(async () => {
		database = await createDatabase()
	})

	after(() => {
		rimraf.sync(databaseDir)
	})

	context('create index', () => {
		before(async () => {
			await database.createIndex({
  				index: {
					fields: ['name'],
    				name: 'myindex',
    				ddoc: 'myindex'
  				}
			})
			await database.put({
				 _id: '_design/index',
  				views: {
  					anotherIndex: {
      					map: function (doc) { emit(doc.name) }.toString()
    				}
  				}
			})
		})

		it('creates a mr-view directory', () => {
			let directories = fs.readdirSync(databaseDir)
			let hasItem = directories.some((item) => {
				return item.match(/-mrview-/)
			})
			expect(hasItem).to.be.true
		})

		it('is able to search by query', async () => {
			
			await database.query('index/anotherIndex')
		})

		context('delete and recreate index', () => {
			before( async () => {
				await database.close()
				rimraf.sync(path.join(databaseDir, '*-mrview-*'))
			})

			it('deletes the view directory', () => {
				let directories = fs.readdirSync(databaseDir)
				let hasItem = directories.some((item) => {
					return item.match(/-mrview-/)
				})
				expect(hasItem).to.be.false
			})

			context('database reopen', () => {
				before(async () => {
					database = await createDatabase()

				})

				it('creates the view directories', () => {
					let directories = fs.readdirSync(databaseDir)
					let hasItem = directories.some((item) => {
						return item.match(/-mrview-/)
					})
					expect(hasItem).to.be.true
				})
			})
		})
	})

})
```

Package.json
```
{
  "dependencies": {
    "chai": "^4.1.2",
    "mocha": "^5.2.0",
    "pouchdb": "6.4.3",
    "pouchdb-find": "6.4.3",
    "rimraf": "^2.6.2"
  }
}
```

rasgo-cc:
This fixes #7331 so "mrview" folders can be deleted/repaired after closing the db and without having to restart the app.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,5 @@ lerna-debug.log
 tests/integration/utils-bundle.js
 *.heapsnapshot
 /pouchdb-server-install
+package-lock.json
+yarn.lock
--- a/packages/node_modules/pouchdb-adapter-leveldb-core/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-leveldb-core/src/index.js
@@ -1182,7 +1182,7 @@ function LevelPouch(opts, callback) {
           eventEmitter.removeAllListeners();
           eventEmitter.close();
           adapterStore.delete(key);
-        })
+        });
         
         callback();
       }
--- a/packages/node_modules/pouchdb-adapter-leveldb-core/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-leveldb-core/src/index.js
@@ -134,7 +134,7 @@ function fetchAttachments(results, stores, opts) {
       return;
     }
     var attNames = Object.keys(row.doc._attachments);
-    attNames.forEach(function (attName) {
+    attNames.forEach(function pa(attName) {
       var att = row.doc._attachments[attName];
       if (!('data' in att)) {
         atts.push(att);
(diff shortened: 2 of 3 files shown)


daleharvey (approved):
Very nice, I think ideally we have a test which shows this breakage, but considering its a tricky test to write I dont think we need to block fixing.

PR #8274, opened 2021-02-24
rev() should clone only when needed

pubkey:
Doing a deep clone over an object is very expensive. We do this very often when creating new revision keys.
I changed the code to:
- Only clone when needed for deterministic revision
- Do a flat clone via Object.assign because we anyway only need the cloned object for the md5-hash without `_rev_tree`

Question: Shouldnt't we also strip `_rev` from the document data before hashing it?


daleharvey (approved):
Good catch

As for whether to include the rev, I think thats on purpose as if we didnt include it a document updating itself without changing anything would get the same revision back, which seems like it could have advantages or disadvantages. 

But worth checking on couch behaviour here

PR #8278, opened 2021-03-06
Removed #8225

ucusita:
That´s it?


daleharvey (changes requested):
Thats all the code needed, however the tests will fail, for assertions that are testing the contents of `.type` they will just need deleted, there may be other tests that need ammended.

daleharvey:
As mentioned, needs the tests fixed, will close until theres a PR with those addressed

Issue #8307, opened 2021-06-05
Adding TypeScript and Storybooks to PouchDB core development


mqtik:
PouchDB needs types and storybooks, for easier development, better documentation for folks that want to hop in in the development cycle, and even improve starts for new beginners.

I'm happy to start working on it, gradually, with small things first, if you're all okay with it.

dheimoz:
hey @mqtik thanks for your suggestion. I believe we can start with types.
Any particular idea you have for documentation?

mqtik:
Absolutely.

Storybooks are a great tool, not only for development, but also for documenting things afterwards, in an automatical manner.
The good thing, is that we don't need to actually explicitly document, since it all comes from types.

So, we define the types, and the documentation will generate on its own.

I'm not sure if taking pouchdb-core just yet, I'd start with a smaller repo as a proof of concept.

https://storybook.js.org/

It's focused on the UI, however, we don't need to actually write UI components, but rather write use-cases there, so everyone can hop in on the development and easily help out.

dheimoz:
All right, let's see the proof of concept please.

mqtik:
@dheimoz @daleharvey see https://github.com/pouchdb/pouchdb/pull/8310

dheimoz:
Hey @mqtik 

Thank you very much for your efforts. I am a little bit overwhelm with a project. Do you mind if I take a look this weekend?

Once again. Thanks. This kind of help is what keeps Pouchdb rolling.

mqtik:
Of course! I completely understand, we all have jobs (:

No rush, review it when you have time.

Thanks!

sureshvv:
Correct me if I am wrong but Storybooks (just googled it) seems to be for building/prototyping UIs quickly. How would it help pouchdb development?

mqtik:
If you use Storybooks for only building UI, then it will be just that.

PouchDB doesn't need UI for the end user, true.
The UI will be just for us, the developers, to build different scenarios, share test/cases easily, try the main component in a lot of different ways and test that.

If I found an error in X PouchDB plugin, for example, I can just fork it, run it, and replicate the problem with a Storybook, work on a fix easily; or share it to you so you can run it and help me out.
Then we have those stories to run tests, to assure that it doesn't get broken again.

UI it's really helpful, not only for us, but for new beginners too.

If I was a beginner in PouchDB, running Pouch with a lot of written scenarios, it would be a BIG plus.

[...]

PR #8310, opened 2021-06-13
Adding TypeScript and Storybooks to PouchDB

mqtik:
This is a proof of concept (see #8307) for:

- Adding TypeScript to PouchDB
- Add Storybooks
- Modularize and refactor
- Types
- Test per component/package / Jest
- Refactored PouchDB-Errors to demostrate the change.

As I couldn't properly run `pouchdb-errors`, I tried to replicate as much as I could, with a few things in mind.

- Ease of development of each component
- Add Storybooks to: 
   - Allow to test components in a isolated environment. Storybooks are not included in the final bundle, and are only for development porpuses.
   - Allow new developers to make changes and contributions easily, with components to actually see, and hotload environment for fast development and more.
   - Automatically generate Documentation based on the typings (TS) for each component.
   - Use Stories to replicate new scenarios, for the component to be well built and tested.
   - Use tests to run those stories to see which conflicts.
- Better documentation for running and debugging the project.

- Rollup and so on for building and publishing porpuses are already added and tested.
- Separate ErrorStatus into a new isolated constant, with its actual types.
- Added Prettier for the code to be beautifully formatted, and ESLint

The structure for each component, is:
-/ **src**
--**/modules**
--**/shared**
--**/components**
---**/defaultComponent** _(or any other name, like Error, Find, Utils, ...)_
----**/Default.tsx** _(where the component actually is, and bundled into production)_
----**/Default.types.tsx** _(where types for Default.tsx are)_
----**/Default.stories.tsx** _(where we create new stories, using Default.tsx component to build new scenarios)_
----**/Default.tests.tsx** _(where we write tests using those stories/scenarios)_
----**/index.tsx** _(where the component gets exported)_

Beware that React it's added, but only for development porpuses, for Storybooks. It will, of course, not be loaded into the final bundle, since this must work everywhere.

In order to complete this PR, we should make a call; since a lot of the code from PouchDB-Errors it's not documented, and it's quite confusing to run.

If you all want, we can start to gradually implement this skeleton for each component (pouchdb-find, pouchdb-utils and so on), and ultimately, PouchDB itself.

I left instructions on how to run it.
I'm open to any questions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "test-webpack": "bash bin/test-webpack.sh"
   },
   "dependencies": {
+    "abort-controller": "3.0.0",
     "argsarray": "0.0.1",
     "buffer-from": "1.1.1",
     "clone-buffer": "1.0.0",
@@ -52,7 +53,6 @@
     "ltgt": "2.2.1",
     "memdown": "1.4.1",
     "node-fetch": "2.6.1",
-    "abort-controller": "3.0.0",
     "promise-polyfill": "8.1.3",
     "readable-stream": "1.1.14",
     "request": "2.88.2",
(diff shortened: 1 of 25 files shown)


daleharvey:
Closed as discussed in the bug

Issue #8309, opened 2021-06-09
Errors when aborting replication


hermitdemschoenenleben:
### Issue
I'm using pouchdb in a mobile application. When the app is in background mode for some time, replication doesn't work anymore (which is to be expected, I think). Therefore, after app resume I check whether replication still works, and if it doesn't work I cancel the old replication and start a new one.

This solution seems to work fine. However I sometimes see

* `Cannot read property 'filter' of undefined` thrown in  https://github.com/pouchdb/pouchdb/blob/master/packages/node_modules/pouchdb-core/src/adapter.js#L887
* or `undefined is not an object (evaluating 'r.length')` in https://github.com/pouchdb/pouchdb/blob/master/packages/node_modules/pouchdb-core/src/adapter.js#L893

when cancelling the replication.

In both cases, the callback is called from here: https://github.com/pouchdb/pouchdb/blob/master/packages/node_modules/pouchdb-adapter-indexeddb/src/bulkDocs.js#L375

Apparently, `error` in `onabort` is undefined when cancelling replication manually, which then causes the error in the callback function.

### Possible fix

I think the issue should be quite simple to fix:
 ```
txn.onabort = function () {
  callback(error || someDefaultError);
}
```

Problem is: I don't know enough of pouchdb's internals to know what `someDefaultError` should be. Any suggestions?

sureshvv:
> Therefore, after app resume I check whether replication still works, and if it doesn't work I cancel the old replication and start a new one.
> 

How do you check if the replication "still works"?

hermitdemschoenenleben:
> > Therefore, after app resume I check whether replication still works, and if it doesn't work I cancel the old replication and start a new one.
> 
> How do you check if the replication "still works"?

I'm  using [https://github.com/hermitdemschoenenleben/pouchdb-csg/](pouchdb-csg) in order to retrieve the changes feed using websockets. When opening the application I send a "ping" doc to the server and expect a websocket message within the next few seconds. If I don't receive a message I assume that replication doesn't work anymore, cancel the old one and start a new one.

As the problem is quite hard to debug (happens only when the app has been in background for a long time) I did not investigate in detail what exactly happens in the case of an unresponsive replication; as android / ios put the application in a network-isolated sandbox when in background, I suppose that either a HTTP request or the changes feed can't recover and just hangs without an error mesage when re-opening the app.

[...]

PR #8316, opened 2021-06-28
properly handle transaction abort

hermitdemschoenenleben:
This PR fixes the errors of #8309: if the transaction is aborted for some reason, it creates an error using `createError` which fixes the problems in the callbacks.


daleharvey:
Hey sorry for the delay

daleharvey:
We had some issues with travis, the test should run now, I think this PR looks good, as a tiny nit we usually use " instead of ' for strings, will see what the tests say but if you wanted to update that would be ideal

daleharvey:
If we dont have eslint not block on quote types then happy to merge (although if someone wanted to fix the eslint confict, I think standardising on " is just a bit nicer)

daleharvey (approved):

dheimoz:
@daleharvey thanks man.

PR #8319, opened 2021-07-06
Fix node build ordering

jcoglan:
We have been observing that when running `npm test`, the tests don't
seem to run the current version of the code. It's possible to make edits
to the code, and the first time you run the tests, your edits won't have
any effect. They only apply the second time you run the tests.

This seems to be because of the order things are done in
`bin/run-test.sh`. Depending on the value of `SERVER`, it might start a
server using this command:

    node ./tests/misc/pouchdb-express-router.js

This process will import `packages/node_modules/pouchdb`, which in turn
imports other packages, whose `main` file is usually `lib/index.js`.
This file is not source code but is generated by the build process.

After it's started the server, `bin/run-test.sh` then runs one of the
test suites, with `test-node` being the default because `CLIENT`
defaults to `node`. The npm `test-node` command runs the `build-node`
command, which is what triggers the build of all packages and updates
their `lib` directories.

Because this happens after the server is started, the server is then
running out-of-date code. We fix this by putting code in
`bin/run-test.sh` to run `npm run build-node` if it's not already been
run, before any command that needs the node.js packages up-to-date, and
removing the call to `build-node` from the `test-node` command. This
should be safe because the only place the `test-node` task is invoked is
in `bin/run-test.sh`.

### DIFF
--- a/bin/run-test.sh
+++ b/bin/run-test.sh
@@ -52,6 +52,13 @@ pouchdb-link-server-modules() {
   cd ..
 }
 
+pouchdb-build-node() {
+  if [[ $BUILD_NODE_DONE -ne 1 ]]; then
+    npm run build-node
+    BUILD_NODE_DONE=1
+  fi
+}
+
 if [[ ! -z $SERVER ]]; then
   if [ "$SERVER" == "pouchdb-server" ]; then
     export COUCH_HOST='http://127.0.0.1:6984'
@@ -69,10 +76,12 @@ if [[ ! -z $SERVER ]]; then
       export COUCH_HOST="http://127.0.0.1:5984"
     fi
   elif [ "$SERVER" == "pouchdb-express-router" ]; then
+    pouchdb-build-node
     node ./tests/misc/pouchdb-express-router.js &
     export SERVER_PID=$!
     export COUCH_HOST='http://127.0.0.1:3000'
   elif [ "$SERVER" == "express-pouchdb-minimum" ]; then
+    pouchdb-build-node
     node ./tests/misc/express-pouchdb-minimum-for-pouchdb.js &
     export SERVER_PID=$!
     export COUCH_HOST='http://127.0.0.1:3000'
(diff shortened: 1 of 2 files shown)


dheimoz (commented):
hey @daleharvey , what do you think about this? I also having problems running tests.

dheimoz:
Hey @jcoglan thanks for the PR. Definitely will take a look, because as you stated, there are issues with the test.

jcoglan:
This was the smallest functional change I could see to make this work but reading the code I can see several distinct things happening that could potentially be better separated, if we wanted to refactor further:

- Building the node packages
- Starting a server, if necessary
- Waiting for the server to come up
- Running the tests
- Stopping the server

Most of what this script does is starting and stopping the server, and there's a little bit in the middle to run the build and tests. It's hard to extract the server functionality into a distinct build task because we need to remember `SERVER_PID` in order to kill it once the tests are done.

The complexity here comes from the fact that starting the server and running the tests _might_ require a build step, but you don't want to run the build multiple times. It's possible that using a task runner might deal with this nicely but I didn't want to make bigger changes to the build system without discussing them first.

daleharvey:
Awesome thanks for the PR, so the first thing we need to do is fix the travisCI integration which seems to have stopped, around a month ago looks like the last run which is strange https://www.travis-ci.com/github/pouchdb/pouchdb/pull_requests, anyone happy to know off the top of their head any changes?

daleharvey:
ok seems like some type of auth problem @ https://travis-ci.com/github/pouchdb/pouchdb/requests

janl:
@daleharvey maybe something has to be done to re-enable this on travis-ci.com // as I assume this was running on travis-ci.org  before, which was recently shut down IIRC. I’m super hazy on the details.

daleharvey:
Its grand we already did the switch to .com however travis did some update recently that meant some button needed pressed to pick a plan for them to run, done and passing now :)

daleharvey (approved):
ok this looks good to me now and all passing, thanks :)

dheimoz:
@daleharvey and @janl thank you both for sort this out,

PR #8320, opened 2021-07-06
Let caller set the change batch size when updating views

jcoglan:
Some users need to be able to increase or reduce the size of changes
batches used when updating view indexes, to make the best use of the
storage adapter they're running.

For example, memory constraints or processing capacity might necessitate
using smaller batch sizes.

This change lets the caller set a DB option named `index_batch_size`,
which changes the batch size we use for updating views in
`abstract-mapreduce`. It defaults to 50, the value of the existing
`CHANGES_BATCH_SIZE` constant in this module.

We'd appreciate some guidance on how this change should be tested. This
change was originally done by @chrisekelley on their fork so they may
be able to provide some more context.

### DIFF
--- a/docs/_includes/api/create_database.html
+++ b/docs/_includes/api/create_database.html
@@ -16,6 +16,7 @@ This method creates a database or opens an existing one. If you use a URL like `
 * `adapter`: One of `'indexeddb'`, `'idb'`, `'leveldb'`, or `'http'`.
 * `revs_limit`: Specify how many old revisions we keep track (not a copy) of. Specifying a low value means Pouch may not be able to figure out whether a new revision received via replication is related to any it currently has which could result in a conflict. Defaults to `1000`.
 * `deterministic_revs`: Use a md5 hash to create a deterministic revision number for documents. Setting it to false will mean that the revision number will be a random UUID. Defaults to true.
+* `view_update_changes_batch_size`: Specify how many change records will be consumed at a time when rebuilding view indexes when the `query()` method is used. Defaults to 50.
 
 **Options for remote databases:**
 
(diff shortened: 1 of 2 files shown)


janl:
Preliminary performance testing (I don’t want to call our method benchmarking), shows that increasing this value also allows users trading off memory used and indexing speed, up to a factor of 10 between the default 50 and a rather large 1000 docs per batch.

The main goal here though is to _reduce_ memory usage and _increase_ stability for the `cordova-sqlite` adapter that has been enhanced with the SQLCipher plugin.

janl:
And one last note: this is already being used to great success (stability) in production with the [Tangerine](https://github.com/Tangerine-Community/Tangerine) project.

dheimoz:
hey @daleharvey any concerns for not merging? mind if I do?

daleharvey:
So the code here looks good, while ideally everything has tests I cant think of a simple one for this level of implementation detail change and I dont think its worth mocking everything out and testing that way so I think we are fine given coverage tells us we are hitting the code paths and not breaking.

I do think however we should be documenting options that are exposed to the end user if for no other reason than to keep track of them. In the process of documenting it we might want to bikeshed the name a little, `changes_batch_size` I am not sure is specific enough `view_update_changes_batch_size` is unweidly but less likely to get confused, any other suggestions?

jcoglan:
@daleharvey I agree the name should be more specific, given this is a `PouchDB()` constructor option so there's not a strong cue from context about which operations it affects. I'll rename it.

jcoglan:
I've just rebased and amended the commit that adds the user-facing option.

dheimoz (approved):
Thank you very much for your contribution

PR #8321, opened 2021-07-07
Add an indexing event to show progress of view updates

AlbaHerrerias:
The motivation behind this PR is to show progress of view updates during the indexing. This PR is related to @chrisekelley 's [fork](https://github.com/chrisekelley/pouchdb/pull/1). Our goal is to extract and provide test for these functionalities. 

In this PR:
- Add emit event `indexing` inside the package `pouchdb-abstract-mapreduce` that sends a `progress` object. 
- Add test inside `test.design_docs.js`. But:
1. We are unsure if this is the correct suite to test it.
2. We can only test it if the adapter is `local`, as the `http` adapter does not emit the event. We are guessing that we can't listen to the event on a remote DB over HTTP, hence the test fails.
3. The `progress` object has the following properties: `view`,`last_seq`,`results_count`,`indexed_docs`. Would these follow the `PouchDB` variable name conventions?

Any help would be highly appreciated. Thank you.

### DIFF
--- a/tests/integration/test.design_docs.js
+++ b/tests/integration/test.design_docs.js
@@ -96,6 +96,39 @@ adapters.forEach(function (adapter) {
       });
     });
 
+    it('Indexing event', function (done) {
+      var docs1 = [
+        doc,
+        {_id: 'dale', score: 3},
+        {_id: 'mikeal', score: 5},
+        {_id: 'max', score: 4},
+        {_id: 'nuno', score: 3},
+      ];
+      var db = new PouchDB(dbs.name);
+      // Test invalid if adapter doesnt support mapreduce
+      if (!db.query || adapter !== 'local') {
+        return done();
+      }
+
+      let indexingEvents = [];
+
+      db.on('indexing', function (result) {
+        indexingEvents.push(result);
+      });
+
+      db.bulkDocs({ docs: docs1 }, function () {
+        db.query('foo/scores', { reduce: false }, function () {          
+          indexingEvents.length.should.equal(2);
+          indexingEvents[0]['indexed_docs'].should.equal(0);
+          indexingEvents[1]['last_seq'].should.equal(5);
+          indexingEvents[1]['results_count'].should.equal(5);
+          indexingEvents[1]['indexed_docs'].should.equal(5);
+
+          done();
+        });
+      });
+    });
+
     it('Concurrent queries', function (done) {
       var db = new PouchDB(dbs.name);
       // Test invalid if adapter doesnt support mapreduce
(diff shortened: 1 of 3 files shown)


tests/integration/test.design_docs.js
```diff
@@ -96,6 +96,39 @@ adapters.forEach(function (adapter) {
...
+        return done();
+      }
+
+      let indexingEvents = [];
+
+      db.on('indexing', function (result) {
+        indexingEvents.push(result);
+      });
+
+      db.bulkDocs({ docs: docs1 }, function () {
```
daleharvey, line 119: Its a slow process however we have started moving over to using `await` for the tests instead of using the callbacks and `done` so new tests should probably do the same

tests/integration/test.design_docs.js
```diff
@@ -96,6 +96,39 @@ adapters.forEach(function (adapter) {
...
+      ];
+      var db = new PouchDB(dbs.name);
+      // Test invalid if adapter doesnt support mapreduce
+      if (!db.query || adapter !== 'local') {
+        return done();
+      }
+
+      let indexingEvents = [];
+
+      db.on('indexing', function (result) {
```
daleharvey, line 115: This seems like a brittle assumption, are we guaranteed these events will fire by the time the bulkdocs then query is complete?

daleharvey (dismissed):
This looks good cheers, I think we need to be documenting these events however and there are some other comments, plus I think the tests maybe caught something. 

If you could rebase with master before you update that would be ideal so we dont get a full travis run, Thanks

tests/integration/test.design_docs.js
jcoglan: Hi @daleharvey, I work with @AlbaHerrerias at Neighbourhoodie. I'm still learning about how this code works, but my assumption going into this was that calling `db.query()` first causes Pouch to make sure the underlying index is up to date with the primary data store, so any `indexing` events would necessarily be emitted before `query()` returns a result.

If that's not the case, can you give an example of when this ordering would be violated and suggest a better place for us to wait until all the expected `indexing` events have been emitted? My only other idea was to put a counter in the `indexing` event listener and invoke the assertions and `done()` when it reaches `2`.

tests/integration/test.design_docs.js
AlbaHerrerias: Thank you, done! Please note that we had to upgrade eslint `ecmaVersion` in order to use arrow functions and async/await syntax

dheimoz (approved):
Async/Await addressed

dheimoz (commented):
thanks

tests/integration/test.design_docs.js
dheimoz: hey @daleharvey , thanks for you previous efforts, Travis is works and all tests are passing. Any other concern you see with the rewrite?

PR #8324, opened 2021-07-09
Add checkpoint events during replication

AlbaHerrerias:
The motivation behind this PR is to emit checkpoint events during the replication process. This PR is related to @chrisekelley 's [fork](https://github.com/chrisekelley/pouchdb/pull/1). Our goal is to extract and provide test for these functionalities. 

In this PR:
- Add emit event `checkpoint` inside the package `pouchdb-replication`. 
We emit an object, which can contain the following property names: `checkpoint`, `revs_diff`, `start_next_batch` and `pending_batch`. Would these follow `PouchDB` variable name conventions? We will be happy to refactor them if someone comes up with better naming for these.
- Add test inside `test.replication_events.js`, but we are unsure if this is the correct suite to test it.

Any help would be highly appreciated. Thank you.

### DIFF
--- a/packages/node_modules/pouchdb-replication/src/replicate.js
+++ b/packages/node_modules/pouchdb-replication/src/replicate.js
@@ -132,6 +132,7 @@ function replicate(src, target, opts, returnValue, result) {
     writingCheckpoint = true;
     return checkpointer.writeCheckpoint(currentBatch.seq,
         session).then(function () {
+      returnValue.emit('checkpoint', { 'checkpoint': currentBatch.seq });
       writingCheckpoint = false;
       /* istanbul ignore if */
       if (returnValue.cancelled) {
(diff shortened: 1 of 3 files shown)


AlbaHerrerias:
We have updated eslint `ecmaVersion` to 8 in order to use async/await in tests. This commit is from [this PR](https://github.com/pouchdb/pouchdb/pull/8321).

packages/node_modules/pouchdb-replication/src/replicate.js
```diff
@@ -149,6 +150,7 @@ function replicate(src, target, opts, returnValue, result) {
   function getDiffs() {
     var diff = {};
     currentBatch.changes.forEach(function (change) {
+      returnValue.emit('checkpoint', { 'diffing': change });
```
janl, line 153: slight preference for calling this `revs_diff` as that’s what the CouchDB API for this is called.

janl (changes requested):

janl (approved):

Issue #8325, opened 2021-07-10
Reduce travis runs


daleharvey:
The new travis setup seems like it may be a hassle for the amount of jobs we run, we can reduce to the minumum for now and look at switching to github actions or similiar

dheimoz:
hey @daleharvey I am not well versed in travis, but happy to help in any capacity.

daleharvey:
No worries all done cheers

PR #8326, opened 2021-07-10
(#8325) - Reduce the number of travis runs

daleharvey:
Just commenting them out for now so we can see what the impact for travis looks like and whether we need to cull further or can reenable some.


daleharvey:
Gonna merge without review because I would like to avoid someone doing another full travis PR

PR #8335, opened 2021-08-04
Add view adapter option

janl:
(I’m posting this for @ninetteadhikari, who is on vacation atm)

## Current setup
Currently PouchDB uses a single adapter to store all application data in one place including the core data as well as the view index data. The adapter for dependent databases is registered in `registerDependentDatabase` in the `adapter.js` file.

## Issue
The use of a single adapter per database becomes a problem when the user wants to store their core data and view index data separately in multiple databases. The current setup does not provide the option to do so. 

## Use case
One of the use cases for using separate adapters can include using the `memory` adapter for indexes to speed up queries. Separate adapters is also helpful in taking advantage of the encryption options of the different databases.

## Approach
This PR provides an option for the user to specify a separate adapter to store their view index in addition to the main adapter for their core data.
For example the user will be able to define their main adapter using `adapter` and view index adapter using `view_adapter`. Here the core data will be stored in IndexedDB and view index data in WebSQL.
```
const db = new PouchDB(DB_NAME, {adapter: 'idb', view_adapter: 'websql'})
```

* * *

We’re mainly looking for guidance from the PouchDB team on the following points:

1. Should this be added? We believe the patch is small enough for a special-interest feature that is annoying to maintain outside of PouchDB, but others might disagree :)
2. Is `view_adapter` a sensible option name. A little more precise might be `index_adapter`, happy to bikeshed this.
3. other users might want to pass additional options to the view adapter, and so far they’d have to pass them into their local `new PouchDB()` options, which is “good enough™” for us, but we could see a future PR where `view_adapter` takes an object with per-adapter infos. We wanted to keep this simple for now.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -144,6 +144,15 @@ following values:
 
 Set this to `1` to skip the migration tests.
 
+#### `VIEW_ADAPTERS` (default: `memory`)
+
+Comma-separated list of preferred view adapter backends that PouchDB will use. 
+This variable overrides the default choice and causes additional adapters to
+be loaded if they're not part of the default distribution.
+
+On Node.js the available adapters are `leveldb` and `memory`. In the
+browser they're `idb`, `indexeddb` and `memory`.
+
 
 ## Other sets of tests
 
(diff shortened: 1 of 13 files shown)


janl:
I’ll look at the CI issues later this week.

dheimoz:
hey @janl I would love to have this great contribution merged. Do you think it is possible to have the latest version with Github Actions from you and your team?
Thank you very much for the great efforts.

AlbaHerrerias (approved):

esurface:
Hi @janl, I'm with the Tangerine folks at RTI. We are looking for a way to speed up indexing. In the notes above, you mention that using 'memory' may speed up queries. Do you have any benchmarks on the different options for `view_adaptor`?

PR #8336, opened 2021-08-06
(#8225)  - Added check for blob.type setter

pturner1989:
From Issue #8225:
### Issue
When synchronizing documents with attachements (replicateFrom) in our Electron application we get a TypeError on a Blob.
The code is running in the RenderProcess with NodeIntegration enabled (webPreferences: {nodeIntegration: true, enableRemoteModule: true, webSecurity: false}).

### Cause
In pouch-adapter-http the "api.getAttachment" function tries to set blob.type even if there is no setter (i.e. in the browser).

### Fix Information
On line 517 the same operation is conducted but a check is made using Object.getOwnPropertyDescriptor() to ensure blob.type has a setter. This same check has been copied to avoid trying to set blob.type incorrectly.


dheimoz (approved):

AlbaHerrerias:
Hey @pturner1989, could you rebase from master? It will allow me to trigger Github actions. Thanks!

pturner1989:
Hey,

Will do it later today.

Cheers, Pete

AlbaHerrerias:

Issue #8291, opened 2021-04-20
Add Doc id/ current keys to the error message of a failing map reduce function


paolosanchi:
### Issue
When a map/reduce function throws an error, the function name and the doc which caused the error are not provided with the error, making very hard to spot what caused the issue.

I get this error: (please, ignore console-log-viewer.js, it's a log wrapper I use)
```
console-log-viewer.js:158 The user's map/reduce function threw an uncaught error.
You can debug this error by doing:
myDatabase.on('error', function (err) { debugger; });
Please double-check your map/reduce function.
```
and from `myDatabase.on('error', function (err) { debugger; });`
```
console-log-viewer.js:158 TypeError: Cannot read property '' of null
    at searchEmitRecursive (eval at scopeEval (index-browser.es.js:868), <anonymous>:20:52)
    at searchEmitRecursive (eval at scopeEval (index-browser.es.js:868), <anonymous>:28:41)
    at searchEmit (eval at scopeEval (index-browser.es.js:868), <anonymous>:13:33)
    at eval (eval at scopeEval (index-browser.es.js:868), <anonymous>:69:29)
    at tryMap (index-browser.es.js:7973)
    at createDocIdsToChangesAndEmits (index-browser.es.js:8427)
    at processBatch (index-browser.es.js:8410)
    at ZoneDelegate.invoke (zone-evergreen.js:364)
    at Object.onInvoke (core.js:28301)
    at ZoneDelegate.invoke (zone-evergreen.js:363)
    at Zone.run (zone-evergreen.js:123)
    at zone-evergreen.js:857
    at ZoneDelegate.invokeTask (zone-evergreen.js:399)
    at Object.onInvokeTask (core.js:28289)
    at ZoneDelegate.invokeTask (zone-evergreen.js:398)
```
When the error is rethrown there is the document, or the keys involved. It could be a good idea to attach them to the comment. 
This wouldn't affect performance, since it's a point where this seems important. The function which caused the problem could be spotted easily activating the debug mode, or with a custom log before the command, so the doc id should be enough for further  analisys.

https://github.com/pouchdb/pouchdb/blob/c3f066930c62045783e7c0f51f9cfb690104d595/packages/node_modules/pouchdb-abstract-mapreduce/src/index.js#L98
https://github.com/pouchdb/pouchdb/blob/c3f066930c62045783e7c0f51f9cfb690104d595/packages/node_modules/pouchdb-abstract-mapreduce/src/index.js#L110

### Info
[...]

PR #8343, opened 2021-08-23
(#8291) - Add contextual data to the error message of failing tryMap and tryReduce

paolosanchi:
When a map/reduce function throws an error, there is not data available in the console to spot what caused the issue.
The only way is to get it, is via debugger, but in a production environment it's hard to understand what document caused the issue and where.

I've added an object attached to the error.

https://github.com/pouchdb/pouchdb/issues/8291


dheimoz (approved):

jcoglan (approved):

AlbaHerrerias:

PR #8344, opened 2021-08-25
chore: remove unused dependency

TomerAberbach:
A search for [`double-ended-queue` in the repo only shows `package.json`](https://github.com/pouchdb/pouchdb/search?q=double-ended-queue).


janl:
looks like we need this during tests, I’ll see about cleaning this up later. Reverting here: https://github.com/pouchdb/pouchdb/pull/8361

Issue #8345, opened 2021-08-25
Improving the state of PouchDB CI


jcoglan:
Me and the team at [Neighbourhoodie][NH] (including @AlbaHerrerias, @garbados,
@janl) are trying to ascertain the state of the new `indexeddb` adapter, to get
its functionality and performance to a point where we're confident deploying it
in production and potentially making it a core/default adapter in PouchDB. In
order to do this we've been looking into the current state of the test suite and
CI on Travis. Our first aim is to make sure `indexeddb` has feature parity with
`idb` and passes all the same functional tests, and we're opening this issue to
start a conversation among ourselves and other PouchDB maintainers to get the
build into a state where we're confident this is the case.

We have a [branch][ci-branch] that addresses a few problems we've found with the
build so far:

- Update `pouchdb-express-router` to `0.0.11`, which [fixes the heartbeat
  logic][heartbeat-fix] so that our integration tests work reliably without
  timing out in cases where `changes()` requests need to wait before receiving
  data via long-polling.

- Correct the path used to load PouchDB in
  `express-pouch-minimum-for-pouchdb.js`.

- The `ADAPTER` option appears to do nothing for browser tests; we've unified
  the `ADAPTER` and `ADAPTERS` options and changed `webrunner.js` so that it
  loads any requested adapters if they're packaged as plugins rather than part
  of the core distribution.

- To allow `indexeddb` to be tested, we've turned it into a plugin so that it
  compiles into a file in `packages/node_modules/pouchdb/dist` and can be loaded
  via the `ADAPTERS` option. This might not be desired for including in releases
  but it was the easiest way to get the tests to run with this adapter.

On this branch, I have run the `npm run test` task with various combinations of
options and recorded the results, setting `BAIL=0` so the full suite runs. The
first column in this table indicates whether this combination is currently
configured to run on Travis. Rows marked `y*` are configured, but we believe
that these actually execute the `idb` adapter because of either the `ADAPTER`
variable having no effect, or `indexeddb` not being loaded.

CI  | CLIENT           | ADAPTERS     | PLUGINS      | TYPE        | result
--- | ---------------- | ------------ | ------------ | ----------- | ---------------------------------------
y   | node             |              |              | integration | pass: 1668, pend: 45
y   | node             |              |              | mapreduce   | pass: 415, pend: 2, fail: 16
n   | node             |              | pouchdb-find | find        | pass: 343, pend: 156, fail: 228
[...]

PR #8351, opened 2021-08-27
Fix the `ADAPTERS` option

jcoglan:
This addresses an issue I mentioned in #8345. While running the tests we were confused about the behaviour of the `ADAPTER` and `ADAPTERS` options, particularly when running in the browser. We observed that:

- `ADAPTER` (singular) seems to have no effect, we can't find anything in the browser build scripts that makes use of it other that adding it to the query string of `tests/integration/index.html`; we believe this means that CI tasks using `ADAPTER=indexeddb` will have had no effect, and will actually have been running the `idb` adapter.
- `ADAPTERS` (plural) has two responsibilities: it sets the value of `PouchDB.preferredAdapters` and so changes the default adapter choice, _and_ it causes any requested adapters to be loaded from `packages/node_modules/pouchdb/dist`.

The problem this causes is that setting e.g. `ADAPTERS=idb` causes the build to fail, because `idb` is baked into the core PouchDB bundle, it's not a separate file. Trying to [load it dynamically](https://github.com/pouchdb/pouchdb/blob/9f5da871ac5748bca187f97b3926326cdbddcce5/tests/integration/webrunner.js#L15-L24) fails, which ultimately causes `window.testEvents()` to not be defined, which causes an error when the test script tries to poll for results.

This PR addresses these problems by unifying the two options, keeping the plural form which was added much earlier in the project history. Requested adapters are only dynamically loaded if we know they are a plugin compiled into the `pouchdb/dist` directory.

The final commit here makes the new `indexeddb` adapter into such a plugin, in order that it can be loaded in a browser-compatible bundle and activated via the `ADAPTERS` option. This might require further work if we don't wish to distribute this adapter as part of the `pouchdb` package but it's currently the simplest way to be able to compile a browser bundle and run the tests against it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,29 +31,29 @@ env:
   - CLIENT=node COMMAND=test
 
   # Test in-memory in Node
-  #- CLIENT=node ADAPTER=memory COMMAND=test
+  #- CLIENT=node ADAPTERS=memory COMMAND=test
 
   # Test in firefox running on travis
-  #- CLIENT=selenium:firefox ADAPTER=idb COMMAND=test
-  - CLIENT=selenium:firefox ADAPTER=idb POUCHDB_SRC="../../packages/node_modules/pouchdb/dist/pouchdb.min.js" COMMAND=test
-  #- CLIENT=selenium:firefox ADAPTER=indexeddb COMMAND=test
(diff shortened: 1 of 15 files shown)


dheimoz (approved):
Thank you very much for your contributions @jcoglan 
Definitely I owe you a beer or two :)

janl:
I created a follow-up issue so we don’t forget to revisit whether the `indexeddb` adapter should be part of the default distribution.

PR #8358, opened 2021-09-08
Fix escaping of key path segments in indexeddb adapter

jcoglan:
The test `deeper values can be escaped` in `tests/find/test-suite-1/test.escaping.js` is broken on `indexeddb` because its `sanitise()` function does not deal correctly with key paths like `foo.0bar`, where a path segment is not a valid JS name. This is because the `PATH_INVALID` regex only handles the special case where a digit appears at the very beginning of a string, not at the beginning of a path segment following a `.` character.

Here @AlbaHerrerias and I have fixed this by replacing `^` with `(^|\.)` in the regex. We considered using a lookbehind like `(?<=\.)` but this isn't supported in Safari. That means we also need to handle `.` characters we don't want to escape, e.g. in the key path `foo.0bar` the regex will match `.0` and we only want to escape the `0`, but leave the `.` as-is to be interpreted by IndexedDB. So the resulting string should be `foo._c48_bar`.

We handle this by noting the pattern has three branches:

- `\\.`: this will always match a literal `\` followed by a single char
- `(^|\.)[^a-zA-Z_$]`: this will either be a single char that's not a legal first char for a JS name, or it will be a `.` followed by such a char; it could therefore match `'..'`
- `[^a-zA-Z0-9_$.]+`: this will never contain a `.` char

So, we special-case `.` chars appearing as the first char in the match, since these must be produced by the second branch. This is tightly coupled to the structure of the pattern though, and might be easily broken, so I'd appreciate any suggestion for a better way to handle this.

### DIFF
--- a/packages/node_modules/pouchdb-adapter-indexeddb/src/rewrite.js
+++ b/packages/node_modules/pouchdb-adapter-indexeddb/src/rewrite.js
@@ -4,6 +4,19 @@ var IDB_NULL = Number.MIN_SAFE_INTEGER;
 var IDB_FALSE = Number.MIN_SAFE_INTEGER + 1;
 var IDB_TRUE = Number.MIN_SAFE_INTEGER + 2;
 
+// These are the same as bellow but without the global flag
+// we want to use RegExp.test because it's really fast, but the global flag
+// makes the regex const stateful (seriously) as it walked through all instances
+var TEST_KEY_INVALID = /^[^a-zA-Z_$]|[^a-zA-Z0-9_$]+/;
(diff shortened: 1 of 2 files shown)


jcoglan:
Note that to test this requires the changes from #8351. You can then run this command to verify the tests failing and passing before/after this change:

```
$ BAIL=0 TYPE=find PLUGINS=pouchdb-find CLIENT=selenium:firefox ADAPTERS=indexeddb COUCH_HOST='http://admin:admin@localhost:5984' npm run test
```

PR #8359, opened 2021-09-09
Run "find" tests against a single implementation at a time

jcoglan:
A problem we've run into with running the test suite is figuring out which combinations of settings for the `ADAPTERS` and `SERVER` options pass or fail, and which combinations we actually care about supporting. This is partly a documentation problem, and I intend to improve the testing docs after a few of our PRs have landed, but it's also not clear from the Travis CI config which combinations are supported.

We also found it problematic that the `find` tests work by running the entire test suite twice (once with a local adapter and once with an http one), for a couple of reasons. First, the default choice for the `SERVER` variable is `pouchdb-express-router`, which doesn't support `find()`, and so half the tests fail. To figure out what was supposed to work, we were trying the tests with different combinations of `ADAPTERS` and `SERVER` and the results are confusing and made us think there was something about the _combination_ of these settings that was important, when actually the tests are just being run twice with two different adapters. Second, it makes the tests take a lot longer than necessary -- if you're just interested in checking that `idb` works, you waste a lot of time also running the tests against an http adapter.

You get around this using `GREP` _if you know this is how the tests work_, but I think it would be clearer if the `find` tests ran against a single adapter at a time, and we specify in the CI config all the adapters we're interested in supporting. As far as we're able to tell, these are the builds we want:

- Test against CouchDB, since this is the reference implementation that PouchDB adapters must be compatible with
- Test with `pouchdb-server`, which is supposed to be API-compatible with CouchDB
- Test the local adapters for each environment, i.e. `leveldb` and `memory` on Node.js, and `idb`, `indexeddb` and `memory` in the browser

This PR essentially removes the dynamic generation of tests in the `find` suite that causes all tests to be defined twice. Instead of running one http adapter and one local one in each run, the tests run just once and the environment variables determine which implementation we're testing. I've listed the settings we should be testing in `.travis.yml`.

We may also want to apply a similar transform to the `mapreduce` tests but we'll do that as a separate PR if this one is accepted. The main motivation for this was we were trying to isolate failing tests in the `indexeddb` adapter, and running a single implementation per test run let us do this faster.

You may need to pull in the changes from #8351 to verify that the test configs here actually work.


jcoglan:
A note on the diff, since it's enourmous: most of the changed lines here are just removing one level of indentation, after removing the outer `testCases.push(function () { ... })` wrapper. All I did we remove that, remove the `dbType` part from the `describe`, move any local definitions inside the `describe`, and then move the whole body one tab to the left.

dheimoz:
@jcoglan hey James. I would love to merge this, however I am not comfortable with the conflicts detected. Could you help me to address them?

jcoglan:
Hey @dheimoz I've rebased and resolved the conflict -- this branch was conflicting with the changes I made when I replaced `ADAPTER` with `ADAPTERS` everywhere; the whole `find` block of the CI config should take take the matrix from this PR.

dheimoz (approved):
AMAZING JOB!!!!!!

PR #8363, opened 2021-09-15
Quote arguments to mocha(1) to make GREP work correctly

jcoglan:
Because the arguments in this script are not quoted, using multi-word values for `GREP` to select a particular test doesn't work when running tests on Node.js. For example if you set `GREP='should query correctly with skip'`, the script ends up running `mocha --grep=should query correctly with skip`, so only the word `should` is used for the `--grep` option and the other words become positional arguments and ignored.

### DIFF
--- a/bin/test-node.sh
+++ b/bin/test-node.sh
@@ -33,20 +33,20 @@ if [ $PERF ]; then
 elif [ ! $COVERAGE ]; then
     ./node_modules/.bin/mocha \
         $BAIL_OPT \
-        --timeout $TIMEOUT \
+        --timeout "$TIMEOUT" \
         --require=./tests/integration/node.setup.js \
-        --reporter=$REPORTER \
-        --grep=$GREP \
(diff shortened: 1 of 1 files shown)


dheimoz (approved):

dheimoz:
hey @jcoglan , as always, thank you very much for the contributions.
Do you have idea why coverage/coveralls has always the status of "Expected"?

jcoglan:
@dheimoz I'm not sure but I think it's something to do with access credentials to coveralls which are owned by the pouchdb org. We'll shortly be opening a PR to migrate to GitHub Actions to that would be a good opportunity to resolve the coveralls integration.

dheimoz:
@jcoglan hey James. you guys are doing an incredible work with your PR's.
I would love to assist to anything you need.

dheimoz:
thanks @janl I was planning to leave the branches at least for a month, before deleting. But you know better than I.

Once again, thanks for the efforts of your team.

PR #8365, opened 2021-09-16
Fix mapreduce tests

jcoglan:
This is another PR relating to the work described in https://github.com/pouchdb/pouchdb/issues/8345. This collection of commits gets the `mapreduce` tests back into working shape. To explain the commits...

First we make the tests run against a single adapter, similar to https://github.com/pouchdb/pouchdb/pull/8359. We have some uncertainty around the test structure here, which is more ambiguous than the `find` tests in terms of the intent of running tests against multiple adapters:

- `test.persistent.js` contains only tests that use design documents for view queries
- `test.views.js` contains only tests that make ad-hoc queries without persistent views/ddocs
- `test.mapreduce.js` contains a mixture of these

Since CouchDB no longer supports temporary views, we've skipped the whole of `test.views.js` for `http` adapters. Some tests in `test.mapreduce.js` are selectively skipped, and the `createView()` function changes behaviour on `http` adapters to define a design doc instead of running an ad-hoc query. We're not sure if this matches the intent of how these tests are organised, but we figured we should still test ad-hoc queries as PouchDB local adapters still support them. So we continue to run `test.views.js` unmodified for local adapters, but we've changed some `test.mapreduce.js` tests to use ddocs instead of ad-hoc queries. We'd appreciate feedback on whether this matches the tests' intent.

Those structural changes are the first three commits. Following those, we have:

- CouchDB can sometimes return a 500 error instead of 400 or 404 so we adjust some tests to allow this
- Fix a bug in `indexeddb` so that a doc with ID `hasOwnProperty` doesn't break `bulkDocs()`
- Stop destroying the DB before tests; this shouldn't be necessary and sometimes makes CouchDB return an error
- Create databases with `q=1` when using `http`. We discovered some inconsistencies in view results when CouchDB has multiple shards, e.g. rows with equal `key`/`id` having undefined sort order, and metadata like `total_rows` being absent. We've filed an [issue against CouchDB](https://github.com/apache/couchdb/issues/3750) relating to this. We also have a commit that we've saved on the branch [`fix/mapreduce-sort-order`](https://github.com/neighbourhoodie/pouchdb/tree/fix/mapreduce-sort-order) that updates tests affected by this, should we later decide to run tests against multi-shard DBs, but it's not included in this PR. We can include this instead of the `q=1` change if desired.
- Generate DB names consistently so that they are URLs when using the `http` adapter and we don't have a mix of local and remote DBs. Again the intent here is unclear -- is it a necessary part of these tests that local vs remote DBs are used, or do we just need to replicate from one DB to another and we don't care what backends those DBs use as long as all of them get tested via CI?
- Update the CI config to list all the configs we want to test.

We'd really appreciate feedback around the ambiguities we've identified.

If/when this gets merged, if we squash-merge it can we make sure author information is retained so my colleagues retain credit on their contributions? :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,9 +45,15 @@ env:
   #- AUTO_COMPACTION=true CLIENT=selenium:firefox ADAPTERS=indexeddb COMMAND=test
 
   # Test map/reduce
-  - TYPE=mapreduce CLIENT=node COMMAND=test
-  #- TYPE=mapreduce CLIENT=selenium:firefox ADAPTERS=idb COMMAND=test
-  - TYPE=mapreduce CLIENT=selenium:firefox ADAPTERS=indexeddb COMMAND=test
+  - TYPE=mapreduce CLIENT=node ADAPTERS=http SERVER=couchdb-master COUCH_HOST=http://127.0.0.1:3001
+  - TYPE=mapreduce CLIENT=node ADAPTERS=http SERVER=pouchdb-server
+  - TYPE=mapreduce CLIENT=node ADAPTERS=leveldb
+  - TYPE=mapreduce CLIENT=node ADAPTERS=memory
+  - TYPE=mapreduce CLIENT=selenium:firefox ADAPTERS=http SERVER=couchdb-master COUCH_HOST=http://127.0.0.1:3001
+  - TYPE=mapreduce CLIENT=selenium:firefox ADAPTERS=http SERVER=pouchdb-server
+  - TYPE=mapreduce CLIENT=selenium:firefox ADAPTERS=idb
+  - TYPE=mapreduce CLIENT=selenium:firefox ADAPTERS=indexeddb
+  - TYPE=mapreduce CLIENT=selenium:firefox ADAPTERS=memory
 
   # Test pouchdb-find
   - TYPE=find PLUGINS=pouchdb-find CLIENT=node ADAPTERS=http SERVER=couchdb-master COUCH_HOST=http://127.0.0.1:3001
--- a/packages/node_modules/pouchdb-adapter-indexeddb/src/bulkDocs.js
+++ b/packages/node_modules/pouchdb-adapter-indexeddb/src/bulkDocs.js
@@ -79,17 +79,17 @@ export default function (api, req, opts, metadata, dbOpts, idbChanges, callback)
       var newDoc;
 
       // The first document write cannot be a deletion
-      if ('was_delete' in opts && !(oldDocs.hasOwnProperty(doc.id))) {
+      if ('was_delete' in opts && !(Object.prototype.hasOwnProperty.call(oldDocs, doc.id))) {
         newDoc = createError(MISSING_DOC, 'deleted');
 
       // The first write of a document cannot specify a revision
       } else if (opts.new_edits &&
-                 !oldDocs.hasOwnProperty(doc.id) &&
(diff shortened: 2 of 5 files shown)


jcoglan:
I will also note that there are some outstanding failures on `pouchdb-server`, which you can see if you run:

```
$ TYPE=mapreduce CLIENT=node ADAPTERS=http SERVER=pouchdb-server TRAVIS_REPO_SLUG=pouchdb/pouchdb BAIL=0 npm run test
```

These seem like issues with `pouchdb-server` itself since the tests pass on all other backends. Since we're primarily focussed on working on `indexeddb`, and getting CI into shape to support that, we've not investigated these `pouchdb-server` failures. We could skip it in CI and open separate issues for them if desired.

dheimoz (approved):
Astonished work!!!! Thanks guys.

dheimoz:
> I will also note that there are some outstanding failures on `pouchdb-server`, which you can see if you run:
> 
> ```
> $ TYPE=mapreduce CLIENT=node ADAPTERS=http SERVER=pouchdb-server TRAVIS_REPO_SLUG=pouchdb/pouchdb BAIL=0 npm run test
> ```
> 
> These seem like issues with `pouchdb-server` itself since the tests pass on all other backends. Since we're primarily focussed on working on `indexeddb`, and getting CI into shape to support that, we've not investigated these `pouchdb-server` failures. We could skip it in CI and open separate issues for them if desired.

hey James @jcoglan , how many tests will failed should we merge the PR? I would be afraid of losing control.

jcoglan:
@dheimoz These are the results we're currently getting:

    TYPE=mapreduce CLIENT=node ADAPTERS=leveldb npm test
    => 227 passing, 2 pending

    TYPE=mapreduce CLIENT=node ADAPTERS=memory npm test
    => 227 passing, 2 pending

    TYPE=mapreduce CLIENT=node ADAPTERS=http COUCH_HOST='http://admin:admin@localhost:5984' npm test
    => 201 passing

    TYPE=mapreduce CLIENT=node ADAPTERS=http SERVER=pouchdb-server npm test
    => 195 passing, 6 failing

    TYPE=mapreduce CLIENT=selenium:firefox ADAPTERS=idb npm test
    => 226 passing, 1 pending

    TYPE=mapreduce CLIENT=selenium:firefox ADAPTERS=indexeddb npm test
    => 226 passing, 1 pending

    TYPE=mapreduce CLIENT=selenium:firefox ADAPTERS=memory npm test
    => 226 passing, 1 pending

    TYPE=mapreduce CLIENT=selenium:firefox ADAPTERS=http COUCH_HOST='http://admin:admin@localhost:5984' npm test
    => 201 passing

    TYPE=mapreduce CLIENT=selenium:firefox ADAPTERS=http SERVER=pouchdb-server npm test
    => 196 passing, 5 failing

So the tests pass against my local CouchDB (v3.1.1) but not against `pouchdb-server`. (I ran these with `export TRAVIS_REPO_SLUG=pouchdb/pouchdb` so that `pouchdb-server` gets rebuilt during `npm test`.)

I can't get the tests to run with the `localstorage` adapter but that seems to be a setup issue -- the tests just fail to run at all for the most part. I'm not sure this is related to our work; I've never got any of the test suites to work with `localstorage` but fixing that is not a priority for us right now.

dheimoz:
@jcoglan hey James, thanks for the report.

PR #8368, opened 2021-09-21
Migrate CI to GitHub Actions

janl:
This PR introduces a CI setup for GitHub Actions (GHA). 🎉 It tries to faithfully copy what was tested on Travis and improve upon that.

Notable changes:
- remove support for Travis CI (RIP)
- configures an expandable matrix of running tests in different configurations:
  - Node.js 12 (later Node versions can easily be added later)
  - CouchDB 2.3.1 and 3.1.1 (later CouchDB versions can easily be added when they come out)
- run Selenium tests in Chrome in addition to Firefox, other browsers can be added later
- run find and map reduce tests in configurations where they make sense only
- re-introduces (but disabled by default for now) running test suites against real CouchDB versions
- run Selenium tests against `idb` and `indexeddb`

There are a few open items that we are keen to close out, but also hope that other folks in the community can help finish those up, as our focus is on getting to a reliable CI situation and what we have is better than the nothing that currently exists :)

Open items:
- the coverage test is disabled for now as it didn’t pass on our test repo due to missing credentials. This should be as easy as configuring GHA with the right Coveralls credentials (cc @daleharvey)
- the `nom run test-webpack` job is currently commented out, we haven't gotten to this yet.
- there are a handful of tests that reliably but randomly fail in the replication and sync test suites. We have added copious `console.log`-debug output to the ones we’ve found so far to see what the problem there is. These tests currently only fail on GHA but we haven’t gotten them to fail locally. It’d be interesting to know if those tests were flakey on Travis as well?

Misc items:
- we added support for handling responses from clustered CouchDBs/sharded databases to tests to make it easier to run tests on a dev machine or other existing CouchDB installation.
- local dev couches from 3.x onwards have enforced user/pass, we augmented some test utils to handle this better, as well as updated the test scripts to configure CORS correctly for that configuration.


dheimoz:
hey @janl kudos to you and the whole team !!! @AlbaHerrerias , @garbados and @jcoglan.
I assume what is missing is what is documented, or have you found anything else? 

Hey @daleharvey , what do you think of this great collaboration? Would love to have your feedback.

Issue #8384, opened 2021-10-19
Find query does not work when indexing on _id with indexeddb adapter 


paultranvan:
### Issue
After creating an index including `_id` in the fields, trying to query on this index with `find` always return `docs: []`, even though there are documents matching the query.

### Info
- Environment: browser
- Platform: Chrome
- Adapter: IndexedDB

### Reproduce

```
await db.post({name: 'test'})
await db.createIndex({index: { fields: ['name', '_id']}, ddoc: 'by-name-and-id'})
await db.find({selector: { name: 'test', _id: { $gt: null}}, use_index: '_design/by-name-id', sort: [{'name': 'asc'}]})
```
This will always return: `docs: []`

Note I reproduce the issue with the `indexeddb` adapter but not `idb`.

dheimoz:
hey @paultranvan , would you mind creating a repo?
Thanks.

PR #8389, opened 2021-10-25
_id was mappped to data._id instead of id

jlami:
Indexes with toplevel fields were mapped incorrectly.
This might need a version upgrade of the index names by going from `_find_idx/...` to `find_idx2/...` in `naturalIndexName` or something if we want to fix existing databases.

### DIFF
--- a/packages/node_modules/pouchdb-adapter-indexeddb/src/util.js
+++ b/packages/node_modules/pouchdb-adapter-indexeddb/src/util.js
@@ -71,7 +71,7 @@ function correctIndexFields(fields) {
   // we need to filter out deleted documents.
   return ['deleted'].concat(
     fields.map(function (field) {
-      if (field in ['_id', '_rev', '_deleted', '_attachments']) {
+      if (['_id', '_rev', '_deleted', '_attachments'].includes(field)) {
         // These properties are stored at the top level without the underscore
         return field.substr(1);
       } else {
--- a/tests/find/index.html
+++ b/tests/find/index.html
@@ -59,6 +59,7 @@
     <script src='./test-suite-2/test.kitchen-sink.js'></script>
     <script src='./test-suite-2/test.kitchen-sink-2.js'></script>
     <script src='./test-issues/test.issue7810.js'></script>
+    <script src='./test-issues/test.issue8389.js'></script>
     <script type="text/javascript" src="../integration/webrunner.js"></script>
   </body>
 </html>
--- a/tests/find/test-issues/test.issue8389.js
+++ b/tests/find/test-issues/test.issue8389.js
@@ -1,0 +1,95 @@
+"use strict";
+
+describe("test.issue8389.js", function () {
+  var adapter = testUtils.adapterType();
+  var db = null;
+  var dbName = null;
+  
+  const docData = {
+    _id: "foobar",
+    indexedField: "foobaz",
+  };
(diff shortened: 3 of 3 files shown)


jlami:
Could somebody look at this very simple fix and the test that proves that this works (in #8395).
I keep having to fix this in my local installs ;-)

jlami:
Can you take a look at this fix and the PR for the test (#8395) @janl ?

broerse:
@janl we also need this fix. What is the merge plan?

garrensmith (commented):
This looks good to me. Do the CI tests use this adapter? Once we know that, we can merge this.

broerse:
> This looks good to me. Do the CI tests use this adapter? Once we know that, we can merge this.

Thanks!

broerse (approved):
Looks good to me

jlami:
I don't think it would be a problem. The old indexes should still work the same way. 

Only if you want some way to fix old databases do you have to migrate them. But like I stated in the opening post, we could also rename the indexes by adding a version number, then the old indexes should be replaced with the new ones.

janl:
would you be able to spell out here (for a casual reviewer of this) what happened before, what state exiting databases are in, what happens with the fix and what the new code does with databases that were created before?

jlami:
I'll try to elaborate on this later today

jlami:
### Old situation:
The native indexes are just mappings. And the mapping for _id and others was wrong.
- Indexes without those toplevel fields work.
- Indexes with toplevel fields are wrong.

### New situation:
The mapping has changed to fix the toplevel fields. But without a name change in `naturalIndexName` the mappings wont be updated.
https://github.com/pouchdb/pouchdb/blob/f7901fcb3e218de8ecb952035beff6ce01258924/packages/node_modules/pouchdb-adapter-indexeddb/src/setup.js#L129
This function builds the required indexes on `onupgradeneeded`. So it would be run on a new pouchdb version. But it only checks the indexes on the names.

https://github.com/pouchdb/pouchdb/blob/f7901fcb3e218de8ecb952035beff6ce01258924/packages/node_modules/pouchdb-adapter-indexeddb/src/setup.js#L62
https://github.com/pouchdb/pouchdb/blob/f7901fcb3e218de8ecb952035beff6ce01258924/packages/node_modules/pouchdb-adapter-indexeddb/src/setup.js#L73-L90
#### Without changing index names:
- Old indexes that worked before will still work, since nothing will be changed.
- But old indexes that did have toplevel fields will also not be updated, and thus also not work. 
- New databases would create correct indexes

#### With index name change
If we change the index versioning in `naturalIndexName` all the indexes will be recreated.
- The already working indexes will be recreated bu actually remain the same.
- Only the indexes with toplevel fields will actually change and start working.

janl:
@jlami thanks so much this is an excellent rundown. I now agree with merging this as is. We should make a note in the release notes for the version that this ships in that folks who need this will have to delete/recreate their indexes to get the fix. Forcing an index rebuild for all existing databases is not really an option. Sorry for going down the wrong path here, but I’m glad we talked about this :)

janl (approved):

jlami:
no problem, glad to see this merged

paultranvan:
Very nice, I think this also solves https://github.com/pouchdb/pouchdb/issues/8384

PR #8395, opened 2021-10-26
Failing test for PR #8389

jlami:
This test uses an index with 2 fields, where one is _id
This fails in indexeddb but works in idb


jlami:
This test is failing as expected: https://github.com/pouchdb/pouchdb/runs/4034130603?check_suite_focus=true
So is the firefox one, but there are 2 other spurious errors that I don't think I caused.
Should I rebase to trigger another run?

PR #8396, opened 2021-10-26
Adjust revs_limit in existing test to reproduce RangeError

jlami:
This should fail for indexeddb on the following test:
`test.bulk_docs.js-local`
`2839 implement revs_limit`


janl:
also todo for us: look into running tests on windows

alxndrsn:
I don't understand the rationale for the suggested test change.

Should there be an additional test which deliberately generates `RangeError` cross-platform and ensures that it's handled as expected?

Issue #8370, opened 2021-09-28
`limit` and `skip` view query options are ignored when `keys` are used, if the adapter is not http


dianabarsan:
### Issue
When running a `query` request, passing `keys` will always return the full set of results that match these keys, regardless of the passed `skip` and `limit` parameters. 

Relevant code: https://github.com/pouchdb/pouchdb/blob/master/packages/node_modules/pouchdb-abstract-mapreduce/src/index.js#L757

### Info
- Environment: node.js
- Platform: Linux
- Adapter: memory

### Reproduce

Setup:
- create a database using the memory adapter
- create a simple view and add some docs 
- query the view with `keys` and `limit` and notice that results are not limited
- query the view with `keys` and `skip` and notice the results are not skipped

Running the same code over a remote database works as expected. 

gist: https://gist.github.com/dianabarsan/ab9b903e3eae42089e730438e384e1da

dianabarsan:
This has been already reported several times: https://github.com/pouchdb/pouchdb/issues/7976 and https://github.com/pouchdb/pouchdb/issues/7573

ramblin-rose:
No stale-bot!

ramblin-rose:
No stale-bot, no!

AlbaHerrerias:
This issue was addressed by @ramblin-rose  in https://github.com/pouchdb/pouchdb/pull/8404. It has been merged to master and a new release will land soon. Thank you

Issue #8390, opened 2021-10-25
When querying a view the options variable is mutated when specifying an empty keys array


ramblin-rose:
### Issue
When querying a view, the `options` variable is mutated when specifying an empty keys array by deleting the `keys` property and adding the property `limit = 0`. This side effect taints the test space.

### Info
- Package: pouchdb-abstract-mapreduce

### Reproduce
```lang-js
(async () => {
  const db = new PouchDB("database", { adapter: "memory" });
  await db.put({
    _id: "_design/main",
     views: {
       main: {
	 map: `function(doc){ emit(doc.key) }`,
	},
      },
  });
  let opts = { keys: [] };
  console.log(`\nopts before: ${JSON.stringify(opts, undefined, 2)}`);
  let result = await db.query("main/main", opts);
  console.log(`\nopts after: ${JSON.stringify(opts, undefined, 2)}`);
})();
```
Console output:
>$ node index.js
>
>opts before: {
>  "keys": []
>}
>
>opts after: {
>  "limit": 0
>}
		
### Description, or Why This Matters
I am close to submitting a PR that fixes #8370. However testing fails on the 'returns one doc' expectation

>1) test.mapreduce.js-persisted Testing query with keys:
>     AssertionError: returns one doc: expected [] to have a length of 1 but got 0
>      at tests\mapreduce\test.mapreduce.js:2080:33

The germane test.mapreduce.js code:
```lang-js
var opts = {include_docs: true};
. . .
}).then(function (data) {
  data.rows.should.have.length(7, 'returns all docs');
  opts.keys = [];
  return db.query(queryFun, opts);
}).then(function (data) {
  data.rows.should.have.length(0, 'returns 0 docs');
  opts.keys = [0];
  return db.query(queryFun, opts);
}).then(function (data) {      
  data.rows.should.have.length(1, 'returns one doc');  
. . .
```
With my change the problem is revealed: the `opts` parameter is mutated when `opts.keys = []`
by the following pouchdb-abstract-mapreduce code 
```lang-js
function queryViewInQueue(view, opts) {
  var totalRows;
  var shouldReduce = view.reduceFun && opts.reduce !== false;
  var skip = opts.skip || 0;
  if (typeof opts.keys !== 'undefined' && !opts.keys.length) {
    // equivalent query
    opts.limit = 0;
    delete opts.keys;
  }
}
```
[...]

PR #8400, opened 2021-10-30
#8370 Implement skip and limit for keys query

ramblin-rose:
The limit and skip parameters were not implemented for map/reduce
keys queries.


ramblin-rose:
Regarding modification of existing map-reduce tests see #8390

PR #8402, opened 2021-11-01
Add JWT authentication

### LINKED ISSUE #8387: Add support for JWT authentication
Currently pouchdb supports Basic Authentication, and https://github.com/pouchdb-community/pouchdb-authentication provides a higher level wrapper around couchdb's cookie authentication, but other than https://github.com/gyetvan-andras/pouchdb-adapter-http-jwt (which is unmaintained), there seems to be no support for Json Web Tokens. Whilst it is possible to inject the header via pouchdb's fetch hook, it would be preferable to be able to specify the token in a similar way to the username/password pair, and treat the token as an opaque string, leaving the management of the token to the pouchdb-using application.

I'm happy to write a PR adding this functionality, but it would be good to confirm that this would be accepted, and if there were any opinions on the design of the API (I'd plan to follow the current Basic Authentication API, but if people have other suggestions, I'm willing to look at them).

aragilar:
Closes #8387.

Currently there are no tests, I'm not sure what the test suite looks like, or how to use it, are there some docs for that?

### DIFF
--- a/packages/node_modules/pouchdb-adapter-http/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-http/src/index.js
@@ -173,6 +173,8 @@ function HttpPouch(opts, callback) {
       var str = nAuth.username + ':' + nAuth.password;
       var token = btoa(unescape(encodeURIComponent(str)));
       options.headers.set('Authorization', 'Basic ' + token);
+    } else if (opts.jwt_token) {
+      options.headers.set('Authorization', 'Bearer ' + opts.jwt_token);
     }
 
     var headers = opts.headers || {};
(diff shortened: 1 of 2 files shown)


aragilar:
Sorry, I meant to set this as draft until the tests were added.

packages/node_modules/pouchdb-adapter-http/src/index.js
```diff
@@ -173,6 +173,8 @@ function HttpPouch(opts, callback) {
       var str = nAuth.username + ':' + nAuth.password;
       var token = btoa(unescape(encodeURIComponent(str)));
       options.headers.set('Authorization', 'Basic ' + token);
+    } else if (opts.jwt_token) {
```
mqtik, line 176: This could easily be set by using `fetch` property, and I'd advice to use such property since allows you more freedom to do  operations (i.e. the token expired, and you have a refresh token to refresh credentials before proceeding with the fetch)

PR #8414, opened 2021-11-15
Add idb global database failure error as reason to transaction errors

romang-at-guarda-co:
When you exceed storage quota with pouchdb and idb (on chrome) this happens in order: 
1. A `DOMException` with `name: "QuotaExceededError"` is logged to console https://github.com/pouchdb/pouchdb/blob/a33bc407407f97b9f2382db364f9e49a0fa03b62/packages/node_modules/pouchdb-adapter-idb/src/index.js#L696
2. The database connection is closed https://github.com/pouchdb/pouchdb/blob/a33bc407407f97b9f2382db364f9e49a0fa03b62/packages/node_modules/pouchdb-adapter-idb/src/index.js#L697
3. Pouchdb operations that triggered the quota overflow or happened to coincide with the quota overflow are rejected with a `CustomPouchError` with `name: "indexed_db_went_bad", reason: "QuotaExceededError"`
4. Other operations that happen after the quota overflow and database closure (from step 2) are rejected with a `DOMException` with `name: "InvalidStateError", message: "Failed to execute 'transaction' on 'IDBDatabase': The database connection is closing."`. Thrown on this line: https://github.com/pouchdb/pouchdb/blob/a33bc407407f97b9f2382db364f9e49a0fa03b62/packages/node_modules/pouchdb-adapter-idb/src/utils.js#L220

Exception from step 1 are discarded by idb adapter (user code can't access them). This is a problem but this PR does not address it fully.

Exception from step 3 propagate to user code and have a meaningful reason string that can be used by user code to acknowledge the quota error. This is ok but IMO could be improved by using the original error object instead of just a string as a reason. This PR does not address it.

Exceptions from step 4 propagate to user code but do not have any data that can be used to link them to the original quota error from step 1. This is bad. **This PR fixes this** by attaching the exception from step 1 to exceptions from step 4 as a `reason` property.

You can find many stale pouchdb issues with `InvalidStateError`s of unknown origin. This could help users understand, debug and possibly programmatically recover from them or show meaningful error messages to user.


dheimoz (approved):
Thanks for the contribution.

AlbaHerrerias:
Hey @romang-at-guarda-co, could you rebase from master? It will allow me to re-trigger Github actions. I suspect the failing tests were due flakiness. Thanks!

AlbaHerrerias:
In order to retrigger GHA, we created a new PR with these changes https://github.com/pouchdb/pouchdb/pull/8473 Since that was merged, I'm closing this one. Thank you for your contribution!

PR #8421, opened 2021-11-30
Fix: replace replication suite call from www.infiniterequest.com to local down server

AlbaHerrerias:
Hello,

Currently, our Github Actions CI is failing in the following replication test: `suite2 test.replication.js-down-test`. As I understand, this test calls `www.infiniterequest.com` to receive a server error (not an infinite request as the name suggest):
image (3)

But _sometimes_ the request is successful as it redirects to a domain seller, which break our test:
image (2)

For that, we have created a local down server to test against. I also refactored the test to remove unused code and updated it to use modern JS.

Thank you

### DIFF
--- a/bin/test-node.sh
+++ b/bin/test-node.sh
@@ -12,6 +12,8 @@ else
 fi
 
 if [ $TYPE = "integration" ]; then
+    node bin/down-server.js 3010 & export DOWN_SERVER_PID=$!
+
     TESTS_PATH="tests/integration/test.*.js"
 fi
 if [ $TYPE = "fuzzy" ]; then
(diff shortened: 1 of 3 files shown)


janl:
this looks great, thanks! My only nitpick is that this server isn’t really “slow” and instead responds with a 500 error. Maybe we should clean up the naming while we are at it?

bin/test-node.sh
```diff
@@ -12,6 +12,8 @@ else
 fi
 
 if [ $TYPE = "integration" ]; then
+    node bin/down-server.js 3010 & export SLOW_SERVER_PID=$!
```
janl, line 15: this still mentions SLOW :)

AlbaHerrerias: Thank you for catching this!

janl: looks good to me, wanna squash this into your first commit? It wouldn’t make much sense to retain as a separate commit once this hits `master`.

janl (approved):

PR #8433, opened 2022-01-12
feat: add retries to failed jobs in gha ci

AlbaHerrerias:
We have been experiencing many flaky tests in our runs, an unknown cause for now. We have added two retry steps. Ideally, we should turn these steps into a composite action to avoid repetition but at the moment [it does not support](https://github.com/actions/runner/issues/1457) the `continue-on-error` property.

Thanks!


janl (approved):
excellent!

Issue #8422, opened 2021-12-01
🌟  Feature Proposal/Discussion: ActiveTasks


espy:
Hello PouchDB maintainers 👋

We (neighbourhood.ie) would like to propose a new PouchDB feature and use this space to discuss details with you.

## What
We would like to implement an `activeTasks` feature within PouchDB, analogous to the `_active_tasks` feature in CouchDB. 

## Why
Currently, there is little visibility into what PouchDB is doing with regard to (potentially) long-running tasks such as index building and replication. While it is possible to subscribe to sync events, we feel it would be helpful for developers to have a place where they can get instant answers to the question “what is PouchDB busy with and how long is it actually taking?” CouchDB’s Fauxton offers the active tasks page which is helpful in answering these questions, and we feel it would be useful if PouchDB did something similar.

## How
As a baseline, we want to make this as tiny and unobtrusive with as little performance overhead as possible. Our current idea is that there should be a registry on `PouchDB` that other modules can CRUD tasks to. This is purposely not an event emitter, but just an in-memory collection of task objects that you can read (poll) whenever you need to. In addition, we would make it opt-in, meaning that modules should only use `activeTasks` if it is explicitly enabled via `PouchDB.activeTasks.enable()`, which is similar to how `PouchDB.debug` is activated. 

If people want this to be subscribable, this functionality can be added via a plugin later.

### What is an active task?
A pojo with the following properties:
- `id`: a string set automatically on creation and passed back to the creator, a UUID for example
- `name`: a required string settable on creation, eg. “Building index `by-director` for database `studio-ghibli`”
- `started_at`: an ISO 8601 date string, set automatically on creation
- `updated_at`: an ISO 8601 date string, automatically updated on update
- `total_items`: a positive integer or zero, settable on create/update
- `completed_items`: a positive integer or zero, settable on create/update

### What can you do with an active task?
The registry is public on the `PouchDB` global, any module, plugin or adapter can use it. However, any consumers of this API should respect when `PouchDB.activeTasks.enabled` is `false` and not use it in that case. Additionally, the implementation can also ignore attempts to use it.

You can:

**1. Create** an active task by passing an object without an id to `create()`: 
```javascript
const myTaskId = PouchDB.activeTasks.create({
  name: 'Building index ABC for database XYZ',
  total_items: 133
})
```
[...]

PR #8441, opened 2022-01-25
activeTasks API

jankaszel:
This PR brings an implementation of the activeTasks API sketched in https://github.com/pouchdb/pouchdb/issues/8422, similar to what the `_active_tasks` API does for CouchDB. Further discussion can also be found in our development PR, https://github.com/neighbourhoodie/pouchdb/pull/8.

### Long-running tasks

We made several long-running tasks report to the activeTasks API:

- database compaction
- replication (single & continuous)
- view indexing
- find indexing (via view indexing/`abstract-mapreduce` when not using the `indexeddb` adapter)
  - the `indexeddb` adapter brings its own find indexing/querying implementation based on native IDB indexes, which don't support precise progress monitoring

### Tests

Run tests for this feature: 
```$ GREP="active_tasks" && npm test```

### Demo

We've built a demo application for trying out (and visualizing) the activeTasks API for monitoring long-running tasks, which helps with quickly spinning up dummy databases and running each kind of task: https://github.com/neighbourhoodie/pouchdb-activetasks-demo

### DIFF
--- a/docs/_includes/api/active_tasks.html
+++ b/docs/_includes/api/active_tasks.html
@@ -1,0 +1,45 @@
+{% include anchor.html edit="true" title="List active tasks" hash="active_tasks" %}
+
+{% highlight js %}
+PouchDB.activeTasks.list()
+{% endhighlight %}
+
+List all active database tasks. There are three types of internal tasks: `database_compaction`, `view_indexing`, and `replication`. PouchDB will report progress of these tasks to the active tasks API and remove tasks as soon as they are completed or have failed.
+
+#### Example Usage:
+
+{% highlight js %}
+var tasks = PouchDB.activeTasks.list()
+{% endhighlight %}
+
+#### Example Result:
+
+{% highlight js %}
+[{
+  "id": "d81fea92-8ce4-42df-bb2b-89a4e67536c3",
+  "name": "database_compaction",
+  "created_at": "2022-02-08T15:38:45.318Z",
+  "total_items": 12,
+  "completed_items": 1,
+  "updated_at": "2022-02-08T15:38:45.821Z"
+}]
+{% endhighlight %}
+
+### Real-time updates
+
+You can use [JavaScript Proxies](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) to monitor calls to the active tasks API. For the `PouchDB.activeTasks.add()` function, which is used internally to announce new tasks to PouchDB, you can monitor calls as follows:
+
+{% highlight js %}
+PouchDB.activeTasks.add = new Proxy(PouchDB.activeTasks.add, {
+  apply: (target, thisArg, argumentsList) => {
+    const task = argumentsList[0];
+    const id = Reflect.apply(
+      target,
+      PouchDB.activeTasks,
+      [task]
+    );
+    console.log('Added task', id, task.name);
+    return id;
+  },
+});
+{% endhighlight %}
--- a/packages/node_modules/pouchdb-core/src/active-tasks.js
+++ b/packages/node_modules/pouchdb-core/src/active-tasks.js
@@ -1,0 +1,48 @@
+import {v4 as uuidv4} from "uuid";
+
+export default class ActiveTasks {
+  constructor() {
+    this.tasks = {};
+  }
+
+  list() {
+    return Object.values(this.tasks);
+  }
+
+  add(task) {
+    const id = uuidv4();
+    this.tasks[id] = {
+      id,
+      name: task.name,
+      total_items: task.total_items,
+      created_at: new Date().toJSON()
+    };
+    return id;
+  }
+
+  get(id) {
+    return this.tasks[id];
+  }
+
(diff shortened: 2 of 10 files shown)

### CODE REVIEW

**Review by 92Pranjal — APPROVED** [review]

**Comment by jankaszel** [comment]
@janl thanks, docs are available now! we've covered just the public-facing API—`activeTasks.list()`—since `add()`, `update()`, and `remove()` are used internally for now. @espy and I have decided to hint to the [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) method for real-time updates, but kept it brief.

File: docs/_includes/api/active_tasks.html [hunk 1]
```diff
@@ -0,0 +1,45 @@
...
+#### Example Usage:
+
+{% highlight js %}
+var tasks = PouchDB.activeTasks.list()
+{% endhighlight %}
+
+#### Example Result:
+
+{% highlight js %}
+{
```
janl, line 18: this is not a list?

jankaszel: it's a list and shouldn't be an array :-) forgot the brackets.

packages/node_modules/pouchdb-core/src/setup.js
```diff
@@ -127,4 +128,6 @@ PouchDB.fetch = function (url, opts) {
   return fetch(url, opts);
 };
 
+PouchDB.prototype.activeTasks = PouchDB.activeTasks = new ActiveTasks();
```
jcoglan, line 131: What's the effect of assigning to `PouchDB.activeTasks` if you have multiple `PouchDB` instances around, or if a new instance is implicitly created to store a view index?

packages/node_modules/pouchdb-replication/src/replicate.js
```diff
@@ -131,6 +136,22 @@ function replicate(src, target, opts, returnValue, result) {
       returnValue.emit('change', outResult);
     }
     writingCheckpoint = true;
+
+    var task = src.activeTasks.get(taskId);
```
jcoglan, line 140: I think there's a possible race condition leading to lost updates in this section of code, because the call to `src.info()` is not synchronised with any of the surrounding code. We shouldn't block on the call to `info()` because this is just done for information purposes and it shouldn't hold up the replication process, but it does mean the task might be updated with a state `completed_items` value.

This would happen if the call to `info()` doesn't resolve quickly, but resolves some time during the next batch, or even later. What we have here is the following:

1. Read the `task` from `activeTasks.get()`
2. An arbitrary amount of time passes
3. Update task via `activeTasks.update()`

It's possible for multiple instances of this process to happen concurrently, meaning the `task` read in step 1 is stale by the time step 3 executes, so multiple instances of step 3 run with the same value of `completed`, causing a lost update.

This can be fixed by making sure you read and update the task in the same event loop tick, i.e. rewrite this section as:

```js
      src.info().then(function (info) {
        if (!currentBatch) {
          return;
        }
        var task = src.activeTasks.get(taskId);
        var completed = task.completed_items || 0;
        var total_items = info.update_seq - initial_last_seq;
        src.activeTasks.update(taskId, {
          completed_items: completed + currentBatch.changes.length,
          total_items
        });
      });
```

The calls to ``activeTasks.get()` and `activeTasks.update()` are now synchronous so the result of the `activeTasks.get()` can't be invalidated by this code executing concurrently.

packages/node_modules/pouchdb-core/src/active-tasks.js
```diff
@@ -0,0 +1,48 @@
...
+    return this.tasks[id];
+  }
+
+  /* eslint-disable no-unused-vars */
+  remove(id, reason) {
+    delete this.tasks[id];
+    return this.tasks;
+  }
+
+  update(id, updatedTask) {
```
jcoglan, line 33: This implementation is a nice protection against race conditions because it replaces the task with a _new object_. That means any code that's holding onto a task it got via `activeTasks.get()` won't mess anything up by mutating that object, and updates need to go via the `update()` method in order for other code to see them.

jcoglan (changes requested):

packages/node_modules/pouchdb-core/src/setup.js
jankaszel: I agree that this approach is a bit confusing; the original sketch had the active tasks API reside in `PouchDB.activeTasks`, but that way we can't access the object in `AbstractPouchDB` and other elements that compose the `PouchDB` object. So for technical reasons, we use `PouchDB.prototype.activeTasks` to make it accessible to other parts of the inheritance hierarchy of the PouchDB class.

On a semantic level, I find it sensible to monitor all PouchDB activity of a particular application rather than monitoring only certain instances.

Technically, we could drop the `PouchDB.activeTasks` reference and just keep the prototype reference in order to keep it simple.

packages/node_modules/pouchdb-replication/src/replicate.js
jankaszel: Good point, thanks! I've fixed this with your approach.

PR #8445, opened 2022-01-28
Remove argsarray dependency

AlbaHerrerias:
Remove `argsarray` package and use rest parameters instead.

These changes were initially proposed [here](https://github.com/pouchdb/pouchdb/pull/7779), but, as I understand, were rejected due to a CI failure. I tried them again, and the GHA tests passed. 

@lindner, as you were the first to propose these changes, please let me know if you’d like to be a co-author of these commits. 

Also noticed that `packages/node_modules/pouchdb-find/src/utils.js` had an inlined implementation of this dependency, so I removed it and used rest parameters as well.

Thank you.

### DIFF
--- a/bin/build-pouchdb.js
+++ b/bin/build-pouchdb.js
@@ -14,7 +14,6 @@ var rollupPlugins = require('./rollupPlugins');
 var rimraf = denodeify(require('rimraf'));
 var mkdirp = denodeify(require('mkdirp'));
 var all = Promise.all.bind(Promise);
-var argsarray = require('argsarray');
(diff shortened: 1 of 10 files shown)


lindner:
Sure, please feel free to add me.

janl:
Is this a PR we should hold for the next major release? We have a few more fixes and features pending for folks that might not want to opt into rest params yet?

AlbaHerrerias:
Sounds reasonable 👍  This PR can wait

janl (approved):

Issue #8444, opened 2022-01-27
[meta] Usage of ES6 classes instead of prototypes and inherits


AlbaHerrerias:
Since the usage of `inherits` it's [deprecated](https://nodejs.org/api/util.html#utilinheritsconstructor-superconstructor), and ES6 has native classes, I was wondering if we would be interested in switching. I made a small [POC refactor in the `changes` handler](https://github.com/neighbourhoodie/pouchdb/commit/3a345be7127ee007b124a63138c12350b64cebee) and the [GHA tests passes](https://github.com/neighbourhoodie/pouchdb/actions/runs/1757315227).

I am aware that [similar proposals](https://github.com/pouchdb/pouchdb/pull/7662/files) in the past have been rejected, but I was hoping the consensus on the usage of ES6 in PouchDB it’s more favourable now than then. 

Please let me know what you think. Thank you.

daleharvey:
Hi! So the previous reluctance around Es6 was primarily around introducing core dependencies, at the time I think babel would have been a hard requirement in what was already a complex build chain.

If I remember we actually have deps to implement `inherits` in the browser so this could potentially be removing deps, in that case its a huge +1, your proposed patch looks clean and idiomatic and a huge improvement so I think it would be great to go in that direction

AlbaHerrerias:
We have done some progress on this topic, and we have raised a PR [here](https://github.com/pouchdb/pouchdb/pull/8450).

Please note that two packages still use `inherits`:
1. `sublevel-poucdhb`
- [`readStream`](https://github.com/pouchdb/pouchdb/blob/9b51141a4a54ec79ac774c6117b5aa4c986d785b/packages/node_modules/sublevel-pouchdb/src/readStream.js#L27)

2. `pouchdb-core`
- [`adapter`](https://github.com/pouchdb/pouchdb/blob/b1742b1d9373f19d0ad070779839a9d033f2b39d/packages/node_modules/pouchdb-core/src/adapter.js#L166)
- [`constructor`](https://github.com/pouchdb/pouchdb/blob/e324499dce4fd3f96d9e8db2aed6875358cdad37/packages/node_modules/pouchdb-core/src/constructor.js#L36)
- [`setup`](https://github.com/pouchdb/pouchdb/blob/55c24df7d4e136b401f6981d378d6f6708885096/packages/node_modules/pouchdb-core/src/setup.js#L110)

[...]

PR #8450, opened 2022-02-15
(#8444) Refactor: use es6 classes

AlbaHerrerias:
Please read more info in this related issue #8444

* Use `classes` instead of `prototypes` when possible
* Deprecate the usage of the package `inherits` and use ES6 classes instead.


louwers:
This is a good start, but it may be a bit premature to close this issue.

PR #8453, opened 2022-02-15
Purging revs

jankaszel:
This PR will introduce new functionality for [purging documents similar to how CouchDB does it](https://docs.couchdb.org/en/stable/api/database/misc.html#db-purge) from a PouchDB with the `indexeddb` adapter.

### Description
The `db.purge(docId, rev, callback)` function directly updates the IndexedDB document and calculates a new winning rev, as well as updates the respective data fields. You can only purge a leaf revision, and doing so will remove all ancestors of that leaf up to the next branch, just like CouchDB does it. If no revisions remain, the document is removed entirely. Purge will also remove any attachments from the affected revisions, as well as delegate purge removals to views. Purges will trigger view updates as soon as they are queried.

Purge returns either an error (document doesn't exist, revision doesn't exist, revision isn't a leaf node), or a results object with the format:

```js
{
  ok: true,
  deletedRevs: [
    "6-3a24009a9525bde9e4bfa8a99046b00d",
    "5-df4a81cd21c75c71974d96e88a68fc2f"
  ],
  documentWasRemovedCompletely: false
}
```

Purge parameters (`docId`, `rev`) are being stored in a local document, `_local/purges`, alongside a `purgeSeq`. With that, on-demand consumers such as views can reflect the purges at a later time.

A purge should only be considered complete after subsequent database compaction and after view indexes have been updated (the latter is part of this PR).

#### Purge constructor option

[Similar to CouchDB](https://docs.couchdb.org/en/stable/api/database/misc.html#put--db-_purged_infos_limit), we've added a `purged_infos_limit` option to control how many purges PouchDB will keep track off in order to e.g. delegate them to views. This option defaults to `1000` (same as CouchDB).

```js
new PouchDB({
  purged_infos_limit: 1500
})
```

#### Utility functions

There are two new utility functions that operate on a document's rev tree. They are:

* `findPathToLeaf()` to calculate the path from the last branching within a tree up to a leaf in order to determine the path of revs that have to be deleted during a purge. It is called by the new purge implementation in `pouchdb-core`, and passes the resulting array of revisions down to the adapter-specific implementation, which expects them in this format and order. `findPathToLeaf()` does the checks for whether a rev exists and is a leaf, and the adapter-level implementations of `purge` can rely on this. There are explanatory comments in the code
* `removeLeafFromTree()` to then remove a leaf from a rev tree data structure (`revs_info`) from within the adapter-level implementations of `purge` 

Run the new unit tests with `npm run test-unit`.

#### Adapter support

If purge is called on an adapter that doesn't support it, it will throw an error. Currently, it is only supported by the `indexeddb` adapter. This implementation also contains explanatory comments, in case anyone wants to implement this for other adapters.

### DIFF
--- a/packages/node_modules/pouchdb-core/src/adapter.js
+++ b/packages/node_modules/pouchdb-core/src/adapter.js
@@ -20,7 +20,8 @@ import {
   rootToLeaf,
   collectConflicts,
   isDeleted,
-  isLocalId
+  isLocalId,
+  findPathToLeaf
 } from 'pouchdb-merge';
 import {
   MISSING_BULK_DOCS,
@@ -958,5 +959,42 @@ class AbstractPouchDB extends EventEmitter {
     return (typeof this._type === 'function') ? this._type() : this.adapter;
   }
 }
+
+AbstractPouchDB.prototype.purge = adapterFun('_purge', function (docId, rev) {
+  if (typeof this._purge === 'undefined') {
+    throw new Error('Purge is not implemented by the ' + this.adapter + ' adapter.');
+  }
+  var self = this;
+
+  // TODO: this should probably return some information
+  // about the completed purge
+  return new Promise((resolve) => {
+    self._getRevisionTree(docId, async (whatever, revs) => {
+      const path = findPathToLeaf(revs, rev);
+      console.log("purge path", path);
+
+      for (const revHash of path) {
+        await (async () =>
+          new Promise((resolve) =>
+            self._purge(docId, revHash, (err) => {
+              if (err) {
+                console.error("error", err);
+              } else {
+                console.log("purge successful");
+              }
+              resolve();
+            })
+          ))();
+      }
+
+      console.log(
+        "revs_info after purge",
+        await self.get(docId, { revs_info: true })
+      );
+
+      resolve();
+    });
+  });
+});
 
 export default AbstractPouchDB;
--- a/packages/node_modules/pouchdb-core/src/adapter.js
+++ b/packages/node_modules/pouchdb-core/src/adapter.js
@@ -973,10 +973,9 @@ AbstractPouchDB.prototype.purge = adapterFun('_purge', function (docId, rev) {
       const path = findPathToLeaf(revs, rev);
       console.log("purge path", path);
 
-      for (const revHash of path) {
         await (async () =>
           new Promise((resolve) =>
-            self._purge(docId, revHash, (err) => {
+            self._purge(docId, path, (err) => {
               if (err) {
                 console.error("error", err);
               } else {
@@ -985,12 +984,13 @@ AbstractPouchDB.prototype.purge = adapterFun('_purge', function (docId, rev) {
               resolve();
             })
           ))();
-      }
 
-      console.log(
-        "revs_info after purge",
-        await self.get(docId, { revs_info: true })
-      );
+      try {
+        console.log(
+          "revs_info after purge",
+          await self.get(docId, { revs_info: true })
+        );
+      } catch (err) {}
 
       resolve();
     });
--- a/packages/node_modules/pouchdb-core/src/adapter.js
+++ b/packages/node_modules/pouchdb-core/src/adapter.js
@@ -974,6 +974,46 @@ AbstractPouchDB.prototype.purge = adapterFun('_purge', function (docId, rev) {
       // `findPathToLeaf` throws if rev is invalid
       const path = findPathToLeaf(revs, rev);
       console.log("purge path", path);
+
+      async function appendPurgeSeq(docId, rev) {
+        let doc;
+
+        try {
+          doc = await self.get('_local/purges');
+          doc.purges.push({
+            docId,
+            rev,
+            purgeSeq: doc.purgeSeq + 1,
+          });
+          doc.purgeSeq += 1;
+        } catch (err) {
+          if (err.status !== 404) {
+            // something else? what could be reasons that this should fail,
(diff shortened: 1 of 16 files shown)


packages/node_modules/pouchdb-merge/src/removeLeafFromTree.js
```diff
@@ -0,0 +1,52 @@
...
+
+// this method removes a leaf from a rev tree, independent of its status.
+// e.g., by removing an available leaf, it could leave its predecessor as
+// a missing leaf and corrupting the tree.
+function removeLeafFromRevTree(tree, leafRev) {
+  const _tree = clone(tree);
+  const toVisit = _tree.slice();
+
+  // we will store branchings (node.branches > 1) as a LIFO stack
+  const branchings = [];
```
TheSlimvReal, line 11:
```suggestion
  const branchingNodes = [];
```
To distinct the purpose of this array a little more from the `branches` array.

TheSlimvReal (commented):
Thank you for taking this up. We highly need this feature in our project. 

Even for somebody new to the project I could quite fast understand how it works. I left some comments for smaller improvements.

packages/node_modules/pouchdb-merge/src/findPathToLeaf.js
```diff
@@ -0,0 +1,43 @@
+function findPathToLeaf(revs, targetRev) {
```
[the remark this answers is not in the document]
espy: Hi there, thanks for your comments. I’m afraid that due to this being a work in progress PR much of what you commented on is going to change anyway, so I’m not going to reply to each one. However, we’ve read the comments and will consider them, so thanks for taking the time 🙏

To answer this specific question, we’ve extracted some `rev_tree` structures from an `indexeddb` PouchDB for use in our [unit tests](https://github.com/pouchdb/pouchdb/blob/723df4127d8d689019dd535be3413291dcf2002b/tests/unit/test.purge.js), you can check out what they look like there. There are some fun cases like multi-root trees 🥳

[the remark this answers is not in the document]
TheSlimvReal: Thanks for the clarification. I am looking forward to following the progress of this PR. Let me know once any help is needed here.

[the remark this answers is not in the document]
espy: Will do, thanks for the offer!

packages/node_modules/pouchdb-core/src/adapter.js
```diff
@@ -508,6 +509,51 @@ AbstractPouchDB.prototype._compact = function (opts, callback) {
...
+
+  return new Promise((resolve) => {
+    self._getRevisionTree(docId, async (whatever, revs) => {
+      if (!revs) {
+        throw new Error('The target document doesn’t exist (has no revisions)');
+      }
+      // `findPathToLeaf` throws if rev is invalid
+      const path = findPathToLeaf(revs, rev);
+      console.log("purge path", path);
+      await (async () =>
```
TheSlimvReal, line 526: More out of curiosity: What is the reason for wrapping the `_purge` call into this `await (async () => {...})()` call? Is this done in order to push the execution on another thread or what is the background of this?

espy: A misunderstanding on my part, this has all been changed to use `callback()` correctly, and the awaits and promises removed.

tests/integration/test.attachments.js
```diff
@@ -3256,6 +3256,38 @@ adapters.forEach(function (adapter) {
...
+      }
+
+      const doc = { _id: 'foo' };
+      const base64 =
+        'iVBORw0KGgoAAAANSUhEUgAAAhgAAAJLCAYAAAClnu9J' +
+        'AAAgAElEQVR4Xuy9B7ylZXUu/p62T5nOMAPM0BVJICQi' +
+        'ogjEJN5ohEgQ';
+
+      return db.put(doc).then(function (res) {
+        return db.putAttachment('foo', 'foo.bin', res.rev, base64, 'image/png');
```
TheSlimvReal, line 3274: I guess you are not a big fan of arrow functions or the `async/await` syntax but these would help the readability of this (and the following tests) a lot.

E.g. arrow function:
```javascript
return db.put(doc)
   .then((res) => db.putAttachment('foo', 'foo.bin', res.rev, base64, 'image/png'))
   .then(() => db.get(doc._id))
   .then((_doc) => db.purge(doc._id, _doc._rev))
   .then(...);
```
or with `async/await`
```javascript
const res = await db.put(doc);
await db.putAttachment('foo', 'foo.bin', res.rev, base64, 'image/png');
const _doc = await db.get(doc._id);  // this step can also be skipped because you can use the `res.rev` in the next line again
await db.purge(doc._id, _doc._rev);
...
```

TheSlimvReal (commented):
Thank you for finishing this up so quickly. 
I just left a few notes.

tests/integration/test.attachments.js
janl: We’ll start introducing ES6 syntax soonish :)

jcoglan (approved):

goltra:
Hello, I'm not sure if this is the right site to make this question. If not, please let me know where can we discuss the next. 
Now a days if we have an app with pouch sync with couchdb, if we purge in couch, app will upload again deleted documents because pouch doesn't know purge. Does mean that with this new feat in pouch when we purge in couchdb it will purge in pouch too? if not, what will be the best way to purge same documents to avoid sync deleted documents ?

PR #8464, opened 2022-03-02
(8463) - check if docFieldValue is also null value.

paolosanchi:
The fix checks if docFieldValue is also null value.
typeof null returns object in javasctipy. Fixes a bug when querying against and array with a null value, $elemMatch and $allMatch.
Two tests are added for this case. One for $elemMatch and one for $allMatch


paolosanchi:
Sorry.. I made a mistake on the test.

Issue #8463, opened 2022-03-01
Cannot read properties of null (reading '$eq') with null value in array and $elemMatch or $allMatch


paolosanchi:
### Issue
I found an issue with the .find() function for the local database.
If the value of an array is null, the operator $elemMatch and $allMatch don't work.

example:
document:
`       
{
	_id: '1',
	values: [null, null]
}
`
queries:
```
{
	selector: {
        	values: {$elemMatch: {$eq: null}},
	}
}

{
	selector: {
        	values: {$allMatch: {$eq: null}},
	}
}
```

Obtained error:
```
TypeError: Cannot read properties of null (reading '$eq')
    at getFieldFromDoc (http://127.0.0.1:8000/packages/node_modules/pouchdb/dist/pouchdb.find.js:3225:18)
    at http://127.0.0.1:8000/packages/node_modules/pouchdb/dist/pouchdb.find.js:3972:25
    at Array.every (<anonymous>)
    at rowFilter (http://127.0.0.1:8000/packages/node_modules/pouchdb/dist/pouchdb.find.js:3969:25)
    at http://127.0.0.1:8000/packages/node_modules/pouchdb/dist/pouchdb.find.js:4154:16
    at Array.every (<anonymous>)
    at Object.$allMatch (http://127.0.0.1:8000/packages/node_modules/pouchdb/dist/pouchdb.find.js:4153:28)
    at match (http://127.0.0.1:8000/packages/node_modules/pouchdb/dist/pouchdb.find.js:4049:32)
    at http://127.0.0.1:8000/packages/node_modules/pouchdb/dist/pouchdb.find.js:3993:16
    at Array.every (<anonymous>)
```

I found that the issue is caused by in-memory-filter.js because the value null is mistaken for an object.
I've already fixed it and added the relative tests. I'll add a pull request soon.

AlbaHerrerias:
Your fix https://github.com/pouchdb/pouchdb/pull/8465 has been merged to master, so I'm closing this issue. Thank you again!

PR #8465, opened 2022-03-02
(#8463) - check if docFieldValue is also null value.

paolosanchi:
The fix checks if docFieldValue is also null value.
typeof null returns object in javasctipy. Fixes a bug when querying against and array with a null value, $elemMatch and $allMatch.
Two tests are added for this case. One for $elemMatch and one for $allMatch

Related issue: https://github.com/pouchdb/pouchdb/issues/8463


jcoglan (approved):

AlbaHerrerias:

PR #8471, opened 2022-04-03
BUG: Querying with $gt over a nested field returns the wrong results

pubkey:
This PR contains a test case to reproduce a bug in the pouch-find plugin.

Using the `$gt` operator over a nested field will return the wrong query results.

```js

await pouch.put({
    _id: 'hundred',
    nr: 100,
    nes: {
        ted: 100
    }
  });

const result = await pouch.find({
        selector: {
            'nes.ted': {
                $gt: 100
            }
        },
        sort: [{ 'nes.ted': 'asc' }]
    });
```

Because we have `$gt` in the query, the document with `nes: {ted: 100}` must not be part of the query result because 100 is not `$gt` 100.
The problem does not appear when querying with `$gt` over a top-level field.

### DIFF
--- a/tests/find/test-suite-1/test.ltgt.js
+++ b/tests/find/test-suite-1/test.ltgt.js
@@ -411,6 +411,69 @@ describe('test.ltgt.js', function () {
     });
   });
 
+  it('$gt on nested field', function(){
+    var db = context.db;
+    return db.createIndex({
+      name: 'nestedIndex',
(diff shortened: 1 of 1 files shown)


bgervan (approved):
This is a good catch. My solution is coming

bgervan:
Fixed in https://github.com/pouchdb/pouchdb/pull/8552 PR.

PR #8483, opened 2022-04-14
Docs: update PouchDB on page

Terreii:
Gratulation to the new release!

I was checking on the Page the PouchDB version, but did see that it is still PouchDB v6.2.0.

After some digging I found [where it was loaded](https://github.com/pouchdb/pouchdb/blob/140896cbdff064b5308cc3d55c481f4034855145/docs/_layouts/default.html#L229). It should be always the latest version, but isn't.
On the [JsDeliver Features Site](https://www.jsdelivr.com/features) the used method isn't listed anymore.
So here is my patch: Switch to the always up to date npm based URL.


AlbaHerrerias:
Thank you for noticing and for your contribution!

Issue #8451, opened 2022-02-15
[meta] Usage of async/await and ES6 syntax in general


AlbaHerrerias:
As I understand, we are interested in modernising the codebase syntax to use ES6 features. We started looking at places where we could refactor and introduce ES6 syntax such as `async`/`await` (to make the code flatter and more readable, reduce the pyramid of doom, deprecate some callbacks and so on). 

We made a [POC in this branch](https://github.com/neighbourhoodie/pouchdb/commit/cb7f6f60e0c7acbee813fe458e48d08aa9bc8b8c) where we refactor the http adapter:
- Use async/await whenever possible
- Deprecate some callbacks
- Use const/let instead of var, template strings
- [Upgrade the webpack’s version](https://github.com/neighbourhoodie/pouchdb/commit/01ef56197c1ccbd9cc90237ff9145ba662870530) in our test to be able to use async/await in the codebase

Please let us know what you think about this refactor. If positive we would like to create a PR with these changes.

We have also started working on the [`abstract-mapreduce` package](https://github.com/neighbourhoodie/pouchdb/commit/9d4102c83e93fea973b221f496cbaee675099bf6) as well.

Are there any other places that would benefit from a refactor that we should be aware of?

Thank you

PR #8489, opened 2022-04-18
Es6 refactor

AlbaHerrerias:
In an effort to modernize the codebase we started doing some refactors in the `http-adapter` and `abstract-mapreduce`. We also needed to modify the webpack testing to be able to use `async/await` in the codebase.

More context in https://github.com/pouchdb/pouchdb/issues/8451

### DIFF
--- a/packages/node_modules/pouchdb-abstract-mapreduce/src/index.js
+++ b/packages/node_modules/pouchdb-abstract-mapreduce/src/index.js
@@ -701,16 +701,15 @@ function createAbstractMapReduce(localDocName, mapper, reducer, ddocValidator) {
       return indexableKeysToKeyValues;
     }
 
-    return createTask().then(function () {
-      return processNextBatch();
-    }).then(function () {
-      return queue.finish();
-    }).then(function () {
+    try {
+      await createTask();
+      await processNextBatch();
+      await queue.finish();
       view.seq = currentSeq;
       view.sourceDB.activeTasks.remove(taskId);
-    }).catch(function (err) {
-      view.sourceDB.activeTasks.remove(taskId, err);
-    });
+    } catch (error) {
+      view.sourceDB.activeTasks.remove(taskId, error);      
+    }
   }
 
   function reduceView(view, results, options) {
--- a/packages/node_modules/pouchdb-abstract-mapreduce/src/index.js
+++ b/packages/node_modules/pouchdb-abstract-mapreduce/src/index.js
@@ -388,7 +388,7 @@ function createAbstractMapReduce(localDocName, mapper, reducer, ddocValidator) {
     const indexableKeysToKeyValues = docData[0];
     const changes = docData[1];
 
-    async function getMetaDoc() {
+    function getMetaDoc() {
       if (isGenOne(changes)) {
         // generation 1, so we can safely assume initial state
         // for performance reasons (avoids unnecessary GETs)
--- a/packages/node_modules/pouchdb-abstract-mapreduce/src/index.js
+++ b/packages/node_modules/pouchdb-abstract-mapreduce/src/index.js
@@ -113,7 +113,7 @@ function createAbstractMapReduce(localDocName, mapper, reducer, ddocValidator) {
   }
 
   function sortByKeyThenValue(x, y) {
-    var keyCompare = collate(x.key, y.key);
+    const keyCompare = collate(x.key, y.key);
     return keyCompare !== 0 ? keyCompare : collate(x.value, y.value);
   }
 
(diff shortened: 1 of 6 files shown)


TheSlimvReal:
@AlbaHerrerias I love what you are doing here!
 Are there some things that you would need some help with? I am very well familiar with the ES6 syntax and I would like to get started contributing to PouchDB. Therefore, I think I could help you out with this refactor if there are any parts that still need work.

garbados:
Just chiming in that I'm loving this so far. Thank you for this work!

janl:
@TheSlimvReal I think you are welcome to contribute other module rewrites-to-ES6 as separate PRs :)

packages/node_modules/pouchdb-abstract-mapreduce/src/index.js
```diff
@@ -378,14 +381,14 @@ function createAbstractMapReduce(localDocName, mapper, reducer, ddocValidator) {
...
-
-    function getMetaDoc() {
+  async function getDocsToPersist(docId, view, docIdsToChangesAndEmits) {
+    const metaDocId = '_local/doc_' + docId;
+    const defaultMetaDoc = {_id: metaDocId, keys: []};
+    const docData = docIdsToChangesAndEmits.get(docId);
+    const indexableKeysToKeyValues = docData[0];
+    const changes = docData[1];
+
+    async function getMetaDoc() {
```
jcoglan, line 391: This function doesn't strictly need to be marked `async`; it doesn't use the `await` keyword and it returns a `Promise` explicitly. It doesn't do any harm to leave this keyword here though.

packages/node_modules/pouchdb-abstract-mapreduce/src/index.js
```diff
@@ -445,57 +448,61 @@ function createAbstractMapReduce(localDocName, mapper, reducer, ddocValidator) {
...
 
-  function updateView(view, opts) {
+  async function updateView(view, opts) {
     return sequentialize(getQueue(view), function () {
       return updateViewInQueue(view, opts);
     })();
   }
 
-  function updateViewInQueue(view, opts) {
+  async function updateViewInQueue(view, opts) {
```
jcoglan, line 498: This function also currently doesn't need to be marked `async`, but we could change the code at the end of this function from `return createTask().then(function () {` onwards to use `await`. A lot of the internal structure of this function is very much designed around callbacks and can probably be simplified further but I wouldn't want to expand the scope of this PR right now.

packages/node_modules/pouchdb-adapter-http/src/index.js
```diff
@@ -680,65 +698,69 @@ function HttpPouch(opts, callback) {
...
+      callback(null, result.data);
+    } catch (error) {
+      callback(error);
+    }
   });
 
   // Update/create multiple documents given by req in the database
   // given by host.
-  api._bulkDocs = function (req, opts, callback) {
+  api._bulkDocs = adapterFun('_bulkDocs', async function (req, opts, callback) {
```
jcoglan, line 716: What's the reason that this method has been wrapped in `adapterFun()` but others like `_put()` have not?

packages/node_modules/pouchdb-abstract-mapreduce/src/index.js
AlbaHerrerias: This was removed here https://github.com/pouchdb/pouchdb/pull/8489/commits/cd47d29ff90506541b5bf030998dca635a91d720

janl (approved):

janl:
reminder for @AlbaHerrerias to add some more context about the comments made on the PR

packages/node_modules/pouchdb-abstract-mapreduce/src/index.js
AlbaHerrerias: This was not addressed as we didn't want to expand the scope of the PR

packages/node_modules/pouchdb-adapter-http/src/index.js
AlbaHerrerias: We found two problems here.
First, `adapterFun()` should be used on public methods only, so we removed it.
Second, when we attempted to refactor this into async-await, it failed. James pointed out it was [this `return` keyword](https://github.com/pouchdb/pouchdb/pull/8489/commits/3fb18acf9b5d12c888702f7e660890db448cea51#diff-304890c65c36e613ccffc19b7ae5376489e99355d5c84fb318e4ac2c41c75e2aL833), which shouldn’t be needed anymore as all the `_bulkDocs()` implementations in various adapters return nothing, and [this call](https://github.com/pouchdb/pouchdb/blob/701f96dc7c4ba2a598f46381b645a5c4c399b154/packages/node_modules/pouchdb-core/src/adapter.js#L833-L851) is the last statement in the function where it appears. 

This error was caused because the refactor makes `_bulkDocs` to return a promise. The [adapterFun](https://github.com/pouchdb/pouchdb/blob/2eca392bfae8c76bfb8d25b75414e5b47381a653/packages/node_modules/pouchdb-core/src/adapter.js#L769) calls the wrapped function [here](https://github.com/pouchdb/pouchdb/blob/2eca392bfae8c76bfb8d25b75414e5b47381a653/packages/node_modules/pouchdb-utils/src/adapterFun.js#L47) and then that result is wrapped by `toPromise()`. `toPromise()` calls the function and checks if the result is a promise [here](https://github.com/pouchdb/pouchdb/blob/2eca392bfae8c76bfb8d25b75414e5b47381a653/packages/node_modules/pouchdb-utils/src/toPromise.js#L26-L29), and if so, calls the wrapping promise's `fulfill()`, which causes the callback's `fulfill()` to be ignored.

Issue #8491, opened 2022-04-21
$or with $regex  broken after 7.2.2 -> 7.3.0


JanGoerss:
### Issue

BugFix [0e27e23f](https://github.com/pouchdb/pouchdb/commit/0e27e23f4437bce134d0780a543f08a3020fa4de) ([#8222](https://github.com/pouchdb/pouchdb/issues/8222)) - fix multiple $regex conditions on same field when using $and

has implement error with $or and $regex

Error => userValue.every is not a function

in 
  '$regex': function (doc, userValue, parsedField, docFieldValue) {
    return fieldExists(docFieldValue) &&
      typeof docFieldValue == "string" &&
      userValue.every(function (regexValue) {
        return regexMatch(docFieldValue, regexValue);
      });
  },

with Selector

{
	"$and": [
		{
			"$or": [
				{
					"field1": {
						"$regex": "/searchtext/i"
					}
				},
				{
					"field2": {
						"$regex": "/searchtext/i"
					}
				},
				{
					"field3": {
						"$regex": "/searchtext/i"
					}
				},
				{
					"field4": {
						"$regex": "/searchtext/i"
					}
				}
			]
		},
		{
			"fieldA": {
				"$exists": true
			},
			"fieldB": {
				"$exists": true
			}
		}
	]
}

### Info
- Environment: Browser
- Platform: all
- Adapter: indexeddb
- Server: CouchDB

### Reproduce
see up

filionf:
I have the fix ready for this. I need to write the unit tests and send the PR.

pete-hotchkiss:
@filionf looks like there was a build error on your PR. Is this something that can be sorted - seems simple enough as it's largely a linting error

filionf:
From what I can see, the latest build went through. What seems to be missing is a reviewer.
https://github.com/pouchdb/pouchdb/pull/8493

maxnowack:
This is still an issue. #8493 needs to be merged.

gwen1230:
When new version will be available ?

gwen1230:
When this fix will be available ?

garethbowen:
I'm not planning a release. I suggest reaching out on Slack.

pete-hotchkiss:
@garethbowen ok. I cant seem to find details of the slack channel. Is that something you can post at all ?

garethbowen:
[...]

PR #8493, opened 2022-04-25
(#8491) - Fix $regex and $ne within $or operator

filionf:


packages/node_modules/pouchdb-selector-core/src/utils.js
```diff
@@ -308,22 +307,41 @@ function massageSelector(input) {
...
 }
 
+//
+// The $ne and $regex values must be placed in an array because these operators can be used multiple times on the same field.
+// When $and is used, mergeAndedSelectors takes care of putting some of them into arrays, otherwise it's done here.
+//
+function normalizeArrayOperators(selector) {
+  var fields = Object.keys(selector);
+  for (var i = 0; i < fields.length; i++) {
+    var field = fields[i];
```
garethbowen, line 326: It'd be better to use a forEach loop here...

```suggestion
  Object.keys(selector).forEach(function (field) {
    ...
  });
```

garethbowen:
Original issue: https://github.com/pouchdb/pouchdb/issues/8491

garethbowen (changes requested):
LGTM. Just one small readability change and I'll sign off on it. Thanks!

packages/node_modules/pouchdb-selector-core/src/utils.js
filionf: Agreed! The forEach is more of my style, I must have been trying to replicate code style from above.

garethbowen (approved):
Thanks!

Issue #8494, opened 2022-04-25
New IndexedDB Adapter broken when used simultaneously on multiple browser pages


filionf:
### Issue
If database updates are done to the same IndexedDB database through different browser pages, one will eventually break due to the metadata `seq` not being properly updated.

### Info
- Environment: Browser
- Platform: Chrome
- Adapter: indexeddb
- Server: Not required.

### Reproduce

On page load, create the Pouch Database:
```js
var db = new PouchDB('test', { adapter: 'indexeddb' });
```

On demand (using a button), create documents in the database
```js
await db.post({ title: 'Ziggy Stardust' });
```

1. Open the served pages on 2 different tabs
2. By alternating between pages, click on the button.

The following error will appear in the console: `Database has a global failure DOMException: Unable to add key to index 'seq': at least one key does not satisfy the uniqueness requirements.`
When looking at the IndexedDB state, `metadata.seq` is not updated between updates from other pages.

dmitryshostak:
Hello 👋  When can we expect Francis's commit to be merged? 👀  thanks!

filionf:
Unstaling... Waiting for #8495 to be reviewed.

hpop:
This issue is quite severe because it appears to create a broken local database for users. 

Until the fix is merged:
-  Is there a workaround for this issue?
-  Can we recover from this error state, or is deleting the database the only option?

PR #8495, opened 2022-04-25
(#8494) - Refreshes seq before indexeddb bulkDocs

filionf:
There is not unit test yet. I would need some guidance on where to create those, and perhaps how to create a multi-page scenario as well.


broerse (approved):

broerse:
We need this fix to get PouchDB with this new adapter working in multiple browser tabs.

packages/node_modules/pouchdb-adapter-indexeddb/src/bulkDocs.js
```diff
@@ -314,6 +314,17 @@ export default function (api, req, opts, metadata, dbOpts, idbChanges, callback)
     };
   }
 
+  function refreshMetadataSeq() {
+    return new Promise(function (resolve) {
+      txn.objectStore(DOC_STORE).index('seq').openCursor(null, 'prev').onsuccess = function (evt) {
```
alxndrsn, line 319: Would this have the same effect as:
```js
metaStore.get(META_STORE).onsuccess = function (e) {
  metadata.seq = e.target.result.seq;
  ...
}
```
?

filionf: I will have to spend some time analyzing it. It's been too long for me to remember the technical details.

filionf: Yes it does. It seems to be a better solution so I've merged and updated the code to match this technique.

packages/node_modules/pouchdb-adapter-indexeddb/src/bulkDocs.js
```diff
@@ -306,6 +306,18 @@ export default function (api, req, opts, metadata, dbOpts, idbChanges, callback)
     };
   }
 
+  function refreshMetadataSeq() {
+    return new Promise(function (resolve) {
```
alxndrsn, line 310: Is there any chance of an error in this Promise, requiring a `reject()` call?

filionf:
> > There is not unit test yet. I would need some guidance on where to create those, and perhaps how to create a multi-page scenario as well.
> 
> Given the nature of the bug, a unit test would probably not be useful, even if it were possible.
> 
> A multi-tab/window test scenario would probably require deeper or different integration with playwright then any current test suites.
> 
> I would favour including tests before merging this PR.
> 
> I suspect there are also performance implications, so including perf test results for this branch vs the current `master` would also be interesting.

I've added one unit test, for what it's worth.

I've also ran the performance tests and found no major impacts. Results are within standard deviation to me. See below.

## Test Methology
1. Switch branch.
2. npm ci
3. npm run build
4. npm run dev
5. Open Chrome Browser at: http://127.0.0.1:8000/tests/performance/?adapters=indexeddb&couchHost=http%3A%2F%2F127.0.0.1%3A3000

## master
Testing PouchDB version 7.0.0-prerelease, using adapter(s): indexeddb

Starting suite: basics

Starting test: basic-inserts with 1 assertions and 1000 iterations... median: 0.2999999523162842 ms
Starting test: bulk-inserts with 1 assertions and 100 iterations... median: 6.200000047683716 ms
Starting test: bulk-inserts-large-docs with 1 assertions and 100 iterations... median: 9.25 ms
Starting test: bulk-inserts-massive-docs with 1 assertions and 10 iterations... median: 623.1499999761581 ms
Starting test: basic-updates with 1 assertions and 100 iterations... median: 14.99999988079071 ms
Starting test: basic-gets with 1 assertions and 1000 iterations... median: 0 ms
Starting test: all-docs-skip-limit with 1 assertions and 10 iterations... median: 7 ms
Starting test: all-docs-startkey-endkey with 1 assertions and 10 iterations... median: 1.2999999523162842 ms
Starting test: all-docs-keys with 1 assertions and 100 iterations... median: 0.40000009536743164 ms
Starting test: all-docs-include-docs with 1 assertions and 100 iterations... median: 0.9000000953674316 ms
Starting test: pull-replication-one-generation with 1 assertions and 1 iterations... median: 778 ms
Starting test: pull-replication-two-generation with 1 assertions and 1 iterations... median: 1358.2999999523163 ms
Starting suite: views

Starting test: temp-views with 1 assertions and 1 iterations... median: 763.1000001430511 ms
Starting test: build-secondary-index with 1 assertions and 1 iterations... median: 121.89999985694885 ms
Starting test: persisted-views with 1 assertions and 10 iterations... median: 2.299999952316284 ms
Starting test: persisted-views-stale-ok with 1 assertions and 10 iterations... median: 1.3000000715255737 ms
Starting suite: find

Starting test: create-index with 1 assertions and 1 iterations... median: 157.70000004768372 ms
Starting test: simple-find-query with 1 assertions and 5 iterations... median: 0.2999999523162842 ms
Starting test: simple-find-query-no-index with 1 assertions and 5 iterations... median: 63.200000047683716 ms
Starting test: complex-find-query with 1 assertions and 5 iterations... median: 76.80000019073486 ms
Starting test: complex-find-query-no-index with 1 assertions and 5 iterations... median: 62 ms
Starting test: multi-field-query with 1 assertions and 5 iterations... median: 74.79999995231628 ms
Starting suite: attachments

Starting test: basic-attachments with 1 assertions and 1000 iterations... median: 1.6999998092651367 ms
Starting test: many-attachments-base64 with 1 assertions and 100 iterations... median: 5.799999952316284 ms
Starting test: many-attachments-binary with 1 assertions and 100 iterations... median: 0.7999999523162842 ms

Tests Complete!


## refresh-seq-before-update
Testing PouchDB version 7.0.0-prerelease, using adapter(s): indexeddb

Starting suite: basics

Starting test: basic-inserts with 1 assertions and 1000 iterations... median: 0.2999999523162842 ms
Starting test: bulk-inserts with 1 assertions and 100 iterations... median: 6.099999904632568 ms
Starting test: bulk-inserts-large-docs with 1 assertions and 100 iterations... median: 9.299999952316284 ms
Starting test: bulk-inserts-massive-docs with 1 assertions and 10 iterations... median: 622 ms
Starting test: basic-updates with 1 assertions and 100 iterations... median: 17.75000011920929 ms
Starting test: basic-gets with 1 assertions and 1000 iterations... median: 0 ms
Starting test: all-docs-skip-limit with 1 assertions and 10 iterations... median: 6.850000023841858 ms
Starting test: all-docs-startkey-endkey with 1 assertions and 10 iterations... median: 1.2999999523162842 ms
Starting test: all-docs-keys with 1 assertions and 100 iterations... median: 0.39999985694885254 ms
Starting test: all-docs-include-docs with 1 assertions and 100 iterations... median: 1.0999999046325684 ms
Starting test: pull-replication-one-generation with 1 assertions and 1 iterations... median: 759.8999998569489 ms
Starting test: pull-replication-two-generation with 1 assertions and 1 iterations... median: 482.40000009536743 ms
Starting suite: views

Starting test: temp-views with 1 assertions and 1 iterations... median: 738.2999999523163 ms
Starting test: build-secondary-index with 1 assertions and 1 iterations... median: 122.09999990463257 ms
Starting test: persisted-views with 1 assertions and 10 iterations... median: 2 ms
Starting test: persisted-views-stale-ok with 1 assertions and 10 iterations... median: 1.1500000953674316 ms
Starting suite: find

Starting test: create-index with 1 assertions and 1 iterations... median: 138.10000014305115 ms
Starting test: simple-find-query with 1 assertions and 5 iterations... median: 0.2999999523162842 ms
Starting test: simple-find-query-no-index with 1 assertions and 5 iterations... median: 73.19999980926514 ms
Starting test: complex-find-query with 1 assertions and 5 iterations... median: 94.30000019073486 ms
Starting test: complex-find-query-no-index with 1 assertions and 5 iterations... median: 70.20000004768372 ms
Starting test: multi-field-query with 1 assertions and 5 iterations... median: 78.69999980926514 ms
Starting suite: attachments

Starting test: basic-attachments with 1 assertions and 1000 iterations... median: 1.7000000476837158 ms
Starting test: many-attachments-base64 with 1 assertions and 100 iterations... median: 5.900000095367432 ms
Starting test: many-attachments-binary with 1 assertions and 100 iterations... median: 0.7000000476837158 ms

Tests Complete!

filionf:
When I submit a PR to PouchDB, there are tests that fails randomly, unrelated to my PR. I sometimes push empty commits to re-trigger the flow but it feels like there should be a better way. Should I keep pushing empty commits until it passes?

SourceR85:
> When I submit a PR to PouchDB, there are tests that fails randomly, unrelated to my PR. I sometimes push empty commits to re-trigger the flow but it feels like there should be a better way. Should I keep pushing empty commits until it passes?

Hi, we know, there are some flaky test cases.
I've re-run the failing test, this will be quicker as if you trigger a complete execution via force-push,
and the potential for another test to break is eliminated.

jcoglan:
Hi folks 👋  Thank you for identifying this issue and all the work you've already done on it. We're doing some work on testing and fixing this now, starting with adding automated multi-tab tests here: [9212](https://github.com/apache/pouchdb/pull/9212)

We have confirmed the issue exists and we can reproduce it reliably with two `PouchDB` instances in different pages, not in the same page -- that's why a unit test running inside a single page will not detect the problem. The test we've added in the above PR does reliably fail on `indexeddb`.

The changes made here stop `put()` from throwing an error, but the doc does not seem to actually get written. The `update_seq` changes, but no actual doc is added to the database. The test we've written still fails, because even though `put()` returns `{ ok: true }`, `doc_count` is still lower than expected and instances still see stale `info()` output sometimes.

Any insight you can provide into why this might be the case, or suggestions for properly fixing it, would be greatly appreciated :)

jcoglan:
The other thing I'll add is that this problem seems entirely deterministic, not a race condition -- if you add docs one at time, alternating between two tabs while doing so, the problem always shows up. This should give us some confidence if we actually find a fix. It also means that if `put()` behaves correctly in this situation, then the problems with replication should also go away.

Currently, applications calling `put()` directly can catch the error and retry, but if you're doing replication you can't intervene in this manner. But hopefully if we fix `put()`, then replication will also be fixed.

Issue #993, opened 2013-11-03
Use allDocs to batch generation 1 docs in replication


jo:
from [Replication Algorithm by Jens Alfke](https://github.com/couchbaselabs/TouchDB-iOS/wiki/Replication-Algorithm):

> — From my limited testing, the performance bottleneck in the current algorithm seems to be in fetching the new revisions from the source. I think this is due to the overhead of handling many separate HTTP requests. It should be possible to speed up replication by introducing a new API call that fetches revisions in bulk. (The _all_docs call can fetch a list of revisions, but currently can’t be told to include revision histories.)
> 
> — A limited case of the above-mentioned bulk-get optimization is possible with the current API: revisions of generation 1 (revision ID starts with “1-”) can be fetched in bulk via _all_docs, because by definition they have no revision histories. Unfortunately _all_docs can’t include attachment bodies, so if it returns a document whose JSON indicates it has attachments, those will have to be fetched separately. Nonetheless, this optimization can help significantly, and is currently implemented in TouchDB.

This can be implemented using the same technique as `writeQueue`, if there is only one generation 1 revision.

PR #8499, opened 2022-05-02
chore: remove obsolete replication revision-one optimization

jo:
With #993, we introduced an optimization to retrieve documents via `_all_docs` if they had a revision of 1, since there was no `_bulk_get` API at that time. This optimization allowed us to retrieve many revision-one documents with a single request, instead of requesting each document individually. Documents with higher revisions still had to be fetched one by one.

Since then, `_bulk_get` has been implemented in CouchDB and is used by PouchDB's replicator, so this optimization is no longer necessary. On the contrary, it now leads to additional requests that slow down replication, although not that much.

This PR removes this optimization. The tests have also been adjusted.

### DIFF
--- a/packages/node_modules/pouchdb-replication/src/getDocs.js
+++ b/packages/node_modules/pouchdb-replication/src/getDocs.js
@@ -1,9 +1,5 @@
 import { clone, flatten, isRemote } from 'pouchdb-utils';
 
-function isGenOne(rev) {
-  return /^1-/.test(rev);
-}
-
 function fileHasChanged(localDoc, remoteDoc, filename) {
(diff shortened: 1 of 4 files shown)


janl:
@jo this is great, can you look at the CI test fails?

jo:
Oh I did and it's... mysterious. Have also restarted them multiple times, and its reproducible. I have no clue what it is yet. Could you also have a look?

jo:
Update on the test failures: https://github.com/pouchdb/pouchdb/pull/8507 fixes a race condition when cancelling replications. This also fixes the failing test here.

PR #8503, opened 2022-05-15
Align active-tasks entries with CouchDB's endpoint response.

SourceR85:
I refactored the ActiveTasks class so it matches more the `/_active_tasks` endpoint response of CouchDB.

For `pid` i chose `Symbol` to mimic the uniqueness of couchdb's response (there should no need for custom uuid-string generation).

### DIFF
--- a/bin/test-webpack.sh
+++ b/bin/test-webpack.sh
@@ -7,9 +7,7 @@
 #
 
 npm run build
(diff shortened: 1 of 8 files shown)


SourceR85:
Linter errors are fixed... I've missed to turn on eslint in my editor.
It's now ready to run/merge.

PR #8507, opened 2022-05-20
fix: replication cancelling race condition

jo:
This fixes a race condition that was observed in a presumed flaky test that was ultimately not flaky: If a replication is aborted at the exact time when a checkpoint was written at the end of a changes batch without changes, then the replication was not really aborted.


jo:
Yes, while I am generally in favor of more testing rather than less, I think we have sufficient test coverage here. The exception is also not new, it is already thrown everywhere else in the code at this point when a replication is aborted. Besides, this not-flaky test indirectly tests the behavior.
But I'll always happily review a pull that tests this.

janl (approved):

PR #8508, opened 2022-05-20
fix: active task total items calculation

jo:
Update sequences can be strings, e.g. `2342-g1AAAACheJzLYWBgYMpgTmEQTM4vTc5ISXIwNDLXMwBCwxyQVB4LkGRoAFL_gSArgzmJgYFlQi5QjN0wydgsMS0Jm1Y8BiYyJNUjTJoGNsnMMC0lLckCm54sAOFdKgw` (and are always strings in clustered CouchDB since v2) and therefore cannot be subtracted. This uses `parseInt(seq, 10)` to get an integer from the sequence number or sequence string. Note that `parseInt` will ignore the appendix (is that even specified anywhere?) and also works with integers.

In reality, this is not even accurate. Because eg changing the `_security` object in an Apache CouchDB database also changes the sequence, even though no document change has been written. Therefore, the sequence arithmetic is insufficient for counting document changes. But normally the active task is used for progress display purposes, so this might suffice as a rough approximation.

### DIFF
--- a/packages/node_modules/pouchdb-replication/src/replicate.js
+++ b/packages/node_modules/pouchdb-replication/src/replicate.js
@@ -144,7 +144,7 @@ function replicate(src, target, opts, returnValue, result) {
       }
 
       var completed = task.completed_items || 0;
-      var total_items = info.update_seq - initial_last_seq;
+      var total_items = parseInt(info.update_seq, 10) - parseInt(initial_last_seq, 10);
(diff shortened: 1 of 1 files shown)


jo:
yes very true.

janl:
yeah I think, we’ll need to go with “good enough” here

janl (approved):

PR #8509, opened 2022-05-21
Support IPv6 addresses as hostnames

aaronhuggins:
Implements a naive but fast pattern for matching an IPv6-style hostname, while preserving the original pattern for anything not matching IPv6.

This relates to a bug found by myself and @CanRau. For discussion of the bug, please see [our twitter thread](https://twitter.com/CanRau/status/1527808747837571072).

If you'd like me to add tests to specifically catch this issue and regressions in the future, please let me know and I'll add them.


AlbaHerrerias:
Hello! Would you be able to add some [tests](https://github.com/pouchdb/pouchdb/blob/master/tests/unit/test.parse-uri.js)? Thank you!

aaronhuggins:
@AlbaHerrerias I'll have time for this over the holiday before end of month. Does that fit with your overall plans?

AlbaHerrerias:
@aaronhuggins Sure! This PR won't make it to the release we're planning to do this week but it will for the next release, therefore I'm adding the `post-8.0.0` label if this is ok to you. Thanks!

janl:
we should probably rather switch to standard URI parsing APIs

PR #8513, opened 2022-05-25
fix: Race condition for pouchdb hanging indefinitely when multiple instances are used

PaulMest:
This is a self-contained repro for a problem discovered in PouchDB 7.3.0 where 2 databases using the `memory` adapter could impact each other and ultimately become non-responsive.

This was originally discovered in the RxDB test suite. Discussion here: https://github.com/pubkey/rxdb/pull/3807.

```
  134 passing (2s)
  3 pending
  1 failing

  1) test.memory-adapter.js Race condition initially discovered with PouchDB in-memory-adapter 7.3.0:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
      at listOnTimeout (node:internal/timers:559:17)
      at processTimers (node:internal/timers:502:7)
```

### DIFF
--- a/packages/node_modules/pouchdb-adapter-leveldb-core/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-leveldb-core/src/index.js
@@ -1176,7 +1176,7 @@ function LevelPouch(opts, callback) {
 
         var adapterName = functionName(leveldown);
         var adapterStore = dbStores.get(adapterName);
-        var keys = [...adapterStore.keys()].filter(k => k.includes("-mrview-"));
+        var keys = [...adapterStore.keys()].filter(k => k.includes(name)).filter(k => k.includes("-mrview-"));
         keys.forEach(key => {
           var eventEmitter = adapterStore.get(key);
           eventEmitter.removeAllListeners();
--- a/packages/node_modules/pouchdb-adapter-leveldb-core/src/index.js
+++ b/packages/node_modules/pouchdb-adapter-leveldb-core/src/index.js
@@ -4,6 +4,7 @@ import { obj as through } from 'through2';
 import getArguments from 'argsarray';
 import Deque from 'double-ended-queue';
 import bufferFrom from 'buffer-from'; // ponyfill for Node <6
+import PouchDB from 'pouchdb-core';
 import {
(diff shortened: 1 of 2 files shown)


PaulMest:
I've updated my PR:
1) Rebased this branch on top of the latest from `master`. 
2) Addressed ESLint warning
3) Pulled in fix from #8515 to pair with my repro from 6 months ago.

Hope this gets included into a future release! 🤞🏼

packages/node_modules/pouchdb-adapter-leveldb-core/src/index.js
```diff
@@ -1176,7 +1176,7 @@ function LevelPouch(opts, callback) {
 
         var adapterName = functionName(leveldown);
         var adapterStore = dbStores.get(adapterName);
-        var keys = [...adapterStore.keys()].filter(k => k.includes("-mrview-"));
+        var keys = [...adapterStore.keys()].filter(k => k.includes(name)).filter(k => k.includes("-mrview-"));
```
garethbowen, line 1179: I think there's a small issue here if you have two dbs with the same prefix, for example, "test-db" and "test-db-2". If you close "test-db" the same race condition will occur. I think we can make the matching string more restrictive to avoid these edge cases, eg:

```suggestion
        var viewNamePrefix = PouchDB.prefix + name + "-mrview-";
        var keys = [...adapterStore.keys()].filter(k => k.includes(viewNamePrefix));
```

You'll have to import PouchDB too, and I haven't actually run this code, but I think it'll work.

garethbowen (changes requested):
Thanks for the test, that's really helpful. I've added a suggestion.

packages/node_modules/pouchdb-adapter-leveldb-core/src/index.js
PaulMest: Thanks for looking this over! Ran this locally and it seemed to work. Updated the tests and used the code snippet you provided. Just pushed the commit. CI should complete in about 10 minutes to verify.

garethbowen (approved):
Thanks!

Issue #3062, opened 2014-11-28
Regression in leveldb.js according to Map/reduce tests


nolanlawson:
So we're lucky; apparently some issue not caught by our own tests is being caught by the map/reduce tests. I'm hoping this is related to https://github.com/pouchdb/pouchdb/issues/3026.

Only appears in leveldb currently. Can be triggered by current pouchdb master.

PR #8523, opened 2022-06-17
fix: store partial_filter_selector in ddoc

### LINKED ISSUE #8522: pouchdb-find createIndex doesn't write partial_filter_selector
### Issue
The `partial_filter_selector` field is not stored when pouchdb-find's `createIndex` is used on a local db.

### Info
- Environment: Node.js
- Platform: FF
- Adapter: idb/indexeddb/memory/leveldb/etc.
- Server: none

### Reproduce
```javascript
pouchdb.createIndex(index: {
    fields: ['hello'],
    partial_filter_selector: { type: 'x' }
  },
  ddoc: 'test-partial',
  name: 'type-x'
})
  .then(result => pouchdb.get('_design/test-partial'))
  .then(ddoc => {
    console.log(ddoc.views['type-x'].map.partial_filter_selector) // logs undefined
  })
```

Terreii:
This pull request fixes #8522.

### DIFF
--- a/docs/_includes/api/create_index.html
+++ b/docs/_includes/api/create_index.html
@@ -212,6 +212,17 @@ db.createIndex({
 
 If you want to index a subset of the documents in the database, you can use `partial_filter_selector`. This has the same syntax as the selector you'd pass to `find()`, and only documents matching the selector will be included in the index.
 
+{% include alert/start.html variant="warning"%}
+{% markdown %}
+
+**When using _pouchdb-adapter-indexeddb:_** On the `indexeddb`-adapter, indexes with
+a`partial_filter_selector` will fall back to using [map-reduce](#query_database),
+instead of the native indexes, negating all index performance gains from the
+`indexeddb`-adapter over the `idb`-adapter.
+
+{% endmarkdown %}
+{% include alert/end.html%}
+
 {% include code/start.html id="create_idx5" type="callback" %}
 {% highlight js %}
 db.createIndex({
--- a/packages/node_modules/pouchdb-adapter-indexeddb/README.md
+++ b/packages/node_modules/pouchdb-adapter-indexeddb/README.md
@@ -50,3 +50,11 @@ This will manifest in two ways:
 
 This is highly unlikely to be fixed in the future by PouchDB, as it would require improvements to IndexedDB. See:
  - https://github.com/pouchdb/pouchdb/blob/master/tests/find/test-suite-1/test.data-type-order.js#L66
+
+### Mango Queries with `partial_filter_selector`
+
+Mango indexes with a `partial_filter_selector` are using map-reduce views. Simually to the `idb`-adabter. All performance gains of native IndexedDB indexes will be lost.
+
+Also ordering is CouchDB combatible.
+
+This fallback is required because IndexedDB indexes only work on all db entries. And not on a subsection.
--- a/packages/node_modules/pouchdb-adapter-indexeddb/src/find.js
+++ b/packages/node_modules/pouchdb-adapter-indexeddb/src/find.js
@@ -4,6 +4,7 @@ import {
   DOC_STORE,
   idbError,
   rawIndexFields,
+  isPartialFilterView,
   naturalIndexName
 } from './util';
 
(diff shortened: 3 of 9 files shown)


Terreii:
Now only indexeddb adapter fails. I will look into it, on Sunday.

Terreii:
OK, indexeddb adapter implements its own [query interface](https://github.com/pouchdb/pouchdb/blob/9519c748c48cbac41451039765c758e5987bbc57/packages/node_modules/pouchdb-adapter-indexeddb/src/index.js#L129) (which is used by find).
It only uses the [fields to find an index](https://github.com/pouchdb/pouchdb/blob/9519c748c48cbac41451039765c758e5987bbc57/packages/node_modules/pouchdb-adapter-indexeddb/src/find.js#L139). Those indexes are [native indexes](https://github.com/pouchdb/pouchdb/blob/9519c748c48cbac41451039765c758e5987bbc57/packages/node_modules/pouchdb-adapter-indexeddb/src/setup.js#L43), and they only [don't support filtering](https://developer.mozilla.org/en-US/docs/Web/API/IDBObjectStore/createIndex).
So I might have to create a fields array with the `partial_filter_selector` included and combine it with the user-provided fields-array.

Terreii:
Sorry for the long silence.
I fixed it by adding to `_customFindAbstractMapper` a third argument: The standard/original `findAbstractMapper`. And then in the _indexeddb adapter_, I added a check if a view has a `partial_filter_selector`. And if so use the standard mapper (a fallback).

Terreii:
@pouchdb OK Pull request is refactored, squashed, and all tests pass!
I think it is ready to review/merge.

Terreii:
@AlbaHerrerias I think this pull request is ready for merging.

garbados (approved):
LGTM!

janl:
> In other words: find indexes with a partial_filter_selector are on indexeddb adapter db (like on all other adapters) just like map-reduce-views.

We should flag that in the docs then, that using partial filters on `indexeddb` will negate any of the perf benefits of native indexes (arguable one of the two prime reasons for `indexeddb`’s existence :)

Terreii:
I did add a warning.
Once this pr is merged, I will open an Issue, for finding a way to fix performance in the indexeddb-adapter. Which will be complicated, a reason why I didn't do it in a bug fix.

Terreii:
OK, I think __test.bulk_docs.js-http #3062 bulkDocs with staggered seqs__ is on pouchdb-server unstable.

docs/_includes/api/create_index.html
```diff
@@ -212,6 +212,17 @@
 
 If you want to index a subset of the documents in the database, you can use `partial_filter_selector`. This has the same syntax as the selector you'd pass to `find()`, and only documents matching the selector will be included in the index.
 
+{% include alert/start.html variant="warning"%}
+{% markdown %}
+
+**When using _pouchdb-adapter-indexeddb:_** On the `indexeddb`-adapter, indexes with
+a`partial_filter_selector` will fallback to using [map-reduce](#query_database),
```
janl, line 219:
```suggestion
a`partial_filter_selector` will fall back to using [map-reduce](#query_database),
```

docs/_includes/api/create_index.html
```diff
@@ -212,6 +212,17 @@
...
+instead of the native indexes. Negating all index performance gains from the
```
janl, line 220:
```suggestion
instead of the native indexes, negating all index performance gains from the
```

janl (changes requested):

janl:
>  for finding a way to fix performance in the indexeddb-adapter.

I’d be curious to know how you think this can be fixed :)

Terreii:
> > for finding a way to fix performance in the indexeddb-adapter.
> 
> I’d be curious to know how you think this can be fixed :)

At the moment only by trying to translate filters into indexes, and some into the selectors on query-time. With a fall back to map-reduce if that doesn't work.

But the more I think about it, the more I remember why I droped that path in this fix.

...or it may could act as an use case for a proposal to add partial indexes to IndexedDB indexes (after all SQLite [has them](https://www.sqlite.org/partialindex.html), too).

janl:
I like your thinking :D

PR #8531, opened 2022-07-14
Ignore deleted docs in sorted find

### LINKED ISSUE #8438: Adding a sort option to a query includes deleted documents in the indexeddb adapter
### Issue

When sending a query to PouchDB using this adapter, deleted documents show up in the results if any sort option is added to the query. E.g, using the following query:

```json
{
  "selector": {
    "rank": {
      "$gt": ""
    }
  },
  "sort": [
    {
      "rank": "asc"
    }
  ]
}
```
If I remove the `sort` property, deleted documents will not be included in the results. With the `sort` present, deleted documents appear. This is a change in behavior from _pouchdb-adapter-idb_. For the time being, I need to add my own selector to exclude deleted documents any time I want to sort values.

### Info
- Environment: Browser
- Platform: Firefox
- Adapter: indexeddb
- Server: None
- Version: 7.2.2

Terreii:
This pull request will fix #8438.

When using the indexeddb adapter, `find` will include deleted docs, if the sort field is set.

### DIFF
--- a/tests/find/test-suite-1/test.sorting.js
+++ b/tests/find/test-suite-1/test.sorting.js
@@ -261,4 +261,37 @@ describe('test.sorting.js', function () {
       });
     });
   });
+
+  it('#8438 exclude deleted docs', async function () {
+    const db = context.db;
+
+    await db.createIndex({
+      "index": {
+        "fields": ["foo"]
+      },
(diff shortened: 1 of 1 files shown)


janl (approved):

janl:
nice!

PR #8552, opened 2022-11-01
Fix nested field $gt filter behaviour

bgervan:
BUG: Querying with $gt over a nested field returns the wrong results, [PR with the tests](https://github.com/pouchdb/pouchdb/pull/8471)

The nested field did result undefined docs, so no filtering worked at all, when filterInclusiveStart called. (asc and gt in the same time)


AlbaHerrerias:
We're closing this PR in favor of https://github.com/pouchdb/pouchdb/pull/8573, thank you for your contribution!

PR #8568, opened 2022-12-10
eslint: use cache

alxndrsn:
For me this speeds up `npm run eslint` from 10+ seconds to ~1 second.


garethbowen (approved):
I have no idea why this isn't eslint's default value...

PR #8570, opened 2022-12-10
ci: use node 14

alxndrsn:
node 12 is EOL: https://nodejs.dev/en/about/releases/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ on:
     branches: [master]
 
 env:
-  NODE_VERSION: 12
+  NODE_VERSION: 14
   SELENIUM_HUB_HOST: hub
   TEST_HOST: localhost
 
@@ -39,7 +39,7 @@ jobs:
       fail-fast: false
       matrix:
         couchdb: ['2.3', '3.1']
-        node: [12]
+        node: [14]
         cmd:
           - npm test
           - TYPE=find PLUGINS=pouchdb-find ADAPTERS=http npm test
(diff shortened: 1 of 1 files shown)


alxndrsn:
A couple of tests needed to be re-run in my fork's Actions, and one failed on pouch's Actions.  Is this normal?

garethbowen:
@alxndrsn It's not uncommon.

garethbowen:
@alxndrsn Definitely an improvement. Is it worth jumping to Node 18? What about using the Actions matrix to test a range of Node versions?

alxndrsn:
> Is it worth jumping to Node 18?

Tempting.  I haven't tried it, but potentially more incidental updates would be required for dependencies.

> What about using the Actions matrix to test a range of Node versions?

I think that would be very helpful.  Currently I'm not sure the impact of the current line, which looks like it ignores the matrix version of node:

https://github.com/pouchdb/pouchdb/blob/55c2743e4b17a85b21f4f888ff8f02a2b31beaaa/.github/workflows/ci.yml#L17

garethbowen:
It seems like some steps use `${{ env.NODE_VERSION }}` and some use `${{ matrix.node }}`. I think it would be less confusing to have all use `matrix.node`. It would be good to have multiple node versions in the matrix but there's already 60 builds so I'm reluctant to add too many more. Node 14 is fine for now...

garethbowen (approved):

janl:
this looks great, thanks! We are about to do the 8.0.0 release and since this doesn’t affect the actual code, we’re holding this PR until after the release, hope that’s ok!

PR #8572, opened 2022-12-12
use shells specified in shebangs

alxndrsn:
Is there a subtle reason for the current syntax?  Maybe Windows compatibility?

It seems like a potential source of bugs where different shells are specified, e.g.:

https://github.com/pouchdb/pouchdb/blob/55c2743e4b17a85b21f4f888ff8f02a2b31beaaa/package.json#L22

vs.

https://github.com/pouchdb/pouchdb/blob/55c2743e4b17a85b21f4f888ff8f02a2b31beaaa/bin/release.sh#L1


technicool (approved):
This seems much more consistent, and passes all checks. Should help in cases like Mac OS, where the default shell is not `bash`.

garethbowen (approved):

PR #8576, opened 2022-12-14
Recover if publishing packages fails part-way through

jcoglan:
This is motivated by some pain that @AlbaHerrerias and I have run into publishing packages since npm started requiring two-factor authentication. Each time we hit an `npm publish` command, we have to type in a two-factor code. If this is rejected, or if anything else goes wrong during publish, the release script halts and the remaining packages need to be published by hand.

This lets us recover from this situation by writing all the package names to disk first. The `publish-packages.sh` script will try to published the items in the "todo" file until it hits an error, at which point it writes the remaining packages back to disk. The script can then be run again to publish the remaining packages, resuming from the one that failed last time.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,4 @@ tests/integration/utils-bundle.js
 package-lock.json
 yarn.lock
 /.eslintcache
+release-todo.txt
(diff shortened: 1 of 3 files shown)


janl (approved):
Nice!

PR #8577, opened 2022-12-15
browser-tests: remove saucelabs support

alxndrsn:


alxndrsn:
Test failure:

```
  1) test.bulk_docs.js-http #3062 bulkDocs with staggered seqs:

      AssertionError: expected [ Array(5) ] to deeply equal [ Array(5) ]
      + expected - actual

       [
         {
      -    "error": true
      +    "error": false
           "id": "doc-19"
      -    "rev": "undefined"
      +    "rev": "string"
         }
         {
      -    "error": true
      +    "error": false
           "id": "doc-17"
      -    "rev": "undefined"
      +    "rev": "string"
         }
         {
      -    "error": true
      +    "error": false
           "id": "doc-15"
      -    "rev": "undefined"
      +    "rev": "string"
         }
         {
           "error": false
           "id": "doc-13"
      
      at Assertion.assertEqual (node_modules/chai/lib/chai/core/assertions.js:485:19)
      at Assertion.ctx.<computed> (node_modules/chai/lib/chai/utils/addMethod.js:41:25)
      at doAsserterAsyncAndAddThen (node_modules/chai-as-promised/lib/chai-as-promised.js:296:33)
      at Assertion.<anonymous> (node_modules/chai-as-promised/lib/chai-as-promised.js:255:21)
      at Assertion.ctx.<computed> [as equal] (node_modules/chai/lib/chai/utils/overwriteMethod.js:49:33)
      at tests/integration/test.bulk_docs.js:572:24
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

Does this look related?

garethbowen (commented):
Can you document (commit message, issue, PR comment, whatever) why this is being removed? Was it not working?

alxndrsn:
@garethbowen here's some more context:

* Saucelabs tests were disabled in November 2020 (04d2a45872d11e57e25fd1bfdaa0011234ac1fb6).
* Removing support for running on Saucelabs was discussed in #pouchdb on CouchDB's Slack last year, although conversations there are discarded after 90 days.  There was support and no pushback that I remember.
* In August 2021, @daleharvey suggested it would be OK to remove support:

> I think I removed this relatively recently, this is needed for iphone and android mostly, I think its fine to take out for now while its not being used.
> -- _https://github.com/pouchdb/pouchdb/issues/8345#issuecomment-906239977_

* I'd like to move the test runner from Selenium to Playwright (PR at https://github.com/pouchdb/pouchdb/pull/8569), and trying to support Saucelabs during this transition will be difficult.

garethbowen (approved):
Thanks for the context. It feels like a shame but if they're not working anyway let's not let them hold us up.

PR #8579, opened 2022-12-15
perf tests: enable in-browser performance profiling

alxndrsn:
I've found this very helpful while working on improving performance in a specific adapter function.


garethbowen (approved):
Nice.

Issue #8581, opened 2022-12-27
The error `this.emit is not a function` has been raised while local storage has changed from other windows.


vrtmrz:
### Issue
Hi, team! I am always thankful to you for this grateful and excellent database!

I noticed that the error that I wrote in the title became shown on the browser. (It happens both with using `pouchdb-browser` and custom build from the core and some required plugins.). After I checked some behaviour and realise a concern, therefore I made a patch.
The reference to `this` looks has been captured in the function at [packages/node_modules/pouchdb-utils/src/changesHandler.js](https://github.com/pouchdb/pouchdb/blob/0bdb3423acf0cc4fae8747c1bd2cea0b1cf23339/packages/node_modules/pouchdb-utils/src/changesHandler.js#L13).

I made [the patch](https://github.com/vrtmrz/pouchdb/commit/181d3b6dd8adcec9940f60c5bec70540622d316a) and tested it, and looks like works fine.

Would you mind if I make make a pull request?

### Info
- Environment: Browser
- Platform: Chrome
- Adapter: idb
- Server: N/A

### Reproduce
1. Open two windows of pouchdb.com
2. Open devTools in both windows. perform localStorage.setItem("Hello", World")
3. The error appears in the opposite window.

steodor:
+1, we found this issue as well and came up with this exact same bugfix. We are currently monkey-patching the `index.js` and `index-browser.js` files in production so a pull request and official fix would be more than welcome. Thanks!

vrtmrz:
I am relieved to hear that I am not the only one.

I made a pull request, hope it will be reviewed and processed, and PouchDB becomes perfect again!

mbret:
Can confirm the issue, it broke my synchronization as well.

ishaiavrahami:
Hi I’m also getting this error. Is there a solution for this?

mbret:
> 

only downgrading at this point

vrtmrz:
In my opinion, as a temporary solution, we could utilize the temporarily modified package located in our local drive.
Like this: https://github.com/vrtmrz/obsidian-livesync/blob/2c36ec497c5f57093794bbf14914614297670c9d/package.json#L38. 

It might be better to be idempotent.

vrtmrz:
Thank you for merging the PR!

AlbaHerrerias:
The fix was released in 8.0.1. Thank you!

PR #8583, opened 2022-12-29
(#8581) -  Fix this of changesHandler

vrtmrz:
Fix `this` of changesHandler.

The test is also written but works only for browsers.

### DIFF
--- a/packages/node_modules/pouchdb-utils/src/changesHandler.js
+++ b/packages/node_modules/pouchdb-utils/src/changesHandler.js
@@ -10,7 +10,7 @@ export default class Changes extends EventEmitter {
     this._listeners = {};
     
     if (hasLocalStorage()) {
-      addEventListener("storage", function (e) {
+      addEventListener("storage", (e) => {
         this.emit(e.key);
       });
     }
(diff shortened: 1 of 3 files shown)


janl:
@vrtmrz thank you for this, but the test doesn’t actually pass. Could you have a look?

janl:
@vrtmrz thank you for this. Unfortunately the test fails in CI: https://github.com/pouchdb/pouchdb/actions/runs/4063298839/jobs/6995899992 — I’m not sure I understand the error message about number of assertions either. Is there something you could do?

vrtmrz:
Thank you for pointing it out! 
I am not sure about the value that assertion detected. However, we may also have timing problems.
I have rewritten the function to a callback style and will test it again.

vrtmrz:
I have checked this test again. The assertion error you pointed out was probably not a problem with this test itself.
Perhaps, `test.viewadapter.js` is not waiting for both of `onsuccess`, and this test detects it.
I have not tested these assumptions, but I think we need to address and call `done()` or wrap events to Promise and await.

The browser-side tests became can be passed. May I ask you to review it again?

janl (approved):

PR #8592, opened 2023-01-31
perf-tests: allow running specific suite(s)

alxndrsn:


tests/performance/index.js
```diff
@@ -11,19 +18,24 @@ function runTestSuites(PouchDB) {
...
   }
 
-  require('./perf.basics')(PouchDB, checkDone);
-  require('./perf.views')(PouchDB, checkDone);
-  require('./perf.find')(PouchDB, checkDone);
-  require('./perf.attachments')(PouchDB, checkDone);
+  if (suites.includes('basics')) { require('./perf.basics')(PouchDB, checkDone); }
+  if (suites.includes('views')) { require('./perf.views')(PouchDB, checkDone); }
+  if (suites.includes('find')) { require('./perf.find')(PouchDB, checkDone); }
+  if (suites.includes('attachments')) { require('./perf.attachments')(PouchDB, checkDone); }
```
garethbowen, line 38: Consider iterating over ALL_SUITES? For example, change ALL_SUITES to an obj with name key and path value?

```js
const ALL_SUITES = {
  basics: './perf.basics',
  views: './perf.views',
  find: './perf.find',
  attachments: './perf.attachments',
};

for (const [name, filepath] of Object.entries(ALL_SUITES)) {
  if (suites.includes(name)) {
    require(filepath)(PouchDB, checkDone);
  }
}
```

garethbowen (approved):
Approved - definite improvement as is.

Once optional suggestion inline.

tests/performance/index.js
alxndrsn: This looks much clearer, and was similar to my initial approach, but my memory is that dynamic `require(...)` calls were breaking bundling.  

I've tried recreating the issue at https://github.com/alxndrsn/pouchdb/compare/perf-tests-suite-filter...alxndrsn:pouchdb:perf-tests-suite-filter-indirect-requires?expand=1, but it seems to build ok: https://github.com/alxndrsn/pouchdb/actions/runs/4890727694

However, I'm nervous to make the change as I'm not confident that it will work in all situations :grimacing:

garethbowen: Ok. Feel free to merge as is!

PR #8601, opened 2023-02-15
Fix tests for local doc IDs

alxndrsn:
Closes https://github.com/pouchdb/pouchdb/issues/8602

This changes tests for local doc IDs inline with documented behaviour to test that the id starts with "_local/" instead of just "_local".


alxndrsn:
This introduces a couple of subtle changes:

1. `String.startsWith()` has not always been available: https://caniuse.com/?search=startsWith
2. if `id` is `null` or `undefined`, this code can now throw an Error where previously an ID would have been marked as non-local

lucidNTR (approved):
This is more readable and modern, the changes in behaviour seem also more consistent to me. @janl in a case like this where all tests pass, but potentially older browsers could become unsupported, is our oldest supported browser part of the integration tests and we can always modernize if tests are ok?

Issue #8598, opened 2023-02-09
Support Node v16


DougReeder:
### Issue
running `npm test` under Node v16 (MacOS Ventura v13.1 x86) generates these errors:

```
  1) test.all_docs.js-http Testing all docs:
     Error: Timeout of 50000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
      at listOnTimeout (node:internal/timers:559:17)
      at processTimers (node:internal/timers:502:7)

  2)  "after each" hook for "Testing all docs":
     ReferenceError: PromiseRejectionEvent is not defined
      at Context.<anonymous> (tests/integration/test.setup_global_hooks.js:18:33)
      at processImmediate (node:internal/timers:466:21)
```
in CI, with Node v16 added to the matrix, the errors are, for [couchdb-nodejs (3.1, 16, npm test)](https://github.com/DougReeder/pouchdb/actions/runs/4135077492/jobs/7147092144#logs)
```
  1) test.close.js-http test unref for coverage:
     Error: Timeout of 50000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
      at listOnTimeout (node:internal/timers:559:17)
      at processTimers (node:internal/timers:502:7)

  2)  "after each" hook for "test unref for coverage":
     ReferenceError: PromiseRejectionEvent is not defined
      at Context.<anonymous> (tests/integration/test.setup_global_hooks.js:18:33)
      at processImmediate (node:internal/timers:466:21)
```

### Info
- Environment: Node v16
- Platform: MacOS Ventura v13.1 x86, GitHub Actions ubuntu-latest

### Context
I am not presently using PouchDB, so I'm not asking anyone to work on this at this time. However, there are only 51 days until Node v14 is EOL, so I thought to add node v16 to the CI. I'm aware of the effort involved in maintaining open-source software, and I created this issue in the hope it will be useful in planning where effort should be focused.

PR #8616, opened 2023-03-30
Adds Node v16 to CI testing matrix

DougReeder:
Node v14 goes EOL at the end of April, so it's essential that everything work with Node v16.

Also updates GitHub checkout  action

The problems I saw in February are not manifesting now, possibly because of a more recent version of Node v16, or because of the updated checkout actions.


DougReeder:
Fixes #8598

alxndrsn (approved):
Looks good, although I'd be in favour of dropping node 14 completely.

PR #8620, opened 2023-04-06
Update changesHandler.js

lemanschik:
Switching over to namedExports but keeping the old behavior for incremental upgrade


SourceR85:
To much changes for one PR...
Please provide much smaller chunks.

PR #8623, opened 2023-04-12
ci: upgrade github actions to v3

alxndrsn:
Current versions are EOL as per https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

This should resolve the warning in current Actions jobs:

Screenshot_2023-04-12_18-51-06


garethbowen (approved):
LGTM!

PR #8624, opened 2023-04-13
Clarify the semantics of `since`

fischerman:
For reference: https://github.com/pouchdb/pouchdb/blob/30fd84f1ae88ca8687c2842bd131e67e7cf7c41c/packages/node_modules/pouchdb-adapter-leveldb-core/src/index.js#L1063


lucidNTR (approved):
good point, confirmed this is the current behaviour

PR #8629, opened 2023-05-01
Added null check for value in getFieldFromDoc,  closes #8027

lucidNTR:
Re-opened PR on latest master to close #8027


garethbowen:
@lucidNTR Any idea why these builds are failing?

PR #8634, opened 2023-05-05
Remove legacy script: bundle-browserify-test.sh

alxndrsn:
This script looks like it's no longer used:

* it was introduced in 2013 in cdc9af212a007ab5cc59c4b39b58fcc4cd4e9b04
* it looks like it's been superseded by doBrowserify() in bin/build-utils.sh
* it is never called in existing scripts
* there are no references to it being called previously, either in docs or other scripts


garethbowen (approved):
Thanks.

PR #8633, opened 2023-05-05
Remove legacy script: baldrick-test.sh

alxndrsn:
This script relies on `grunt`, which was removed in 2014: 70ec20be2018442bce72d6a57461778e06be7f2a

PR #8635, opened 2023-05-05
Remove legacy script: start_standalone_couch.sh

alxndrsn:
This looks like it was only ever used by `baldrick-test.sh`.

Depends on https://github.com/pouchdb/pouchdb/pull/8633


garethbowen (approved):
Sure.

PR #8636, opened 2023-05-07
update "integration test only" deps and move to devDependencies, closes #8600

lucidNTR:


alxndrsn (approved):
I wonder if `promise-polyfill` is still required - will check.

package.json
```diff
@@ -111,6 +110,7 @@
     "throw-max-listeners-error": "1.0.1",
     "ua-parser-js": "0.7.24",
     "watch-glob": "0.1.3",
+    "whatwg-fetch": "3.6.2",
```
alxndrsn, line 113: @lucidNTR can you fix the order here?  I think `wd` should come before `whatwg-fetch`, and `npm` changes the order if I install/remove other deps.

lucidNTR:
@alxndrsn the polyfill is only used at /tests/integration/index.html#L10 afaik, there seems to be an integration test testing pouch with ie and the polyfill.

SourceR85:
Close, due to promise-polyfill and whatwg-fetch removed by #8825 and #8643.

PR #8643, opened 2023-05-26
Remove promise polyfill dependency

alxndrsn:
This dependency was only used in integration tests, and is no longer required.

Closes #8600


lucidNTR (approved):
great lets get rid of thiss

PR #8646, opened 2023-05-26
Standardise shell shebangs

alxndrsn:
* always use `/bin/bash` instead of `/usr/bin/env bash` or `/bin/sh`
* always `set -e`


SourceR85 (approved):
LGTM

garethbowen (approved):

PR #8648, opened 2023-05-31
(#8594) - plugin override built-in methods

### LINKED ISSUE #8594: Cannot override bulkDocs anymore in custom Plugin with version 8.0.0
### Issue
According to the documentation: 

https://pouchdb.com/api.html#plugins

Example Plugin: Intercept Updates
A useful feature of plugins is to intercept updates before they are stored in PouchDB. In this way, a plugin might validate that the data is correct for the application, or even alter documents before they are committed to the database.

The best way to intercept all updates to a PouchDB database is to override the bulkDocs() method. All changes to PouchDB documents ultimately pass through the bulkDocs() method. For example, a call to put() will become a bulkDocs() call with a “batch” of one document.

Because PouchDB guarantees to plugin authors that all data changes ultimately happen via bulkDocs(), it is the ideal place for an application or plugin to intercept updates.

```
// Keep a reference to the "upstream" function.
var pouchBulkDocs = PouchDB.prototype.bulkDocs;
PouchDB.plugin({bulkDocs: validBulkDocs});

function validBulkDocs(body, options, callback) {
  if (typeof options == 'function') {
    callback = options
    options = {}
  }

  if (Array.isArray(body)) {
    var docs = body;
  } else {
    var docs = body.docs;
  }

  // All documents must have a .name field.
  for (var i = 0; i < docs.length; i++) {
    if (!docs[i].name) {
      var id = doc._id || '(no _id given)';
      return callback(new Error('Document is missing .name field: ' + id));
    }
  }

  // All documents check out. Pass them to PouchDB.
  return pouchBulkDocs.call(this, docs, options, callback);
}

```

The problem is that now with version 8.0.0 
```
var pouchBulkDocs = PouchDB.prototype.bulkDocs is undefined.
```
### Info
- Environment: Browser
- Platform: Chrome
- Adapter: idb
- Server: CouchDB

EthanDawes:
I added some test cases for the issue. I implemented the solution by storing the prototype of `PouchDB` at the start of the constructor, and then redeclaring the plugin methods from the stored prototype at the end of the constructor.

This resolves #8594

This is the continuation of #8631 which was accidently closed when I rebased. I rebased from a commit that didn't have CI errors. However, local `npm test` still throws errors when running from master. 🤷


garethbowen:
@funblaster22 The build is failing consistently with this change. Can you have another look?

louwers:
@garethbowen The problem is probably that the plugin override is globally registered.

Let me open a new PR with another test.

garethbowen:
@louwers That'd be my guess too. It may need an after/aftereach to unregister the plugin after the test.

PR #8649, opened 2023-06-04
fix build and test warnings

SourceR85:
## Intention:
I got annoyed by these warnings, so I patched 'em.

## Changes:
1. updates the rollup-plugin-node-resolve config fields "jsnext" and "main" to "mainFields".
2. replace package.json's field "jsnext:main" to modern standard "module".
3. replace deprecated Buffer() constructor call with equivalent Buffer.from()/Buffer.alloc().
4. replace module "buffer-from" with Buffer.from (obsolete, since node versions below 14 aren't supported by PouchDB).


garethbowen (approved):
Thanks!

PR #8651, opened 2023-06-07
test.viewadapter: restructure async handling

alxndrsn:
Reinstate async test handling using `done()` which was removed in 25db22fb0ff025b8d2c698da30c6c409066baa0c, and add more direct handling of errors.

-----

I think the current structure of tests mean that assertion failures in IndexedDB event handlers may not be detected.

This seems to introduce instability in the tests, see in places like https://github.com/pouchdb/pouchdb/pull/8569#issuecomment-1523449019.

## Example

```js
require('chai').should();

describe('async handling', () => {
  it('may not fail if error is thrown in async handler', async function() {
    setTimeout(() => {
      const x = 1;
      x.should.equal(2);
    }, 1000);
  }); 

  it('should fail if done() is used', function(done) {
    setTimeout(() => {
      const x = 1;
      x.should.equal(2);
    }, 1000);
  }); 
});
```

### Result

```
$ npx mocha mytest.js


  async handling
    ✓ may not fail if error is thrown in async handler
    1) should fail if done() is used


  1 passing (1s)
  1 failing

  1) async handling should fail if done() is used:

      Uncaught AssertionError: expected 1 to equal 2
      + expected - actual

      -1
      +2
      
      at Timeout._onTimeout (mytest.js:7:16)
      at listOnTimeout (internal/timers.js:557:17)
      at processTimers (internal/timers.js:500:7)
```

PR #8652, opened 2023-06-07
test.viewadapter: only run for expected adapters

alxndrsn:
These tests are written explicitly for `leveldb` and `idb`, but currently also run in CI with the primary adapter as `memory`.

The tests do not fail in CI because of a race condition, which is eliminated by https://github.com/pouchdb/pouchdb/pull/8651.

### DIFF
--- a/tests/integration/test.viewadapter.js
+++ b/tests/integration/test.viewadapter.js
@@ -55,6 +55,10 @@ viewAdapters.forEach(viewAdapter => {
         return;
       }
 
+      if (db.adapter !== 'leveldb' && db.adapter !== 'idb') {
+        return;
+      }
+
       await db.bulkDocs(docs);
       await db.query('index', { key: 'abc', include_docs: true });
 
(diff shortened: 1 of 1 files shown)


garethbowen:
I think checking `ADAPTERS` is right. It's confusing that it's plural, but if you check that then it should be skipped completely for everything except `leveldb` and `idb` test runs. You could also go one step further and also use this technique for the `viewAdapter` check but that's an optional extra.

Incidentally, why does this test not work with `indexeddb`?

alxndrsn:
> Incidentally, why does this test not work with `indexeddb`?

The test is asserting on the internal structure of the `IndexedDB` object stores, which is adapter-specific and not specified anywhere.

alxndrsn:
@garethbowen I'm keen to get this merged as even in the current form it's a significant improvement to the tests, and unblocks other work.  Would it be acceptable to keep the current implementation, and add a `TODO` for future improvement?  E.g.

```js
// TODO this currently marks tests as passing when they have not run.  Ideally
// the tests would not be created in the first place when they do not apply.
```

PR #8655, opened 2023-06-08
additional deprecation fix

SourceR85:
As the title imply:
I missed one Buffer() instantiation in #8649


garethbowen (approved):
Thanks.

SourceR85:
Three attempts to green mark...

We should consider upgrading to a more stable (more modern?) test runner.

PR #8659, opened 2023-06-13
Update npm

lemanschik:
This is the fundamental fix to read the npm based workflows and remove the hardcoded workaround that implemented a workspaces like environment to easy incremental adopt this new flows
i did create a additional package.json inside the package directory

so that npm will create packages/node_modules until all is ported to use the packages directly
after that works we can deprecate the package.json in favor of adding its content to the ../package.json.


janl:
@lemanschik apologies for the long silence here and thank you for spending a substantial amount of time on the project.

We fully agree that our source organisation and build system could do with modernisation, but we cannot accept a 20kLOC PR which touches many files without prior discussion about what the target situation is going to look like. We all need to be happy with a proposal before any work can actually happen.

We are putting this work on the back burner for now, but we might come back and pick out bits and pieces once we have a concrete proposal that we are happy with moving forward. For the moment, we are going to close this PR. We hope you understand.

lemanschik:
@janl thats not a problem i already have forked the needing parts it was still a good training to refactor by hand today with ai this takes minuts when you at any time want smaller pr's or anything you can ping me. I improved in the last years and can now easy write codeMods for all that and they can refactor package by package and so on.

Issue #5798, opened 2016-10-24
PouchDB source is not UTF-8 encoded when minified


slhck:
### Issue

Chrome does not load PouchDB's minified file because it is not UTF-8 encoded.

When I try to use PouchDB's `.min.js` version in my extension, I get:

```
Could not load file 'src/js/pouchdb-6.0.7.min.js' for content script. It isn't UTF-8 encoded.
```
### Info
- Environment: Chrome Extension
- Platform: Chrome Version 54.0.2840.71 (64-bit) on OS X 10.10
### Reproduce
- Download `pouchdb-6.0.7.min.js`
- Try to load it in Google Chrome extension `manifest.json` file under `content_scripts`

The non-minified version works without problems.

nolanlawson:
The reason this is happening is because Uglify converts e.g. `"\u0000"` to a literal `\0` (null terminator) character. I have often tried to find a way to disable this behavior or undo it but wasn't able to.

slhck:
I see, thanks.

Do you think it makes sense to file a feature request for Uglify, or a bug report for Chrom(e|ium)?

nolanlawson:
I could reasonably see it as either one. Alternatively we could write our own UTF8 encoder on top of our Uglify process but it's a lot of work just to support Chrome extensions.

slhck:
Hm. Perhaps a naïve suggestion, but I suppose a simple `s/\0/\u0000/g` doesn't do it?

daleharvey:
I would give it a try :) A bug upstream to either chrome or uglify sounds like a good idea though, this has come up more than once

nolanlawson:
There's both 0, 1, and 2 characters in teh source.

nolanlawson:
Closing because this is really an Uglify bug rather than a PouchDB bug, and I don't think we need to go to the effort of fixing it since users can fix it themselves e.g. by using their own minifier on `pouchdb.js`.

katlimruiz:
fyi, this is still happening when I tried to insert a pouchdb-8.0.1.min.js into a chrome extension.

Using VSCODE, I saved it with encoding Western ISO 8859-1, and it worked.

garethbowen:
[...]

PR #8660, opened 2023-06-14
#5798 set terser to minify with option ascii_only=true

katlimruiz:
Using this setting will fix the encoding problems occurring when using the minified version in a Chrome extension and other places.

### DIFF
--- a/bin/build-utils.js
+++ b/bin/build-utils.js
@@ -29,7 +29,7 @@ function writeFile(filename, contents) {
 }
 
 function doUglify(pkgName, code, prepend, fileOut) {
-  var miniCode = prepend + terser.minify(code).code;
+  var miniCode = prepend + terser.minify(code, { output: { ascii_only: true }}).code;
   return writeFile(addPath(pkgName, fileOut), miniCode);
 }
 


katlimruiz:
For ticket #5798

katlimruiz:
I confirm that:
- If I build with terser options as is in my local environment, the minified file makes my Chrome extension fail to load.
- If I build with the ascii_only option, my Chrome extension loads correctly.

katlimruiz:
cc @garethbowen

katlimruiz:
@garethbowen here I would need some help. A firefox test failed with this:
```
  1) test.retry.js-http-local target doesn't leak "destroyed" event:
     AssertionError: numListeners should never increase by +1/-1: expected 5 to be within 2..4
      at Hn.<anonymous> (test.retry.js:283:40)
      at d (http://127.0.0.1:8000/packages/node_modules/pouchdb/dist/pouchdb.min.js:7:1629)
      at s.emit (http://127.0.0.1:8000/packages/node_modules/pouchdb/dist/pouchdb.min.js:7:4615)
      at C (http://127.0.0.1:8000/packages/node_modules/pouchdb/dist/pouchdb.min.js:7:133666)
```

And also I found these files tests/integration/deps/pouchdb-8.0.1-postfixed.js that were in the repo with the failing Unicode characters

image

Should these files also be updated?

garethbowen:
@katlimruiz I've seen that test fail before so it's probably not because of your change. I'll re-run the test and see if it goes green.

Those postfixed files could be regenerated but they're only used in testing so let's just leave them be.

garethbowen:
@katlimruiz All green here. Are you happy for me to merge?

garethbowen (approved):
LGTM.

katlimruiz:
> @katlimruiz All green here. Are you happy for me to merge?

sure go ahead, I don't have the "Merge button" available.

PR #8663, opened 2023-06-17
replace the last occurrence of jsnext:main

SourceR85:
This patch removes the last "jsnext:main"-field within the PouchDB project.

(addition to #8649)


bin/update-package-json-for-publish.js
```diff
@@ -57,10 +57,10 @@ modules.forEach(function (mod) {
       './lib/index.es.js': './lib/index-browser.es.js',
     };
   }
-  // Update "jsnext:main" to point to `lib/` rather than `src/`.
+  // Update "module" to point to `lib/` rather than `src/`.
   // `src/` is only used for building, not publishing.
   // Also add "module" member: https://github.com/rollup/rollup/wiki/pkg.module
-  pkg['jsnext:main'] = pkg.module = './lib/index.es.js';
+  pkg['module'] = pkg.module = './lib/index.es.js';
```
garethbowen, line 63: This is setting the `pkg.module` twice. You can just remove the bit to the left of the "equals". However, I'm not sure we should merge this. Having both `jsnext:main` and `module` means pouch is more compatible with a range of module patterns, so this may break some projects that rely on `jsnext:main`. I don't see much downside leaving it in.

SourceR85: Thanks for the hint.

jsnext:main was originally designed to add import/export statement support.

I think also: nobody, who currently uses pouch, get in trouble if we let both in the package...
(Because the module property get priority)

On the other hand: jsnext:main didn't explicitly require support for es2015 environments (class/let/const/strict mode only).
So my thoughts are also against this flag and I want to let it R.I.P.

janl (approved):

PR #8668, opened 2023-06-20
Fix typo in attachments guide

PhearZero:
"Allow users to store an attachment" guide uses `type` instead of `content_type`


garethbowen (approved):
Thanks!

PR #8676, opened 2023-06-21
browser.worker test: dereference globals safely

alxndrsn:
The following line implies `window` may be `undefined`:
```js
var isNodeWebkit = typeof window !== 'undefined' && ...
```
If that's the case, then it doesn't make sense to be referring to `window.location`, `window.Worker`, `window.chrome`, or `navigator` without guards.


garethbowen (approved):
I think it could be simpler still, but this is definitely an improvement. Merging.

Issue #8680, opened 2023-06-23
tests/integration/browser.worker.js is never run


alxndrsn:
The tests in `tests/integration/browser.worker.js` are never run in Github Actions.

alxndrsn:
Also demonstrated by https://github.com/alxndrsn/pouchdb/actions/runs/5354111597.

Issue #8673, opened 2023-06-21
Selenium tests do not fail if an Error is thrown outside the scope of a test


alxndrsn:

alxndrsn:
Still a problem with `playwright`; PR at #8768

PR #8682, opened 2023-06-23
Explicitly skip browser.worker tests

alxndrsn:
As per https://github.com/pouchdb/pouchdb/issues/8680, these tests are not currently running.

### DIFF
--- a/tests/integration/browser.worker.js
+++ b/tests/integration/browser.worker.js
@@ -1,22 +1,18 @@
 'use strict';
 
+// Historically:
 // only running in Chrome and Firefox due to various bugs.
 // IE: https://connect.microsoft.com/IE/feedback/details/866495
 // Safari: doesn't have IndexedDB or WebSQL in a WW
 // NodeWebkit: not sure what the issue is
-
-var isNodeWebkit = typeof window !== 'undefined' &&
-  typeof process !== 'undefined';
-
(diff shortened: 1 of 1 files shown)


garethbowen (commented):
Do these tests fail if run?

Can we skip without the whitespace changes?

alxndrsn:
> Do these tests fail if run?

With playwright, IIRC, the tests pass except with `test-webpack.sh`.  This is due to `leveldown`/`pouchdb-adapter-leveldb` being included in the webpack bundle.  I think this should be investigated and corrected after changing the test browser to playwright.  This should also resolve https://github.com/pouchdb/pouchdb/issues/8683.

> Can we skip without the whitespace changes?

The whitespace changes are essentially to move `before()` and `after()` inside `describe()`, and have `mocha` fully in control of whether the tests are run or not.  This works around https://github.com/pouchdb/pouchdb/issues/8673, and will mean we see skipped tests in environments where the tests do not run, instead of the current behaviour (`browser.worker` does not appear in test reports if not run).

If the intention is explicitly to avoid indentation changes, then an extra level of `describe()` could probably be added, but I think the currently-proposed changes are the best structure for the tests going forward.

garethbowen (approved):
Ok, given that we're going from silently skipping to loudly skipping this is an incremental improvement.

Approved. Merge if you're ready!

Issue #8702, opened 2023-06-26
Intermittent test failure: test.sync.js-http-http 5007 sync 2 databases


alxndrsn:
https://github.com/pouchdb/pouchdb/actions/runs/5323768087/jobs/9643765670#step:5:4529
```
  1) test.sync.js-http-http 5007 sync 2 databases:
     Error: Timeout of 1200000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

alxndrsn:
Closing as ancient.  Open new issues if there are intermittent failures in the test suite.

Issue #8695, opened 2023-06-26
Frequent test failure: suite2 test.sync_events.js-local-http #4251 Should fire paused and active on sync


alxndrsn:
https://github.com/pouchdb/pouchdb/actions/runs/5339252762/jobs/9713921054#step:5:1429
```
   1) suite2 test.sync_events.js-local-http #4251 Should fire paused and active on sync:
     Error: Timeout of 1200000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

alxndrsn:
Closing as ancient.  Open new issues if there are intermittent failures in the test suite.

Issue #8691, opened 2023-06-26
Frequent test failure: suite2 test.attachments.js-local #3074 live changes()


alxndrsn:
https://github.com/pouchdb/pouchdb/actions/runs/5339179002/jobs/9677579675#step:5:1772
```
  1) suite2 test.attachments.js-local #3074 live changes():
     Error: Timeout of 1200000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

https://github.com/pouchdb/pouchdb/actions/runs/5328992647/jobs/9654247802#step:5:1780
```
   1) suite2 test.attachments.js-local #3074 live changes():
     Error: Timeout of 1200000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

https://github.com/pouchdb/pouchdb/actions/runs/5328992647/jobs/9654247802#step:5:1780
```
  1) suite2 test.attachments.js-local #3074 live changes():
     Error: Timeout of 1200000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

https://github.com/pouchdb/pouchdb/actions/runs/5328503152/jobs/9653604774#step:5:1854
```
  1) suite2 test.attachments.js-local #3074 live changes():
     Error: Timeout of 1200000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

https://github.com/pouchdb/pouchdb/actions/runs/5328992647/jobs/9654248604#step:5:1934
```
   1) suite2 test.attachments.js-local #3074 live changes():
     Error: Timeout of 1200000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

alxndrsn:
Closing as ancient.  Open new issues if there are intermittent failures in the test suite.

PR #8703, opened 2023-06-26
tests/integration: decrease default test timeout

alxndrsn:
browser: from 20 minutes to 20 seconds
node: from 50 seconds to 5 seconds

Various integration tests currently time out intermittently, e.g.:

* https://github.com/pouchdb/pouchdb/issues/8702
* https://github.com/pouchdb/pouchdb/issues/8695
* https://github.com/pouchdb/pouchdb/issues/8691

This timeout can safely be decreased by a significant amount without affecting build stability, but increasing build speeds.

### DIFF
--- a/bin/test-node.sh
+++ b/bin/test-node.sh
@@ -1,6 +1,6 @@
 #!/bin/bash -e
 
-: ${TIMEOUT:=50000}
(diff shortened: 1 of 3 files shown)


garethbowen (approved):
Awesome

Issue #8701, opened 2023-06-26
Intermittent test failure: test.close.js-http test unref for coverage


alxndrsn:
https://github.com/pouchdb/pouchdb/actions/runs/5316742697/jobs/9626593871#step:4:1035
```
  1) test.close.js-http test unref for coverage:
     Error: Timeout of 50000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
      at listOnTimeout (internal/timers.js:557:17)
      at processTimers (internal/timers.js:500:7)

  2)  "after each" hook for "test unref for coverage":
     ReferenceError: PromiseRejectionEvent is not defined
      at Context.<anonymous> (tests/integration/test.setup_global_hooks.js:18:33)
      at processImmediate (internal/timers.js:464:21)
```

https://github.com/pouchdb/pouchdb/actions/runs/5316742697/jobs/9626597098#step:4:1035
```
   1) test.close.js-http test unref for coverage:
     Error: Timeout of 50000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
      at listOnTimeout (internal/timers.js:557:17)
      at processTimers (internal/timers.js:500:7)

  2)  "after each" hook for "test unref for coverage":
     ReferenceError: PromiseRejectionEvent is not defined
      at Context.<anonymous> (tests/integration/test.setup_global_hooks.js:18:33)
      at processImmediate (internal/timers.js:464:21)
```

alxndrsn:
Ref https://github.com/pouchdb/pouchdb/issues/8687

alxndrsn:
The second failure, `ReferenceError: PromiseRejectionEvent is not defined`, should now be resolved by https://github.com/pouchdb/pouchdb/pull/8686.

alxndrsn:
Closing as ancient.  Open new issues if there are intermittent failures in the test suite.

PR #8704, opened 2023-06-26
unit tests: decrease default test timeout

alxndrsn:
From 50 seconds to 5 seconds.

Some unit tests currently time out intermittently, e.g.:

* https://github.com/pouchdb/pouchdb/issues/8701

This timeout can safely be decreased by a significant amount without affecting build stability, but increasing build speeds.


alxndrsn:
These are also integration tests.  This should be part of https://github.com/pouchdb/pouchdb/pull/8703.

Issue #8689, opened 2023-06-26
Frequent test failure: test.retry.js-http-local target doesn't leak "destroyed" event


alxndrsn:
https://github.com/pouchdb/pouchdb/actions/runs/5372105205/jobs/9745435702#step:5:1465
```
  1) test.retry.js-http-local target doesn't leak "destroyed" event:
     AssertionError: numListeners should never increase by +1/-1: expected 6 to be within 2..4
      at Hn.<anonymous> (test.retry.js:283:40)
      at d (http://127.0.0.1:8000/packages/node_modules/pouchdb/dist/pouchdb.min.js:7:1629)
      at s.emit (http://127.0.0.1:8000/packages/node_modules/pouchdb/dist/pouchdb.min.js:7:4615)
      at C (http://127.0.0.1:8000/packages/node_modules/pouchdb/dist/pouchdb.min.js:7:133666)
```

https://github.com/pouchdb/pouchdb/actions/runs/5372099418/jobs/9745289725#step:5:1458
```
  1) test.retry.js-http-local target doesn't leak "destroyed" event:
     AssertionError: numListeners should never increase by +1/-1: expected 6 to be within 2..4
      at Hn.<anonymous> (test.retry.js:283:40)
      at d (http://127.0.0.1:8000/packages/node_modules/pouchdb/dist/pouchdb.min.js:7:1629)
      at s.emit (http://127.0.0.1:8000/packages/node_modules/pouchdb/dist/pouchdb.min.js:7:4615)
      at C (http://127.0.0.1:8000/packages/node_modules/pouchdb/dist/pouchdb.min.js:7:133666)
```

https://github.com/pouchdb/pouchdb/actions/runs/5372105205/jobs/9745435702#step:6:1458
```
   1) test.retry.js-http-local target doesn't leak "destroyed" event:
     AssertionError: numListeners should never increase by +1/-1: expected 6 to be within 2..4
      at Hn.<anonymous> (test.retry.js:283:40)
      at d (http://127.0.0.1:8000/packages/node_modules/pouchdb/dist/pouchdb.min.js:7:1629)
      at s.emit (http://127.0.0.1:8000/packages/node_modules/pouchdb/dist/pouchdb.min.js:7:4615)
      at C (http://127.0.0.1:8000/packages/node_modules/pouchdb/dist/pouchdb.min.js:7:133666)
```

https://github.com/pouchdb/pouchdb/actions/runs/5372099418/jobs/9745289428#step:5:1461
[...]

PR #8710, opened 2023-07-24
test.retry: only check for leaked listeners at the end

alxndrsn:
Possible fix for https://github.com/pouchdb/pouchdb/issues/8689


alxndrsn:
Probably not the correct approach.

PR #8715, opened 2023-07-26
test.replication: remove "down-server"

alxndrsn:
The naming of the replication test which accessed the "down server" implies it's attempting to connect to a server which is down.

That test previously passed for any non-pouchdb server, including:

* a hostname which doesn't resolve
* a hostname which resolves but server does not respond
* a server which responds, but not as a couch server

This commit:

1. replaces the "down-server" with a localhost HTTP address which is very unlikely to respond, and
2. tightens up assertions against the thrown Error


tests/integration/test.replication.js
```diff
@@ -4248,6 +4248,8 @@ adapters.forEach(function (adapters) {
 // This test only needs to run for one configuration, and it slows stuff
 // down
```
alxndrsn, line 4249: But... it's currently run for quite a few configurations.  And behaves very differently with different configurations.  Perhaps this comment should be updated.

janl:
> The naming of the replication test which accessed the "down server" implies it's attempting to connect to a server which is down.

if you look at the down server script, this is especially a server that returns a 500 error. I’m not sure removing this tests helps, but I recall we got here because this used to rely on an external public server

alxndrsn:
> > The naming of the replication test which accessed the "down server" implies it's attempting to connect to a server which is down.
> 
> if you look at the down server script, this is especially a server that returns a 500 error. I’m not sure removing this tests helps, but I recall we got here because this used to rely on an external public server

If you look at the original commit that introduced this test (https://github.com/pouchdb/pouchdb/commit/8ec27a78e7bf1035f72bea24b0458717728e723c), it was actually a non-existent server rather than a server that returned 500.

Maybe both tests have value?

PR #8718, opened 2023-07-27
(#8717) run-test.sh: always shut down server on exit

### LINKED ISSUE #8717: pouchdb-express-router not shutting down after tests fail
# Reproduction

## Run

```shell
# deliberately make integration tests fail
sed -i.bak 's/should.equal(0)/should.equal(1)/' $(git grep -l 'should.equal(0)' -- tests/integration)

# check initial count of servers
echo -n 'pouchdb-express-router instances (before): '; ps aux | grep '[n]ode ./tests/misc/pouchdb-express-router.js' | wc -l

# run tests, doomed to fail
ADAPTERS=indexeddb CLIENT=chromium npm run test >/dev/null 2>&1

# check final count of servers
echo -n 'pouchdb-express-router instances  (after): '; ps aux | grep '[n]ode ./tests/misc/pouchdb-express-router.js' | wc -l
```

## Output

```
pouchdb-express-router instances (before): 13
pouchdb-express-router instances  (after): 14
```

# Notes

possibly made worse by: https://github.com/pouchdb/pouchdb/pull/8656/files#diff-89213709eb0feffc94d8151169c26f4eebedd714781deaf8d7df9348900f698dR61-R67

I think shutdown happens here: https://github.com/pouchdb/pouchdb/blob/f1a9b8969f26a950756cbf8b5ef5e11e88335a2c/bin/run-test.sh#L138-L143

...and it should be trapped rather than only happening if the script gets that far

alxndrsn:
Closes #8717

### DIFF
--- a/bin/run-test.sh
+++ b/bin/run-test.sh
@@ -1,5 +1,12 @@
 #!/bin/bash -e
 
+cleanup() {
+  if [[ ! -z $SERVER_PID ]]; then
+    kill $SERVER_PID
+  fi
+}
+trap cleanup EXIT
+
 # Run tests against a local setup of pouchdb-express-router
 # by default unless COUCH_HOST is specified.
 [ -z "$COUCH_HOST" -a -z "$SERVER"  ] && SERVER="pouchdb-express-router"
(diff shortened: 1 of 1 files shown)


garethbowen (approved):
Makes sense...

Issue #8706, opened 2023-07-17
PouchDB.get(): TypeError when optional argument 'options' is null


regnete:
When the API-method 'get' is called  without a callback an unexpected error is thrown:
`TypeError: Cannot read properties of null (reading 'open_revs')`

Environment:
 - Browser: Chrome
 - PouchDB: https://github.com/pouchdb/pouchdb/releases/download/8.0.1/pouchdb-8.0.1.js

Reproducer:

```js
var db = new PouchDB({name: "test"});
var id = "12345678";
db.put({_id: id, content: "test"}).then(function(){

  db.get(id, function(error, result){
    if(error)
      console.error("cannot get with callback", error);
    else
      console.log("get with callback", result)
  });

  db.get(id, null) .then(
  	function(result){
    	  console.log("get with promise", result)
       }, 
       function(error){
  	   console.error("cannot get with promise", error);
	});
});

```

The following line of code should be inserted at https://github.com/pouchdb/pouchdb/blob/980970331b232969824df71df7bf0bf8cf19053f/packages/node_modules/pouchdb-core/src/adapter.js#L496 to avoid the error:

```js
opts = opts || {};
```

regnete:
Thank you for the fast response on this issue. A PR is a large effort for a single line of code. Would be great if someone who modifies the code anyway could apply this little change.

garethbowen:
A PR is the only way code is changed by anyone, even someone who modifies the code anyway.

regnete:
OK. May take a while til I find the time ...

AmalChandru:
@garethbowen I was just going through the project and came across this conversation. Submitted a PR with the suggested modifications. Please let me know if any further details are needed for the commit

garethbowen:
Merged. Thanks @AmalChandru

PR #8720, opened 2023-08-07
(#8706) - Handle null options in 'get' method

AmalChandru:
When calling the `get` method without a callback and passing `null` as the options parameter, an unexpected error is thrown due to an unhandled case. This fix ensures that the options parameter is properly initialized when it is null, preventing the error from occurring.


garethbowen (approved):
Thanks!

PR #8736, opened 2023-08-25
Remove Promise polyfill mention from docs

louwers:
It is no longer true that Promises are not supported by some browsers.

Also it seems no polyfill is used in PouchDB itself anymore.

Probably this entire guide can be removed since Promises are basic JavasScript knowledge by now.

### DIFF
--- a/docs/_guides/async-code.md
+++ b/docs/_guides/async-code.md
@@ -206,23 +206,6 @@ db.get('charlie').then(function (charlie) {
 });
 ```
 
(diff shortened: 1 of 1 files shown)


alxndrsn (approved):

alxndrsn:
> Probably this entire guide can be removed since Promises are basic JavasScript knowledge by now.

:+1: for the sentiment.  I haven't read the guide, so I don't know if it has useful pouchdb-specific content.

PR #8738, opened 2023-08-28
Fix problematic test missing done()

louwers:
Test is failing in #8648 but the test itself seems problematic.

cc @garethbowen


garethbowen:
I don't think this test is the problem. The test has been unchanged for 7 years and it hasn't been failing until now. It's a bit weird through because it's passing an error handling callback to db.put, but returning as though it's a promise. I'd really like to know if the test hasn't been running properly all this time, and if so, what about your PR made it start complaining now.

PR #8739, opened 2023-08-28
(#8594) - fix plugin override built-in methods - Take 2

louwers:
#8648 with better test

cc @funblaster22 

Now the overridden `bulkDocs()` should not mess with the other tests.


louwers:
Alright, I tried. :/

I have no idea where / how `bulkDocs()` is even defined.

PR #8741, opened 2023-09-02
 test.retry: make leaked-listener test more robust

alxndrsn:
Possible fix for https://github.com/pouchdb/pouchdb/issues/8689

Alternative to https://github.com/pouchdb/pouchdb/pull/8710


garethbowen (commented):
I'm worried this is an actual bug/leak and not just a test flake, otherwise it'd be strange that it only affects a single adapter.

alxndrsn:
> I'm worried this is an actual bug/leak and not just a test flake, otherwise it'd be strange that it only affects a single adapter.

That's definitely possibly.  I think a likely alternative is that it's a race condition.

PR #8748, opened 2023-09-08
Fixing some non-function links in documentation + some http -> https

eklem:
Found out that the script-tag example for using pouchdb through jsdelivr wasn't fully correct, so fixed what I found and also changed some http:// to https://

There are a lot of http but saw that some Python-pages wasn't set up with https, so didn't go through the whole library, and didn't want to break localhost-stuff.

I don't get the `npm run build-site` to work, so only find-and-replace. Let me know if you want these changes in another form.

### DIFF
--- a/docs/_layouts/default.html
+++ b/docs/_layouts/default.html
@@ -223,11 +223,11 @@
       </div>
 
     </footer>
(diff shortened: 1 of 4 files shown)


garethbowen (approved):
Thanks!

eklem:
Nice it was of some help.

PR #8750, opened 2023-09-11
test.migration: simplify scenario/constructor mapping

alxndrsn:
Merge scenarios & constructors lists.

This should simplify adding future PouchDB versions to the migration tests as no longer need to separately declare:

* eslint globals exclusions
* cleanup via `delete window`
* extra mapping between scenario and constructor


alxndrsn:
Migration tests are not currently being run on this branch.

lucidNTR (approved):
great

PR #8751, opened 2023-09-11
tests/integration: use asyncLoadScript()

alxndrsn:
Replace duplicated script-loading code with existing `commonUtils.asyncLoadScript()` function.

Also removes a timeout-based implementation in `tests/integration/test.attachments.js`.

### DIFF
--- a/tests/common-utils.js
+++ b/tests/common-utils.js
@@ -120,7 +120,7 @@ commonUtils.loadPouchDBForBrowser = function (plugins) {
 
 // Thanks to http://engineeredweb.com/blog/simple-async-javascript-loader/
 commonUtils.asyncLoadScript = function (url) {
-  return new commonUtils.Promise(function (resolve) {
+  return new commonUtils.Promise(function (resolve, reject) {
     // Create a new script and setup the basics.
     var script = document.createElement("script");
     var firstScript = document.getElementsByTagName('script')[0];
@@ -128,6 +128,7 @@ commonUtils.asyncLoadScript = function (url) {
     script.async = true;
     script.src = url;
 
+    script.onerror = reject;
     script.onload = function () {
       resolve();
 
--- a/tests/common-utils.js
+++ b/tests/common-utils.js
@@ -123,7 +123,6 @@ commonUtils.asyncLoadScript = function (url) {
   return new commonUtils.Promise(function (resolve, reject) {
     // Create a new script and setup the basics.
     var script = document.createElement("script");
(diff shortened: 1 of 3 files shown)


tests/integration/browser.migration.js
```diff
@@ -46,13 +46,7 @@ describe('migration', function () {
       if (!match) {
         return testUtils.Promise.resolve();
       }
-      return new testUtils.Promise(function (resolve, reject) {
-        var script = document.createElement('script');
-        script.onload = resolve;
-        script.onerror = reject;
```
lucidNTR, line 52: the asyncLoadScript does not have the error handler, would you mind adding it there with the promise rejection as here?

lucidNTR: in addition, i think asyncLoadScript will fail if there is no existing script tag due to prepending document.getElementsByTagName('script')[0], if we replace existing code we might also add the fallback behaviour there to fall back to append child to body if there is no existing script tag.

tests/integration/test.attachments.js
```diff
@@ -3113,15 +3113,9 @@ adapters.forEach(function (adapter) {
             var bigimage = require('./deps/bigimage.js');
             cb(null, bigimage);
           } else { // browser
-            var script = document.createElement('script');
-            script.src = 'deps/bigimage.js';
-            document.body.appendChild(script);
-            var timeout = setInterval(function () {
```
lucidNTR, line 3119: great this code will be gone

lucidNTR (changes requested):

tests/integration/browser.migration.js
alxndrsn: Good catch - thanks!  I'll update.

alxndrsn: > we might also add the fallback behaviour there to fall back to append child to body if there is no existing script tag

I've updated to just append all scripts to `document.body`.  It seems to work 🫣

tests/common-utils.js
```diff
@@ -141,9 +141,7 @@ commonUtils.asyncLoadScript = function (url) {
       }
     };
 
-    // Attach the script tag to the page (before the first script) so the
-    //magic can happen.
-    firstScript.parentNode.insertBefore(script, firstScript);
+    document.body.append(script);
```
lucidNTR, line 144: good catch, for script async = true the order does not matter as per spec, and even if , the order would mostly depend on call order of asyncLoadScript, not tag order...

lucidNTR (approved):

PR #8752, opened 2023-09-11
test.migration: add test for indexeddb adapter

alxndrsn:
Test version upgrades with the `indexeddb` adapter.

Might be helpful to ignore whitespace when reviewing: https://github.com/pouchdb/pouchdb/pull/8752/files/?w=1


tests/integration/test.attachments.js
```diff
@@ -3113,15 +3113,9 @@ adapters.forEach(function (adapter) {
...
-            document.body.appendChild(script);
-            var timeout = setInterval(function () {
-              if (window.bigimage) {
-                clearInterval(timeout);
-                cb(null, window.bigimage);
-              }
-            }, 500);
+            testUtils.asyncLoadScript('deps/bigimage.js')
+                .then(() => cb(null, window.bigimage))
+                .catch(err => cb(err));
```
alxndrsn, line 3118: Already submitted as separate PR.

tests/common-utils.js
```diff
@@ -140,6 +140,7 @@ commonUtils.asyncLoadScript = function (url) {
         script.onload();
       }
     };
+    script.onerror = reject;
```
alxndrsn, line 143: Not directly relevant to this PR.  This needs testing separately, and including as its own PR if it works.

packages/node_modules/pouchdb-adapter-indexeddb/src/setup.js
```diff
@@ -127,6 +132,32 @@ function openDatabase(openDatabases, api, opts, resolve, reject) {
...
+        doc.data.migrated = true;
+        Object.values(doc.revs).forEach(rev => {
+          rev.data.migrated = true;
+        });
+
+        cursor.update(doc).onsuccess = () => {
+          cursor.continue();
+        };
+      };
+    }
```
alxndrsn, line 160: This all needs to be rolled back.

tests/integration/browser.migration.js
```diff
@@ -1,27 +1,34 @@
...
+  const scenarios = [
+    { scenario: 'PouchDB v1.1.0', constructorName: 'PouchDBVersion110'} ,
+    { scenario: 'PouchDB v2.0.0', constructorName: 'PouchDBVersion200'} ,
+    { scenario: 'PouchDB v2.2.0', constructorName: 'PouchDBVersion220'} ,
+    { scenario: 'PouchDB v3.0.6', constructorName: 'PouchDBVersion306'} ,
+    { scenario: 'PouchDB v3.2.0', constructorName: 'PouchDBVersion320'} ,
+    { scenario: 'PouchDB v3.6.0', constructorName: 'PouchDBVersion360'} ,
+    { scenario: 'PouchDB v7.3.1', constructorName: 'PouchDBVersion731'} ,
+    { scenario: 'PouchDB v8.0.1', constructorName: 'PouchDBVersion801'} ,
+    { scenario: 'websql',         constructorName: 'PouchDB'} ,
```
alxndrsn, line 31: Already submitted in separate PR.

tests/integration/browser.migration.js
```diff
@@ -631,523 +741,508 @@ describe('migration', function () {
...
+              .then(()    => oldPouch.get('norm-1'))
+              .then((doc) => oldPouch.putAttachment('norm-1',       'att.txt', doc._rev,  attFor('norm-1'), 'text/plain'))
+
+              .then(()    => oldPouch.putAttachment('norm-2',       'att.txt', undefined, attFor('norm-2'), 'text/plain'))
+
+              .then(() => oldPouch.close())
+
+              .then(() => {
+                newPouch = new dbs.second.pouch(dbs.second.local, { auto_compaction: false, adapter: 'indexeddb' });
+              })
```
garethbowen, line 1197: I can't help but think this sort of setup should be common to all migration tests but this implementation looks quite different from the "idb" one. Ideally as it's an integration test it would be more black box style testing where the "given" and "then" is the same, it's just the parameters for the "when" (adapter, pouch version, etc) that change. Is there a clean way to pull this out so we're testing the same things between adapters?

garethbowen (changes requested):
Nice one! One suggestion inline but it may be too much work for this PR.

tests/integration/browser.migration.js
alxndrsn: > Is there a clean way to pull this out so we're testing the same things between adapters?

How about https://github.com/pouchdb/pouchdb/pull/8766?

alxndrsn:
Now #8766 has been merged, this will need re-working to just introduce a new test around local docs & attachments.

alxndrsn:
Superseded by https://github.com/pouchdb/pouchdb/pull/8766

PR #8767, opened 2023-10-05
tests/browser.migration: compare version parts as numbers not strings

alxndrsn:
### Current `master` implementation:

```js
> versionGte('PouchDB v9.0.0', '8.0.0')
true
> versionGte('PouchDB v10.0.0', '8.0.0')
false
```

### This PR implementation:

```js
> versionGte('PouchDB v9.0.0', '8.0.0')
true
> versionGte('PouchDB v10.0.0', '8.0.0')
true
```

### DIFF
--- a/tests/integration/browser.migration.js
+++ b/tests/integration/browser.migration.js
@@ -1170,9 +1170,9 @@ describe('migration', function () {
 function versionGte(scenario, minimumRequired) {
   const match = scenario.match(/^PouchDB v([.\d]+)$/);
   if (!match) { return false; }
-  const actual = match[1].split('.');
+  const actual = match[1].split('.').map(Number);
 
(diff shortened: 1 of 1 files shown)


garethbowen (approved):
Nice catch!

Might be worth using [semver](https://www.npmjs.com/package/semver), but I'll leave that up to you.

alxndrsn:
> Might be worth using [semver](https://www.npmjs.com/package/semver), but I'll leave that up to you.

Possibly, but we can't just `require()` deps in the integration tests & I'd rather not add extra dependencies as they are currently minimal: https://github.com/pouchdb/pouchdb/blob/2dfadd174b58f6fe439a97e62cf3dacea010198d/tests/integration/index.html#L10-L13

PR #8768, opened 2023-10-05
test-browser: catch-and-fail on unhandled error

alxndrsn:
This highlights a bug introduced in e6b8c9371a6e5682e35b368b2ababb4a288a17a5

Closes #8673

PR #8769, opened 2023-10-05
test.issue3179: fix: allow test to run in browsers

alxndrsn:
Introduced in e6b8c9371a6e5682e35b368b2ababb4a288a17a5, use of `require()` silently fails tests in browser environments.  Also `should` is already declared as a global.


alxndrsn:
Merging this should allow https://github.com/pouchdb/pouchdb/pull/8768 to pass.

garethbowen:
> Merging this *should* allow https://github.com/pouchdb/pouchdb/pull/8768 to pass.

I see what you did there.

PR #8775, opened 2023-10-09
test.changes: fix case ddoc without filters+key

SourceR85:
This catches an unstable behaviour in case <code>Changes with &grave;filters&grave; key not present in ddoc</code>
caused by a [TypeError in ci](https://github.com/pouchdb/pouchdb/actions/runs/6460640280/job/17538820092?pr=8664#step:6:1324)
```
TypeError: Cannot read properties of undefined (reading 'should')
    at Changes$1.<anonymous> (http://127.0.0.1:8000/tests/integration/test.changes.js:358:22)
    2) "after each" hook for "Changes with `filters` key not present in ddoc"
    at emitOne (http://127.0.0.1:8000/packages/node_modules/pouchdb/dist/pouchdb.js:115:13)

    at Changes$1.emit (http://127.0.0.1:8000/packages/node_modules/pouchdb/dist/pouchdb.js:193:7)

    at http://127.0.0.1:8000/packages/node_modules/pouchdb/dist/pouchdb.js:4347:16

    at http://127.0.0.1:8000/packages/node_modules/pouchdb/dist/pouchdb.js:2897:11
```

Root of this error seems to be an unpolluted (chai's .should getter/setter on Object.prototype) `err` argument.


SourceR85:
P.S: I've applied the same logic to test.changes.js-`Changes with filter not present in ddoc`

tests/integration/test.changes.js
```diff
@@ -343,24 +360,43 @@ adapters.forEach(function (adapter) {
...
-          done();
-        });
-      });
+      const caughtErrors = [];
+      const db = new PouchDB(dbs.name);
+
+      testUtils.promisify(testUtils.writeDocs)(db, docs)
+        .then(async () => {
+          try {
+            await db.changes({
```
alxndrsn, line 374: What does the `await` do here?  Is it equivalent to using the `complete` event?

alxndrsn:
@SourceR85 if the problem is just a missing `should` rather than a missing `err.status` at https://github.com/pouchdb/pouchdb/blob/eafc0f148ea4cc4979f455c0dba198e5abaf8b9c/tests/integration/test.changes.js#L358, are the extra layers of error catching required?  Would replacing `err.status.should...` with `chai.assertEqual(err.status, ...)` be sufficient?

tests/integration/test.changes.js
SourceR85: Indeed, if i run the test without the `await` keyword, the test didn't pass inside the function and therefore a warning (uncaught exception) was thrown.
As a bonus: the `await` is also a guard to ensure that other exceptions (if any) get caught, so we know when an unhandled side effect pollutes this test.

SourceR85:
> are the extra layers of error catching required?
For now it's not required, but without catching, the runtime warning "unhandled Exception" comes up and the advice to handle it.

> Would replacing `err.status.should...` with `chai.assertEqual(err.status, ...)` be sufficient?
That's what i did :wink: (chai isn't directly available in the global context, but `assert` is set by `node.setup.js`)

SourceR85:
> That's my expectation too. Could `done()` have been called twice, first _without_ the error?

That also was my first thought, but then mocha [will complain](https://github.com/mochajs/mocha/pull/1337) about it.

SourceR85:
Ah, now I got it... 🤦

> I Expected that the test fails if an error is passed to `done()`, but instead there is only a log message:
> 
> ```
>     ✓ changes-filter without filter (46ms)
> Error in .on("change", function): Error: deliberate error in changes
>     at Changes$1.<anonymous> ([...]test.changes.js:1735:17)
> ```

The log message is a side effect from `#3539 - Exception in changes is fine`

I'll `try ... catch` it asap, so no one else get tricked by this.

SourceR85:
And so I got tricked again.

The message comes from a guardedConsole call...
Totally random, unpredictable and unobvious.

SourceR85:
This one is much harder...

In [pocuhdb-server (node, npm test)](https://github.com/pouchdb/pouchdb/actions/runs/6472560136/job/17573550077?pr=8775#step:4:3780) i got this warning:
```
     ✓ Changes with filter not present in ddoc
(node:6308) UnhandledPromiseRejectionWarning: not_found: missing
    at Object.<anonymous> (/home/runner/work/pouchdb/pouchdb/packages/node_modules/pouchdb-errors/lib/index.js:26:19)
    at Module._compile (internal/modules/cjs/loader.js:1114:14)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1143:10)
    at Module.load (internal/modules/cjs/loader.js:979:32)
    at Function.Module._load (internal/modules/cjs/loader.js:819:12)
    at Module.require (internal/modules/cjs/loader.js:1003:19)
    at require (internal/modules/cjs/helpers.js:107:18)
    at Object.<anonymous> (/home/runner/work/pouchdb/pouchdb/packages/node_modules/pouchdb-utils/lib/index.js:9:21)
    at Module._compile (internal/modules/cjs/loader.js:1114:14)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1143:10)
(Use `node --trace-warnings ...` to show where the warning was created)
(node:6308) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 6)
(node:6308) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
(node:6308) UnhandledPromiseRejectionWarning: not_found: missing
    at Object.<anonymous> (/home/runner/work/pouchdb/pouchdb/packages/node_modules/pouchdb-errors/lib/index.js:26:19)
    at Module._compile (internal/modules/cjs/loader.js:1114:14)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1143:10)
    at Module.load (internal/modules/cjs/loader.js:979:32)
    at Function.Module._load (internal/modules/cjs/loader.js:819:12)
    at Module.require (internal/modules/cjs/loader.js:1003:19)
    at require (internal/modules/cjs/helpers.js:107:18)
    at Object.<anonymous> (/home/runner/work/pouchdb/pouchdb/packages/node_modules/pouchdb-utils/lib/index.js:9:21)
    ✓ Changes with `filters` key not present in ddoc
    at Module._compile (internal/modules/cjs/loader.js:1114:14)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1143:10)
(node:6308) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 7)
```

In the <code>Changes with &grave;filters&grave; key not present in ddoc</code> test case,  the result seems to be injected into the message...

I can't reproduce/fix this strange behaviour on my machine,
so I think I leave this one for someone smarter than me (or my future self).

tests/integration/test.changes.js
```diff
@@ -309,47 +309,42 @@ adapters.forEach(function (adapter) {
...
+          'correct error status returned'
+        );
+        assert.equal(err.name, 'not_found');
+        caughtErrors.push(err);
+      });
+
+      // await the end of changes and confirm the error,
+      // otherwise it will be a dangeling or false negative test result.
+      await assert.isRejected(changes, "changes is rejected");
+      assert.lengthOf(caughtErrors, 1, "changes emitted the expected error");
```
SourceR85, line 344: Due to my lack of English: Is it clear/understandable enough?

SourceR85:
Well then ... at least this test seems to be stabilized.

I'll say it's good enough for a merge.

alxndrsn:
Is the significant change here just from `.should...` to `assert...`, or is there something more subtle going on?

alxndrsn (approved):

tests/integration/test.changes.js
```diff
@@ -309,32 +309,41 @@ adapters.forEach(function (adapter) {
...
+          err.status,
+          testUtils.errors.MISSING_DOC.status,
+          'correct error status returned'
+        );
+        assert.equal(err.name, 'not_found');
+        caughtErrors.push(err);
+      });
+
+      await assert.isRejected(changes, "completes with an exception");
+      assert.lengthOf(caughtErrors, 1, "changes emitted the expected error");
```
garethbowen, line 342:
```suggestion
      await assert.isRejected(changes, 'completes with an exception');
      assert.lengthOf(caughtErrors, 1, 'changes emitted the expected error');
```

tests/integration/test.changes.js
```diff
@@ -343,24 +352,34 @@ adapters.forEach(function (adapter) {
...
```
garethbowen, line 382:
```suggestion
      await assert.isRejected(changes, 'completes with an exception');
      assert.lengthOf(caughtErrors, 1, 'changes emitted the expected error');
```

garethbowen (commented):
Looks good to me. A couple of nitpicks for consistency but otherwise good to go.

SourceR85:
Nice catch, @garethbowen.

I've totally miss the surrounding string-style.

PR #8776, opened 2023-10-10
tests: remove commonUtils.Promise

alxndrsn:
* simplifies code
* avoids including the `pouchdb-for-coverage` module unless it's explicitly required elsewhere - e.g. halves the LoC of `performance-bundle.js`


garethbowen (approved):
Nice one.

PR #8777, opened 2023-10-10
tests/performance: fix use of this.iterations in setup/tearDown

alxndrsn:
Currently, setting `ITERATIONS` to anything except 0 will silently break `pull-replication-perf-skimdb` because of a mismatch between `this.iterations` in the test `setup()` function and the actual number of iterations the test is run for.

This failure is surfaced by https://github.com/pouchdb/pouchdb/pull/8768.

As noted in `TODO` comment, the fix in this commit is not ideal, as it mutates the `testCase` objects created in `tests/performance/perf.basics.js` and other perf test files.


alxndrsn:
See:

https://github.com/pouchdb/pouchdb/blob/b0f2e57df165368deaffdc4a8765bff4555bdbea/tests/performance/perf.basics.js#L263-L265

tests/performance/utils.js
```diff
@@ -21,15 +21,21 @@ if (global.window && global.window.location && global.window.location.search) {
 var adapterUsed;
 
 exports.runTests = function (PouchDB, suiteName, testCases, callback) {
+  if (typeof iterations === 'number') {
+    // Re-write to allow test.iterations use inside setup & tearDown.
+    // TODO a better approach would be to pass the number of iterations into
+    // setup & tearDown functions.
+    testCases.forEach((testCase) => {
+      testCase.iterations = iterations;
```
garethbowen, line 29: This seems like a risky property name as it's likely to conflict but I can't figure out if you're internationally overwriting a property that's in the testCase? If not, can we call it something that's clearly bespoke?

tests/performance/utils.js
```diff
@@ -21,15 +21,21 @@ if (global.window && global.window.location && global.window.location.search) {
 var adapterUsed;
 
 exports.runTests = function (PouchDB, suiteName, testCases, callback) {
+  if (typeof iterations === 'number') {
```
garethbowen, line 24: Let's default `testCase.iterations` to 0 rather than leaving it undefined.

garethbowen (changes requested):
It feels a bit confusing right now. A few suggestions inline...

alxndrsn:
@garethbowen maybe worth another look: I've updated to pass `iterations` as an argument to the performance test `setup()` function.

garethbowen (approved):
Thanks! Ship it.

PR #8779, opened 2023-10-11
tests/common-utils: parse query string w/ URLSearchParams

alxndrsn:
* previous parsing was manual
* switch from using `document.location` to more-standard `window.location`
* currently this PR re-implements surprising details of old implementation

Original mapping of empty strings to `true` looks like it was introduced in 2f7b57baca50e893a61c33f7de15835796f083a6.

### DIFF
--- a/tests/common-utils.js
+++ b/tests/common-utils.js
@@ -14,14 +14,15 @@ commonUtils.params = function () {
   if (commonUtils.isNode()) {
     return process.env;
   }
-  const usp = new URLSearchParams(document.location.search);
-  return usp.entries().reduce((acc, [k, v]) => {
+  const usp = new URLSearchParams(window.location.search);
+  const acc = {};
+  for (const [k, v] of usp) {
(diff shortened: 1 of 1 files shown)


tests/common-utils.js
```diff
@@ -14,15 +14,15 @@ commonUtils.params = function () {
...
-  return paramStr.split('&').reduce(function (acc, val) {
-    if (!val) {
-      return acc;
-    }
-    var tmp = val.split('=');
-    acc[tmp[0]] = decodeURIComponent(tmp[1]) || true;
-    return acc;
-  }, {});
+  const usp = new URLSearchParams(window.location.search);
+  const acc = {};
```
garethbowen, line 18: Let's rename `acc` as it doesn't make sense now that this isn't a reduce. Just `params` or `result` would be fine.

garethbowen (approved):
One less reduce! Love it.

One nitpick inline.

PR #8787, opened 2023-10-12
adapter: simplify compareByIdThenRev()

alxndrsn:
Remove one layer of functions.

### DIFF
--- a/packages/node_modules/pouchdb-core/src/adapter.js
+++ b/packages/node_modules/pouchdb-core/src/adapter.js
@@ -75,9 +75,9 @@ function cleanDocs(docs) {
 // compare two docs, first by _id then by _rev
 function compareByIdThenRev(a, b) {
   if (a._id === b._id) {
-   const aStart = a._revisions ? a._revisions.start : 0;
-   const bStart = b._revisions ? b._revisions.start : 0;
-   return aStart - bStart;
+    const aStart = a._revisions ? a._revisions.start : 0;
+    const bStart = b._revisions ? b._revisions.start : 0;
+    return aStart - bStart;
   }
   return a._id < b._id ? -1 : 1;
 }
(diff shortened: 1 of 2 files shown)


packages/node_modules/pouchdb-core/src/adapter.js
```diff
@@ -78,13 +74,12 @@ function cleanDocs(docs) {
 
 // compare two docs, first by _id then by _rev
 function compareByIdThenRev(a, b) {
-  var idCompare = compare(a._id, b._id);
-  if (idCompare !== 0) {
-    return idCompare;
+  if (a._id === b._id) {
+   const aStart = a._revisions ? a._revisions.start : 0;
+   const bStart = b._revisions ? b._revisions.start : 0;
+   return aStart - bStart;
```
garethbowen, line 80: The indentation is off I think...

```suggestion
    const aStart = a._revisions ? a._revisions.start : 0;
    const bStart = b._revisions ? b._revisions.start : 0;
    return aStart - bStart;
```

garethbowen (commented):
Is this function used a lot? Does this yield a measurable improvement?

alxndrsn:
> Is this function used a lot? Does this yield a measurable improvement?

I suspect not - it's specific to `bulkDocs(..., { ..., new_edits: true, ...})`

SourceR85:
> > Is this function used a lot? Does this yield a measurable improvement?
> 
> I suspect not - it's specific to `bulkDocs(..., { ..., new_edits: true, ...})`

bulkDocs get called by every api-call to post and put... true is the default value for new_edits.
I think the relative performance impact depends on the use case.

garethbowen (approved):
I'm a little on the fence here... I think it's a little less readable, but should be a little faster. But I'm not one to stand in the way of progress, so approved!

PR #8789, opened 2023-10-15
adapter.computeHeight(): remove unused return value

alxndrsn:
The return value from the callback passed to `traverseRevTree()` is not used for anything.


alxndrsn:
Not valid!

PR #8790, opened 2023-10-16
core/adapter: use forEach() instead of map() when result not used

alxndrsn:


AbdullahWins (commented):
Good Job!

garethbowen (approved):

packages/node_modules/pouchdb-core/src/adapter.js
```diff
@@ -403,7 +403,7 @@ class AbstractPouchDB extends EventEmitter {
         });
       }
 
-      ids.map(function (id) {
+      ids.forEach(function (id) {
```
SourceR85, line 406: In my opinion, a for ... of loop suites this block better:
because `this` is bound to `this` of the outer scope.

SourceR85: Alt.: an arrow-fn

alxndrsn: :+1: both nicer options, but I want to keep incidental changes minimal.

PR #8791, opened 2023-10-16
core/adapter: use findIndex()

alxndrsn:
Avoid creating extra arrays via `map()` when `findIndex()` exists.

* may effect IE11 support: https://caniuse.com/?search=findindex / https://github.com/pouchdb/pouchdb/issues/7906
* already in use elsewhere - in https://github.com/pouchdb/pouchdb/blob/43e6160af35c94564056ea4ce0c7148f80bd0e22/packages/node_modules/pouchdb-abstract-mapreduce/src/index.js#L621 since https://github.com/pouchdb/pouchdb/pull/8453 /  (2022-12-07)


alxndrsn:
Ref https://github.com/pouchdb/pouchdb/pull/8791

PR #8797, opened 2023-10-19
tests/perf: fix doc creation in basic-gets

alxndrsn:
If test param ITERATIONS is specified as larger than 1000, after the 1000th iteration this test currently changes behaviour from fetching existing docs to fetching non-existent docs.

The original intention seems to be that the number of docs created should be the same as the number of iterations.  See e.g. 7089d39dbf85611543b224d3e3dce7af3a7407e2


alxndrsn:
This will only fix the problem if https://github.com/pouchdb/pouchdb/pull/8777 or a similar fix is merged.

garethbowen (approved):

PR #8802, opened 2023-10-23
test-browser: work around weird webkit behaviour

alxndrsn:
Currently, the test suite fails in webkit:

```
$ BROWSER_CONSOLE=1 GREP=test.replication COUCH_HOST='http://admin:password@localhost:5984' CLIENT=webkit npm run test
...
Unhandled error in test page: [Fetch API cannot load http: /0.0.0.0:13370// due to access control checks.]
```


garethbowen (commented):
If this works, great. Can you provide any more details like a stacktrace of where this was being thrown, or a more complete log?

alxndrsn:
> Can you provide any more details like a stacktrace of where this was being thrown, or a more complete log?

There aren't any more details.

Here is how it looks in webkit dev tools' console:

Screenshot_2023-11-07_11-05-02

The error message cannot be expanded.

garethbowen (approved):
Ok, let's go for it.

The only other thing I could find was about turning cors mode on, eg: https://stackoverflow.com/questions/63141448/safari-fetch-api-cannot-load-due-to-access-control-checks-after-reload

But without knowing where this error is coming from that may be completely spurious...

SourceR85:
> Here is how it looks in webkit dev tools' console:
> 
> Screenshot_2023-11-07_11-05-02
> 
> The error message cannot be expanded.

I slightly remember having a bad time with a double slash-path `http://localhost//` and Safari...
Maybe you hit the same issue here.

alxndrsn:
> I slightly remember having a bad time with a double slash-path (e.g.: `http://localhost//`) and Safari...
> Maybe you hit the same issue here.

Sadly not - removing it gives:

```
Unhandled error in test page: [Fetch API cannot load http: /0.0.0.0:13370/ due to access control checks.]
```

Unclear why it's there in the first place, though...

alxndrsn:
> Unclear why the 2nd `/` is there in the first place, though...

Perhaps related to one of the hand-rolled URL parsers/builders.

PR #8805, opened 2023-10-23
ci: run tests in webkit

alxndrsn:
* skipped test: `test.sync.js`: 5007 sync 2 databases - this test times out consistently on webkit

Blocked by:

* https://github.com/pouchdb/pouchdb/pull/8804
* https://github.com/pouchdb/pouchdb/pull/8832


garethbowen (approved):
Nice, thanks!

SourceR85:
I have that feeling, that the replication/sync errors may be a side-effect result of some race condition or some state persists between test-cases:
in isolated runs, I haven't seen these random crashes.
Only when the servers (ci-runners) are under pressure.

PR #8807, opened 2023-10-25
indexeddb: prevent collisions in find()

alxndrsn:
Closes https://github.com/pouchdb/pouchdb/issues/8808


garethbowen (commented):
Won't this be a breaking change if anyone has a key/path that already has an underscore in it?

alxndrsn:
> Won't this be a breaking change if anyone has a key/path that already has an underscore in it?

Good catch!  Updating indexes should be handled in https://github.com/pouchdb/pouchdb/blob/04a1735cef289c2dd6f4dd05b6e65ae6bfac1267/packages/node_modules/pouchdb-adapter-indexeddb/src/setup.js#L43-L92, but only if the database version number has changed.  I've updated the PR to increase the version number.

alxndrsn:
Actually the version number _may_ not be required, as it implicitly includes a timestamp: https://github.com/pouchdb/pouchdb/blob/04a1735cef289c2dd6f4dd05b6e65ae6bfac1267/packages/node_modules/pouchdb-adapter-indexeddb/src/setup.js#L36-L38

However, I'm confused why [`IDBFactory.open()`](https://developer.mozilla.org/en-US/docs/Web/API/IDBFactory/open#version) would ever be called _without_ the 2nd (`version`) arg: https://github.com/pouchdb/pouchdb/blob/04a1735cef289c2dd6f4dd05b6e65ae6bfac1267/packages/node_modules/pouchdb-adapter-indexeddb/src/setup.js#L107-L109

alxndrsn:
@garethbowen I think indexes will be updated correctly without an adapter version change.

Ref. https://github.com/pouchdb/pouchdb/pull/8812, timestamp-based version change should happen on page load, and `maintainNativeIndexes()` should then be called.

garethbowen (approved):
Thanks for the explanation!

SourceR85:
> However, I'm confused why [`IDBFactory.open()`](https://developer.mozilla.org/en-US/docs/Web/API/IDBFactory/open#version) would ever be called _without_ the 2nd (`version`) arg:
> 
> https://github.com/pouchdb/pouchdb/blob/04a1735cef289c2dd6f4dd05b6e65ae6bfac1267/packages/node_modules/pouchdb-adapter-indexeddb/src/setup.js#L107-L109

The `version` parameter is optional.
Omitting it, `open` must instantiate IDBDatabase **without** an `upgrade`. [MDN](https://developer.mozilla.org/en-US/docs/Web/API/IDBFactory/open#version) [W3C](https://w3c.github.io/IndexedDB/#ref-for-dom-idbfactory-open%E2%91%A1)

Issue #8641, opened 2023-05-25
pouchdb-browser fails importing from 'events'


jdfwarrior:
### Issue
I just started on my project and still haven't actually used pouchdb at all yet. I'm in the middle of trying to upgrade a bunch of dependencies that are a few years old and don't even have the app running yet. One such update was migrating from webpack to vite as the build tool and I've come across a few issues.

We're using pouchdb-browser.

As defined in the documentation under the plugins section, says that pouchdb comes default with the pouchdb-replication plugin. pouchdb-replication however imports EE from 'events' which isn't available in the browser. It's throwing an error for me when the Changes class attempts to extend EE, because EE is null or not a constructor. Maybe this could be resolved by replacing imports for 'events' with something like 'eventemitter3'.

I also initially ran into an issue where the `immediate` lib (one of pouchdb-browser's dependencies) was throwing an error because the `mutation` lib within it attempts to initialize a `Mutation` variable to `global.MutationObserver` and vite doesn't provide shims for node like webpack does (reference: https://github.com/vitejs/vite/issues/2778#issuecomment-810086159). This can be resolved by importing a module that sets `window.global = window` and make it your first import. I know this is outside the scope of pouchdb, but if someone is running into these issues, they're potentially doing something similar and wanted to document the cause.

### Info
- Environment: browser
- Platform: electron

### Reproduce
I didn't go make a base case for this but I would think it should be easily reproducible by creating a new vite project and importing pouchdb-browser

jatin-2708:
Please assign me this issue, very much excited to mark may first contribution to this.

twitwi:
Thanks for the issue and the work in progress.
If anyone has the problem with import_events in the browser, I managed to work around it thanks to this issue, with `npm install events`.
And for the missing global, I updated my vite config

```
  define: { // https://stackoverflow.com/questions/72114775/vite-global-is-not-defined
    global: 'window',
  },
```

alxndrsn:
> I also initially ran into an issue where the `immediate` lib (one of pouchdb-browser's dependencies) was throwing an error because the `mutation` lib within it attempts to initialize a `Mutation` variable to `global.MutationObserver`

This may be caused by https://github.com/calvinmetcalf/immediate/issues/39.

zaynekomichi:
> Thanks for the issue and the work in progress. If anyone has the problem with import_events in the browser, I managed to work around it thanks to this issue, with `npm install events`. And for the missing global, I updated my vite config
> 
[...]

Issue #8607, opened 2023-02-27
Unable to use pouchDB 8.0.1 on a Vite 4.1.0 React Typescript + SWC application


monsieurpigeon:
### Issue
Unable to use pouchDB 8.0.1 on a Vite 4.1.0 React Typescript + SWC application

### Info
- Environment: browser
- Platform: Chrome
- Adapter: indexedDB
- Server: PouchDB Server

### Reproduce
Create a new vite app using react / typescript + SWC, add pouchDB
You should get the following error : `Uncaught ReferenceError: global is not defined`

If you update vite.config.ts and add ```define: {
    global: "window"
  }```
You get a new error : `index-browser.es.js:457 Uncaught TypeError: Class extends value [object Object] is not a constructor or null`

monsieurpigeon:
apparently fixed when adding "events" to dependencies
'npm i events'

enhariharan:
Hi. Did you get a solution to this problem? I am facing the same issue and created a similar question for this in StackOverflow - https://stackoverflow.com/q/75808603/5447891 

I tried "npm i events" but that did not help. I still get the same error as above - Uncaught TypeError: Class extends value [object Object] is not a constructor or null.

ehsanghorbani190:
For those who are seeking an answer for this tedious problem, you first have to add this script tag in your `index.html`:
```html
<script>
	var global = global || window;
</script>
```
Now, install `events` package in your project, and you should be good to go.

There's another way of using pouchdb. You can just install `events` package, and just don't import pouchdb in your JS files, it should work ok.

alxndrsn:
> You should get the following error : `Uncaught ReferenceError: global is not defined`

This may be caused by https://github.com/calvinmetcalf/immediate/issues/39.

garethbowen:
Fixed by #8849

Issue #8516, opened 2022-06-06
PouchDB with Nuxt 3: Uncaught ReferenceError: global is not defined


Extarys:
### Issue
Getting `Uncaught ReferenceError: global is not defined` with the line: `var Mutation = global.MutationObserver || global.WebKitMutationObserver;` in file `node_modules/immediate/lib/mutation.js` when using Nuxt 3.

Not sure I provided enough details... I also created a [SO question](https://stackoverflow.com/questions/72483284/how-to-use-rxdb-with-pouchdb-adapter-in-a-nuxt-3-project) for RxDB when used with PouchDB.

My main goal is to use RxDB, but I need a way to make PouchDB works first.

### Info
- Environment: Browser, Nuxt 3 RC 3
- Platform: FF
- Adapter: idb
- Server: CouchDB

### Reproduce

1. `npx nuxi init`
2. `npm install pouchdb`
3. `plugin/pouchdb.ts`:
```
import PouchDB from "pouchdb"

export default defineNuxtPlugin(async nuxtApp => {

    var db = new PouchDB('dbname');

    db.put({
        _id: 'dave@gmail.com',
        name: 'David',
        age: 69
    });

    db.changes().on('change', function () {
        console.log('Ch-Ch-Changes');
    });

    return {
        provide: {
            db: () => db
        }
    }
});
```

I tried adding the following to `nuxt.config.ts` 
```
build: {
		transpile: [
			'pouchdb-browser',
			'pouchdb-utils',

		],
	},
```
I'm far from an expert with transpiling and webpack and such. I saw some polyfill with webpack for the `global` error [here](https://github.com/pouchdb/pouchdb/issues/7984#issuecomment-1003587487).

I tried adding it with Nuxt 3 but either Nuxt didn't add it or this isn't the right fix for my issue...

```
	app: {
		head: {
			script: [
				{
					children: `window.process = window.process || { env: { NODE_DEBUG: undefined, DEBUG: undefined } }; window.global = window;`
				}
			]
		}
	},
```

Nuxt config API [here](https://v3.nuxtjs.org/api/configuration/nuxt.config)

stjoe4:
They can't fix this issue.

Extarys:
Oh ok, thanks for your reply. Quite funny, libraries designed for client-side database sync stuff not working with recent tech, PWA and the like :/

[...]

PR #8816, opened 2023-11-07
Fix "global is not defined" from immediate dep

alxndrsn:
Untested, but I _think_ this will fix the `ReferenceError: global is not defined` issue reported in:

* https://github.com/pouchdb/pouchdb/issues/8607
* https://github.com/pouchdb/pouchdb/issues/8641
* https://github.com/pouchdb/pouchdb/issues/8516

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -24,6 +24,7 @@
     "FileReaderSync": true,
     "sqlitePlugin": true,
     "emit": true,
+    "globalThis": false,
     "PouchDB": true,
     "should": true,
     "assert": true,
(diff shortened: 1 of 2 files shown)


garethbowen:
`immediate` hasn't been updated for 10 years. What do we need it for and can we drop it?

SourceR85:
> `immediate` hasn't been updated for 10 years. What do we need it for and can we drop it?

As far as I have discovered: the standardized `queueMicrotask` API has taken over, so there is no longer a demand for `immediate`.

The charts tell me `queueMicrotask` has been fully established:
https://developer.mozilla.org/en-US/docs/Web/API/queueMicrotask#browser_compatibility

A bit better visible on caniuse, if you switch to "Date relative":
implemented in the major browsers somewhere between 2019 and 2020
https://caniuse.com/?search=queueMicrotask

If we drop `immediate`, it seems logic to switch to queueMicrotask too.

garethbowen:
Fixed in a different way in #8849

PR #8818, opened 2023-11-07
eslint: set globals readonly

alxndrsn:
Depends on https://github.com/pouchdb/pouchdb/pull/8817


.eslintrc.json
```diff
@@ -9,27 +9,19 @@
...
-    "FileReaderSync": true,
-    "sqlitePlugin": true,
-    "emit": true,
-    "PouchDB": true,
-    "should": true,
-    "assert": true,
-    "testUtils": true,
-    "importScripts": true,
-    "testCases": true
+    "Map": "readonly",
```
garethbowen, line 16: Shouldn't we just bump the ecmaVersion rather than allowing a selection of globals?

alxndrsn: That would probably be a useful change too, but I think we'd still need quite a few of the globals.  This change isn't that drastic - it's making the existing eslint globals config more restrictive, and updating to modern syntax.

garethbowen (approved):

PR #8823, opened 2023-11-17
test-browser: use native event handling

alxndrsn:
Previously the browser test integration with Selenium was written to poll the browser for logs and test events at a fixed interval.  With playwright, we can hook directly into the browser events, and handle them immediately.

This reduces lag for general logging, result reporting, and test completion.

This change will also allow deeper integration with playwright, e.g. checking for open network connections on individual test completion.


bin/test-browser.js
```diff
@@ -72,46 +72,49 @@ class RemoteRunner {
...
-    });
+  handleFailed() {
+    if (bail) {
+      try {
+        this.handlers['end'].forEach(function (handler) {
+          handler();
+        });
 
-    this.completed = true;
+        this.handleEnd(true);
```
alxndrsn, line 111: Move this to `finally`

bin/test-browser.js
```diff
@@ -72,46 +72,49 @@ class RemoteRunner {
...
+      } catch (e) {
+        // Temporary debugging of bailing failure
```
alxndrsn, line 113: Remove this comment

bin/test-browser.js
```diff
@@ -72,46 +72,49 @@ class RemoteRunner {
...
+        console.log('An error occurred while bailing:');
+        console.log(e);
```
alxndrsn, line 115: Combine these 2 log messages

PR #8825, opened 2023-11-20
test: remove whatwg-fetch polyfill

alxndrsn:
* only used in tests
* not required by modern browsers, including all tested against


garethbowen (commented):
What is used in the browser? Native fetch?

SourceR85:
> What is used in the browser? Native fetch?

Correct.
The package packages/node_modules/pouchdb-fetch includes fetch-browser.js when the compile target is "browser".

garethbowen (approved):

PR #8826, opened 2023-11-20
test-browser: improve unhandled error debugging

alxndrsn:
### `master`

```
Unhandled error in test page: {"name":"Error","message":""}
```

### This PR

```
Unhandled error in test page: Unhandled Promise Rejection: {"name":"Error","message":""}
    at unknown (http://127.0.0.1:8000/packages/node_modules/pouchdb/dist/pouchdb.min.js:7:4434)
    at unknown (http://127.0.0.1:8000/packages/node_modules/pouchdb/dist/pouchdb.min.js:7:141513)
```


bin/test-browser.js
```diff
@@ -170,7 +170,7 @@ async function startTest() {
       return;
     }
 
-    console.log('Unhandled error in test page:', err);
+    console.log('Unhandled error in test page:', err.stack);
```
garethbowen, line 173: Maybe do both. The message _might_ have useful info sometimes?

alxndrsn: Seems possible.. apparently `.stack` isn't specced.  Adding `.cause` might be useful too.

bin/test-browser.js
```diff
@@ -220,6 +220,8 @@ async function startTest() {
       }
 
       console.log('Unhandled error in test page:', err);
+      console.log('  stack:', err.stack);
+      console.log('  cause:', err.cause);
```
SourceR85, line 224:
```suggestion
      if("cause" in err) {
        console.log('  cause:', err.cause);
      }
```

Since it raises the min. required version by a lot:
Error .cause compatibility

I would argue like @garethbowen did in [PR 8849](https://github.com/pouchdb/pouchdb/pull/8849#issuecomment-1916306988).

SourceR85 (changes requested):

SourceR85 (approved):
Just changed my mind: It's OK for me.

In case of old environments, `console.log('  cause:', err.cause)` will print
``` console
  cause: undefined
```

Everyone who sees this will come to the conclusion, that the environment is to old for the `cause` property.

Issue #8525, opened 2022-06-22
db.compact() runs much slower than auto_compaction


DarrenCook:
In a large web app we switched on `auto_compaction`. It seemed to work (e.g. chrome disk usage went from 102M to 89M on my test setup, and from 2.3GB to 1.6GB for the guinea pig production user) with no reports of long delays. But we did start seeing 409s on about 10-20% of the DB puts.

So, instead of auto_compaction, I set it up to do a `db.compact()` just when the app first connects to the DB. I expected it to have negligible run-time, especially for an already compacted DB, but for the test setup this adds takes 20 seconds to start-up time; for the same production user it takes at least 3hrs 50 (it eventually had to be killed).

### More Info

Initially, assuming it would add negligible run-time, I did `await db.compact()` immediately after the `new PouchDB()` call, and then again once the `db.replicate.from(remoteDB)` had completed. This added a 40 second delay to startup for my test setup. I commented out one of those calls, and it "only" added 20 seconds.

So I tried removing the `await`, letting it run in the background. But now the app hangs for 20 seconds if we try to immediately do anything with those databases. (If we wait 20 seconds, before doing anything, everything is fine.)

So I moved it to a button, for more controlled testing, and to do one DB at a time. Same behaviour.

We have 5 databases. In my test setup, 4 of them were basically finishing immediately, and the 5th (with 174 documents and possibly some high revision counts) was taking about 20 seconds.

In the production system mentioned above two of the five databases are large: 34205 documents in one, 25544 in another. And some of those documents will have over 100 revisions.  (Assuming it was proportional to document count I'd estimated it might take 2-3hrs, but looks like it is even more...)

Possibly Related: we use worker-pouch as the adapter for those databases, so all the couchDB actions are kept in their own thread. I'm not able to easily test without that, though.

Other information: main testing done in Chromium 101, on Linux and Mac; also on Electron (so basically Chromium again). Reasonably fast machines. The remote server is CouchDB (3.1.x); I checked logs during the app startup, and could see the sync requests come through, nothing else, nothing unusual.

### The Question/Bug/Feature Request

Is `auto_compaction` using the same code as `db.compact()`? If so, what could explain the hugely different run times?

If not, what are they doing differently, and why?

Can I create PouchDB with `auto_compaction:true` (as that seems to do what we wanted), but then switch it off again? I.e. just before heavy use of the DB starts, so I can avoid those 409s?

DarrenCook:
[...]

PR #8834, opened 2023-11-23
core: Fix compaction from the last last_seq (#8525)

H--o-l:
In 2014, commit 4f74ea5 made global database compaction smarter by avoiding
compacting documents that weren't modified since the last time the compaction
run.

But in 2015, commit b4f79e2 fixed an issue in that code and at the same time
removed the `since = last_seq` parameter that was given to `self.changes()`.
From what I see this makes `self.changes` emit for each database document
instead of only for the one modified since `since = last_seq`.

I saw in my browser that putting back the `since = last_seq` parameters allows
the `onChange()` callback to be called only for documents with a
`doc.lastSeq > since`, so I think it fixes the feature well.


alxndrsn:
This sounds very good!

IMO it would be helpful to have a new test case which failed before this change and passed afterwards to:

1. prevent future re-regression, and
2. help reviewer understand what's going on

H--o-l:
Hey @alxndrsn, thanks for the feedback.
I'd be happy to add a test on the subject but I'm unfamiliar with the structure of the project, could you pinpoint me to the right place to add it?

H--o-l:
PS: @nolanlawson you made the change back then, I suppose it was a mistake to remove the `since = last_seq`, but maybe it wasn't. Do you remember why you removed it?

alxndrsn:
> could you pinpoint me to the right place to add it?

I'm guessing: `tests/integration/test.changes.js` :slightly_smiling_face:

H--o-l:
Hi @alxndrsn, thanks for pushing for adding a test, it was doable by overwriting the `_local/compaction` normally created by PouchDB itself.
I included it inside the commit. I cannot wait to have your feedback on it!

Note: I have tested that the test fails before the commit and succeeds after.

tests/integration/test.compaction.js
```diff
@@ -507,6 +507,18 @@ adapters.forEach(function (adapter) {
       });
     });
 
+    it('#8525 - Only compact document with seq > last_seq', async function () {
+      const db = new PouchDB(dbs.name, {auto_compaction: false});
+      const rev = (await db.put({_id: 'foo'})).rev;
+      await db.put({_id: 'foo', _rev: rev});
+      // Hack compaction last_seq to make like a previous compaction already
```
alxndrsn, line 514: Is it possible to actually run the compaction instead of hacking the local doc?  This test seems implementation-specific and might be brittle.

H--o-l: Hey @alxndrsn!

Thanks for the quick feedback.

I have thought of three ways to automatically test that compaction is not done from the start anymore:

- Thanks to a performance test that would test the performance gain of the PR. That sounds cumbersome and brittle.
- Thanks to a [JS Proxy](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/Proxy) that would be plugged on the document object and check that the compaction method of that object is not called. It doesn't sound straightforward and it's also implementation-specific
- The solution I added to the commit: create a fake `_local/compaction` which is implementation-specific too, indeed, but less than the second solution I think.

The third solution was the best looking for me, and somehow easy to fix if it break one day.
I also doubt a bit that the implementation of the `_local/compaction` will ever change, because in hasn't changed since 2014, and also because it would re-trigger the compaction from the start in all the databases out there, which, from my point of view, would be bad.

Do you have any idea of a test that would not be implementation-specific (because I exhausted the ideas I had in mind)?

H--o-l:
Hi @alxndrsn!

I don't understand why the test I added to the PR fails on the CI.
Locally on my computer, runing `GREP='#8525 - Only compact document with seq > last_seq' npm test` works well, and, if I understand CI results correctly, it looks like it works in some of the CI env, but not all (maybe I'm wrong?).

What I understand by looking at the error is that the compaction of the document is made no matter what in some CI environments, like if the code I fixed was not the one used (or if my hack of `_local/compaction` had no effect). I guess it can happen if PouchDB is used only as a proxy for a remote CouchDB? But I'm not sure. 
If this is the case I think my test makes no sense in these environments, as my test is here to explicitly test that the compaction is NOT made. 
So I was thinking:

- Could a test be run in only some environments? And if yes, which one makes sense?
- Maybe my test should be just removed from the PR? Testing that something is not done is always a bit hard and strange.

tests/integration/test.compaction.js
```diff
@@ -507,6 +507,18 @@ adapters.forEach(function (adapter) {
       });
     });
 
+    it('#8525 - Only compact document with seq > last_seq', async function () {
```
SourceR85, line 510: To skip tests for remote databases, you could call skip on `this`.
E.g:
```js
// check for remote target:
if(dbs.name.startsWith("http")) {
    // skip the test:
    this.skip()
}

SourceR85: Sry, my code was just meant as an example.
Please add a return statement after the `this.skip();` line, so it exits early.

The current implementation unnecessarily runs the rest of the test function even in the remote case.

H--o-l: OK done!
Thanks for the help on the subject.

SourceR85: Nothing to thanx... everyone get stuck from time to time.

So my brain tricked me sometimes...
I've just looked up the implementation of Mocha's .skip(), and it turns out:
The return stmt isn't required, because .skip() throws an internal object (this way, .skip always exits the runner).
But it also the didn't hurt, to have a keyword in the line. (Keep praising the syntax highlighter 😁)

garethbowen:
@H--o-l How are you getting on with this?

H--o-l:
@garethbowen Tests won't pass, I don't know why, I relaunch them multiple times. If you want to give it a try don't hesitate.
(For me the code is good, it's a test issue I think)

SourceR85:
Hi @H--o-l,
if you update your branch and let the CI run, it will pass (CI will automatically re-run periodically, if it didn't pass and the branch is up to date).
The errors, occurred in recent run, are known false-negatives.

garethbowen:
@SourceR85 @alxndrsn The build is now passing. Can one or both of you review given you've been commenting along with this change?

H--o-l:
Thanks!

SourceR85 (approved):
I'm fine with it.

Even the `_local/compaction` hack seems reasonable, since this tests an internal miss-behaviour.
As @alxndrsn commented: It would be nice if we find another way to check for this, but for now it should be OK (I'll left an issue as a reminder).

H--o-l:
Thanks for the approval @SourceR85, I'm glad you like it.

Just asking, could the PR be merged (and released maybe?)

SourceR85:
I can't,
@alxndrsn  and @garethbowen are the ones with member status.
I'm just a contributor with interest in a performant running DB like you ;)

garethbowen:
@alxndrsn Because you commented on this PR merging is blocked until you approve it. Would you mind?

garethbowen (approved):
Approved to make GH happy

garethbowen:
@H--o-l Merged. Releasing will take a little longer - there's a large change improving the idb adapter I want to get merged and then do a release after that.

PR #8840, opened 2023-12-05
tests: wait for express servers to close

alxndrsn:
> The app.listen() method returns an [http.Server](https://nodejs.org/api/http.html#http_class_http_server) object and (for HTTP)
> - _https://expressjs.com/en/4x/api.html#app.listen_

> Stops the server from accepting new connections. See [net.Server.close()](https://nodejs.org/docs/latest-v14.x/api/net.html#net_server_close_callback).
> - _https://nodejs.org/docs/latest-v14.x/api/http.html#http_server_close_callback_

>  The optional callback will be called once the 'close' event occurs. 
> - _https://nodejs.org/docs/latest-v14.x/api/net.html#net_server_close_callback_


tests/component/test.deletion_error.js
```diff
@@ -29,8 +29,8 @@ describe('test.deletion_error.js', function () {
     server = app.listen(0);
   });
 
-  after(function () {
-    return server.close();
+  after(function (done) {
+    server.close(done);
```
garethbowen, line 33: I think this is safer because the docs say the callback will be called with an Error if the server was not open when it was closed. That would end up failing the `after` which would not be right in this case.

```suggestion
    server.close(() => done());
```

tests/component/test.read_only_replication.js
```diff
@@ -38,8 +38,8 @@ describe('test.read_only_replication.js', function () {
     server = app.listen(0);
   });
 
-  after(function () {
-    return server.close();
+  after(function (done) {
+    server.close(done);
```
garethbowen, line 42: As above.

tests/component/test.replication_perf_regression.js
```diff
@@ -34,8 +34,8 @@ describe('test.replication_perf_regression.js', function () {
     };
   });
 
-  after(function () {
-    return server.close();
+  after(function (done) {
+    server.close(done);
```
garethbowen, line 38: As above.

garethbowen (changes requested):

tests/component/test.deletion_error.js
alxndrsn: I think it's reasonable for the `after()` to fail if the server has already stopped - presumably it means that the server has crashed, which would probably need investigating, especially if it wasn't caught elsewhere.

garethbowen: Fair point. This is an improvement either way so approved, and I'll let you pick which you prefer.

garethbowen (approved):

PR #8843, opened 2023-12-30
Update README.md

codesmith-emmy:


garethbowen:
I think it's debatable whether this is an improvement or not, grammatically speaking. Even if it were it's too minor to spend any time on.

Issue #8526, opened 2022-06-27
M1 support


sirily11:
### Issue
Hi I encountered the following issue when i trying to install pouchdb.
`node_modules/.pnpm/leveldown@5.6.0/node_modules/leveldown: Running install script, failed in 766ms`

### Info
- Environment: Browser on MacOS M1 Max with pnpm
- Node Version: v16.3.0

Full error log
```
node_modules/.pnpm/leveldown@5.6.0/node_modules/leveldown: Running install script, failed in 766ms
.../node_modules/leveldown install$ node-gyp-build
│ node:events:371
│       throw er; // Unhandled 'error' event
│       ^
│ Error: spawn node-gyp ENOENT
│     at Process.ChildProcess._handle.onexit (node:internal/child_process:282:19)
│     at onErrorNT (node:internal/child_process:480:16)
│     at processTicksAndRejections (node:internal/process/task_queues:83:21)
│ Emitted 'error' event on ChildProcess instance at:
│     at Process.ChildProcess._handle.onexit (node:internal/child_process:288:12)
│     at onErrorNT (node:internal/child_process:480:16)
│     at processTicksAndRejections (node:internal/process/task_queues:83:21) {
│   errno: -2,
│   code: 'ENOENT',
│   syscall: 'spawn node-gyp',
│   path: 'node-gyp',
│   spawnargs: [ 'rebuild' ]
│ }
└─ Failed in 766ms
Progress: resolved 1035, reused 1011, downloaded 0, added 0, done
 ELIFECYCLE  Command failed with exit code 1.
```
### Fix

I tried manually installing the leveldown v5 and shows exactly the same error. However,  after I install the latest version of leveldown, the error is gone.

anmavrid:
Similar issue: Pouchdb does not run on Apple M1 architectures. To run it successfully we need leveldown version ^6.0.3 but pouchdb requires leveldown version 5.6.0 (as listed in the package.json file).

rmdstudio:
Same here; we have been having problems getting PouchDB to work on our new Mac M1 machines.

taratatach:
This is still relevant.

garethbowen:
I don't have an M1 to test. If one of you could get it working on your hardware and submit a PR that'd be very helpful.

taratatach:
I don't have that either sadly.

adueck:
It's working fine for me using a M1 MacBook Air with `node v18.13.0` and yarn.

[...]

PR #8727, opened 2023-08-14
fix: upgrade leveldown from 5.6.0 to 6.1.1

madeindjs:
Same as https://github.com/pouchdb/pouchdb/pull/8529 

I had many issue using pouchDB and Next.js _(it seems I'm not the only one https://github.com/vercel/vercel/issues/3635 )_

So I upgraded le library from master, I ran tests locally and it worked. I hope the CI will pass :crossed_fingers:

PR #8666, opened 2023-06-19
leveldown upgraded to 6.1.1

rasgo-cc:
This is the same as #8529. Just wanted to see if I can do something regarding passing tests. Locally, all tests pass.

PR #8529, opened 2022-07-03
fix: upgrade leveldown from 5.6.0 to 6.1.1

vapormusic:
Fix PouchDB not working on arm64 Apple Silicon machines (#8416)

PR #8845, opened 2024-01-04
(#8526): Upgrade leveldown from 5.6.0 to 6.1.1

an9949an:
Leveldown needs to be upgraded to support Apple M1. https://github.com/pouchdb/pouchdb/issues/8526

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "level": "6.0.1",
     "level-codec": "9.0.2",
     "level-write-stream": "1.0.0",
-    "leveldown": "5.6.0",
+    "leveldown": "6.1.1",
     "levelup": "4.4.0",
     "localstorage-down": "0.6.7",
     "ltgt": "2.2.1",


alxndrsn:
This PR is currently identical to:

* https://github.com/pouchdb/pouchdb/pull/8727/files
* https://github.com/pouchdb/pouchdb/pull/8666/files
* https://github.com/pouchdb/pouchdb/pull/8529/files

It looks like the only CI failures are when testing with webpack.

an9949an:
@alxndrsn yes, I saw. But all this PRs outdated and looks abandoned with all CI logs cleared. I just saw that something fails and this is it.

> It looks like the only CI failures are when testing with webpack.

Yes, but very strange. After a brief investigation I see that it fails at `acorn` parse of `node_modules/abstract-level/abstract-chained-batch.js` on the spread operator. I have no idea what's going on for now. Probably some dependencies incompatibility.

I will try to take a look a bit later. Or maybe somebody has ideas why it happens...

alxndrsn:
My guess is it's related to the ecmascript version that webpack is expecting.

an9949an:
@alxndrsn It fails because of `bundle-collapser` which used at `bin/build-utils.js:53`:
```js 
bundler = browserify(addPath(pkgName, filepath), opts)
  .transform('es3ify')
  .plugin('bundle-collapser/plugin');
```
`bundle-collapser` depends on `falafel 2.2.5` which depends on `acorn ^7.1.1` (resolved as 7.4.1).
Both `bundle-collapser`(last commit 4 years ago) and `falafel` (last commit 3 years ago) look abandoned.

Removing of `bundle-collapser` fixing it at least locally. But I don't sure what to do with this, probably somebody with project knowledge needed here. 

Also for some strange reason it doesn't fail during the normal build (exact same line executed during the `npm run build` many times, even with exact same file) and fails when `CLIENT=firefox npm run test-webpack`.

garethbowen:
@an9949an It looks like all bundle-collapser does is help minify the output a little. Could you run with and without and see what the difference in bundle size is?

an9949an:
@garethbowen Here is the differences

<details>
  <summary>With collapser</summary>
  
</details>

<details>
  <summary>Without collapser</summary>
  
</details>

Sometimes it's near nothing (~1kb difference), sometimes not. The worst case is something like `pouchdb.localstorage.min.js`: 288kb against 297kb (~3% increase)

garethbowen:
That seems acceptable to me. Would you mind updating this PR to remove those libs so we can ensure the tests pass?

an9949an:
@garethbowen It's already removed in this PR.

Also one more topic to think about: the `level` package. Right now we have `level:6.0.1` which depends on `leveldown:^5.4.0`. The latest version is `level:8.0.0` which depends on `leveldown:6.1.1` (I updated `leveldown` to `6.1.1` in direct dependencies). So, should we update the level package as well maybe?

garethbowen:
@an9949an Let's run this as it is and then raise a new PR for updating `level` otherwise we risk making everything too complicated again...

bin/build-utils.js
```diff
@@ -50,7 +50,6 @@ function doBrowserify(pkgName, filepath, opts, exclude) {
     } else {
       bundler = browserify(addPath(pkgName, filepath), opts)
         .transform('es3ify')
```
garethbowen, line 52: This needs a semicolon to pass linting...

an9949an: Yes, sorry, fixed.

garethbowen:
@an9949an Unfortunately there are quite a few test failures. Can you run the tests locally and see if you can reproduce the issue?

an9949an:
@garethbowen I can't reproduce locally.

As I can see from the job logs it runs fine on `node:16` (the same locally) and fails on `node:14`. I tried to run on the exact same node version and locally no any errors. Also `firefox` version the same. The only difference is distro. Job use `ubuntu:22.04.3`, I use `fedora:38`. The same test which fails by `20s` timeout complete locally during `2s` with no errors.

Another interesting aspect: as I can see in logs, the job fails on the different tests on the first and second try but both by 20s timeout error. Also the same test (_suite2 test.replication_events.js-http-http -> #4627 Test no duplicate changes in live replication_) which failed by timeout during the second retry was complete in 2108ms during the first one.

So, it looks like something random and maybe env related. Did you see something like that before?

garethbowen:
The tests are sometimes unreliable... I'm re-running the failed builds to check.

an9949an:
@garethbowen one fail less this time :) I tried to run this test locally and it's also fine. But I don't completely sure what is the right way to run the same `indexeddb` test. I use this command:  `CLIENT=firefox ADAPTERS=indexeddb npm test`. Is it the right one?

garethbowen:
@an9949an I re-ran again and it passed.

@alxndrsn I think we're ok to merge this now. What do you think? Are these failures within the bounds of normal?

alxndrsn:
I suspect the correct approach is to remove `es3ify` as well - both of them seem old & unmaintained, and neither of them seem to cope with spread operators:

```
> require('es3ify').transform('(...args) => console.log(...args)')
'(...args) => console.log(...args)'
```

and https://github.com/browserify/bundle-collapser/issues/25

I'd vote for removing them in a separate PR.

garethbowen:
@alxndrsn So that's a yes for merging this as is?

alxndrsn:
> @alxndrsn So that's a yes for merging this as is?

No, I'd rather see bundler-collapser removed first, in a separate PR - to me it's removal seems incidental to upgrading leveldown.

alxndrsn:
The spread operator is used elsewhere in the pouchdb source (and likely in other dependencies), so I don't understand _why_ removing `bundle-collapser` fixes things.  Ideally this would be understood before changing the build process.

an9949an:
> The spread operator is used elsewhere in the pouchdb source (and likely in other dependencies), so I don't understand _why_ removing `bundle-collapser` fixes things. Ideally this would be understood before changing the build process.

@alxndrsn very good question. Unfortunately it fails somewhere inside the build libs without any meaningful output. It was even complicated to figure out where exactly it failed. I did it by adding logs to the `acorn` package source code. But why only this place and why firefox test only - I have no idea honestly.

an9949an:
@alxndrsn for now I prepared PR with removing of `es3ify` and `bundle-collapser`.

> Ideally this would be understood before changing the build process

Yes, but as I can see from your comment at least `es3ify` do nothing meaningful. Spread operator introduced in ECMAScript 6. If `es3ify` don't remove spread it mean that result is still incompatible with ES3. Am I wrong?

alxndrsn:
> Yes, but as I can see from your comment at least `es3ify` do nothing meaningful. Spread operator introduced in ECMAScript 6. If `es3ify` don't remove spread it mean that result is still incompatible with ES3. Am I wrong?

I think you're correct, and I'm 100% in favour of removing `es3ify` immediately.

alxndrsn:
@an9949an does this PR pass with `bundle-collapser`, but with `es3ify` removed?

an9949an:
> @an9949an does this PR pass with `bundle-collapser`, but with `es3ify` removed?

@alxndrsn yes, it fails with the same `SyntaxError: Unexpected token`.

As I can see `bundle-collapser` is also not something very important. It just replaces `require('path/to/something')` with `require(1)` to save a bit of bytes.

garethbowen:
@an9949an I've merged the other PR which removes the outdated libs: https://github.com/pouchdb/pouchdb/pull/8847

Can you rebase this one so it's just bumping leveldown?

an9949an:
@garethbowen done

garethbowen:
@an9949an Looks good to me. Waiting for the build to pass.

I haven't tested this as I don't have an M1 - have you tested this build to ensure it actually fixes the issue?

an9949an:
@garethbowen  I don't have it unfortunately. But we tested it on our project by replacing leveldown binaries to ARM one from leveldown 6.1 and everything worked fine on apple M1.

garethbowen (approved):
Let's do it.

PR #8846, opened 2024-01-04
eslint: comma-dangle: only allow for multiline declarations

alxndrsn:
Ref: https://eslint.org/docs/latest/rules/comma-dangle#options

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -25,6 +25,7 @@
   },
 
   "rules": {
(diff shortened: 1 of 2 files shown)


alxndrsn:
I'd be in favour of `"always-multiline"` as it simplifies diffs, but there are currently 4444 violations...

garethbowen (approved):

PR #8847, opened 2024-01-08
Remove outdated bundle-collapser and es3ify

an9949an:
Both bundle-collapser and es3ify are abandoned. Removing is required to upgrade leveldown. https://github.com/pouchdb/pouchdb/issues/8526


garethbowen (approved):
I'm comfortable with this. They provide little value (and probably even less once the bundle is gzipped) and they're holding us back.

@alxndrsn Can you live with this?

alxndrsn (approved):

PR #8849, opened 2024-01-23
replace `immediate` with queueMicrotask

SourceR85:
This PR unifies pouchdb-utils/src/nextTick and pouchdb-utils/src/nextTick-browser by replacing the dependency on `immediate` and node-specific `process.nextTick` with standardised queueMicrotask-API.

Since the `queueMicrotask`-API has been standardized and established (see [compatibility list at mdn](https://developer.mozilla.org/en-US/docs/Web/API/queueMicrotask#browser_compatibility)), the `immediate` package is obsolete.

As mentioned in https://github.com/pouchdb/pouchdb/pull/8816#issuecomment-1906134713


SourceR85:
As mentioned in #8816

SourceR85:
Are you OK with change @garethbowen?

garethbowen:
Honestly I'm a little reluctant. I think it raises the bar on browser support above where it currently is which always gives me pause. Otherwise as far as I can tell it's better in every way. How do you feel about keeping the next tick files around and updating the browser one to use queue microtask but falling back to a polyfill if it doesn't exist?

eg: 

```
if (typeof queueMicrotask !== 'undefined') {
  queueMicrotask(fn);
} else {
  Promise.resolve().then(fn)
}
```

Alternatively if you can show Pouch already doesn't support Chrome 71 then it's fine as is...

SourceR85:
> How do you feel about keeping the next tick files around

The Node docs specifies: nextTick, promises and microTask queues are executed in this order, and they seem to be equally optimized.
I don't see a benefit in keeping them separated (as browser and node version) and it didn't affect the program flow of PouchDB.
This way, it feels even better: one less branch of behaviour (node vs browser) and one less non-standardized API.

> Alternatively if you can show Pouch already doesn't support Chrome 71 then it's fine as is...

Nothing more advanced than async/await (available since Chrome v55).

TBH I don't feel comfortable supporting (or enable people to keep) unmaintained versions of Chrome, 'cause PouchDB enables the execution of arbitrary code (views map & reduce functions) and the [list of Chrome CVE's](https://cve.mitre.org/cgi-bin/cvekey.cgi?keyword=chrome) show some [serious problems](https://www.cvedetails.com/product/15031/Google-Chrome.html?vendor_id=1224) within the last years.

garethbowen:
> TBH I don't feel comfortable supporting (or enable people to keep) unmaintained versions of Chrome

I agree in general, but if we drop support for browsers then we should probably bump the major version to make it clear, and notify everyone. We should definitely decide what our policy is and document that somewhere when we publish the next major version. Thanks for taking the time to make this backwards compatible for now!

garethbowen (approved):
Thanks!

garethbowen:
@SourceR85 Merged.

Does this mean we can close https://github.com/pouchdb/pouchdb/pull/8816 and linked issues? Have you tested that this solves the Vite and Nuxt issues?

PR #8856, opened 2024-02-19
Checkpointer shortcut

SourceR85:
Omit checkpoint lookup when `writeSourceCheckpoint` and `writeTargetCheckpoint` set to false.
E.g. replicate with `{ checkpoint: false }` option.

Before this patch, a request to a never written document was made, resulting in a network request that fail every time.
Nice side-effects of this patch: reduced latency, less error prone (due to network failures), and reduced network noise in replications when checkpoints are unnecessary.

These tests already cover my patch:
- test.replication.js `Test disable checkpoints on both source and target` and
- test.read_only_replication.js `Test checkpointer handles error codes` (implicitly tests for default/missing option)

### DIFF
--- a/packages/node_modules/pouchdb-checkpointer/src/index.js
+++ b/packages/node_modules/pouchdb-checkpointer/src/index.js
@@ -121,7 +121,7 @@ class CheckpointerInternal {
       return Promise.resolve(LOWEST_SEQ);
     }
 
-    if (self.opts.writeSourceCheckpoint && !self.opts.writeTargetCheckpoint) {
+    if (self.opts && self.opts.writeSourceCheckpoint && !self.opts.writeTargetCheckpoint) {
       return self.src.get(self.id).then(function (sourceDoc) {
         return sourceDoc.last_seq || LOWEST_SEQ;
       }).catch(function (err) {
@@ -134,7 +134,7 @@ class CheckpointerInternal {
     }
 
     return self.target.get(self.id).then(function (targetDoc) {
-      if (self.opts.writeTargetCheckpoint && !self.opts.writeSourceCheckpoint) {
+      if (self.opts && self.opts.writeTargetCheckpoint && !self.opts.writeSourceCheckpoint) {
         return targetDoc.last_seq || LOWEST_SEQ;
       }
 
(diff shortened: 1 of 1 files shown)


packages/node_modules/pouchdb-checkpointer/src/index.js
```diff
@@ -117,6 +117,10 @@ class CheckpointerInternal {
   getCheckpoint() {
     var self = this;
 
+    if (self.opts.writeSourceCheckpoint === false && self.opts.writeTargetCheckpoint === false) {
```
garethbowen, line 120: Below the code just uses truthyness so this could be changed to...

```suggestion
    if (!self.opts.writeSourceCheckpoint && !self.opts.writeTargetCheckpoint) {
```

garethbowen (changes requested):
Thanks. One minor suggestion.

packages/node_modules/pouchdb-checkpointer/src/index.js
SourceR85: Hi Gareth, this was also my first thought.

I've tried it and the result is that the test `checkpointer handles error codes` of `tests/component/test.read_only_replication.js` breaks.
The reason is: it calls `new Checkpointer` without `opts` parameter; resulting in an empty `opts` (line 81) and therefore both options are `undefined` and `!undefined` evaluates to `true`.

So I came to the conclusion, that without specifying the `opts` parameter, the current behaviour must be kept (otherwise, the test might be wrong).

SourceR85: Or the fallback/default is wrong

garethbowen: Makes sense. Thanks.

@SourceR85 Can I add you as a contributor to the pouchdb org? It might unblock you a bit?

garethbowen (approved):

packages/node_modules/pouchdb-checkpointer/src/index.js
SourceR85: > @SourceR85 Can I add you as a contributor to the pouchdb org? It might unblock you a bit?
You're welcome.

SourceR85: > @SourceR85 Can I add you as a contributor to the pouchdb org? It might unblock you a bit?

Did you mean, that you want to invite/suggest me as a member to the GitHub group PouchDB?
I'd be delighted.

SourceR85: If you merge #8871, we can use the shorter syntax 😉

PR #8858, opened 2024-02-22
add descriptions and comments

SourceR85:
I added some comments (even if the steps are obvious) and missing `description` fields in `.github/actions/install-deps/action.yml` to be GH action schema compliant.


SourceR85:
Close, in favor of #8866

PR #8862, opened 2024-02-26
Fix timeout of test.issue3179.js conflicts synced live sync

SourceR85:
A solution to fix "timeout" exception in `test.issue3179.js` case `conflicts synced live sync`.

This patch eliminates a (potential) infinite loop and refactors the test case to es6.

(In combination with #8863, this has the potential to eliminate 90% of false-negatives I regularly encounter)

### DIFF
--- a/tests/integration/test.issue3179.js
+++ b/tests/integration/test.issue3179.js
@@ -143,7 +143,7 @@ adapters.forEach(function (adapters) {
           }
         }
 
-        // Tried 10 times, throw to let the test fail.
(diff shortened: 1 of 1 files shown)


SourceR85:
@garethbowen would you take a look (review/merge)?

garethbowen (approved):

alxndrsn:
This test is failing again.  It would be helpful to separate refactoring PRs from functional changes, as combining the two makes it harder to understand what the significant change was which fixed the tests.

PR #8863, opened 2024-02-26
fix subsequent fails of `5007 sync 2 databases`

SourceR85:
This patch refactors `test.sync.js` case `5007 sync 2 databases` to es6 and ensure remote2 is destroyed before and after manipulation.

Without this cleanup, CI steps `First retry` and `Second retry` fail because synchronization with `remote2` emit changes for documents inserted in previous runs.

**This patch hasn't the intention to fix `Intermittent test failure`!**

(In combination with #8862, this has the potential to eliminate 90% of false-negatives I regularly encounter)

### DIFF
--- a/tests/integration/test.sync.js
+++ b/tests/integration/test.sync.js
@@ -815,7 +815,7 @@ adapters.forEach(function (adapters) {
       });
     });
 
-    it('5007 sync 2 databases', function (done) {
+    it('5007 sync 2 databases', function () {
       if (testUtils.isSafari()) {
         // FIXME this test fails consistently on webkit.  It needs to be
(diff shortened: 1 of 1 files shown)


SourceR85:
I'm glad to have a prove, that it successfully recovers a failing workflow:
A failure with `Timeout of 20000ms exceeded.` exception: https://github.com/pouchdb/pouchdb/actions/runs/8055226934/job/22001674526#step:4:5671
The recover in `First retry` of the same run: https://github.com/pouchdb/pouchdb/actions/runs/8055226934/job/22001674526#step:5:5643

garethbowen (approved):

PR #8866, opened 2024-02-27
Reduce ci overhead

SourceR85:
This patch reduces overhead and distributes work of CI.

Before this patch, `install-deps` was a "handle-em-all" mess.

Now the preparation is done separated by workload:
- Node.js and npm install: ~30 seconds
- CouchDB install: 20-30 seconds (only relevant in coudb + nodejs/browser workflows)
- Playwright install: 30-90 seconds (omitted in server-only scenarios)
- build PouchDB (extracted as an extra step): ~20 seconds (not necessary in lint, `First retry`, and `Second retry`)

The biggest chunk is still `npm test` and companions (per retry) with ~2m in server-to-server test and ~4m in browser-environments.

### DIFF
--- a/.github/actions/install-node-package/action.yml
+++ b/.github/actions/install-node-package/action.yml
@@ -10,11 +10,11 @@ runs:
   using: composite
 
   steps:
-    - uses: actions/setup-node@v3
+    - uses: actions/setup-node@v4
       with:
         node-version: ${{ inputs.node-version }}
 
-    - uses: actions/cache@v3
+    - uses: actions/cache@v4
       with:
         path: ~/.npm
         key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
--- a/.github/actions/install-node-package/action.yml
+++ b/.github/actions/install-node-package/action.yml
@@ -8,7 +8,7 @@ inputs:
 
   skip-playwright-installation:
     description: Save some time by excluding the installation of playwright
-    default: 'yes'
+    default: ${{ true }}
 
 runs:
   using: composite
@@ -25,7 +25,7 @@ runs:
         restore-keys: |
           ${{ runner.os }}-node-
 
-    - if: ${{ inputs.skip-playwright-installation == 'yes' }}
+    - if: ${{ inputs.skip-playwright-installation == true }}
       run: "sed -i 's/\"playwright\":.*//' $GITHUB_WORKSPACE/package.json"
       shell: bash
 
--- a/.github/actions/install-node-package/action.yml
+++ b/.github/actions/install-node-package/action.yml
@@ -6,9 +6,9 @@ inputs:
     description: Specify the Node.js version to install
     required: true
 
-  skip-playwright-installation:
+  prepare-playwright:
     description: Save some time by excluding the installation of playwright
-    default: ${{ true }}
+    default: ${{ false }}
 
 runs:
   using: composite
@@ -25,7 +25,7 @@ runs:
         restore-keys: |
           ${{ runner.os }}-node-
 
-    - if: ${{ inputs.skip-playwright-installation }}
+    - if: ${{ !inputs.prepare-playwright }}
       run: "sed -i 's/\"playwright\":.*//' $GITHUB_WORKSPACE/package.json"
       shell: bash
 
--- a/.github/actions/install-node-package/action.yml
+++ b/.github/actions/install-node-package/action.yml
@@ -25,7 +25,7 @@ runs:
         restore-keys: |
           ${{ runner.os }}-node-
 
-    - if: ${{ inputs.skip-playwright-installation == true }}
+    - if: ${{ inputs.skip-playwright-installation }}
       run: "sed -i 's/\"playwright\":.*//' $GITHUB_WORKSPACE/package.json"
       shell: bash
 
--- a/.github/actions/install-node-package/action.yml
+++ b/.github/actions/install-node-package/action.yml
@@ -6,6 +6,10 @@ inputs:
     description: Specify the Node.js version to install
     required: true
 
+  skip-playwright-installation:
+    description: Save some time by excluding the installation of playwright
+    default: 'yes'
+
 runs:
   using: composite
 
@@ -20,6 +24,10 @@ runs:
         key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
         restore-keys: |
           ${{ runner.os }}-node-
+
+    - if: ${{ inputs.skip-playwright-installation == 'yes' }}
+      run: "sed -i 's/\"playwright\":.*//' $GITHUB_WORKSPACE/package.json"
+      shell: bash
 
     - run: npm install
       shell: bash
--- a/.github/actions/install-node-package/action.yml
+++ b/.github/actions/install-node-package/action.yml
@@ -1,7 +1,10 @@
 name: Install dependencies
+description: Install npm dependencies from package.json
 
 inputs:
-  node-version: { required: true }
+  node-version:
+    description: Specify the Node.js version to install
+    required: true
 
 runs:
   using: composite
--- a/.github/actions/install-node-package/action.yml
+++ b/.github/actions/install-node-package/action.yml
@@ -6,10 +6,6 @@ inputs:
     description: Specify the Node.js version to install
     required: true
 
-  prepare-playwright:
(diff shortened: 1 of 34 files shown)


SourceR85:
Hi @garethbowen,
I got a big question for you: should we keep pouchdb-server as a build dependency?

The reason for this question: pouchdb-server hasn't seen an update in half a decade and for me, it just seems like it's another failing point (unmaintained, incompatible with CouchDB 3.x, outdated dependencies, CJS only, depends on pouchdb:7.x).

If you agree, I'll write another PR to drop pouchdb-server from PouchDB's CI.

SourceR85:
Sry for the interrupt,
I've moved the question about pouchdb-server to [discussions](https://github.com/pouchdb/pouchdb/discussions/8872) so that everyone can comment there opinion.

SourceR85:
@alxndrsn I would be happy, if you could share your opinion about this patch (maybe review?).
Since it looks like you've been deep into the performance and CI-workflow.

bin/dev-server.js
```diff
@@ -82,23 +82,6 @@ function watchAll() {
...
-  return Promise.all([
-    rebuildPouch(),
-    rebuildTestUtils(),
-    rebuildPerf()
-  ]);
-}).then(function () {
-  console.log('Rebuilt PouchDB/test/perf JS bundles');
-  filesWritten = true;
-  checkReady();
-});
```
alxndrsn, line 100: Does removing this break hot-reloading when running the dev server locally?

SourceR85: I've moved the initial build to bin/run-dev.sh.

To better understand this change, here's the call order, when `npm run dev` is executed:
1. npm run dev -> calls ./bin/run-dev.sh
2. run-dev.sh -> builds via build-pouchdb.js and npm build-test (equivalent to the removed build promise you mentioned) and executes run-test.sh
3. run-test.sh -> invokes npm run launch-dev-server
4. npm run launch-dev-server finally run `node ./bin/dev-server.js`

(I know exactly what you think: it's a mess)

My thought: `bin/dev-server.js` is an implementation detail and isn't meant to invoke directly (and not mentioned in TESTING.md). Also: build artifacts aren't required to start `dev-server.js`.

.github/workflows/docs.yml
```diff
@@ -1,5 +1,8 @@
 name: PouchDB Docs
 
+env:
+  NODE_VERSION: 14
+
```
alxndrsn, line 5: could this be inline?  is it here for consistency with the other workflow?

package.json
```diff
@@ -12,13 +12,12 @@
...
     "test-component": "mocha --exit tests/component # --exit for #8839",
     "test-fuzzy": "TYPE=fuzzy npm run test",
-    "test-browser": "npm run build-test && node ./bin/test-browser.js",
+    "test-browser": "node ./bin/test-browser.js",
     "test-memleak": "mocha -gc tests/memleak",
     "eslint": "eslint --cache bin/ packages/node_modules/**/src tests/",
     "dev": "./bin/run-dev.sh",
     "launch-dev-server": "node ./bin/dev-server.js",
     "test": "./bin/run-test.sh",
-    "prepublish": "npm run build",
```
alxndrsn, line 21: I don't understand this change - I assume publish only affects publishing to npm; is there more to it?

bin/dev-server.js
alxndrsn: OK, I understand re: hot-reloading.

alxndrsn: > Edit: Now I'm thinking; should we separate "utilities" into a sub-folder of bin?
> This makes it clearer which files are intended for execution by users.

If this is planned, I'd prefer it as a separate PR - there's already a lot going on here

package.json
SourceR85: I've removed it, because it build on `npm install`.

The naming of this script isn't very intuitive, so I looked up it's behaviour and found this article:
https://blog.greenkeeper.io/what-is-npm-s-prepublish-and-why-is-it-so-confusing-a948373e6be1

If the author's intention was to build it for npm, the `publish` script is more suited for this.
I guess: it was more convenient for them to build automatically after install than do it as an extra step...

bin/dev-server.js
SourceR85: > If this is planned, I'd prefer it as a separate PR - there's already a lot going on here

Agree 👍️

.github/workflows/docs.yml
SourceR85: ✅

package.json
SourceR85: Just looked a bit deeper: `prepublish` is deprecated and npm has introduced two alternative "events": `prepare` and `prepublishOnly`
https://docs.npmjs.com/cli/v10/using-npm/scripts#prepare-and-prepublish

My recent commit adds back the build step for publishing (now via `prepublishOnly`).

alxndrsn:
LGTM.  What is the impact on build times?

alxndrsn (approved):

.github/actions/install-node-package/action.yml
```diff
@@ -1,8 +1,10 @@
 name: Install dependencies
+description: Install npm dependencies from package.json
```
alxndrsn, line 2: For me, the name of the action is self-explanatory.

.github/actions/install-node-package/action.yml
```diff
@@ -1,8 +1,10 @@
...
 
 inputs:
-  node-version: { required: true }
-  couchdb-version: { required: false, default: '3.1' }
+  node-version:
+    description: Specify the Node.js version to install
+    required: true
```
alxndrsn, line 7: As above - not sure this needed to change.

.github/workflows/ci.yml
```diff
@@ -19,15 +19,19 @@ env:
 
 jobs:
 
+  # Check style compliance of JavaScript code.
+
```
alxndrsn, line 23: This comment also seems to explain the obvious to me, but maybe i've been looking at this code too long :older_adult:

.github/workflows/ci.yml
```diff
@@ -19,15 +19,19 @@ env:
...
     steps:
       - uses: actions/checkout@v4
-      - uses: ./.github/actions/install-deps
+      - uses: ./.github/actions/install-node-package
         with:
           node-version: ${{ env.NODE_VERSION }}
       - run: npm run eslint
 
+  # Check shell scripts.
+
```
alxndrsn, line 34: As above.

bin/dev-server.js
```diff
@@ -82,23 +82,6 @@ function watchAll() {
...
-
```
alxndrsn, line 101: I may have just caused a conflict here (?), but happy to help resolve - shouldn't be too complicated.

.github/actions/install-node-package/action.yml
SourceR85: Agree; the name is very descriptive.
But without description fields, the file violates [github-action.json](https://json.schemastore.org/github-action.json).

It's valid but schema-checks aren't happy
grafik

.github/actions/install-node-package/action.yml
SourceR85: Same Answer as above ;)

.github/workflows/ci.yml
SourceR85: True, it is just a comment describing the intention of this step
and everyone who is looking at the last step of the action may guess it by first sight.

With these comments you didn't have to ;)
it feels more like a headline of an article.

Side note: at first, I've just added it to be consistent with the rest of the file (new line, comment, newline, action).

.github/workflows/ci.yml
SourceR85: As above ;)

bin/dev-server.js
SourceR85: No problem, I've just removed the lines again.

SourceR85:
> What is the impact on build times?

The build times aren't affected. (if you reduce "build" to `npm run build`)
Instead it reduces the overall time spending on preparation (depending on the workload).

For GH workflows, it has a big impact (it depends on the performance of ci runners).
As the [description for this patch](https://github.com/pouchdb/pouchdb/pull/8866/#issue-2157763477) highlights:

This patch:
Full CI run of #8866

Another CI run (without changing PouchDB's source code):
Full CI run of #8882

PR #8871, opened 2024-03-04
Set default opts for `CheckpointerInternal`

SourceR85:
This patch fixes the fallback/default value of `CheckpointerInternal`'s `opts` parameter.

Since `Checkpoint` is the backbone for replication, I've derived the "default" from it's `checkpoint` parameter:
If no `opts` parameter is passed, checkpoints for source and target will be written.

### DIFF
--- a/packages/node_modules/pouchdb-checkpointer/src/index.js
+++ b/packages/node_modules/pouchdb-checkpointer/src/index.js
@@ -73,12 +73,15 @@ function updateCheckpoint(db, id, checkpoint, session, returnValue) {
 }
 
 class CheckpointerInternal {
-  constructor(src, target, id, returnValue, opts) {
+  constructor(src, target, id, returnValue, opts = {
+    writeSourceCheckpoint: true,
+    writeTargetCheckpoint: true,
+  }) {
     this.src = src;
     this.target = target;
     this.id = id;
     this.returnValue = returnValue;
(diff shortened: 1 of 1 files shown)


packages/node_modules/pouchdb-checkpointer/src/index.js
```diff
@@ -73,12 +73,15 @@ function updateCheckpoint(db, id, checkpoint, session, returnValue) {
 }
 
 class CheckpointerInternal {
-  constructor(src, target, id, returnValue, opts) {
+  constructor(src, target, id, returnValue, opts = {
+    writeSourceCheckpoint: true,
+    writeTargetCheckpoint: true,
+  }) {
```
garethbowen, line 79: I don't think this defaults if a value is provided for opts, but one or other of these values aren't provided, eg: `CheckpointerInternal(...., {})`

garethbowen (changes requested):

packages/node_modules/pouchdb-checkpointer/src/index.js
SourceR85: Okay, I've added fallback values for individual properties.
Now you can provide partial options, pass an empty object, pass `undefined`, call without opts, or pass any arbitrary object (except null and frozen/read-only objects).

garethbowen (approved):

PR #8876, opened 2024-03-06
perf-tests: fix JSON reporter

alxndrsn:
Since runner.completed was removed in dc97aedf449a94b87f074f3aa975b5412415706a, env var JSON_REPORTER has not worked.

This also changes the exit status code for perf tests: previously the process would always exit with status 0, even if there was a test failure.  Test failure(s) will now result in an exit status of 0.

### DIFF
--- a/bin/test-browser.js
+++ b/bin/test-browser.js
@@ -72,6 +72,7 @@ class ArrayMap extends Map {
 
 class RemoteRunner {
   constructor(browser) {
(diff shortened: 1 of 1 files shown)


SourceR85 (approved):
LGTM:
Setting a `failed` state instead of passing seems reasonable and
when the runner emits an "end" event, `completed` is implicitly given.

PR #8880, opened 2024-03-20
chore(abstract-mapreduce): replace `var` with `const`

SourceR85:
Purely cosmetic updates for PouchDB's abstract-mapreduce.

- Replace `var` with `const` declarations.
- At Line 620 replaced `let` with `const`, since `purge` never get reassigned in this context.


alxndrsn:
@SourceR85 lots of good stuff here, although this is a lot to review.  I'd personally prefer:

* smaller, more focussed PRs - this reduces the burden on reviewer, and generally leads to quicker turnaround
* less opinion-based formatting changes - these make diffs harder to read (increasing review burden), and make git history harder to navigate, which can be important for understanding historic decisions in this repo

I'm generally in favour of the move from callbacks to `async`/`await`, but it has also historically introduced bugs in this repo (I suspect including some of the current test flake).  Smaller PRs would help here too for both better reviews, and clearer git history.

SourceR85:
@alxndrsn Thank you for your explanation,
I'll split both of my PRs asap (8860 is even worse in this one).

> I'm generally in favour of the move from callbacks to `async`/`await`, but it has also historically introduced bugs in this repo (I suspect including some of the current test flake).

In my experience: switching between promise chains and `async`/`await` keywords is equivalent, because the keywords are just syntactic sugar....

The reason behind the flaky tests must be something else, introduced as a side effect.
I'm pretty sure, someone just forgot to wait for a promise in a critical path or the event emitter misfires (I need to study the side effects a bit more deeply to come to a conclusion).

SourceR85:
Now it's down to `var` replacement only ;)

alxndrsn (approved):

PR #8890, opened 2024-03-26
fix: input isn't defined in array case

SourceR85:
Fixed comparison against an undeclared/undefined variable (`value !== null` always evaluates to `undefined !== null` when `input` is an array)

To me it looks like a copy & past mistake (102 looks similar to line 124)


garethbowen (approved):
LG. Can you add a test that proves this?

SourceR85:
> Can you add a test that proves this?

Included now (and found #8898, #8899, #8900, and #8901 on the way)

PR #8891, opened 2024-03-26
chore(abstract-mapreduce): cleanup taskqueue.js

SourceR85:
Use `Promise.resolve` instead of an self-resolving instantiation.

### DIFF
--- a/packages/node_modules/pouchdb-abstract-mapreduce/src/taskqueue.js
+++ b/packages/node_modules/pouchdb-abstract-mapreduce/src/taskqueue.js
@@ -12,7 +12,7 @@ class TaskQueue {
     this.promise = this.promise
       // just recover
       .catch(function () { })
-      .then(promiseFactory);
+      .then(() => promiseFactory());
     return this.promise;
   }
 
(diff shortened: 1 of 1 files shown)


packages/node_modules/pouchdb-abstract-mapreduce/src/taskqueue.js
```diff
@@ -3,18 +3,16 @@
...
 
   add(promiseFactory) {
-    this.promise = this.promise.catch(function () {
+    this.promise = this.promise
       // just recover
-    }).then(function () {
-      return promiseFactory();
-    });
+      .catch(function () { })
+      .then(promiseFactory);
```
garethbowen, line 15: Technically you've changed the behaviour here because any return value from the `promise` will be passed to `promiseFactory` as a parameter. I'm too lazy to work out if this will break anything, but to be safe let's change it to...

```suggestion
      .then(() => promiseFactory());
```

SourceR85: Sure, I'll provide a follow up pr for this.

garethbowen (approved):

PR #8893, opened 2024-03-26
Add scripts for comparing perf test result files

alxndrsn:


bin/perf-test-results/compare-multiple.js
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env node
+/* eslint-disable curly,max-len */
+
+const { loadResultFile, printComparisonReport, SUITE_FOR } = require('./lib');
+
+const [ , , ...files ] = process.argv;
```
SourceR85, line 6:
```suggestion
const files = process.argv.slice(2);
```
Slightly less noise ;)

alxndrsn: But less readable IMO :shrug:

SourceR85: It depends on the reader preferences.

In the end, the style is for humans and shouldn't matter:
the engines will optimize both styles down to the same op-codes 😉

PR #8895, opened 2024-03-27
Remove legacy workaround for Blob.slice()

alxndrsn:
Common since 2012.

See: https://developer.mozilla.org/en-US/docs/Web/API/Blob/slice#browser_compatibility

### DIFF
--- a/packages/node_modules/pouchdb-utils/src/cloneBinaryObject-browser.js
+++ b/packages/node_modules/pouchdb-utils/src/cloneBinaryObject-browser.js
@@ -14,14 +14,8 @@ function cloneBinaryObject(object) {
   if (object instanceof ArrayBuffer) {
     return cloneArrayBuffer(object);
   }
-  var size = object.size;
-  var type = object.type;
   // Blob
-  if (typeof object.slice === 'function') {
-    return object.slice(0, size, type);
-  }
-  // PhantomJS slice() replacement
-  return object.webkitSlice(0, size, type);
+  return object.slice(0, object.size, object.type);
 }
 
 export default cloneBinaryObject;
--- a/packages/node_modules/pouchdb-md5/src/binaryMd5-browser.js
+++ b/packages/node_modules/pouchdb-md5/src/binaryMd5-browser.js
@@ -7,17 +7,10 @@ function rawToBase64(raw) {
   return btoa(raw);
 }
 
-function sliceBlob(blob, start, end) {
-  if (blob.webkitSlice) {
-    return blob.webkitSlice(start, end);
-  }
-  return blob.slice(start, end);
-}
-
 function appendBlob(buffer, blob, start, end, callback) {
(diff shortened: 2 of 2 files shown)


packages/node_modules/pouchdb-utils/src/cloneBinaryObject-browser.js
```diff
@@ -14,14 +14,8 @@ function cloneBinaryObject(object) {
   if (object instanceof ArrayBuffer) {
     return cloneArrayBuffer(object);
   }
```
SourceR85, line 16:
```suggestion
  if (object instanceof ArrayBuffer) {
    return object.slice();
  }
```
Since we dropped support for IE, `cloneArrayBuffer` can be removed too 😉

SourceR85, line 16:
```suggestion
  if (object instanceof ArrayBuffer) {
    return object.slice();
  }
```
Since we dropped support for IE: `cloneArrayBuffer` can be removed too 😉

Side note: `start` (first parameter) of .slice is optional and has 0 as default, calling without any parameter is equivalent to .slice(0, size).

alxndrsn: is blob.slice() guaranteed to work?  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer#transferring_arraybuffers

SourceR85: It's in the specs:
https://tc39.es/ecma262/multipage/structured-data.html#sec-arraybuffer.prototype.slice

And the compatibility table on Mozilla says, that every browser supports it:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/slice#browser_compatibility

Same  results on caniuse (even was available in IE 11... So this workaround is ancient):
https://caniuse.com/?search=Arraybuffer%20slice

alxndrsn: I was referring to this:

> Detached buffers have the following behaviors:
> ...
> Methods, such as ... [slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/slice), throw a [TypeError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError) when invoked

I don't currently have enough context to confidently make changes to `ArrayBuffer` usage.  I can see how it could be a good follow-up to this PR though.

SourceR85: > I can see how it could be a good follow-up to this PR though.

Sounds like a plan.

SourceR85 (approved):

PR #8897, opened 2024-03-27
Add tests for attachments on local docs

alxndrsn:
These tests describe the current behaviour of different PouchDB adapters & server implementations.  Documenting them via test means that changes in behaviour for local attachments can be recognised and mentioned in release notes.


tests/integration/utils.js
```diff
@@ -175,6 +175,26 @@ testUtils.isCouchDB = function (cb) {
   });
 };
 
+testUtils.getServerType = function (cb) {
+  const knownServers = [
+    'couchdb',
+    'express-pouchdb', // seems to be pouchdb-server
```
SourceR85, line 181:
```suggestion
    'express-pouchdb',
```
It lives in pouchdb-server`s packages, but it just wraps what you feed it with.
https://github.com/pouchdb/pouchdb-server/blob/975ea2377957223dc828bcdb4f5d403a5cc9f709/packages/node_modules/express-pouchdb/lib/index.js#L251

tests/integration/test.attachments.js
```diff
@@ -2031,6 +2040,81 @@ adapters.forEach(function (adapter) {
...
+              } else {
+                done(new Error(`No handling for server type: '${serverType}'`));
+              }
+            });
+            done();
+          } else {
+            throw new Error(`No handling for adapter: '${adapter}'`);
+          }
+        });
+      });
```
SourceR85, line 2115:
```suggestion
    it('Test attachments:true for _local doc', async function () {
      const db = new PouchDB(dbs.name);
      await db.put(binAttDocLocal);

      const doc = await db.get('_local/bin_doc', { attachments: true });

      if (adapter === 'local') {
        doc._attachments['foo.txt'].content_type.should.equal('text/plain');
        doc._attachments['foo.txt'].data.should.equal('VGhpcyBpcyBhIGJhc2U2NCBlbmNvZGVkIHRleHQ=');
      } else if (adapter === 'http') {
        const serverType = await testUtils.getServerType();
        if (serverType === 'couchdb') {
          should.not.exist(doc._attachments);
        } else if (serverType === 'pouchdb-express-router' || serverType === 'express-pouchdb') {
          doc._attachments['foo.txt'].content_type.should.equal('text/plain');
          JSON.parse(decodeBase64(doc._attachments['foo.txt'].data)).should.deep.equal({
            error: 'not_found',
            reason: 'missing',
          });
        } else {
          throw new Error(`No handling for server type: '${serverType}'`);
        }
      } else {
        throw new Error(`No handling for adapter: '${adapter}'`);
      }
    });
```

SourceR85 (commented):
Streamlining tests and reducing the mixing of cb/promise style.

tests/integration/utils.js
```diff
@@ -175,6 +175,26 @@ testUtils.isCouchDB = function (cb) {
...
 };
 
+testUtils.getServerType = function (cb) {
+  const knownServers = [
+    'couchdb',
+    'express-pouchdb',
+    'pouchdb-express-router',
+  ];
+
+  var {url, options} = parseHostWithCreds(testUtils.couchHost());
```
SourceR85, line 185:
```suggestion
  const { url, options } = parseHostWithCreds(testUtils.couchHost());
```

SourceR85: Sorry for the delay...
Just had a rough time to get this running on my machine.

Every other test worked without a problem, so I didn't thought about any configuration problem.

But here it is: when setting the CouchDB port via couchdb.ini,
the authorization header isn't passed correctly by `PouchDB.fetch`, resulting in `401 Unauthorized` errors but `db.fetch` keep working as expected...

tests/integration/utils.js
```diff
@@ -175,6 +175,26 @@ testUtils.isCouchDB = function (cb) {
...
+  PouchDB.fetch(url, options).then(function (response) {
+    return response.json();
+  }).then(function (res) {
+    for (const known of knownServers) {
+      if (res[known]) {
+        return cb(known);
+      }
+    }
+    cb('unknown');
+  });
```
SourceR85, line 195: The last thing I found:

In case of an exception: the callback was never called, resulting in an unreasonable `time out`.
```suggestion
  }).catch(function (_err) {
    cb('unknown');
  });
```

alxndrsn: Good catch - the error definitely needs handling.  I'd rather throw instead of logging I think.

alxndrsn:
> Just a suggestion: Reduce the mix of callbacks and promises.

:+1: I've converted the new tests & util function to use promises.

SourceR85 (approved):

PR #8905, opened 2024-04-02
Remove dev dependency: median

alxndrsn:
This is a tiny dependency, used in one place, and not maintained.


alxndrsn:
OTOH more convenient than having to write and maintain similar functionality.

PR #8906, opened 2024-04-02
Remove dev dependency: mkdirp

alxndrsn:
As recommended in mkdirp README:

> On Node.js v10.12.0 and above, use the native fs.mkdir(p, {recursive:true})

See: https://www.npmjs.com/package/mkdirp#implementation


SourceR85 (approved):
LGTM

PR #8907, opened 2024-04-02
dev dependencies: replace individual lodash functions

alxndrsn:
* it's not clear if these individual functions' releases are kept up-to-date with main lodash module
* keeping multiple dependencies up-to-date is harder than working with a single lodash dependency
* the whole of lodash is available anyway in dev via sub-dependencies


SourceR85:
An idea for another PR: lodash's flatten can be replaced with standard [`Array.prototype.flat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat)

SourceR85 (approved):

PR #8908, opened 2024-04-02
remove unused devDependency: less

alxndrsn:


alxndrsn:
Used implicitly in `bin/build-site.js`.

PR #8909, opened 2024-04-02
remove unused devDependency: query-string

alxndrsn:


SourceR85 (approved):
It was imported for `pouchdb-find/test/test.js` in https://github.com/pouchdb/pouchdb/commit/b309d997797a39813b9ec6fcb1c2c7950aba9e1a

and `pouchdb-find/test/test.js` was then removed by https://github.com/pouchdb/pouchdb/commit/50d6cdd297e396ec366186595301ecdeca5f8e72

query-string wasn't removed with the test file.

PR #8911, opened 2024-04-02
repeat-perf-test: allow specifying adapters

alxndrsn:


bin/repeat-perf-test.sh
```diff
@@ -37,7 +37,7 @@ if [[ "$#" -lt 1 ]]; then
       Repeatedly run the performance test suite against one or more versions of the codebase.
 
     USAGE
-      [PERF_REPEATS=<N>] $0 ...tree-ish
+      [PERF_REPEATS=<N>] $0 ...tree-ish:adapter
```
SourceR85, line 40: Which adapters are allowed?

SourceR85 (changes requested):

bin/repeat-perf-test.sh
alxndrsn: https://github.com/pouchdb/pouchdb/blob/master/TESTING.md#adapters-default-depends-on-client

alxndrsn: > Comma-separated list of preferred adapter backends that PouchDB will use for local
> databases. These are selected automatically based on the execution environment,
> but this variable overrides the default choice and causes additional adapters to
> be loaded if they're not part of the default distribution.                    
>                                                                               
> On Node.js the available local adapters are `leveldb` and `memory`. In the    
> browser they're `idb`, `indexeddb` and `memory`.                              
>                                                                               
> You can also set `ADAPTERS=http` to force all PouchDB databases to be created on
> a remote server, identified by `COUCH_HOST`. This is not necessary for        
> integration tests since they use a mixture of local and remote databases to   
> check compatibility, but it's useful for the `find` and `mapreduce` suites.

SourceR85: Thx, seems to be working well.

SourceR85 (approved):

PR #8914, opened 2024-04-03
chore: PouchDB-find adapters/local refactor to async/await

SourceR85:
Isolated PR from #8860

### DIFF
--- a/packages/node_modules/pouchdb-find/src/adapters/local/find/index.js
+++ b/packages/node_modules/pouchdb-find/src/adapters/local/find/index.js
@@ -24,7 +24,7 @@ function indexToSignature(index) {
   return index.ddoc.substring(8) + '/' + index.name;
 }
 
-function doAllDocs(db, originalOpts) {
+async function doAllDocs(db, originalOpts) {
   var opts = clone(originalOpts);
 
   // CouchDB responds in weird ways when you provide a non-string to _id;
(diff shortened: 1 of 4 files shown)


packages/node_modules/pouchdb-find/src/adapters/local/find/index.js
```diff
@@ -89,119 +87,113 @@ function find(db, requestDef, explain) {
...
+    return Promise.resolve(queryPlan, opts);
+  }
 
-      if (queryPlan.inMemoryFields.length) {
-        // need to filter some stuff in-memory
-        res.rows = filterInMemoryFields(res.rows, requestDef, queryPlan.inMemoryFields);
-      }
+  const res = indexToUse.name === '_all_docs'
+    ? await doAllDocs(db, opts)
+    : await abstractMapper(db).query.call(db, indexToSignature(indexToUse), opts);
```
garethbowen, line 141: I find this hard to follow. I think it's multiline ternary statements with awaits. A cleaner solution for me is to pull out a named function to do this, eg:

```js
async function queryIndex(db, opts, indexToUse) { // TODO come up with a better fn name
  if (indexToUse.name === '_all_docs') {
    return doAllDocs(db, opts);
  }
  return abstractMapper(db).query.call(db, indexToSignature(indexToUse), opts);
}

...

const res = await queryIndex(db, opts, indexToUse);
```

garethbowen (changes requested):
Nice! Just one suggestion for readability...

packages/node_modules/pouchdb-find/src/adapters/local/find/index.js
SourceR85: https://github.com/pouchdb/pouchdb/pull/8914/commits/1edb94a6224e82c01f049eb6d708f67e9f127655 should do the trick ;)

garethbowen (approved):
Thanks!

PR #8918, opened 2024-04-04
chore(PouchDB-find): loop updates

SourceR85:
Improve readability and performance (slightly) of loops in pouchdb-find.

- refactor iterator loops to for ... of: when `i` was only for element lookup.
- flatten: use native `Array.prototype.flat` is 2x faster than polyfill.
- `uniq` utility -> Array.from(Set): slight perf boost (`Set` is guaranteed to be unique).

Isolated from #8860

### DIFF
--- a/packages/node_modules/pouchdb-find/src/adapters/local/find/query-planner.js
+++ b/packages/node_modules/pouchdb-find/src/adapters/local/find/query-planner.js
@@ -31,13 +31,8 @@ const SHORT_CIRCUIT_QUERY = {
 // couchdb second-lowest collation value
 
 function checkFieldInIndex(index, field) {
-  const indexFields = index.def.fields.map(getKey);
-  for (const indexField of indexFields) {
-    if (field === indexField) {
-      return true;
-    }
-  }
-  return false;
+  return index.def.fields
+    .some((key) => getKey(key) === field);
 }
 
 // so when you do e.g. $eq/$eq, we can do it entirely in the database.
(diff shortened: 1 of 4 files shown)


packages/node_modules/pouchdb-find/src/utils.js
```diff
@@ -196,13 +200,7 @@ function arrayEquals(arr1, arr2) {
 }
 
 function uniq(arr) {
-  const obj = {};
-  for (let i = 0; i < arr.length; i++) {
-    obj['$' + arr[i]] = true;
-  }
-  return Object.keys(obj).map(function (key) {
-    return key.substring(1);
-  });
```
garethbowen, line 205: Wow, that's insane...

garethbowen (approved):

packages/node_modules/pouchdb-find/src/adapters/local/find/query-planner.js
```diff
@@ -32,8 +32,7 @@ const SHORT_CIRCUIT_QUERY = {
 
 function checkFieldInIndex(index, field) {
   const indexFields = index.def.fields.map(getKey);
-  for (let i = 0, len = indexFields.length; i < len; i++) {
-    const indexField = indexFields[i];
+  for (const indexField of indexFields) {
```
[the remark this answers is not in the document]
SourceR85: Thanks for the hint 👍️
Just as I was reading this, I had an idea for another shortening:
https://github.com/pouchdb/pouchdb/commit/323e43eabb10bec05151734879f3738ef86a729a

packages/node_modules/pouchdb-find/src/utils.js
SourceR85: Pardon?

garethbowen: The old code, not the new stuff!  :)

garethbowen (approved):
Thanks!

garethbowen (approved):
Thanks!

PR #8923, opened 2024-04-05
test.slash_id: assert getAttachment() result

alxndrsn:
It looks like this assertion was missing.


tests/integration/test.slash_id.js
```diff
@@ -34,6 +34,7 @@ adapters.forEach(function (adapter) {
         db.putAttachment(docId, attachmentId, info.rev, blob, 'text/plain',
                          function () {
           db.getAttachment(docId, attachmentId, function (err, res) {
+            res.should.equal(blobData);
             testUtils.readBlob(res, function () {
```
SourceR85, line 38: `res` is a blob (test fail).
Plain text is available when readBlob is done:
```suggestion
            testUtils.readBlob(res, function (data) {
              data.should.equal(blobData);
```

SourceR85 (approved):

PR #8928, opened 2024-04-12
Replace lodash.flatten() with Array.flat() built-in

alxndrsn:
Follow-up to https://github.com/pouchdb/pouchdb/pull/8907

### DIFF
--- a/bin/update-package-json-for-publish.js
+++ b/bin/update-package-json-for-publish.js
@@ -12,7 +12,7 @@ var path = require('path');
 var glob = require('glob');
 var findRequires = require('find-requires');
 var builtinModules = require('builtin-modules');
-const { flatten, uniq } = require('lodash');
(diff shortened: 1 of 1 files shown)


SourceR85:
This rises the min. Browser Support
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat#browser_compatibility

@garethbowen said [here](https://github.com/pouchdb/pouchdb/pull/8849#issuecomment-1916306988), he want to have a clear way of communicating such requirement changes...

Unfortunately, I don't know how and where we can announce this.

garethbowen:
IMO the recent changes to the indexeddb adapter are significant enough to warrant a major version change, which means it's also the perfect time for things like raising minimum browser requirements.

SourceR85 (approved):

PR #8931, opened 2024-04-12
db.put(): throw INVALID_REV for invalid revs

alxndrsn:
Without the fix, tests fail:

```
  test.basics.js-http
Testing on: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) HeadlessChrome/123.0.6312.4 Safari/537.36
Failed to load resource: the server responded with a status of 400 (Bad Request)
    ✔ Modify a doc with illegal rev value #0 (83ms)
Failed to load resource: the server responded with a status of 400 (Bad Request)
    ✔ Modify a doc with illegal rev value #1 (42ms)
Failed to load resource: the server responded with a status of 400 (Bad Request)
    ✔ Modify a doc with illegal rev value #2
Failed to load resource: the server responded with a status of 500 (Internal Server Error)
    1) Modify a doc with illegal rev value #3
Failed to load resource: the server responded with a status of 500 (Internal Server Error)
    2) Modify a doc with illegal rev value #4
Failed to load resource: the server responded with a status of 500 (Internal Server Error)
    3) Modify a doc with illegal rev value #5
Failed to load resource: the server responded with a status of 500 (Internal Server Error)
    4) Modify a doc with illegal rev value #6
Failed to load resource: the server responded with a status of 500 (Internal Server Error)
    5) Modify a doc with illegal rev value #7
Failed to load resource: the server responded with a status of 400 (Bad Request)
    ✔ Modify a doc with illegal rev value #8
Failed to load resource: the server responded with a status of 400 (Bad Request)
    ✔ Modify a doc with illegal rev value #9 (67ms)
Failed to load resource: the server responded with a status of 400 (Bad Request)
    ✔ Modify a doc with illegal rev value #10
Failed to load resource: the server responded with a status of 400 (Bad Request)
    ✔ Modify a doc with illegal rev value #11
Failed to load resource: the server responded with a status of 400 (Bad Request)
    ✔ Modify a doc with illegal rev value #12
Failed to load resource: the server responded with a status of 500 (Internal Server Error)
    6) Modify a doc with illegal rev value #13
Failed to load resource: the server responded with a status of 500 (Internal Server Error)
    7) Modify a doc with illegal rev value #14
Failed to load resource: the server responded with a status of 500 (Internal Server Error)
    8) Modify a doc with illegal rev value #15
Failed to load resource: the server responded with a status of 500 (Internal Server Error)
    9) Modify a doc with illegal rev value #16
[...]

PR #8933, opened 2024-04-13
isValidRev(): reject revs with more than one dash

alxndrsn:
There are various places in the code where revs are assumed to include a single dash character:

```js
* pouchdb-adapter-idb/src/index.js:533: doc._rev = '0-' + (parseInt(oldRev.split('-')[1], 10) + 1);
* pouchdb-adapter-leveldb-core/src/index.js:1402: oldRev ? '0-' + (parseInt(oldRev.split('-')[1], 10) + 1) : '0-1';
* pouchdb-core/src/adapter.js:249: var parts = doc._rev.split('-');
* pouchdb-core/src/adapter.js:599: var splittedRev = doc._rev.split('-');
* pouchdb-core/src/adapter.js:623: const pathId = doc._rev.split('-')[1];
* pouchdb-merge/src/revExists.js:4: var splitRev = rev.split('-');
```

Follow-up to https://github.com/pouchdb/pouchdb/pull/8931

### DIFF
--- a/packages/node_modules/pouchdb-core/src/adapter.js
+++ b/packages/node_modules/pouchdb-core/src/adapter.js
@@ -191,7 +191,7 @@ function attachmentNameError(name) {
 }
 
 function isValidRev(rev) {
-  return typeof rev === 'string' && /^\d+-/.test(rev);
+  return typeof rev === 'string' && /^\d+-[^-]*$/.test(rev);
 }
 
 class AbstractPouchDB extends EventEmitter {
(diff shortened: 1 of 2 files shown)


SourceR85:
I'm curious: Is `"1-"` also a valid revision ID for CouchDB?
Or was the original check just a hack?

SourceR85 (approved):
There's room for improvement of checking the hash (e.g. check for illegal characters),
but that's a story for another PR.

PR #8936, opened 2024-04-22
(#8927) - Set default limit on find queries to 25

### LINKED ISSUE #8927: Inconsistant/undocumented difference between default `limit` on `find` calls when using http vs local adapter
### Issue
When calling the `find` function, the `request.limit` value is optional.  However, if you do not provide a limit value, the amount of data returned will differ based on if you are using the http adapter (connecting to a CouchDB instance) vs a local adapter.  

The [CouchDB docs for `_find`](https://docs.couchdb.org/en/stable/api/database/find.html#post--db-_find) say:

> limit (number) – Maximum number of results returned. Default is 25. Optional

I cannot find any mention of a default limit size in the Pouch docs:
- https://pouchdb.com/api.html#query_index
- https://pouchdb.com/guides/mango-queries.html

My question is: should the `limit` on Pouch `find` behave the same way as Couch `_find`, where, by default, the limit is `25` on both the local and the http adapter?  If not, can we at least update the docs to call out that the default `limit` value may differ based on which adapter you are using?

### Reproduce

You can demonstrate what I am talking about by adding a new test to [test.limit.js](https://github.com/pouchdb/pouchdb/blob/master/tests/find/test-suite-1/test.limit.js)

```js
  it('should consistently limit the results', function () {
    const  extraDocs = Array.from({ length: 3000 }, (_, i) => ({ name: `Test${i}`, rank: 5, series: 'Test', debut: 1981 }));

    var db = context.db;
    return db.bulkDocs(extraDocs).then(function () {
      return db.find({
        selector: {
          series: 'Test'
        },
        fields: ['_id']
      });
    }).then(function (res) {
      res.docs.length.should.equal(3000);
    });
  });
```

This test will succeed when running with the local adapter since all `3000` results will be returned.  It will fail if you instead connect Pouch to a Couch instance via the http adapter. Only 25 results will be returned in that case.

jkuester:
Closes #8927

As @garethbowen [noted](https://github.com/pouchdb/pouchdb/issues/8927#issuecomment-2060641535) this PR should be blocked waiting for the next major release since this is a breaking change.

The only question I had with the implementation is if we should worry about the http adapter being used with _other_ server implementations that might not implement the default limit of `25`?  Currently, this change only affects the behavior of the _local adapter_ and will not set a default limit when the http adapter is used.

### DIFF
--- a/docs/_includes/api/query_index.html
+++ b/docs/_includes/api/query_index.html
@@ -98,7 +98,7 @@ The above is a simple example. For an in-depth tutorial, please refer to
   * `$elemMatch` Matches all documents that contain an array field with at least one element that matches all the specified query criteria.
 * `fields` (Optional) Defines a list of fields that you want to receive. If omitted, you get the full documents.
 * `sort` (Optional) Defines a list of fields defining how you want to sort. Note that sorted fields also have to be selected in the `selector`.
-* `limit` (Optional) Maximum number of documents to return. Default is `25`.
+* `limit` (Optional) Maximum number of documents to return. Default is `25` when connected to a local database.
 * `skip` (Optional) Number of docs to skip before returning.
 * `use_index` (Optional) Set which index to use for the query. It can be "design-doc-name" or "['design-doc-name', 'name']".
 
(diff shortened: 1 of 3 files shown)


garethbowen (approved):
Looks good. Thanks for including the docs update and a unit test too.

garethbowen:
> The only question I had with the implementation is if we should worry about the http adapter being used with other server implementations that might not implement the default limit of 25?

I think it's ok. CouchDB is by far the most common server implementation and you could argue Pouch shouldn't be overriding the server default in the http adapter. It might be worth making this explicit in the docs?

garethbowen:
@jkuester Looks like there's a linting issue in CI.

jkuester:
Whoops!  Should be good now. Also, I added a bit more context to the docs.

garethbowen:
Thanks @jkuester !

PR #8939, opened 2024-04-28
fix: use universally portable shebang

janl:
this will make macOS and the BSDs happier


alxndrsn (approved):
Works on my machine :+1:

PR #8946, opened 2024-04-29
pouchdb-find: make _design/ prefix removal more specific

alxndrsn:
Current code will match `_design/` anywhere.

I suspect it should only be replacing `_design/` when it's a prefix...

### DIFF
--- a/packages/node_modules/pouchdb-find/src/adapters/local/utils.js
+++ b/packages/node_modules/pouchdb-find/src/adapters/local/utils.js
@@ -18,6 +18,7 @@ function massageSort(sort) {
   });
 }
 
+const ddocIdPrefix = /^_design\//;
 function massageUseIndex(useIndex) {
   let cleanedUseIndex = [];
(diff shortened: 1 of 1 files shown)


packages/node_modules/pouchdb-find/src/adapters/local/utils.js
```diff
@@ -27,7 +27,7 @@ function massageUseIndex(useIndex) {
   }
 
   return cleanedUseIndex.map(function (name) {
-    return name.replace('_design/', '');
```
garethbowen, line 30: Consider caching the regex.

garethbowen (changes requested):

packages/node_modules/pouchdb-find/src/adapters/local/utils.js
alxndrsn: Done

garethbowen (approved):

PR #8957, opened 2024-06-12
chore: remove clone-buffer

SourceR85:
This removes "clone-buffer" dependency.

Since `Buffer.from` is stable, there is no need for it.


janl:
small point, but should we maybe replace all instances of this with `Buffer.from()` just to remove one level of indirection

SourceR85:
> small point, but should we maybe replace all instances of this with `Buffer.from()` just to remove one level of indirection

This will break in pouchdb-utils/clone.js:2
cloneBinaryObject.js is import-mapped from cloneBinaryObject-browser.js, 'cause Buffer is node.js only.

janl:
oops, makes sense!

janl (approved):

PR #8959, opened 2024-06-12
chore: remove Set/Map polyfill

SourceR85:
This patch started as a cleanup of Set/Map polyfill.

Since pouchdb-collections only purpose is to provide this polyfill, I've removed it alongside.

The `rollup-plugin-inject` package is in archived state since august 21 and only referenced to link to pouchdb-collections in rollup builds, so it's fine to remove it too.


janl (approved):

janl:
nice work!

PR #9000, opened 2024-10-31
adapter-indexeddb: fix two rewrite bugs

### LINKED ISSUE #9002: adapter-indexeddb: rewrite overrides subvalues with false when they don’t need rewriting.
### Issue

The function `rewrite` from `rewrite.js` in adapter-indexeddb sometimes replaces objects with `false`.

#### Example
The expression
```javascript
rewrite({ foo: { bar: "baz" }, 2: "value"}})
```
returns
```javascript
{ foo: false, _c50_: "value" }
```
expected would be
```javascript
{ foo: { bar: "baz" }, _c50_: "value"}
```

This has the effect that wrong data is written to the `data` field in the indexeddb which can lead to entries missing from indices.

### Info
- Environment: Browser
- Platform: Chrome
- Adapter: indexeddb
- Server: CouchDB

maralorn:
Fixes #9002 and #9003

### DIFF
--- a/packages/node_modules/pouchdb-adapter-indexeddb/src/rewrite.js
+++ b/packages/node_modules/pouchdb-adapter-indexeddb/src/rewrite.js
@@ -101,7 +101,12 @@ function rewrite(data) {
     } else if (typeof data[key] === 'boolean') {
       clone[safeKey] = data[key] ? IDB_TRUE : IDB_FALSE;
     } else if (typeof data[key] === 'object') {
-      clone[safeKey] = rewrite(data[key]);
+      const val = rewrite(data[key]);
+      if (val) {
+        clone[safeKey] = val;
+      } else {
+        clone[safeKey] = data[key];
+      }
     } else {
       clone[safeKey] = data[key];
     }
--- a/packages/node_modules/pouchdb-adapter-indexeddb/src/rewrite.js
+++ b/packages/node_modules/pouchdb-adapter-indexeddb/src/rewrite.js
@@ -77,8 +77,8 @@ function needsRewrite(data) {
       return true;
     } else if (data[key] === null || typeof data[key] === 'boolean') {
       return true;
-    } else if (typeof data[key] === 'object') {
-      return needsRewrite(data[key]);
+    } else if (typeof data[key] === 'object' && needsRewrite(data[key])) {
+      return true;
     }
   }
 }
--- a/tests/find/test-suite-1/test.escaping.js
+++ b/tests/find/test-suite-1/test.escaping.js
@@ -391,4 +391,15 @@ describe('test.escaping.js', function () {
       res.docs.should.deep.equal([ { _id: 'doc', 'foo/bar': -1, foo_c47_bar: 2 } ]);
     });
   });
+
+  it('#9002: query works without rewrite in nested object', async () => {
+    const db = context.db;
+    await db.bulkDocs([{ _id: "issue9002", foo: { bar: "baz" }, 2: "value"}]);
+    await db.createIndex({ index: { "fields": ["foo.bar"] }});
+    const resp = await context.db.find({
+      selector: { "foo.bar": "baz" },
+      fields: ["_id"]
+    });
+    resp.docs.should.deep.equal([{ _id: "issue9002"}]);
+  });
 });
--- a/tests/find/test-suite-1/test.escaping.js
+++ b/tests/find/test-suite-1/test.escaping.js
@@ -402,4 +402,15 @@ describe('test.escaping.js', function () {
     });
     resp.docs.should.deep.equal([{ _id: "issue9002"}]);
   });
+
+  it('#9003 index works with rewrite in later field', async () => {
+    const db = context.db;
+    await db.bulkDocs([ { _id: "issue9003", anobject: { without: "need to rewrite" }, year: { 2024: [] }} ]);
+    await db.createIndex({ index: { "fields": ["year.2024"] }});
+    const resp = await context.db.find({
+      selector: { "year.2024": { $exists: true }},
+      fields: ["_id"]
+    });
+    resp.docs.should.deep.equal([{ _id: "issue9003"}]);
+  });
 });


garethbowen:
@maralorn Thanks for this. Can you create an issue and provide reproduction steps? Also a some unit/integration tests to cover this change would be helpful!

maralorn:
Yeah, I am still debugging a bit more and now think that this is probably not the right fix, but I already have a better one.

maralorn:
I need help writing the test. It should be super easier to write one, the logic is quite trivial. 

But I haven’t been writing javascript in the last 10 years and this whole new module system stuff is beyond me.

I have tried writing a unit test in `tests/unit` which apparently uses CommonJS, but the rewrite.js file I want to test is apparently ES6, so I don’t know how to import it from within the test. The internet suggest I rename the file to rewrite.mjs, but that is clearly not the intended setup. I can’t just import the whole indexeddb-adapter module because that does not export the rewrite function. Writing a whole integration test to reproduce the error here would be even more work.

Please guide me.

maralorn:
Okay, with with renaming rewrite.js to rewrite.mjs I am able to write these tests, which all three fail on master but are fixed by this PR and #9001.

```
describe('test.indexeddb-adapter.js-rewriting', () => {
  it('rewrites key starting with a number correctly', async () => {
    const rewrite = await import('../../packages/node_modules/pouchdb-adapter-indexeddb/src/rewrite.mjs');
    const unsanitized = { year: { 2024: [] } };
    const sanitized = { year: { _c50_024: [] } };
    rewrite.rewrite(unsanitized).should.deep.equal(sanitized);
  })
  it('rewrites correctly with another value which needs no rewrite', async () => {
    const rewrite = await import('../../packages/node_modules/pouchdb-adapter-indexeddb/src/rewrite.mjs');
    const unsanitized = { anobject: { without: "need to rewrite" }, year: { 2024: [] } };
    const sanitized = { anobject: { without: "need to rewrite" }, year: { _c50_024: [] } };
    rewrite.rewrite(unsanitized).should.deep.equal(sanitized);
  })
  it('rewrites correctly when some values need rewrite and others don’t', async () => {
    const rewrite = await import('../../packages/node_modules/pouchdb-adapter-indexeddb/src/rewrite.mjs');
    const unsanitized = { foo: { bar: "baz" }, 2: "value"};
    const sanitized = { foo: { bar: "baz" }, _c50_: "value"};
    rewrite.rewrite(unsanitized).should.deep.equal(sanitized);
  })
});
```
But the plumbing here is obviously wrong, so I will wait for your feedback.

garethbowen:
@maralorn I think the solution is to copy the plumbing, location, and structure of an existing unit test, for example: https://github.com/pouchdb/pouchdb/blob/master/tests/unit/test.memory-adapter.js 

Let me know how you get on!

maralorn:
Ah, well. Copying that file to one next to it is exactly how I arrived at my test.
The problem is that all unit tests I could find test a `PouchDB` object or a some of the functions exposed from `PouchDB.utils` the `rewrite` function I want to test is not exposed like any of that.

garethbowen:
You should just be able to require it directly, eg:

```
const rewrite = require('../../packages/node_modules/pouchdb-adapter-indexeddb/src/rewrite');
rewrite.rewrite(...);
```

Does that not work?

maralorn:
When I do that I get

```

/home/maralorn/git/pouchdb/packages/node_modules/pouchdb-adapter-indexeddb/src/rewrite.js:113
export {
^^^^^^

SyntaxError: Unexpected token 'export'
```

garethbowen:
Of course! Makes sense. I'll update my dev environment and see if I can figure it out...

garethbowen:
Ok I think your best bet is to write an integration test. If you can create a map/reduce index which breaks because of this code then it should be straightforward. Most of the pouchdb tests are integration tests so it works better with the suite as it is.

For example, I think it would fit in this file: https://github.com/pouchdb/pouchdb/blob/master/tests/mapreduce/test.mapreduce.js

maralorn:
Yeah, I feared that. Since I encountered this issue using the find api, I will rather try adding a test to the find tests.

garethbowen:
Fair enough. The underlying implementation is the same so whatever's easiest to reproduce for you will work fine!

maralorn:
Okay. I managed to write an integration test and I have verified that both tests fail on master and every fix fixes one test. Decided to combine both fixes in this PR. I hope that’s alright.

garethbowen:
@alxndrsn Would you mind reviewing this?

tests/find/index.html
```diff
@@ -39,6 +39,7 @@
     <script src='./test-suite-1/test.exists.js'></script>
     <script src='./test-suite-1/test.fields.js'></script>
     <script src='./test-suite-1/test.issue66.js'></script>
+    <script src='./test-suite-1/test.issue9002f.js'></script>
```
alxndrsn, line 42: Why `9002f` instead of `9002`?

maralorn: Because it tests 9002 and 9003. :laughing: I admit its a bit weird.

alxndrsn: :laughing:  is `f` the plural of `2`?

maralorn: I stole it from citations. There you sometimes say `see page 50f` to say its on page 50 and page 51.

alxndrsn: TIL :upside_down_face: 

Would `'./test-suite-1/test.issue9002-9003.js'` be more obvious?

Also sorry for the superficial review - I'll try to look at the actual details tomorrow.

maralorn: Don’t worry. Sure, can do.

packages/node_modules/pouchdb-adapter-indexeddb/src/rewrite.js
```diff
@@ -77,8 +77,8 @@ function needsRewrite(data) {
       return true;
     } else if (data[key] === null || typeof data[key] === 'boolean') {
       return true;
-    } else if (typeof data[key] === 'object') {
-      return needsRewrite(data[key]);
+    } else if (typeof data[key] === 'object' && needsRewrite(data[key])) {
+      return true;
```
alxndrsn, line 81: Did this need to change?  It's not immediately obvious why it would need to.

packages/node_modules/pouchdb-adapter-indexeddb/src/rewrite.js
```diff
@@ -101,7 +101,12 @@ function rewrite(data) {
     } else if (typeof data[key] === 'boolean') {
       clone[safeKey] = data[key] ? IDB_TRUE : IDB_FALSE;
     } else if (typeof data[key] === 'object') {
-      clone[safeKey] = rewrite(data[key]);
+      const val = rewrite(data[key]);
+      if (val) {
+        clone[safeKey] = val;
+      } else {
+        clone[safeKey] = data[key];
+      }
```
alxndrsn, line 109: This code looks surprising.  Would it also work to change the top of the function from

```js
function rewrite(data) {
  if (!needsRewrite(data)) {
    return false;
  } 
  ...
```
to e.g.

```js
function rewrite(data) {
  if (!needsRewrite(data)) {
    return data;
  } 
  ...
```

?

maralorn: That was the first fix I had. But this change seems more defensive because in

https://github.com/pouchdb/pouchdb/blob/c2d05ade1d23ec06328d8e68328de031bc8f456e/packages/node_modules/pouchdb-adapter-indexeddb/src/bulkDocs.js#L200-L206

the code seems to branch depending on whether rewriting happened or not.

packages/node_modules/pouchdb-adapter-indexeddb/src/rewrite.js
maralorn: The problem here is that the return cancels the for loop. That means when `data[key]` is an object the `needsRewrite` will be evaluated for that object, but if there are any more keys in the outer object they will be ignoring, leading to false negatives.

maralorn (commented):
Thank you for the review!

packages/node_modules/pouchdb-adapter-indexeddb/src/rewrite.js
alxndrsn: :bulb:

packages/node_modules/pouchdb-adapter-indexeddb/src/rewrite.js
alxndrsn: Good spot.  Any idea why calling `delete doc.data._attachments;` depends on whether rewriting happens?!

maralorn: I have no idea. All of these troubles seem to come from this commit:

https://github.com/pouchdb/pouchdb/commit/0cd8304e68693a6c47656c98e8aed188ebdb07f5#diff-5869c15c41cf71179c16a81e421d78dafce21dfe37bc47707513bc2809b49c53

Which introduces those difference branches without explaining it. From my limited understanding of this adapter the data attribute is only used for indexeddb views and those views will most likely not operate on attachments. So I guess we could also drop them always, but I really don’t know.

alxndrsn: > I have no idea. All of these troubles seem to come from this commit:

It looks like `_attachments` may have already been rewritten to `_c95_attachments` at that point.  If that's the case, `delete ...._attachments` would be a no-op.  I'll see if that can be cleaned up a bit.

alxndrsn: Let's see if this passes: https://github.com/pouchdb/pouchdb/pull/9010

alxndrsn: #9010 is passing.  If that gets merged, I think 

```js
    if (rewriteEnabled) {
      // doc.data is what we index, so we need to clone and rewrite it, and clean
      // it up for indexability
      const result = rewrite(theDoc);
      if (result) {
        doc.data = result;
        delete doc.data._attachments;
      } else {
        doc.data = theDoc;
      }   
    } else {
      doc.data = theDoc;
    }
```

can eventually become

```js
    if (rewriteEnabled) {
      // doc.data is what we index, so we need to clone and rewrite it, and clean
      // it up for indexability
      doc.data = rewrite(theDoc);
    } else {
      doc.data = theDoc;
    }

```

Anyway, this PR looks good for now.

alxndrsn:
Code looks good to me.  Would it be appropriate to move the tests to `tests/find/test-suite-1/test.escaping.js`?

alxndrsn:
> Decided to combine both fixes in this PR. I hope that’s alright.

Generally I'd discourage this.  Is it just done here for convenience?  Looking at the tests, their setup could easily be separated, which might make each more readable.

maralorn:
> > Decided to combine both fixes in this PR. I hope that’s alright.
> 
> Generally I'd discourage this, as it can make commit history harder to follow. Is it just done here for convenience? Looking at the tests, their setup could easily be separated, which might make each more readable.

Main reason is that I couldn’t write a test that succeeds with the second fix but without the first. What I can do is first introduce test a + fix a and then do a second PR which adds test b + fix b but depends on PR a.

maralorn:
> Code looks good to me. Would it be appropriate to move the tests to `tests/find/test-suite-1/test.escaping.js`?

Sure I was just following the naming scheme I found.

alxndrsn:
> > Code looks good to me. Would it be appropriate to move the tests to `tests/find/test-suite-1/test.escaping.js`?
> 
> Sure I was just following the naming scheme I found.

This was a genuine question :¬)

I _think_ these new tests are in the same area as those in `test.escaping.js`, and it looks like there's already a bug-specific fix in there.

maralorn:
Closed in favor of #9019 and #9020

PR #9001, opened 2024-10-31
adapter-indexeddb: don’t return too early in needsRewrite

### LINKED ISSUE #9003: adapter-indexeddb: needsRewrite can erroneously return undefined
### Issue

`needsRewrite` in adapter-indexeddb `rewrite.js` sometimes erroneously returns `undefined`.

#### Example
```javascript
needsRewrite({year: { 2024: []}})
```
is `true`, because a key can’t begin with a number.
But
```javascript
needsRewrite({anobject: { without: "need to rewrite" }, year: { 2024: []})
```
gives `undefined`. Expected would be `true` because the key `2024` should be rewritten.

Similar to #9002 this can lead to `data` not being correctly sanitized and thus indices to miss entries.

### Info
- Environment: browser
- Platform: Chrome
- Adapter: indexeddb
- Server: CouchDB

maralorn:
Fixes #9003

### DIFF
--- a/packages/node_modules/pouchdb-adapter-indexeddb/src/rewrite.js
+++ b/packages/node_modules/pouchdb-adapter-indexeddb/src/rewrite.js
@@ -77,8 +77,8 @@ function needsRewrite(data) {
       return true;
     } else if (data[key] === null || typeof data[key] === 'boolean') {
       return true;
-    } else if (typeof data[key] === 'object') {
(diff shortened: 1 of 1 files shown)


maralorn:
I decided to do this together with #9002 in #9000 because this bug only becomes easily visible when #9002 is fixed and I wrote tests for both of them together.

PR #9007, opened 2024-11-13
add missing heading for nodejs

bburns:
was missing "PouchDB in Node.js" heading


garethbowen (approved):
Thanks! I think there's also a missing "server" heading but looking at the change history I can't see when this was removed.

This is a step in the right direction so merging as is.

PR #9010, opened 2024-12-04
indexeddb: remove no-op delete of _attachments after rewrite

alxndrsn:
If `rewrite()` had an effect, `doc.data._attachments` will have been rewritten as `doc.data._c95_attachments`, so `delete doc.data._attachments` is a no-op.

### DIFF
--- a/packages/node_modules/pouchdb-adapter-indexeddb/src/bulkDocs.js
+++ b/packages/node_modules/pouchdb-adapter-indexeddb/src/bulkDocs.js
@@ -198,12 +198,7 @@ export default function (api, req, opts, metadata, dbOpts, idbChanges, callback)
       // doc.data is what we index, so we need to clone and rewrite it, and clean
       // it up for indexability
       const result = rewrite(theDoc);
-      if (result) {
-        doc.data = result;
-        delete doc.data._attachments;
-      } else {
-        doc.data = theDoc;
-      }
+      doc.data = result || theDoc;
     } else {
       doc.data = theDoc;
     }


SCdF (commented):
I don't remember why we delete `._attachments`, but if we had a reason shouldn't we instead delete `._c95_attachments`?

Sorry I've lost my memory palace for this code :-)

alxndrsn:
> I don't remember why we delete `._attachments`, but if we had a reason shouldn't we instead delete `._c95_attachments`?

True!

To me it implies that _attachments are not wanted in map/reduce views.

But if that's the case, `._attachments` should probably (also) be deleted if there were no rewrites(???)

SCdF (approved):
So the rewrites are happening on all "main" databases (`const rewriteEnabled = dbOpts.name.indexOf("-mrview-") === -1;
`). The mrview databases shouldn't even exist with the new adapter? So this frames this as a weird decision anyway. It may have been a legacy thing I'm not sure.

I'd probably say take a look at the structure in the dev tools to be a bit clearer. On the presumption you've done that and are happy (and the tests pass so we should be good?), approving.

alxndrsn:
> So the rewrites are happening on all "main" databases

Whoops!  I will review this more thoroughly before deciding on the way forward.  It sounds like one or more of:

* _attachments should never be deleted
* _attachments should be deleted _before_ rewrite
* _c95_attachments should be deleted _after_ rewrite
* `rewriteEnabled` should always be true

alxndrsn:
Closing in favour of https://github.com/pouchdb/pouchdb/pull/9027.

PR #9012, opened 2024-12-04
indexeddb/rewrite: separate sanitise*() functions

alxndrsn:
Also no need to redeclare correctCharacters() as an inner function every time sanitise*() is called.


alxndrsn:
Doesn't work.

PR #9014, opened 2024-12-05
pouchdb-errors: remove no-op call to Error constructor

alxndrsn:
Introduced in cef1ded5712dea0256e8fe4d2c9ffaf48c891651.

The original intention is unclear.

### DIFF
--- a/packages/node_modules/pouchdb-errors/src/index.js
+++ b/packages/node_modules/pouchdb-errors/src/index.js
@@ -1,7 +1,6 @@
 class PouchError extends Error {
   constructor(status, error, reason) {
     super();
-    Error(reason);
     this.status = status;
     this.name = error;
     this.message = reason;


packages/node_modules/pouchdb-errors/src/index.js
```diff
@@ -1,7 +1,6 @@
 class PouchError extends Error {
   constructor(status, error, reason) {
     super();
```
SourceR85, line 3:
```suggestion
    super(reason);
```

SourceR85 (changes requested):
Hi, passing `reason` to parent constructor will initiate `message`.

packages/node_modules/pouchdb-errors/src/index.js
SourceR85, line 3:
```suggestion
    super(reason);
```

packages/node_modules/pouchdb-errors/src/index.js
```diff
@@ -1,7 +1,6 @@
...
-    Error(reason);
     this.status = status;
     this.name = error;
     this.message = reason;
```
SourceR85, line 6:
```suggestion
```

packages/node_modules/pouchdb-errors/src/index.js
SourceR85: Passing reason to parent constructor...

packages/node_modules/pouchdb-errors/src/index.js
SourceR85: ... makes message assignment obsolete ;)

SourceR85 (changes requested):
Hi, passing `reason` to parent constructor will initiate `message`.

alxndrsn:
> Hi, passing `reason` to parent constructor will initiate `message`.

Can you help me understand under what circumstances this is true?  I can't see how it would work.  The following looks analogous to me:

```js
class X extends Error {
  constructor() {
    super();
    Error('hello');
  }
};

new X().message
```

which returns empty string...

SourceR85 (approved):
Ok, I just got more input in Slack chat.
This change will make future changes less noisy (coming soon).

SourceR85:
> Can you help me understand under what circumstances this is true? I can't see how it would work. The following looks analogous to me:
> 
> ```js
> class X extends Error {
>   constructor() {
>     super();
>     Error('hello');
>   }
> };
> 
> new X().message
> ```
> 
> which returns empty string...

Of course,
calling `Error('hello');` will create a new Error instance.
That Error isn't bound to the current `this`-context in constructor, so it's a noop.

The "correct" way to bound "hello" to `this.message` in this context is by calling `super('hello')`.

PR #9016, opened 2024-12-05
Remove greenkeeper references

alxndrsn:
greenkeeper.io shut down 4.5 years ago:

> We’ll stop Greenkeeper’s operation as an independent service on June 3rd, 2020.

See: https://greenkeeper.io/


SourceR85 (approved):
Sure, we now use dependabot as an alternative and nobody removed Greenkeeper.

PR #9017, opened 2024-12-05
Don't ignore package-lock.json

alxndrsn:
> reproducible builds are a huge +1 for us
> - https://github.com/pouchdb/pouchdb/pull/6754#issuecomment-340386723

It looks like these files were gitignored completely randomly in 9bf131092bdc3003ab1682bb17d3ed30be93b30e / #7331.


SourceR85 (approved):
Including lock-files in git-repo's is best practice.
So far, I was wondering why PouchDB didn't include 'em.

PR #9020, opened 2024-12-05
fix(indexeddb) needsRewrite can erroneously return undefined

### LINKED ISSUE #9003: adapter-indexeddb: needsRewrite can erroneously return undefined

maralorn:
Depends on #9019

See discussion in #9000 

Closes #9003

### DIFF
--- a/packages/node_modules/pouchdb-adapter-indexeddb/src/rewrite.js
+++ b/packages/node_modules/pouchdb-adapter-indexeddb/src/rewrite.js
@@ -101,7 +101,12 @@ function rewrite(data) {
     } else if (typeof data[key] === 'boolean') {
       clone[safeKey] = data[key] ? IDB_TRUE : IDB_FALSE;
     } else if (typeof data[key] === 'object') {
-      clone[safeKey] = rewrite(data[key]);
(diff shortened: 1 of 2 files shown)


alxndrsn (approved):

maralorn:
The check looks like it needs another restart …

alxndrsn:
> The check looks like it needs another restart …

Always :shrug: 

See https://github.com/pouchdb/pouchdb/issues?q=intermittent; feel free to file more.

PR #9022, opened 2024-12-09
feat(ci, release): use fixed dependencies

alxndrsn:
* CI should alert when `package-lock.json` is out of sync
* release builds should use the exact dependencies recorded in git so that builds are reproducible

From `npm` docs (https://docs.npmjs.com/cli/v10/commands/npm-ci):

> This command is similar to [`npm install`](https://docs.npmjs.com/cli/v10/commands/npm-install), except it's meant to be used in automated environments such as test platforms, continuous integration, and deployment -- or any situation where you want to make sure you're doing a clean install of your dependencies.

Depends on:

* https://github.com/pouchdb/pouchdb/pull/9017
* https://github.com/pouchdb/pouchdb/pull/9023

### DIFF
--- a/.github/actions/install-node-package/action.yml
+++ b/.github/actions/install-node-package/action.yml
@@ -21,5 +21,5 @@ runs:
         restore-keys: |
           ${{ runner.os }}-node-
 
-    - run: npm install
+    - run: npm clean-install
       shell: bash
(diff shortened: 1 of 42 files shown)


SourceR85 (approved):
Very clean indeed ;)

PR #9027, opened 2024-12-11
fix(indexeddb) don't store _attachments in IndexedDB

alxndrsn:
Attachments do not need to be stored in `doc.data._attachments`, as they are stored in `doc.attachments`, and the user-facing `._attachments` is recreated in `util/processAttachments()` when docs are read from IndexedDB.

However, the current code is deleting the wrong path - as `._attachments` is deleted _after_ `rewrite()` has been called, the property will have moved to `._c95_attachments`.  Instead of hard-coding the key for `._c95_attachments`, it is calculated once when the adapter is first loaded (as `sanitisedAttachmentKey`).


SourceR85 (approved):
Nice work ;)
LGTM

PR #9029, opened 2024-12-13
chore(mapreduce-errors) remove no-op stack regeneration

alxndrsn:
Calling `Error.captureStackTrace()` seems to be an archaic way to add stacktraces to custom Error types from before Javascript supported classes.

Now that these error classes extend `Error`, calling `Error.captureStackTrace(this, thisClass)` is effectively a no-op.

See e.g. https://github.com/pouchdb/pouchdb/blob/f5ae8e02a72a03865b9719bd77429b83c09b069d/node_modules/nano/node_modules/errs/README.md#user-content-reusing-types


SourceR85 (approved):
Less code for improved quality.
I like it.

PR #9031, opened 2024-12-16
feat(npm) disable command overhead

alxndrsn:


SourceR85:
IMO: at least, the audit reports are useful.

Did you try to reduce the noise in CI?
In that case, I would suggest: creating the .npmrc in CI on the fly.

alxndrsn:
I've split off a separate PR to remove the `fund` option: https://github.com/pouchdb/pouchdb/pull/9037

I'll look more into `audit` at another time.

PR #9034, opened 2025-01-04
fix(ci): make shellcheck happy

SourceR85:
This patch makes shellcheck pass again


SourceR85:
https://github.com/pouchdb/pouchdb/pull/9033 includes shellcheck fix

PR #9036, opened 2025-01-04
ci!: drop node 18

SourceR85:
Since v18 is EOL in April 25 anyway, I guess: now is the right time to drop it.

BREAKING CHANGE: drop support for Node.js v18

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,7 +69,7 @@ jobs:
       fail-fast: false
       matrix:
         couchdb: ['2.3', '3.1']
-        node: [20, 22]
+        node: [20]
         cmd:
           - npm test
           - TYPE=find PLUGINS=pouchdb-find ADAPTERS=http npm test
@@ -158,7 +158,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node: [20, 22]
+        node: [20]
         adapter: ['leveldb', 'memory']
         cmd:
           - npm test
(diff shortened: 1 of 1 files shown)


alxndrsn:
Looks like the node 22 upgrade is complicated; maybe the 2 changes can be separated?

SourceR85:
> Looks like the node 22 addition is non-trivial; maybe the 2 changes can be separated?

Sure, for v22 I'll add a separate PR when this is merged.

SourceR85:
> > v18 is EOL in April 25
> 
> Should this be merged at the end of April? Then any versions released before then will still have been testing against v18.

Can be merged anytime (now would be the best time for it).

Let me try to convince you:
At first, it enables us to use some modern API's like [URL](https://github.com/pouchdb/pouchdb/pull/9039), or make our life easier with modern syntax (for await ... of).

This patch will only affect the next release: PouchDB 10.0.0

It will be a major release and we have the freedom to do breaking changes as we've did it with Node 16 in PouchDB 9, like for example: drop support for Node 18 (very convenient 😉).
No one is forced to update to PouchDB 10.

PouchDB 9 is stable and nobody complains about security bugs (so far).
Iff someone find a vulnerability before 18 goes EOL, we can branch from https://github.com/pouchdb/pouchdb/tree/9.0.0 at any time to patch up a version 9.0.1.

Also: I expect that everyone has updated there Node runtime by now (or at least, they are working hard on the transition).

alxndrsn:
> Also: I expect that everyone has updated there Node runtime by now (or at least, they are working hard on the transition).

This is a very generous assumption :grin:

alxndrsn (approved):

PR #9039, opened 2025-01-07
 chore(adapter-http): remove custom URL parsing

alxndrsn:
Replace `pouch-utils/parseUri()` with WHATWG `URL.parse()`.

See: https://developer.mozilla.org/en-US/docs/Web/API/URL/parse_static


alxndrsn:
Does not work on node 18, so blocked by e.g. https://github.com/pouchdb/pouchdb/pull/9036

PR #9040, opened 2025-01-07
chore(test) choose suite with consistent env var

alxndrsn:
There are currently two different env vars which control the test suite to run: `TYPE`, which is used for integration, fuzzy, mapreduce etc., and `PERF`, which is used solely for the performance test suite.  This PR removes the additional `PERF` env var, replacing it with `TYPE=performance`.


SourceR85:
We need to update the CI check, before we can merge (`PERF=1 npm test` waiting to be executed)

PR #9042, opened 2025-01-08
docs: fix CSS live rebuild

alxndrsn:
Due to `process.chdir('docs')` call on line 14, the less path should be relative to `./docs/` instead of the project root.


SourceR85 (approved):
👍

PR #9043, opened 2025-01-08
docs: dockerise dev

alxndrsn:
Setting up Ruby v2.7 is getting increasingly difficult on newer systems.

This PR aims to help with that by providing a simple way to launch a documentation dev environment.


SourceR85:
Just my thought: creating a compose file (e.g. `docker-compose.docs.yml`) for `docker compose up` / `docker compose watch` might be a more obvious solution than yet another npm command.

docs-dev.Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+RUN . "$NVM_DIR/nvm.sh" && nvm install 22
+ENV PATH="$NVM_DIR/versions/node/v${NODE_VERSION}/bin/:${PATH}"
+
+WORKDIR /pouchdb-docs
+
+COPY package.json package-lock.json .
+COPY ./bin/ ./bin/
+
+COPY ./docs/ ./docs/
+RUN npm run install-jekyll
```
alxndrsn, line 28: replace with direct call to `bundler`, and move earlier to reduce layer changes

alxndrsn:
> `docker compose up` / `docker compose watch` might be a more obvious solution than yet another npm command

OTOH it may be more obvious and discoverable to keep all command definitions in the same place.

PR #9044, opened 2025-01-08
docs: fix watch mode for JS

### LINKED ISSUE #9041: docs build-site.sh maxes out CPU in continuous mode
This seems unnecessary.

alxndrsn:
* remove `buildJekyll()` path check, because `path` is not the argument that `watchGlob()` passes to the listener
* update watched paths for `buildJekyll()` to only listen for jekyll-relevant paths
* use async watcher functions so that promises are returned

Closes #9041

### DIFF
--- a/bin/build-site.js
+++ b/bin/build-site.js
@@ -25,7 +25,7 @@ async function checkJekyll() {
   }
 }
 
-function buildCSS() {
+async function buildCSS() {
   fs.mkdirSync(__dirname + '/../docs/static/css', { recursive:true });
   const cmd = [ resolvePath('node_modules/less/bin/lessc'), POUCHDB_LESS ].join(' ');
   const { stdout } = await exec(cmd);


bin/build-site.js
```diff
@@ -73,8 +73,15 @@ if (!process.env.BUILD) {
   const http_server = require('http-server');
   const watchGlob = require('glob-watcher');
 
-  watchGlob('**', buildJekyll);
+  watchGlob('!(_site)/**', buildJekyll);
```
alxndrsn, line 76: This should already be handled by:

https://github.com/pouchdb/pouchdb/blob/8f388436e1d0d2f9e4d7bb18c82a9d54864b42d6/bin/build-site.js#L33-L36

It's probably worth understanding why that isn't good enough.

alxndrsn: ...because the arg passed to glob watchers is not the path of the changed file

bin/build-site.js
```diff
@@ -73,8 +69,15 @@ if (!process.env.BUILD) {
...
```
alxndrsn, line 72: this glob doesn't expand as expected

SourceR85:
Can you resolve that conflict in build-site?

SourceR85 (approved):
LGTM

PR #9051, opened 2025-01-09
fix(docs): minify code.js automatically

alxndrsn:
Previously minification seemed to happen "offline", and both source and minified file were tracked in git.

This also prevents the unused source file `code.js` from being served at https://pouchdb.com/static/js/code.js

### TODO

- [x] set ascii-only mode
- [x] set module-level mode


alxndrsn:
`resolvePath()` also introduced in https://github.com/pouchdb/pouchdb/pull/9050/files

SourceR85 (approved):
Sure, JS minification has not much to do with CSS 👍

PR #9052, opened 2025-01-09
fix(docs): don't include less source files in served site

alxndrsn:
For example: http://pouchdb.com/static/less/pouchdb/alerts.less


SourceR85 (approved):
👍

PR #9054, opened 2025-01-09
chore: remove dependency child-process-promise

alxndrsn:
This dependency is ancient, and can be implement easily with the standard library.


SourceR85 (approved):
👍

PR #9055, opened 2025-01-09
chore: move dev-only dependency ltgt to devDependencies

alxndrsn:


alxndrsn:
Used in `sublevel-pouchdb`.

PR #9086, opened 2025-05-26
Modernize example code

tomayac:
This was triggered by seeing a typo on an offline paper flyer. The typo was since corrected, but the code could still be modernized `:-)`.


SourceR85 (approved):
LGTM

PR #9092, opened 2025-06-11
pouchdb-utils: fix nextTick "Illegal invocation" error

### LINKED ISSUE #9091: pouchdb-utils: "TypeError: Invalid invocation" when calling nextTick in Electron renderer process
### Issue
Instantiating a new PouchDB instance with pouchdb v9.0.0 using the leveldb adapter in an Electron renderer process results in "TypeError: Illegal invocation".  The bug appears to have been introduced in #8849 .  It works fine with pouchdb v8.0.1 and earlier.

```
/tmp/electron-fiddle-1000601-xogHr57gkhkn/node_modules/pouchdb-adapter-leveldb-core/lib/index.js:276
            pouchdbUtils.nextTick(function () {
                         ^

TypeError: Illegal invocation
    at /tmp/electron-fiddle-1000601-xogHr57gkhkn/node_modules/pouchdb-adapter-leveldb-core/lib/index.js:276:26
    at /tmp/electron-fiddle-1000601-xogHr57gkhkn/node_modules/sublevel-pouchdb/lib/index.js:198:7
```

### Info
- Environment: Electron 35.0.0 renderer process with nodeIntegration enabled
- Platform: Node.js 22.14.0, Chromium 134.0.6998.44
- Adapter: leveldb
- Server: CouchDB

### Reproduce
Reproducible using the following electron-fiddle: https://gist.github.com/53193f76b88eaa2cb2dd149df7039ae2
- Download the [electron-fiddle](https://www.electronjs.org/fiddle) app
- Enter the above gist URL in the electron-fiddle address bar and load the fiddle
Image

jslippy:
When instantiating a new PouchDB instance using the leveldb adapter in an Electron renderer process, `nextTick()` is called with a different context than is expected, resulting in `TypeError: Illegal invocation`. Fix this by binding `queueMicrotask` to the `globalThis` context before returning it.

Fixes #9091

### DIFF
--- a/packages/node_modules/pouchdb-utils/src/nextTick.js
+++ b/packages/node_modules/pouchdb-utils/src/nextTick.js
@@ -1,5 +1,5 @@
 const nextTick = typeof queueMicrotask === "function"
-  ? queueMicrotask.bind(globalThis)
+  ? queueMicrotask.bind(undefined)
   : function nextTick(fn) {
     Promise.resolve().then(fn);
   };
--- a/packages/node_modules/pouchdb-utils/src/nextTick.js
+++ b/packages/node_modules/pouchdb-utils/src/nextTick.js
@@ -1,5 +1,5 @@
 const nextTick = typeof queueMicrotask === "function"
-  ? queueMicrotask
+  ? queueMicrotask.bind(globalThis)
   : function nextTick(fn) {
     Promise.resolve().then(fn);
   };


SourceR85:
Thanks for your contribution.

Unfortunately eslint is not up to date and complains about undefined `globalThis`.

My suggestion, for now, would be to bind qm to `this` (without a function context, it points to the global scope: `global` in nodejs, `window` / `self` in browser context).
Please test it in electron context.

jslippy:
When I bind `queueMicrotask` to `this`, it gets converted to `queueMicrotask.bind(undefined)` during build.  Somewhat surprisingly that still works in Electron, even though the files are generated in strict mode.  The build output contains several warnings stating `The 'this' keyword is equivalent to 'undefined' at the top level of an ES module, and has been rewritten`.  

Instead of binding `this`, why not just bind `undefined` to begin with?  It seems to fix the problem and works in various contexts.  Otherwise, we could add some code like the example [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis#search_for_the_global_across_environments) to identify and bind the correct global `this` or just import a [polyfill/shim](https://github.com/es-shims/globalThis) to do it for us, but I assume you want to avoid adding new dependencies.

SourceR85 (approved):
LGTM

PR #9093, opened 2025-06-16
Fix: Ensure metadata.seq >= max(doc.seq) to prevent concurrency bug in IndexedDB adapter

Nuu-maan:
This PR fixes a concurrency bug in the IndexedDB adapter where parallel `bulkDocs()` transactions could result in `metadata.seq` being lower than the highest `doc.seq`, breaking replication logic.

### Fix
Tracks `maxSeqThisTxn` during the transaction and ensures `metadata.seq = Math.max(metadata.seq, maxSeqThisTxn)` before writing metadata.

### Test
Added integration test to `test.bulk_docs.js`:
- Runs 2 parallel `bulkDocs()`
- Verifies all `doc.seq` values are unique and strictly increasing
- Verifies `metadata.seq >= max(doc.seq)`

Fix is minimal, scoped, and backward-compatible.

### DIFF
--- a/packages/node_modules/pouchdb-adapter-indexeddb/src/bulkDocs.js
+++ b/packages/node_modules/pouchdb-adapter-indexeddb/src/bulkDocs.js
@@ -38,6 +38,8 @@ export default function (api, req, opts, metadata, dbOpts, idbChanges, callback)
   const rewriteEnabled = dbOpts.name.indexOf("-mrview-") === -1;
   const autoCompaction = dbOpts.auto_compaction;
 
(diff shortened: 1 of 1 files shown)


alxndrsn:
> Added integration test to `test.bulk_docs.js`:

This file doesn't seem to have changed in the PR.

alxndrsn:
I think this is attempting to fix the same bug as https://github.com/pouchdb/pouchdb/pull/8495

If that's correct, then my comment at https://github.com/pouchdb/pouchdb/pull/8495#issuecomment-2919238345 is probably relevant.

Issue #9104, opened 2025-08-03
builtin _stats reducer cannot digest arrays of numbers


rednil:
The builtin _sum reducer properly handles numbers AND arrays of numbers, returning an array of sums.
In couchdb, this is also true for the _stats reducer.
In pouchdb 9.0.0, if the _stats reducer is confronted with values containing arrays of numbers, it returns something like

```
{
 sum: [17, 10, 4, 5],
 count: 3,
 min: NaN,
 max: NaN,
 sumsqr: NaN
}
```

because the functions for min, max and sumsqr are not implemented to take care of arrays.
I fixed it and will open a pull request if I don't hear otherwise.

PR #9108, opened 2025-08-04
feat(pouchdb-mapreduce): (#9104) - enable _stats reducer to digest arrays of numbers

rednil:
The builtin _sum reducer properly handles numbers AND arrays of numbers, returning an array of sums.
In couchdb, this is also true for the _stats reducer, which returns an array of stats objects.

### DIFF
--- a/packages/node_modules/pouchdb-mapreduce/src/stats.js
+++ b/packages/node_modules/pouchdb-mapreduce/src/stats.js
@@ -16,9 +16,9 @@ function getMixedLengthError() {
 }
 
 function sumsqr(values) {
-  var _sumsqr = 0;
-  for (var i = 0, len = values.length; i < len; i++) {
-    var num = values[i];
+  let _sumsqr = 0;
+  for (let i = 0, len = values.length; i < len; i++) {
+    const num = values[i];
     _sumsqr += (num * num);
   }
   return _sumsqr;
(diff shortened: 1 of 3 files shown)


tests/mapreduce/test.mapreduce.js
```diff
@@ -1084,6 +1108,42 @@ function tests(suiteName, dbName, dbType, viewType) {
...
+            { _id: '1', val: [1, 2, 3] },
+            { _id: '2', val: 4 }
+          ]
+        }).then(function () {
+          return db.query(queryFun, {reduce: true, group_level: 999});
+        });
+      }).should.be.rejected;
+    });
+
+		it("Built in _stats reduce function should throw an error when confronted with arrays of inconsistent length",
```
SourceR85, line 1129: Please be consistent with tabs and spaces

packages/node_modules/pouchdb-mapreduce/src/stats.js
SourceR85, line 1: Please don't mix tab and space indentations.

tests/mapreduce/test.mapreduce.js
SourceR85, line 1: In test.mapreduce.js we use space indentations, please follow

SourceR85 (changes requested):

tests/mapreduce/test.mapreduce.js
SourceR85, line 1: We use const and let for new code.
Would you please update the db assignment?
(use const, because db never changes)

SourceR85:
Well, that doesn't run either...

`npm run eslint` will show you some required formattings, that must be followed.

Don't worry, sometimes even I forget to run the linter ;)

rednil:
bare with me, I will get it ;-)

SourceR85 (approved):
LGTM

Thank you for your contribution 👍

PR #9110, opened 2025-08-20
feat(pouchdb-changes-filter): filter `_design/_design`

janl:
~The error is likley in p-s, but this is a faster fix.~

This is actually a feature addition :)

https://docs.couchdb.org/en/stable/api/database/changes.html#design

### DIFF
--- a/packages/node_modules/pouchdb-changes-filter/package-lock.json
+++ b/packages/node_modules/pouchdb-changes-filter/package-lock.json
@@ -1,5 +1,13 @@
 {
   "name": "pouchdb-changes-filter",
   "version": "7.0.0-prerelease",
-  "lockfileVersion": 1
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "pouchdb-changes-filter",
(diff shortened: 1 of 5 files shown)


SourceR85:
Hi,
Can you provide a test for better understanding of that misbehavior?
(CI currently didn't break, because this filter is untested)

Btw., what does that _design/_design filter means in CouchDB? Can't find docs for that.

janl:
> Hi, Can you provide a test for better understanding of that misbehavior? (CI currently didn't break, because this filter is untested)
> 
> Btw., what does that _design/_design filter means in CouchDB? Can't find docs for that.

Done, this is actually a feature addition to get parity with CouchDB

SourceR85:
I've added a test for this.

Because PouchDB normalizes the filter from `"${filter}"` to `"${filter}/${filter}"` if `filter` didn't contain `"/"`:
CouchDB complains with "unknown builtin filter name",

My last commit excludes "_design" from that normalization.

SourceR85 (approved):
LGTM

janl:
nice work, thanks a lot!

PR #9121, opened 2025-10-05
Fix test.issue915.js on a clean environment

### LINKED ISSUE #9120: npm run test fails on test.issue915.js after cloning
**Steps to reproduce:**
1. `git clone https://github.com/pouchdb/pouchdb.git`
2. `cd pouchdb`
3. `npm install`
4. `npm run test`

**Expected outcome:**
Test suite should pass

**Actual outcome**
Test suite fails:
```
  test.issue915.js
    1) Put a file in the db, then destroy it


  1177 passing (18s)
  14 pending
  1 failing

  1) test.issue915.js
       Put a file in the db, then destroy it:
     Error: ENOENT: no such file or directory, open './tmp/_pouch_veryimportantfiles/something'
```

mureinik:
The test attempts to create the file `./tmp/_pouch_veryimportantfiles/something`, and by doing so implicitly assumes that the directory `./tmp/_pouch_veryimportantfiles` already exists, which may not be true in a clean environment.

This patch creates the directory with `{recursive: true}`, meaning that if the directory already exists, it will just succeed silently without doing anything, and if it doesn't, the directory will be created.

Fixes #9120

### DIFF
--- a/tests/integration/test.issue915.js
+++ b/tests/integration/test.issue915.js
@@ -15,22 +15,24 @@ if (!process.env.LEVEL_ADAPTER &&
     });
     it('Put a file in the db, then destroy it', function (done) {
       var db = new PouchDB('veryimportantfiles');
-      fs.writeFile('./tmp/_pouch_veryimportantfiles/something',
-        Buffer.from('lalala', 'utf8'), function () {
-        db.destroy(function (err) {
-          if (err) {
-            return done(err);
-          }
-          fs.readFile('./tmp/_pouch_veryimportantfiles/something',
-                      {encoding: 'utf8'}, function (err, resp) {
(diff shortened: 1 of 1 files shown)


mureinik:
A note to the reviewers: With GitHub's default display, this PR looks much larger than it really is. If you apply the "Hide whitespace" options (or `-w` in the command line), you'll see that all this PR really does is add an `fs.mkdir` call. The rest of the diff is just indenting everything under `mkdir`'s callback.

mureinik:
This PR is based off a false assumption, see https://github.com/pouchdb/pouchdb/issues/9120#issuecomment-3369352862.

Closing.
Sorry for the noise.

PR #9141, opened 2025-11-20
Add browser shim for `node:events` module

### LINKED ISSUE #8968: Dynamic require of "events" is not supported
### Issue
PouchDB cannot be declared when I use it under vue3+vite+ts, columns such as const db = new PouchDB("test"); An error will occur:

Dynamic require of "events" is not supported.

I have installed the npm package events, do you support it in such an environment
### Info
- Environment: (Node.js/browser/hybrid app/etc.)
- Vue3+vite+ts
- Platform: (Chrome/FF/Safari/Edge/iOS/Android/etc.)
- Chrome
- Adapter: (idb/indexeddb/memory/leveldb/etc.)
- Server: (CouchDB/Cloudant/PouchDB Server/etc.)
- CouchDB

### LINKED ISSUE #8985: `pouchdb-utils` depends on `events` npm package, but does not declare it as a dependency.
### Issue

Bundle `pouchdb-utils` using a typical bundler. For example, bundle using `esbuild` with the following commands:

```shell
mkdir -p /tmp/dependency-repro && cd /tmp/dependency-repro
npm init -y
npm install pouchdb-utils esbuild
echo 'export * from "pouchdb-utils";' | npx esbuild --format=esm --target=es6 --bundle --minify
```

This results in:

```
✘ [ERROR] Could not resolve "events"

    node_modules/pouchdb-utils/lib/index-browser.es.js:2:25:
      2 │ import EventEmitter from 'events';
        ╵                          ~~~~~~~~

  The package "events" wasn't found on the file system but is built into node.
  Are you trying to bundle for node? You can use "--platform=node" to do that,
  which will remove this error.

1 error
```

It seems the browser build of `pouchdb-utils` expects the package `events` from `npm` to be installed as a polyfill for `node`'s built-in `node:events` module. This is not good practice, and I think `pouchdb-utils` should explicitly declare this dependency so that its browser build has all required depencies.

(It may also be wise to conditionally load the polyfill only when `node:events` is not available, but that isn't necessary for this issue.)

### Info

- Environment: macOS 14.6.1
- Platform: macOS 14.6.1
- Adapter: N/A
- Server: N/A

### Reproduce

See above

### LINKED ISSUE #8989: Using pouchdb-browser (+ pouchdb-adapter-memory) expects nodeJS APIs
### Info
- Environment: Browser, ESM, vite
- Platform: Chrome/FF
- Adapter: (pouchdb-adapter-memory + default idb?)
- Server: (vite)

### Issue
I have a project and want to use PouchDB initially only in the browser with in-memory DB persisting to IndexedDB. As my starting point for figuring pouchDB out, introduce servers later.

Using `pouchdb-browser`, I've been getting the same issue from #8641 that was marked resolved in #8849: `Uncaught TypeError: import_events.default is not a constructor`

I saw that `index.es.js` has `import EE from 'events';` and as `events` is commonJS I swapped it out for `eventemitter3` and made a vite alias which sorted that problem out. I mention this as it may be a regression based on the issue + PR mentioned.

However now I'm seeing this familiar error: `Uncaught ReferenceError: process is not defined`, which I was not expecting given that I want to use this in the browser and that `pouchdb-browser` and `pouchdb-adapter-memory` both say they work in the browser.

I couldn't find anything in the documentation about this, so is this a bug or is the 'works in the browser' part of pouchdb-adapter-memory conditional on polyfilling a load of Node APIs in the browser in your bundler?

### LINKED ISSUE #9072: Support observable for easy integration in sveltekit/react/vuejs/…
The popularity of React and similar tools (svelte, vuejs…) show how one can elegantly code with reactive data, i.e. changing a variable automatically update the DOM accordingly… and I'd love to do that with PouchDb. Alternatives like Dexie.js already provide such things like:
```
<script>
  import { liveQuery } from "dexie";
  import { db } from "./db";

  let friends = liveQuery(() => db.friends.toArray());

</script>
<ul>
  {#if $friends}
    {#each $friends as friend (friend.id)}
      <li>{friend.name}, {friend.age}</li>
    {/each}
  {/if}
</ul>
```
and unfortunately I can't find such elegant one-line solution in PouchDb. The best alternative that was proposed to me involves a combinaison of `db.find` to first populate `$friends` followed by `db.changes` and some complex code to decide what to do based on the change… but this is not only really verbose (like 26 lines instead of 1), but also not very resilient, for instance the insertion of the new item is not following the same order as the `db.find` list, let alone when using custom sorting options etc.

It would be great to provide instead a simple observable (now observable are well standardize and quite framework-independent), so that I can use them for instance in sveltekit.

### LINKED ISSUE #9078: new PouchDb. TypeError: The superclass is not a constructor.
Hi, cant start with a pouchDb. After installation pouch and types, i cant use it:

```ts
import * as PouchDB from "pouchdb";
const db = new PouchDB("appState"); // TypeError: The superclass is not a constructor.
```

TS, Vite

alxndrsn:
# TODO

- [ ] add a test which fails against existing codebase similar to how linked issues fail, and passes with the updates in this PR
- [ ] add a test for node-js builds to show that `import 'pouchdb-events'` === `import 'node:events'`?

-----

This seems to be a common issue:

* https://github.com/apache/pouchdb/issues/8968
* https://github.com/apache/pouchdb/issues/8985
* https://github.com/apache/pouchdb/issues/8989
* https://github.com/apache/pouchdb/issues/9072
* https://github.com/apache/pouchdb/issues/9078

I think it was introduced by 404346bba500c357c0276ad15762e985fb143eea, which upgrade from webpack 1 to webpack 5.  From the webpack docs:

> Webpack 5 no longer polyfills Node.js core modules automatically which means if you use them in your code running in browsers or alike, you will have to install compatible modules from npm and include them yourself.
> - https://webpack.js.org/configuration/resolve/

Alternatively, it may come from rollup.

Hopefully:

Closes https://github.com/apache/pouchdb/issues/8968
Closes https://github.com/apache/pouchdb/issues/8985
Closes https://github.com/apache/pouchdb/issues/8989
Closes https://github.com/apache/pouchdb/issues/9072
Closes https://github.com/apache/pouchdb/issues/9078


alxndrsn:
~Less convinced now:~


PR has been updated to handle `events` dependency like other browser-specific polyfills via new `pouchdb-events` package.

```
> e1 = require('node:events'); e2 = require('./node_modules/events');
<ref *1> [Function: EventEmitter] {
  once: [Function: once],
  EventEmitter: [Circular *1],
  defaultMaxListeners: [Getter/Setter],
  init: [Function (anonymous)],
  listenerCount: [Function (anonymous)]
}
> e1
<ref *1> [Function: EventEmitter] {
  addAbortListener: [Function: addAbortListener],
  once: [AsyncFunction: once],
  on: [Function: on],
  getEventListeners: [Function: getEventListeners],
  getMaxListeners: [Function: getMaxListeners],
  EventEmitter: [Circular *1],
  usingDomains: true,
  captureRejectionSymbol: Symbol(nodejs.rejection),
  captureRejections: [Getter/Setter],
  EventEmitterAsyncResource: [Getter],
  errorMonitor: Symbol(events.errorMonitor),
  defaultMaxListeners: [Getter/Setter],
  setMaxListeners: [Function (anonymous)],
  init: [Function (anonymous)],
  listenerCount: [Function (anonymous)]
}
> e2
<ref *1> [Function: EventEmitter] {
  once: [Function: once],
  EventEmitter: [Circular *1],
  defaultMaxListeners: [Getter/Setter],
  init: [Function (anonymous)],
  listenerCount: [Function (anonymous)]
}
> e1.EventEmitter === e2.EventEmitter
false
```

espy:
Hello! 👋

We’ve extensively looked at your PR and manually tested it, and it turned out that, surprisingly, `events` is still not included as a dependency in `npm i pouchdb-browser`. We eventually found that the `events` dependency is _actively removed_ by PouchDB’s build process, because it has the same name as a node.js builtin (via `builtin-modules`), even if it is explicitly a dependency in `package.json` and not just a reference to the builtin.

You can test this via `npm run test-webpack` (which composes the individual modules’ `dependency` field in their `package.json`s), and inspect the files in `packages/node_modules/pouchdb-browser`:
- the `lib/index*` files do some form of `import events from 'events';`, so all imports of `pouchdb-events` in the various modules have been transformed into imports for `events`. However:
- the generated `package.json` depends on _neither_ `events` _nor_ `pouchdb-events`, so the `events` package is still not present in the browser.

We came up with a very hardcoded fix, left for your consideration in [this branch](https://github.com/neighbourhoodie/pouchdb/commits/add-explicit-events-dep-review-updates/). This allows `events` to be added to the `dependency` field in `package.json` files, effectively that means it gets added to `pouchdb-events` and `pouchdb-browser`. After some additional testing, we found that node will prefer its builtin, even if `events` is a proper dependency, so this approach seems fine to us. To test this specific behaviour:

```sh
mkdir tmp
cd tmp
mkdir -p node_modules/events
# edit node_modules/events/index.js to print/return something
# edit index.js to contain `const e = require('events') ; console.log(e)`
node index.js
``` 

This is also true for `index.mjs` and `import e from 'events'`.

To test that the fix itself actually works and `events` is now installed with `npm i pouchdb-browser` (`npm pack` is the closest equivalent for local testing, `npm i <FILEPATH>` and `npm link` both don’t do anything useful here):

```sh
# in your pouchdb repo with our change:
$ npm run test-webpack
$ cd packages/node_modules/pouchdb-browser
$ npm pack
# copy resulting tarball to some other empty directory and cd there
$ npm init # and yes all the things
$ npm i < TARBALL_FILENAME >
# inspect the `node_modules` folder, `events` is present
```

Aside from the fix, the branch also contains another commit with a small cleanup change.

We hope this helps to land this fix in the next release! Thanks for your continued work on PouchDB 🙏

alxndrsn:
Thanks @espy, I'll see if that can be turned into a test case for this PR.  Feel free to submit your own if you're motivated though - I think the repro and automated test case are likely much more significant work than the code changes that will result from the test case and fix the issue.

PR #9176, opened 2026-02-19
refactor mapreduce tests with async/await

emilygilberts:
**tests/unit**
- rewrote the unit tests for `mapreduce.js-upsert` using `async/await`
(left the `mapreduce.js-utils` tests as they are, because they are more readable & straightforward with the current syntax)

**tests/mapreduce (in process)**
- started with some ES6 refactoring of the tests by replacing var with const/let
- rewrote promise chains to async/await
- used arrow functions
- rewrote mocha chai should syntax: (before: `should.become ` to `should.deep.equal` for objects or `should.be.true` for boolean)
- replaced `.toString()` with template literals (except on view map function declarations)

**tests/mapreduce/test.mapreduce.js**
in test _Query_ _result_ _should_ _include_ __conflicts_ and _Views should include_ _conflicts_ :
- replaced the use of `testUtils.fin()` with `try/finally block`
- replaced the use of `promisify` with simply awaiting the replication - so that `testUtils.promisify(remote, db.replicate);` becomes  `await db.replicate.from(remote);`

in test _Test view querying with custom reduce function_:
- in custom reduce function: replaced nested unrolling of values with `values.flat()`
- removed unneccesary line pulling out `keyId[0]` 
- ? can the comment  `//const id = keyId[1]; ` be deleted? 

in test _should query correctly after replicating and other ddoc_:
- moved cleanup (`db2.destroy`) that was in `catch` block and after to `finally` block

in the two tests _should handle many doc changes_:
- replaced the whole `nextTask()` logic to for loop and using `async/await`

### DIFF
--- a/tests/mapreduce/test.mapreduce.js
+++ b/tests/mapreduce/test.mapreduce.js
@@ -1,11 +1,11 @@
 /* global sum */
 'use strict';
 
-var viewTypes = ['persisted', 'temp'];
+const viewTypes = ['persisted', 'temp'];
 viewTypes.forEach(function (viewType) {
-  var suiteName = 'test.mapreduce.js-' + viewType;
-  var adapter = testUtils.adapterType();
-  var dbName = testUtils.adapterUrl(adapter, 'testdb');
+  const suiteName = 'test.mapreduce.js-' + viewType;
+  const adapter = testUtils.adapterType();
+  const dbName = testUtils.adapterUrl(adapter, 'testdb');
 
   tests(suiteName, dbName, adapter, viewType);
 });
@@ -13,10 +13,10 @@ viewTypes.forEach(function (viewType) {
 function tests(suiteName, dbName, dbType, viewType) {
 
   describe(suiteName, function () {
-    var createView;
+    let createView;
     if (dbType === 'http' || viewType === 'persisted') {
       createView = function (db, viewObj) {
-        var storableViewObj = {
+        const storableViewObj = {
           map: viewObj.map.toString()
         };
         if (viewObj.reduce) {
@@ -63,7 +63,7 @@ function tests(suiteName, dbName, dbType, viewType) {
     });
 
     it("Test basic view", function () {
-      var db = new PouchDB(dbName);
+      const db = new PouchDB(dbName);
       return createView(db, {
         map: function (doc) {
           emit(doc.foo, doc);
@@ -94,7 +94,7 @@ function tests(suiteName, dbName, dbType, viewType) {
     });
 
     it("Test basic view, no emitted value", function () {
-      var db = new PouchDB(dbName);
+      const db = new PouchDB(dbName);
       return createView(db, {
         map: function (doc) {
           emit(doc.foo);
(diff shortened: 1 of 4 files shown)


janl (commented):
this looks good overall, I’m just wondering if it would make sense to DRY out the `get` and `put` functions a little bit, it’s three lines each when the only difference is the what the promise resolves with.

Spitballing here:

instead of

```js
   const res = await upsert(
      {
        get: function () {
          return Promise.reject({ status: 404 });
        },
        put: function () {
          return Promise.resolve({ rev: 'abc' });
        },
      },
      'foo',
      function () {
        return { difference: "something" };
      },
    );
```

```js
function makeFun (action, value) {
  return function() { return Promise[action](value) };
}

// [...]

    const res = await upsert(
      {
        get: makeFun('resolve', {status: 404}),
        put: makeFun('reject', { rev: 'abc' }),
      },
      'foo',
      function () { return { difference: "something" };
      },
    );
```

Might be too fancy, feel free to shoot me down :)

emilygilberts:
@janl hmm how about using arrow functions for the functions in get and put? I think it could look neat/be readable cause they are all one liners.

emilygilberts:
UPDATE: ahhh my local CouchDB wasn't properly connected ... nvm :)

I'm running the tests locally with `TYPE=mapreduce npm run test-node`. Any idea why they all pass for `test.mapreduce.js-persisted`, but for `test.mapreduce.js-temp` the first test executed jumps out with a timeout error? 

Bildschirmfoto 2026-03-25 um 14 51 41

emilygilberts:
hmm wondering - in some tests we had promise chains that did multiple queries that returned `res` to then be evaluated with `should` expressions. I always reassigned `res` to keep the same names as before, but I'm guessing it would be better style to use `const` and `res, res2,... `?

emilygilberts:
The test `should handle many doc changes` is here twice, with the difference that the second one uses `expectedResponseKeys.sort((a, b)=>  a - b) `whilst the first one only uses `expectedResponseKeys.sort();`the second one is correct, cause it sorts numerically, can i remove the first test that sorty alphabetically? it looks like someone added the second one to correct the first and forgot to delete it, or am I missing something?

tests/mapreduce/test.mapreduce.js
```diff
@@ -12,134 +12,122 @@ viewTypes.forEach(function (viewType) {
...
     let createView;
     if (dbType === 'http' || viewType === 'persisted') {
-      createView = function (db, viewObj) {
+      createView = async (db, viewObj) => {
         const storableViewObj = {
-          map: viewObj.map.toString()
+          map: `${viewObj.map}`
         };
         if (viewObj.reduce) {
-          storableViewObj.reduce = viewObj.reduce.toString();
```
janl, line 23: in general I think we’d prefer a function call over manual string interpolation, but maybe I am missing a particular reason for this change here?

tests/mapreduce/test.mapreduce.js
```diff
@@ -12,134 +12,122 @@ viewTypes.forEach(function (viewType) {
...
       };
     } else {
-      createView = function (db, viewObj) {
-        return new Promise(function (resolve) {
-          setTimeout(function () {
-            resolve(viewObj);
-          });
-        });
+      createView = async (db, viewObj) => {
+        await Promise.resolve();
```
janl, line 36: would making the function here async not already defer execution, or do we want to do an extra round of that in this context?

tests/mapreduce/test.mapreduce.js
```diff
@@ -12,134 +12,122 @@ viewTypes.forEach(function (viewType) {
...
       };
     }
 
-    beforeEach(function () {
+    beforeEach(async () => {
       if (dbType === 'http') {
         const url = new URL(dbName);
         const dbUrl = `${url.origin}${url.pathname}`;
-        return PouchDB.fetch(dbUrl + '?q=1', {
+        await PouchDB.fetch(dbUrl + '?q=1', {
```
janl, line 45: no need to await, we can just return the `fetch` promise as before.

janl (commented):
reviewed up to line 518

PR #9212, opened 2026-03-27
Multitab tests

jcoglan:
## Overview

This PR adds a testing setup for multi-tab situations, and includes a test that
reproduces a reported bug (see linked issues and PRs) when the `indexeddb`
adapater operates in multi-tab settings.

I have a theory of why multi-tab interactions are broken on `indexeddb` which I
include a couple of possible fixes for. The theory goes like this:

- When you open a `PouchDB` instance on this adapter in tab A, it calls
  `indexedDB.open(dbname, v1)` where `v1` is the database version. In PouchDB
  the version is basically the current unix time.

- When another tab B opens the same database it calls `indexedDB.open(dbname,
  v2)` where `v2 > v1`. This triggers `onversionchange` in tab A since another
  client has opened the database at a higher version.

- PouchDB handles `onversionchange` by doing `indexedDB.open(dbname)` with no
  version, i.e. it opens the database at whatever version is already open. This
  is to prevent multiple tabs interacting with the database from thrashing
  `open()` calls as they keep invalidating each other's connections with higher
  version numbers. (A comment in the code refers to this as an "infinite loop"
  which I think is inaccurate; calling `open()` only happens if you interact
  with the database, so doesn't passively cause an infinite loop, but it does
  cause thrashing if multiple tabs interact with the database.)

- Aside: within a single-tab context, IndexedDB connections are cached by name
  so only one connection per database is needed, so this problem does not happen
  in single-tab contexts even if multiple `PouchDB` instances are active.

- The other ingredient of this problem is that the adapter caches a "metadata"
  object containing `doc_count` and `seq` in memory when the database is first
  opened. This data is stored in the `meta` object store and only updated by the
  `bulkDocs()` function. All other operations read this value from the in-memory
  cache. The `seq` value _must_ be up to date as each doc update in the database
  must have a unique `seq` value, and this is enforced by a unique index in
  IndexedDB. The error that occurs in the multi-tab case is IndexedDB reporting
  a violation of this constraint.

Therefore, the following set of events are possible:

1. Tab A calls `indexedDB.open(name, v1)`
2. Tab B calls `indexedDB.open(name, v2)` where `v2 > v1`
3. This causes tab A to call `indexedDB.open(name)`

The last `open()` does not trigger `onversionchange` in tab B, and so there are
now two active connections to the same IndexedDB database. And both of these
PouchDB instances have an in-memory cache of the `{ doc_count, seq }` metadata.
When one of these clients updates a doc, the cached metadata in the other tab
becomes stale, and this causes a `seq` conflict when it next tries to update a
doc. It also causes the result of `info()` and other methods to be incorrect.

Why does PouchDB use versions in this way?

- IndexedDB puts an integer version on all databases.
- You're only allowed to change the definitions of object stores and indexes
  when the database is first opened, and only when it is opened with a higher
  version than it was last opened at.
- PouchDB uses native IndexedDB indexes to represent Mango indexes so it needs
  to make sure these are up to date when the database is opened.
- Since you can't read anything from the database before opening it, PouchDB
  uses the current time as the version to give itself the opportunity to define
  any necessary indexes.
- When you call `createIndex()`, this also bumps the database's version,
  triggering `onversionchange` in other tabs and forcing them to re-open the
  database with the right indexes defined.

Possible fixes I am considering:

1. _Always give the version when re-opening the database._ i.e. remove the
   mechanism that means `open()` is called without a version number when it's
   triggered via `onversionchange` and always pass in the current time. This
   changes nothing for single-tab contexts, they can continue to cache metadata.
   However, it means any interaction with the database will cause
   `onversionchange` in other tabs and force them to call `open()` and reload
   the metadata again. So in multi-tab contexts, this causes `open()` thrashing.

2. _Stop caching metadata._ If multiple connections can coexist without
   triggering `onversionchange` on each other, then caching metadata in memory
   is simply wrong as there's no way to detect that it's stale. (You _could_
   handle conflicts on `seq` when updating docs but this still allows `info()`
   to return stale data.) This means metadata would always be re-read from the
   `meta` store for all interactions, even for single tabs, which could have
   performance implications.

I have not been able to see any measurable difference in the performance tests
as a result of removing metadata caching. To me, option 1 seems strictly worse
than 2 in the multi-tab case, since re-opening the connection also entails
re-reading the metadata, so in effect it removes metadata caching when `open()`
thrashing occurs. Option 2's possible performance downside in single-tab
situations does not seem measurable so I'm inclined to prefer this solution.

Implementations of both solutions are available in the last 2 commits, so
whichever one we decide to go with, the other can be removed from the history.

I have learned one thing from implementing solution 2: `bulkDocs()` needs the
metadata in memory, because it uses `idb_attachment_format` to decide things
inside `preProcessAttachments()`, which is async and so cannot be done inside an
IndexedDB transaction. We pre-process attachments _and then_ start a transaction
to write the docs and updated metadata.

I have solved this by keeping only some of the the metadata returned by
`setup()` in a var called `connectionInfo` and passing this in to `bulkDocs()`,
while everything else looks the metadata up dynamically. `bulkDocs()` also loads
it dynamically inside the transaction to get up to date `doc_count` and `seq`.
The intended distinction here is that `connectionInfo` only holds information
that will not change when docs are updated, so there's no risk of accidentally
reading a stale `doc_count` or `seq` from it.

- https://github.com/apache/pouchdb/issues/8209
- https://github.com/apache/pouchdb/issues/8494
- https://github.com/apache/pouchdb/pull/8495

## Checklist

- [x] I am not a bot
- [x] Changes are covered by tests

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -21,7 +21,8 @@
     "ArrayBuffer": "readonly",
     "FileReaderSync": "readonly",
     "emit": "readonly",
-    "PouchDB": "readonly"
+    "PouchDB": "readonly",
+    "__pouch__": "readonly"
   },
 
   "rules": {
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -21,8 +21,7 @@
     "ArrayBuffer": "readonly",
     "FileReaderSync": "readonly",
     "emit": "readonly",
-    "PouchDB": "readonly",
-    "__pouch__": "readonly"
+    "PouchDB": "readonly"
   },
 
   "rules": {
--- a/TESTING.md
+++ b/TESTING.md
@@ -22,6 +22,7 @@ help fixing them though.
   - [Test options](#test-options)
 - [Other sets of tests](#other-sets-of-tests)
   - [`find` and `mapreduce`](#find-and-mapreduce)
+  - [multi-tab tests](#multi-tab-tests)
   - ["Fuzzy" tests](#fuzzy-tests)
   - [Performance tests](#performance-tests)
   - [Running tests in the browser](#running-tests-in-the-browser)
@@ -234,6 +235,24 @@ like so:
 
 Note that the default choice for the `SERVER` value (`pouchdb-express-router`)
 does not support `find` or `mapreduce` and does not need to pass these tests.
+
+### Multi-tab tests
+
+We have a few tests to check that interactions with backing stores like
+IndexedDB are synchronized across tabs where multiple PouchDB instances are
+connected to the same database. These have to be scripted via Playwright rather
+than being run inside the normal unit test suite. To run them:
+
+    $ npm run test-multitab
+
+Two environment variables can be used to control how these are run:
+
+- `CLIENT`: may be `firefox`, `chromium` or `webkit` to select the browser where
+  the tests will be run.
+- `ADAPTERS`: may be `idb` or `indexeddb` to set which backend adapter PouchDB
+  will use.
+
+All combinations of these should be tested on CI.
 
 ### "Fuzzy" tests
 
--- a/tests/multitab/user_agent.js
+++ b/tests/multitab/user_agent.js
@@ -1,0 +1,53 @@
+'use strict';
+
+const playwright = require('playwright');
+
+const ADAPTERS = process.env.ADAPTERS || 'indexeddb';
+const CLIENT = process.env.CLIENT || 'firefox';
+const SHELL_URL = 'http://127.0.0.1:8000/tests/multitab/shell.html';
+
+class UserAgent {
+  static async start() {
+    let browser = await playwright[CLIENT].launch();
+    let context = await browser.newContext();
+    return new UserAgent(ADAPTERS, browser, context);
+  }
+
+  constructor(adapter, browser, context) {
+    this._adapter = adapter;
+    this._browser = browser;
+    this._context = context;
+    this._pages = new Map();
+  }
+
+  async stop() {
+    await this._browser.close();
+  }
+
+  async eval(pageId, fn) {
+    let page = await this._getPage(pageId);
+    return page.evaluate(fn);
+  }
+
+  _getPage(id) {
+    if (!this._pages.has(id)) {
+      this._pages.set(id, this._setupPage());
+    }
+    return this._pages.get(id);
+  }
+
+  async _setupPage() {
+    let page = await this._context.newPage();
+    await page.goto(SHELL_URL);
+
+    if (this._adapter === 'idb') {
+      await page.evaluate(() => window.__pouch__ = new PouchDB('testdb', { adapter: 'idb' }));
+    } else if (this._adapter === 'indexeddb') {
+      await page.evaluate(() => window.__pouch__ = new PouchDB('testdb', { adapter: 'indexeddb' }));
+    }
+
+    return page;
+  }
+}
+
+module.exports = UserAgent;
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -158,6 +158,30 @@ jobs:
         if: steps.retry.outcome == 'failure'
         run: ${{ matrix.cmd }}
 
+  # Run the multitab tests in every supported browser
+
+  multitab:
+    needs: lint
+    strategy:
+      fail-fast: false
+      matrix:
+        client: ['firefox', 'chromium', 'webkit']
+        adapter: ['idb', 'indexeddb']
+    runs-on: ubuntu-latest
+    env:
+      CLIENT: ${{ matrix.client }}
+      ADAPTERS: ${{ matrix.adapter }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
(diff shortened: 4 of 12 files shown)


janl (approved):

alxndrsn (changes requested):
This looks great! 👍 

Not sure that the changes to `README.md` are relevant.

tests/multitab/user_agent.js
```diff
@@ -0,0 +1,53 @@
...
+  async stop() {
+    await this._browser.close();
+  }
+
+  async eval(pageId, fn) {
+    let page = await this._getPage(pageId);
+    return page.evaluate(fn);
+  }
+
+  _getPage(id) {
```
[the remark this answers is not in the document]
jcoglan: I was unsure about this. Everyone discussing these issues uses "tab" but Playwright uses "page". Strictly speaking, a "tab" is a particular UI for operating multiple pages, but it's not the only one. Colloquially I think "multi-tab" communicates "several pages operating concurrently" better than "multi-page" which to me sounds like a multi-step sequential workflow.

.eslintrc.json
```diff
@@ -21,7 +21,8 @@
     "ArrayBuffer": "readonly",
     "FileReaderSync": "readonly",
     "emit": "readonly",
-    "PouchDB": "readonly"
+    "PouchDB": "readonly",
+    "__pouch__": "readonly"
```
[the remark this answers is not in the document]
alxndrsn: Not a big issue, but globals can also be scoped via nested `.eslintrc` files, and discouraging project-wide globals is probably helpful.

tests/multitab/user_agent.js
```diff
@@ -0,0 +1,53 @@
...
+
+  async _setupPage() {
+    let page = await this._context.newPage();
+    await page.goto(SHELL_URL);
+
+    if (this._adapter === 'idb') {
+      await page.evaluate(() => window.__pouch__ = new PouchDB('testdb', { adapter: 'idb' }));
+    } else if (this._adapter === 'indexeddb') {
+      await page.evaluate(() => window.__pouch__ = new PouchDB('testdb', { adapter: 'indexeddb' }));
+    }
```
alxndrsn, line 47: What happens for other adapters?  Would it be helpful to throw if `__pouch__` is not going to be instantiated?

tests/multitab/user_agent.js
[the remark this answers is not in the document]
alxndrsn: > `page` is the playwright API language and it is consistent with that. Of course we can change it, but not sure if worth it?

I see that :+1:  Just commenting as it confused me, and I'd expect this to be an area of the code which I'm familiar with.  Hopefully it will be obvious to future readers.

TESTING.md
```diff
@@ -5,14 +5,16 @@ run them. PouchDB has been primarily developed on Linux and macOS, if you are
 using Windows then these instructions will have problems, we would love your
 help fixing them though.
 
-> [!WARNING] VERY IMPORTANT
-> **Always set the `COUCH_HOST` env var to a proper CouchDB for the integration tests!**
+> [!WARNING]
+> VERY IMPORTANT **Always set the `COUCH_HOST` env var to a proper CouchDB for
+> the integration tests!**
```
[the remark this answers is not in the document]
alxndrsn: > @alxndrsn do you prefer these to be separate PRs?

Yes - makes it much easier when reviewing, and especially for repeated reviews, when you don't have to mentally filter out some changes.

It can often simplify digging through old commits too.

.eslintrc.json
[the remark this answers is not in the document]
jcoglan: Thanks, I've moved this rule into `tests/multitab`

tests/multitab/user_agent.js
jcoglan: I have only set CI up to run these tests with the two adapters that are backed by IndexedDB, due to the nature of the bug these tests check for. Would it be better to run all the browser adapters here?

alxndrsn: It definitely seems reasonable to run all browser-based adapters, although you may uncover more bugs :face_with_peeking_eye: 

If that's the case, I'd skip those tests for now, and create separate issues.

Either way I think it's helpful for this setup to throw a clear error if asked to initialise an adapter that it doesn't have code for.

janl: @alxndrsn I’m not sure what the benefit of the things you request here are? This is a test targeted at a specific adapter issue, I’m not sure how other adapters come into play?

alxndrsn: Any browser adapter is likely to have similar requirements for running in multiple tabs simultaneously.  Why would these other adapters be excluded from testing?

janl: right, but that’s not the scope of this PR afaict, can we merge this and open an issue to make more broadly applicable multi-tab tests? Ideally we run a suite of operations and make sure they all return correct results. The minimal repro for just the one adapter here is unlikely to surface anything in other adapters.

PR #9253, opened 2026-06-04
refactor: drop UUID dependency

SourceR85:
## Overview

This replaces the npm dependency [uuid](https://github.com/uuidjs/uuid) with standard [crypto.randomUUID](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID).

randomUUID generates UUID v4: the only version we're using in PouchDB.

Since uuid fixed another CVE in version 14 and introduced a breaking change in v13, dependabot's PR no longer succeeds:
https://github.com/apache/pouchdb/pull/9237

The PR by [mohagan9](https://github.com/mohagan9) (update to 11.1.1) will also be obsolete:
https://github.com/apache/pouchdb/pull/9250

## Checklist

- [x] I am not a bot
- [x] Changes are covered by tests
- [ ] I modified the [types](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/pouchdb) to reflect the changes, if necessary


packages/node_modules/pouchdb-utils/src/uuid-browser.js
```diff
@@ -0,0 +1,3 @@
+export default function uuid() {
+  return crypto.randomUUID();
```
alxndrsn: Does this break http:// support?

SourceR85: You're right: it only works in secure context: https (works despite invalid certificates in FF) or localhost.
That requirement not applied in all browser engines, but strict in specs.
I'll later push a polyfill for that case.

SourceR85: Polyfill for `randomUUID` is now applied:
Firefox and Chrome don't provide the function if context is insecure.

packages/node_modules/pouchdb-utils/src/rev.js
```diff
@@ -7,7 +6,7 @@ import { stringMd5 } from 'pouchdb-md5';
  */
 function rev(doc, deterministic_revs) {
   if (!deterministic_revs) {
-    return uuidV4().replace(/-/g, '').toLowerCase();
+    return crypto.randomUUID().replace(/-/g, '').toLowerCase();
```
alxndrsn: I think you can drop `toLowerCase()` here - the spec says:

> For the steps described in the algorithm to [generate a random UUID](https://w3c.github.io/webcrypto/#dfn-generate-a-random-uuid), the _**hexadecimal representation**_ of a byte _value_ is the two-character string created by expressing _value_ in hexadecimal using [ASCII lower hex digits](https://infra.spec.whatwg.org/#ascii-lower-hex-digit)
> - https://w3c.github.io/webcrypto/#Crypto-method-randomUUID

alxndrsn: Also is there a reason to use `crypto.randomUUID()` directly in preference to the new `uuid` package?

packages/node_modules/pouchdb-utils/src/uuid-browser.js
```diff
@@ -0,0 +1,41 @@
...
+  buffer[6] = (buffer[6] & 0x0f) | 0x40;
+  // and variant: 1
+  buffer[8] = (buffer[8] & 0x3f) | 0x80;
+
+  const str = toHex(buffer);
+
+  return str.slice(0, 8) + "-"
+    + str.slice(8, 12) + "-"
+    + str.slice(12, 16) + "-"
+    + str.slice(12, 16) + "-"
```
alxndrsn: I suspect this means the generated UUID will be the wrong length.  Maybe worth a small test?

packages/node_modules/pouchdb-utils/src/rev.js
SourceR85: Ah, damn... I implemented the polyfill after the hint from you: seems like I missed to revert that line.

packages/node_modules/pouchdb-utils/src/uuid-browser.js
SourceR85: You're right: my implementation is 41 chars long... should be 36 🤦

We already have `tests/test.uuids.js` who should call the uuid, but it doesn't run with polyfill.

I think, I have to export a function to evaluate randomUUID availability at runtime, so I can remove the native function for testing.

SourceR85: I've updated my implementation and add a test for "insecure" mode.

PR #9257, opened 2026-06-14
Fix elemMatch null object array handling

### LINKED ISSUE #9103: elemMatch throws TypeError: value is null  if array has null element on non-first position
pouchdb-find throws `TypeError: value is null`  if array has null element on non-first position.
Example: 
```{ test: [{ foo: "blub"}, null] }```

Exception:
```
TypeError: value is null
    getFieldFromDoc index.es.js:10
    rowFilter index.es.js:396
    rowFilter index.es.js:393
    elemMatch index.es.js:557
    $elemMatch index.es.js:556
    match index.es.js:473
    matchSelector index.es.js:417
    matchSelector index.es.js:413
    rowFilter index.es.js:401
    rowFilter index.es.js:393
```

Code:
```
var matchers = {
  "$elemMatch": function(doc, userValue, parsedField, docFieldValue) {
    if (!Array.isArray(docFieldValue)) {
      return false;
    }
    if (docFieldValue.length === 0) {
      return false;
    }
    if (typeof docFieldValue[0] === "object" && docFieldValue[0] !== null) {
      return docFieldValue.some(function(val) {
        return rowFilter(val, userValue, Object.keys(userValue)); // << val is null
      });
    }
[...]
```

henyanagar:
Fixes #9103

### Summary

This updates `$elemMatch` handling for arrays that contain both objects and `null` values.

Previously, `$elemMatch` decided whether to use `rowFilter()` based only on the first array element. If the first element was an object, later `null` entries were still passed into `rowFilter()`, which could throw a TypeError when resolving nested fields.

The updated logic checks each array element before applying object-field matching, so `null` entries are ignored for object-field `$elemMatch` selectors instead of throwing.

### Testing

* Added a regression test for `$elemMatch` with an object array containing `null`.
* Verified the new test fails before the fix with:
  `TypeError: Cannot read properties of null`
* Verified after the fix:
  `1 passing`
* Verified the full elem-match test file:
  `7 passing`
* Ran ESLint on the changed source and test files.

### DIFF
--- a/packages/node_modules/pouchdb-selector-core/src/in-memory-filter.js
+++ b/packages/node_modules/pouchdb-selector-core/src/in-memory-filter.js
@@ -219,9 +219,15 @@ var matchers = {
       return false;
     }
 
-    if (typeof docFieldValue[0] === 'object' &&  docFieldValue[0] !== null) {
+    var userValueKeys = Object.keys(userValue);
+    var userValueHasNonOperator = userValueKeys.some(function (key) {
+      return key.indexOf('$') !== 0;
+    });
+
+    if (userValueHasNonOperator) {
       return docFieldValue.some(function (val) {
-        return rowFilter(val, userValue, Object.keys(userValue));
+        return typeof val === 'object' && val !== null &&
+          rowFilter(val, userValue, userValueKeys);
       });
     }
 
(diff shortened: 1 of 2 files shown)


janl:
Hi, thank you for your contribution. Would you be able to use our pull request template, so we can process your submission according to our guidelines? Thank you.

